PR #8, opened 2017-05-27
fix python 3 support

tgarc:
+ urllib2 doesn't exist in python 3 - import from urllib.request instead

+ for linux builds open docker process with universal_newlines; this
  opens a pipe in 'text' mode using the default preferred encoding


joerick:
Thanks @tgarc!

PR #9, opened 2017-06-03
add a 'pip install' step before building wheels

tgarc:
This commit makes 3 changes for builds on all platforms:

1) pip install of the package before the `pip wheel` command
2) add `--no-deps` to the `pip wheel` command since (1) already installs dependencies. This also avoids the additional step of building wheels for dependencies (which was the previous behavior)
3) add `--upgrade --force-reinstall --no-deps` to the pip install comand after pip wheel. This makes sure that the build still tests installing from the built wheel (which is consistent with previous behavior).

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -51,7 +51,10 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
                     PATH=$PYBIN:$PATH sh -c {before_build}
                 fi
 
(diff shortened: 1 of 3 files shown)


joerick:
Great! Thanks for this @tgarc, I'll try and get this merged this week once I've had a chance to try it out a bit :)

joerick:
Thanks @tgarc - merged! I just cut another release on PyPI, 0.2.1.

tgarc:
awesome sauce! thanks joe!

PR #14, opened 2017-07-02
minor fixes including python3 fix

tgarc:
+ In linux environment, if there was no matching python configurations for a
  particular architecture (i.e. if that architecture was skipped) the docker
  build script would just crash. The offending line:

    for whl in /tmp/linux_wheels/*.whl; do

  would result in whl being equal to the unexpanded glob expression. Rather
  than checking this in the bash script, python checks whether there are any
  actual wheels to build and skips the script if there are none.

+ added universal_newlines=True when making the call to grab the package name -
  this fix is required for python3 since other a bytes prefix (b'') will be
  added to the package name

+ In the linux/docker environment, added a try/catch statement around the
  docker script for KeyboardInterrupt that will kill the docker subprocess in
  cases where cibuildwheel is being run interactively.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -77,7 +77,8 @@ def main():
 
     try:
         project_setup_py = os.path.join(project_dir, 'setup.py')
-        name_output = subprocess.check_output([sys.executable, project_setup_py, '--name'])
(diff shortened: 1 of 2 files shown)


tgarc:
BTW this has no relation to recent discussions (e.g.  #11)

joerick:
Thanks tgarc!

Issue #16, opened 2017-07-05
Add CIBW_ENVIRONMENT option


YannickJadoul:
Sorry to open yet another issue, but in my efforts to try getting my wheels built on the manylinux docker image, I'm running into the problem that there seems no way in `cibuildwheel` to prepare the container (i.e. install stuff and set environment variables).

In my particular case, for example, I want to install a newer GCC toolchain (https://github.com/squeaky-pl/centos-devtools/releases), set the PATH to use it, and set the `CXXFLAGS` to be picked up by my build.

Am I overlooking a possibilty, or is this currently not possible?

(Probably also related to #13)

joerick:
We discussed in #6 the possibility of a `CIBW_ENVIRONMENT` option. I think that's definitely needed in your case. I'm really busy a the moment, but contributions welcome!

joerick:
#7 would also be useful for this too

YannickJadoul:
Thanks! Hadn't found that issue. I'll have a look at that, and see what I can do to implement something and get a pull request!

YannickJadoul:
Did you have anything in mind on what that CIBW_ENVIRONMENT would do?
Would it be bash code that could install stuff, etc? Or would it be purely a specification of environment variables to be set?

joerick:
CIBW_ENVIRONMENT would purely set environment variables. We'll need format that'll fit into an environment variable itself - I'm thinking space-separated variables, as parsed by `shlex.split()` an example value of it would be `PATH="/usr/local/bin/:$PATH" CXXFLAGS=-Wall`.

YannickJadoul:
Sounds reasonable, and fairly easily implementable. I'll start working on that in a bit.

Three more questions/thougths:
- Should this happen before the main loop over Python version or within that loop?
I mean, there's up- and downsides to both: the former allows a single execution of things and doesn't get into problems with applying e.g. `PATH="/usr/local/bin/:$PATH"` multiple times, but the latter allows for customization per build (in case environment variables depend on the Python version).
- What about installations common for all Python executables (e.g. the GCC suite in my case)? Feels silly to repeat the download and extraction all the time, doesn't it? (You can download before calling cibuildwheel and pass it on, ofc, or you could somehow check, of course, but that gets messy rather quickly.)
- This is actually only necessary in Linux. Since OS X and Windows don't have docker images, you can just do all these things before calling cibuildwheel. So ... should we make this something specific for Windows, or something extra for all platforms for the sake of consistency?

[...]

Issue #13, opened 2017-06-29
How to use cibuildwheel under this "unusual" setup


sgeto:
Hi @joerick

I could use some help.
The wheels I'm trying to build are a binding for a C library. The `setup.py` file is located in a sub folder (python) where the python build takes place (as part of the overall build or seperatly).
However, during build the compiler needs access to additional header and source file located in other sub folders (src include) located on the same level as the python folder.

cibuildwheel seems to only import the folder holding the setup.py file. And the build sadly fails. I tried everything.  Setting global environment variables, adding the whole `src` for as project and adding cibuildwheel pre-build commands.

How can I use cibuildwheel in such a setup?

Thanks a lot

Project: https://github.com/sgeto/libdnet
Travis: https://travis-ci.org/sgeto/libdnet/

joerick:
cibuildwheel does take a project path, like

    cibuildwheel --output-dir wheelhouse python

But in your case, this is going to be a problem because cibuildwheel mounts the project folder on Linux, so the build machine will only have access to the python folder, not the sibling folders.

I can't think of an easy solution we could make... maybe it would work if you moved the setup.py to the root of the repo and build from there?

sgeto:
If cibuildwheel would not look for a setup.py file when attempting to mount the project folder, I could import the whole repo and add e.g. `cd python` as a pre-build command.
Do you think an option to not check for the presents of a setup.py file (at mount time) is a good idea?

sgeto:
By the way moving setup.py to the root of the repo is a possibility

joerick:
Ah, the pre-build command executes in a subshell, so cd'ing won't affect the working dir.

sgeto:
I'm not sure I'm following. Is that good?

joerick:
Sorry, I mean cd'ing won't work the way you're thinking because it won't change cibuildwheel's working dir.

sgeto:
I see.

> I can't think of an easy solution we could make

Can you tell me more about the not so easy solution? Is it related to docker or more of a python limitation?

tgarc:
@sgeto I don't your project setup is uncommon - it just isn't primarily a python package. cibuildwheel is still in early stages so I imagine these kind of use cases likely haven't come up yet. Anyway, just to rephrase the issue you're having:

1) Your setup.py is not in your root folder. cibuildwheel doesn't currently support this configuration as it expects everything required to be contained under the directory where setup.py is located.

[...]

PR #17, opened 2017-07-09
Allowing CIBW_BEFORE_BUILD to modify the shell environment of pip wheel

YannickJadoul:
This is one way #16 could be solved (and which could also solve #13).

Based on @tgarc's suggestion:

> And if the sh -c was removed, would that allow you to do what you wanted?

I basically prepended the content of `CIBW_BEFORE_BUILD` to the `pip` commands, such that they are all evaluated in a `sh -c`. Consequently, the cibuildwheel user could set environment variables, change directory, or do anything else, and this wíll affect the `pip wheel` but won't affect any of the following commands.

Just one path of the many solutions that might work, but this one seems to stick quite close to the original plan, as far as I can tell.
Any thoughts, remarks, complaints?

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -55,14 +55,15 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             cd /project
 
             for PYBIN in {pybin_paths}; do
-                if [ ! -z {before_build} ]; then
-                    PATH=$PYBIN:$PATH sh -c {before_build}
-                fi
+                # Put quoted string in 'ordinary' environment variable
+                CIBW_BEFORE_BUILD_={before_build}
 
-                # install the package first to take care of dependencies
-                "$PYBIN/pip" install .
+                PATH=$PYBIN:$PATH sh -c "$CIBW_BEFORE_BUILD_
 
-                "$PYBIN/pip" wheel --no-deps . -w /tmp/linux_wheels
+                    # install the package first to take care of dependencies
+                    pip install .
+
+                    pip wheel --no-deps . -w /tmp/linux_wheels"
             done
 
             for whl in /tmp/linux_wheels/*.whl; do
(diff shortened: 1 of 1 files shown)


tgarc:
So what is the /host mount for ? I've never used a cache with TravisCI... Is this change required for this PR to work? Or is it unrelated?

YannickJadoul:
It's actually mostly unrelated, probably. It's to allow access to other folders of the host travis ci (such as ccache's cache (a compiler cache, to not rebuild the exact same file multiple times), but just as well something you'd have downloaded for all docker containers, or...). Right now you cannot access something else than the project or wheel output folder. But it's not needed for the rest of this PR. 

Reason I added it is because it's related to why I needed to solve this environment problem, and I wanted to use this PR in my repository before it's merged. But if you want, I could split it up.

tgarc:
Actually this could only ever solve #13 partially; the issue there is that setup.py is not even in the root folder.

YannickJadoul:
@tgarc Hmmm, why? I was assuming that if you would now run `CIBW_BEFORE_BUILD="cd subdir" cibuildwheel` that the `pip wheel --no-deps . -w /tmp/linux_wheels` would run in that subdir, solving the problem of `setup.py` being in that subdirectory?
I haven't tested that though, so where did my reasoning go wrong? :-D

tgarc:
@YannickJadoul because setup.py is checked for in the __main__ script before even calling the CIBW_BEFORE_BUILD

YannickJadoul:
Aha, yes.. Damn, did not see that :-( Thanks for that reality check...

joerick:
Thanks for this @yannickjadoul. I dont think I want to accept this because I think once we have CIBW_ENVIRONMENT, the existing behaviour should work fine for your package I think. Also I think keeping commands isolated is _probably_ a good idea - I can imagine people getting surprised when CIBW_BEFORE_BUILD="cd bin && ./prepare" breaks their build. At least, I'm used to Makefiles where this might look strange.

YannickJadoul:
Hmmm, not sure how flexible things cán get, then. But maybe my use case (copying a different gcc version into that docker) is too complex for what cibuildwheel was meant to do, and I should roll my own...

PR #18, opened 2017-07-09
revert PR9; add --no-deps to pip wheel; address PR11

tgarc:
This is a WIP - there will be some tweaking but the proof of concept is there and should be fully functional. I've tested this with my package on all platforms. This should address the issue raised by PR #11 without needed an additional pip pre-install step. It also removes the step of building wheels for dependencies - a step which adds extra build time and may even cause issues for some packages. BTW this follows more closely the build steps in the https://github.com/matthew-brett/multibuild project.

Revert PR9; it was never actually needed in the first place it just fixed an
issue by side effect.

1) add --no-deps to pip wheel step in order to avoid building wheels for
   dependencies. This also addresses the issues raised in PR #11.

2) Modify the pip install step across platforms to allow installation of
   dependencies from the index. (This is required since we will no longer be
   building wheels for dependencies).

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -53,30 +53,33 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             cd /project
 
             for PYBIN in {pybin_paths}; do
+                # Setup
+                rm -rf /tmp/built_wheel
+                rm -rf /tmp/delocated_wheel
+                mkdir /tmp/built_wheel
+                mkdir /tmp/delocated_wheel
+
                 if [ ! -z {before_build} ]; then
                     PATH=$PYBIN:$PATH sh -c {before_build}
                 fi
 
-                # install the package first to take care of dependencies
-                "$PYBIN/pip" install .
+                # Build that wheel
+                "$PYBIN/pip" wheel . -w /tmp/built_wheel --no-deps
+                built_wheel=(/tmp/built_wheel/*.whl)
 
-                "$PYBIN/pip" wheel --no-deps . -w /tmp/linux_wheels
-            done
-
-            for whl in /tmp/linux_wheels/*.whl; do
-                if [[ "$whl" == *none-any.whl ]]; then
-                    # pure python wheel - just copy to the output
-                    cp "$whl" /output
+                # Delocate the wheel
+                # NOTE: 'built_wheel' here is a bash array of glob matches; "$built_wheel" returns
+                # the first element
+                if [[ "$built_wheel" == *none-any.whl ]]; then
+                    # pure python wheel - just copy
+                    mv "$built_wheel" /tmp/delocated_wheel
                 else
-                    auditwheel repair "$whl" -w /output
+                    auditwheel repair "$built_wheel" -w /tmp/delocated_wheel
                 fi
-            done
+                delocated_wheel=(/tmp/delocated_wheel/*.whl)
 
-            # Install packages and test
-            for PYBIN in {pybin_paths}; do
                 # Install the wheel we just built
-                "$PYBIN/pip" install {package_name} \
-                    --upgrade --force-reinstall --no-deps --no-index -f /output
+                "$PYBIN/pip" install "$delocated_wheel"
 
                 # Install any requirements to run the tests
                 if [ ! -z "{test_requires}" ]; then
@@ -89,6 +92,9 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
                     PATH=$PYBIN:$PATH sh -c {test_command}
                     popd
                 fi
+
+                # we're all done here; move it to output
+                mv "$delocated_wheel" /output
             done
         '''.format(
             package_name=package_name,
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -1,5 +1,5 @@
 from __future__ import print_function
-import os, subprocess, shlex, sys
+import os, subprocess, shlex, sys, shutil
 from collections import namedtuple
 from glob import glob
 try:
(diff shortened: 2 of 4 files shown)


tgarc:
@YannickJadoul would you test this?

YannickJadoul:
@tgarc Yes, please! Thanks. I'll let you know. 

If you want, I'd be curious to have your input on the solution to the other issues in PR #17, too.

YannickJadoul:
Seems to be working! :-)

2 questions, if I may:

1. So now the `install` step before is not happening, I see. So how are the dependencies installed (the ones that PR #9 adressed, such as that 'cffi' you needed) ?

2. Why is the `filter_wheels` needed? Can't you just `pip install -f /output` and let `pip` sort out which versions it can use?

tgarc:
@YannickJadoul thanks for testing! 

> So how are the dependencies installed (the ones that PR #9 adressed, such as that 'cffi' you needed) ?

Unfortunately I caused some confusion with my previous PR. Dependencies aren't usually required to do a wheel build, only to install that wheel. (Although some builds do require 'build-dependencies'). In my case doing a pre- pip install was fixing an issue by side-effect that was actually being caused by including setup_requires=['cffi']. (This seems like a bug in cffi or setuptools etc). IOW, my PR was only ever a workaround for a cffi issue.

> Why is the filter_wheels needed? Can't you just pip install -f /output and let pip sort out which versions it can use?

Okay so now that no wheels are being built for dependencies I had to remove the `--no-index` so that pip can search PyPI for packages. Unfortunately, pip has no way to specify preference with the `--find-links` option so that it installs the local wheel instead of the package found on PyPI. Instead it will just install the listed packages from the first place that it can find it (which happens to be PyPI). Does that make sense? This same issue was actually brought up on the pip issue tracker on github.

tgarc:
Okay, think I've cleaned up as much as I can. @joerick take a look when you have time.

YannickJadoul:
> Unfortunately I caused some confusion with my previous PR. Dependencies aren't usually required to do a wheel build, only to install that wheel. [...]

Ok, perfect :-) That makes things easier than I expected than, indeed :-)

> [...] Instead it will just install the listed packages from the first place that it can find it (which happens to be PyPI). Does that make sense? [...]

Yeps, that does. Weirdly enough though, I had tried that myself, before, and had even been looking up in the pip source (https://github.com/pypa/pip/blame/21be153044a7aa245e12ce3f86793e9b17201519/pip/index.py#L464-L468) and thought that local files díd get precedence. But I was just curious. This is not a monster hack; it's pretty neat, actually. Just a pity you nééded a workaround.

YannickJadoul:
Aaaaaaaaaah, I've actually just been testing (with `-vvv`, extra extra verbose `pip`), and I think I just realized why the tests are failing and why you need that `filter_wheels` command: I think `pip` dóes give precedence to local files, but it gives even more precedence to having the latest version.

In this case, there is already a `spam` package in the PyPI (https://pypi.python.org/pypi/spam), meaning that even though the local `spam` wheel (with `-f`) would be preferred to the PyPI, the PyPI one has a higher version (4.0.9) which shadows the wheel that was just built (0.1.0).

cibuildwheel/windows.py
```diff
@@ -17,11 +17,11 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             request = urlopen('https://github.com/ogrisel/python-appveyor-demo/raw/09a1c8672e5015a74d8f69d07add6ee803c176ec/appveyor/run_with_env.cmd')
             f.write(request.read())
 
-    def shell(args, env=None, cwd=None):
+    def shell(args, env=None, cwd=None, **kwargs):
```
YannickJadoul: `kwargs` does not seem to be used anymore?

cibuildwheel/macos.py
```diff
@@ -58,14 +58,11 @@ def shell(args, env=None, cwd=None):
...
 
-        # install the package first to take care of dependencies
-        shell([pip, 'install', project_dir], env=env)
-
         # build the wheel to temp dir
         temp_wheel_dir = '/tmp/tmpwheel%s' % config.version
         shell([pip, 'wheel', project_dir, '-w', temp_wheel_dir, '--no-deps'], env=env)
-        temp_wheel = glob(temp_wheel_dir+'/*.whl')[0]
 
+        temp_wheel = glob(temp_wheel_dir+'/*.whl')[0]
```
YannickJadoul: Is there a chance that multiple wheels get built?
I don't see how or why one package could do that, but the Linux version is looping over all wheels. So for consistency we can maybe either change this line, or the one in linux.py?

tgarc: This is still an open question. I still don't know of any projects where multiple wheels get built so I decided not to make any changes, at least for this PR.

cibuildwheel/windows.py
tgarc: Good catch! Thanks.

cibuildwheel/macos.py
YannickJadoul: Ok, fair enough. Fine with me.
I have just realised that "for whl in /tmp/linux_wheels/*.whl; do" is probably just easier to write the pattern match, and not per se to do that over all wheels.

cibuildwheel/util.py
```diff
@@ -1,5 +1,11 @@
 from fnmatch import fnmatch
 
+filter_wheels_cmd = '''\
+import sys, glob, itertools, pip; \
+sys.stdout.write(\
+' '.join(f for f in itertools.chain(*map(glob.iglob, sys.argv[1:])) \
+if pip.wheel.Wheel(f).supported()))\
+'''
```
joerick: I'm finding this quite hard to understand. What does this do, and might there be a more elegant way to do it? I'm not averse to separate code for each platform by the way!

cibuildwheel/linux.py
```diff
@@ -77,8 +74,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             # Install packages and test
             for PYBIN in {pybin_paths}; do
                 # Install the wheel we just built
-                "$PYBIN/pip" install {package_name} \
-                    --upgrade --force-reinstall --no-deps --no-index -f /output
+                "$PYBIN/pip" install $("$PYBIN/python" -c "{filter_wheels}" /tmp/linux_wheels/*.whl)
```
joerick: What's the reason for this change? I think the previous one is clearer and also it installs the delocated wheel, which is probably the right thing to do!

cibuildwheel/util.py
joerick: Also I think this is an undocumented Pip API... Might change in the future.

cibuildwheel/linux.py
tgarc: > and also it installs the delocated wheel

What delocated wheel do you mean here? The wheel for the package that's being built or the one(s) for dependencies?

cibuildwheel/util.py
tgarc: Sorry, lack of commenting. I can add some.
This is a short python program that takes as arguments a list of wheels and prints out only those wheels that are supported on the executor's platform. I put it here since all platforms use it and there's no need to have modify it per-platform.

But if you're asking why...hopefully my [previous comment](https://github.com/joerick/cibuildwheel/pull/18#issuecomment-313952707) makes it clear.

> Also I think this is an undocumented Pip API... Might change in the future.

I will try and verify whether that's the case. I tend to assume anything not starting with an underscore is fairly stable but this may not be true in pip world.

tgarc:
@YannickJadoul 
> In this case, there is already a spam package in the PyPI (https://pypi.python.org/pypi/spam), meaning that even though the local spam wheel (with -f) would be preferred to the PyPI, the PyPI one has a higher version (4.0.9) which shadows the wheel that was just built (0.1.0).

That's accurate. In fact it says this in the [pip docs](https://pip.pypa.io/en/stable/reference/pip_install/#satisfying-requirements): 
> it chooses which version of each requirement to install using the simple rule that the latest version that satisfies the given constraints will be installed

cibuildwheel/linux.py
tgarc: Yikes! you're absolutely right, I meant to point to the delocated wheel. See my [previous comment](https://github.com/joerick/cibuildwheel/pull/18#issuecomment-313952707) for why I made this change though.

cibuildwheel/util.py
joerick: Sorry, I had missed that previous comment! Yes I think you're right in that respect.

cibuildwheel/macos.py
joerick: I've not seen any packages that make more than one wheel, so I think this is a good assumption.

tgarc:
> I'm sorry to be slowing things down!

No worries, I'm busy too. Thanks for sharing your code.

So use a temp directory for installing...interesting. I think I can do that. I still think that the filter_wheels functionality may come in handy at some point but if we can do it your way...sounds cleaner.

tgarc:
@joerick here's a shot at doing what you suggested. Note one significant change: the linux build now more closely builds on macos/windows; it uses a single for loop to build/delocate/install/test each wheel. One thing about this that irks me a bit is that if ever `pip wheel` did produce more than one wheel the linux build would produce some cryptic error whereas other platforms only select the first globbed wheel.

Anyway, the tests are failing and I have only tested with windows so I'll need to fix this up.

tgarc:
bah. this has turned out to be more work than anticipated. seems like for some reason the last step (shutil.move(temp_wheel, output)) is failing on windows & macos. Feel free to chime in if you have thoughts. @joerick is this what you envisioned? Does it seem simpler to you ?

YannickJadoul:
@tgarc Just to debug, could you maybe ```shell(['ls', temp_wheel_dir], env=env); shell(['ls', output_dir], env=env)``` before and after that `shutil.move` and maybe also print the values of the dir?
I'd do this myself, but since you have this PR, letting the CI run seems easier :-)

tgarc:
@YannickJadoul found the bug...I needed to add a step to create the output directory. Previously the output directory was getting automatically created by one of the pip commands

tgarc:
@joerick okay I've found the bugs that I had. The TL;DR is that I had to create extra temp directories for delocated wheels, and manually delete any temporary wheels for each build so they don't get picked up on subsequent builds. The complexity I think is about the same as with filter_wheels but I think the code 'looks' simpler. Would you give it a look over for sanity? Then I'll squash the commits and merge it.

YannickJadoul:
@tgarc Nice! My test build is also succeeding: https://travis-ci.org/YannickJadoul/Parselmouth/jobs/251766087

joerick:
Hey @tgarc - this looks good, I just did a few naming tweaks - could you pull from my revert_pr9 branch? Then as long as you're happy, I'm cool to merge!

tgarc:
thanks @joerick 
squashed and rebased. I'll wait until I'm a little more clear headed to give it a last once over before merging.

joerick:
🎉

joerick:
just released 0.4.0 with these changes. 👍

PR #19, opened 2017-08-04
WIP: Enabling a custom environment (cfr. #16)

YannickJadoul:
I've started approaching #16 from another perspective. This PR is not complete yet, but here to open a specific discussion and to get feedback (including feedback from Travis CI and AppVeyor)

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -72,6 +72,7 @@ def main():
     project_dir = args.project_dir
     before_build = get_option_from_environment('CIBW_BEFORE_BUILD', platform=platform)
     skip_config = os.environ.get('CIBW_SKIP', '')
+    environment = os.environ.get('CIBW_ENVIRONMENT', '')
 
     skip = BuildSkipper(skip_config)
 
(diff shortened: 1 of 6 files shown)


YannickJadoul:
(Oh, yeah, Windows and OS X not working makes sense, because I've only implemented the Linux variant, so far)

joerick:
Hey @YannickJadoul. I also did a bit of work on this a couple weeks ago. Have a look at #19, I'd be interested in your comments.

I'm curious how does your solution deal with whitespace in the values and env var expansion?

YannickJadoul:
Hi @joerick, I just had a quick look, and it seems as if your version is a few steps further than mine in solving potential problems.

For example, I am just taking a string with "VARIABLE=value OTHER_VARIABLE=value" syntax, and not splitting it, but directly passing it on to BASH. I however realized a few days ago that this will give inconsistencies between the Windows/Linux syntax, etc, so I really like your `shlex`-solution.

I've got a few minor considerations I've come across while thinking of my implementation. Do you prefer me to add those here, or in your #21?

joerick:
Over on #21 would be great 🙂

YannickJadoul:
Perfect. I'll close this one (for now), give that your proposal seems better :-)

Issue #15, opened 2017-07-03
Error message for "Configuring Windows SDK "v7.0" for Python 2 on a 64 bit architecture"


YannickJadoul:
After printing that `Configuring Windows SDK "v7.0" for Python 2 on a 64 bit architecture`, there are two red errors in my AppVeyor build:

```
The system cannot find the path specified.
The system cannot find the path specified.
```

(cfr. https://ci.appveyor.com/project/YannickJadoul/parselmouth/build/1.0.13/job/4cofnsd0y05jgtb9)

Having had a quick look, it seems as if that's due to the lines
```
"%WIN_SDK_ROOT%\%WINDOWS_SDK_VERSION%\Setup\WindowsSdkVer.exe" -q -version:%WINDOWS_SDK_VERSION%
"%WIN_SDK_ROOT%\%WINDOWS_SDK_VERSION%\Bin\SetEnv.cmd" /x64 /release
```
during the execution of https://raw.githubusercontent.com/ogrisel/python-appveyor-demo/09a1c8672e5015a74d8f69d07add6ee803c176ec/appveyor/run_with_env.cmd

This is not urgent, since my build still continues and finishes building everything, I think. But it might be nice to get rid of it (if possible) since the red color dóes make it seems as if there's a problem.

Besides, the `python --version` call is also highlighted in red, but I'm guessing that's because Python is printing to the stderr. I'll have a look if this is easily resolvable and submit a PR, if I manage?

YannickJadoul:
(Hold your horses on that `python --version`, because my latest build is not marking that in red, somehow: https://ci.appveyor.com/project/YannickJadoul/parselmouth/build/job/t0rcu72ygb4weol3)

joerick:
I don't know what the deal is with Appveyor logs, but they seem to print in the wrong order. Maybe it's related to IO buffering in Python (maybe worth trying an envvar `PYTHONUNBUFFERED=1` to see if that fixes the ordering).

Not sure if these messages are benign or not. Let us know what you find!

YannickJadoul:
I've got an update for you :-)
(https://ci.appveyor.com/project/YannickJadoul/parselmouth/build/1.0.26/job/pyn2gvvuv9ofue48)

[...]

PR #24, opened 2017-08-14
Unbuffered stdout for Windows and displaying the output of subprocess calls

YannickJadoul:
Cfr. side issue in #15:
- Forcing Python on Windows to not buffer its stdout, such that the output of cibuildwheel and that of its subprocess calls is nicely interleaved.
- Changing `subprocess.check_output` to `check_call` such that the stdout of these calls is not swallowed into a never-printed string by cibuildwheel.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,5 @@
 build_script:
   - pip install .
-  - python ./run_tests.py
+  # the '-u' flag is required so the output is in the correct order. 
+  # See https://github.com/joerick/cibuildwheel/pull/24 for more info.
+  - python -u ./run_tests.py
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -136,6 +136,7 @@ def print_preamble(platform, build_options):
         print('  %s: %r' % (option, value))
 
     print('\nHere we go!\n')
+    sys.stdout.flush()
 
 if __name__ == '__main__':
     main()
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -136,7 +136,6 @@ def print_preamble(platform, build_options):
         print('  %s: %r' % (option, value))
 
     print('\nHere we go!\n')
-    sys.stdout.flush()
 
 if __name__ == '__main__':
     main()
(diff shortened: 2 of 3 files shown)


YannickJadoul:
@joerick There are several other workarounds, if you don't like the `sys.stdout = os.fdopen(...)` trick.
- Current solution, reopening `stdout`: https://stackoverflow.com/questions/881696/unbuffered-stdout-in-python-as-in-python-u-from-within-the-program
- Writing a simple wrapper around a stream: https://stackoverflow.com/questions/107705/disable-output-buffering
- Making a custom print function: https://stackoverflow.com/questions/230751/how-to-flush-output-of-python-print (this answer: https://stackoverflow.com/a/35467658)

YannickJadoul:
Seems to be working, if you ask me. Two more little things:

- The lines of the `run_tests.py` (https://ci.appveyor.com/project/joerick/cibuildwheel/build/1.0.91#L1643) are still buffered and only output at the very end, because this solution is only applied to `cibuildwheel` itself. What to do with that? In principle, that is not part of the cibuildwheel library, so we can just as well change `appveyor.yml` to call `python -u` or `set PYTHONUNBUFFERED=1`. But is there then still a point to this PR?

- I've applied the solution only for the `windows.py` file (since OS X and Linux don't seem to have any problems), and applied a single `flush()` at the end of the preamble. The other option is either applying the fix when Windows is detected, or just applying it for all OSes.

joerick:
Excellent PR @YannickJadoul! Thank you for the detailed write-up 👍 

> The lines of the run_tests.py (https://ci.appveyor.com/project/joerick/cibuildwheel/build/1.0.91#L1643) are still buffered and only output at the very end, because this solution is only applied to cibuildwheel itself. What to do with that? In principle, that is not part of the cibuildwheel library, so we can just as well change appveyor.yml to call python -u or set PYTHONUNBUFFERED=1. 

Lets use `python -u` in appveyor.yml for that.

> But is there then still a point to this PR?

Yes! for all the users don't have to worry about that themselves :)

joerick:
Cool, I've made those changes. I'm good to merge and do a release once the lights go green.

YannickJadoul:
Cool, seems to be working! And as far as I can see, the output is 1) there and 2) in the right order :-)

Thanks for the quick reaction and solution!

joerick:
Released as 0.4.1 🎉

PR #25, opened 2017-08-14
Add CI runs for Python3 as well as Python2

joerick:
Since the CI doesn't test cibuildwheel _running_ on Python 3, a bug got through. Here I'm working on CI running the tests with both 2 and 3.

Note that I'm talking about the python that cibuildwheel runs on, not the Python that cibuildwheel is making wheels for!


YannickJadoul:
What about adding `language: python` to the top of the `.travis.yml` file? Python might come installed with pip and such already?

For the OS X build, you probably just need to run `brew install python3`

Issue #23, opened 2017-08-14
The output of commands executed by shell() under Windows is hidden


lelit:
Is there a reason why the `shell()` method in `windows.py` uses `check_output()` instead of `check_call()` to execute subcommands? It was changed in commit f26da27db8cbe565860d432eebfea430f40a7dfb (actually PR #18), but it's not clear if it was fixing a problem or what...

I found this while investigating on one of the problems I annotated in issue #22: [changing it back](https://github.com/lelit/cibuildwheel/commit/2f7ac230f86888d82cd8a4cb960daa2d0c2b2153) to `check_call()` _magically_ solved the fact that the `pytest` call wasn't apparently executed, see https://ci.appveyor.com/project/lelit/python-rapidjson/build/1.0.29.

If @tgarc could check and confirm it wasn't intentional, I will open a PR.

joerick:
Ah, nice catch! I think at one point @tgarc was using the output of the `shell()` call, but that was eventually removed. Yes, a PR to resolve this would be excellent!

joerick:
Looks like @yannickjadoul beat me to the punch! This fix is included in his PR #24

YannickJadoul:
Oops, yeah, didn't see this issue. But I was fixing up output on Windows, already, so I though that was rather on topic in #24.

lelit:
Sigh, trying with latest 0.4.1 I get [ValueError: can't have unbuffered text I/O](https://ci.appveyor.com/project/lelit/python-rapidjson/build/1.0.32#L40): should I change the way cibuildwheel is executed, using `python -u`?

YannickJadoul:
Are you using Python 3?
I can find this thread on SO: https://stackoverflow.com/questions/27067713/why-text-i-o-must-be-buffered-in-python-3

@joerick So maybe one of the solutions that explicitly calls `flush()` would be a better idea then (see #24). I hadn't thought of trying it with Python 3 since 2 is still the default on AppVeyor, but not supporting 3 seems not very future proof...

joerick:
Ah, looks like we need a python3 fix. Also I need to add that to our CI so that it's tested. In the meantime @lelit you could change to running cibuildwheel with Python 2 - it makes no difference to the wheels created.

YannickJadoul:
@lelit By the way, where does it give that error? Is that on the '-u' invocation or within cibuildwheel?

lelit:
[...]

PR #26, opened 2017-08-17
Fixing Unbuffered stdout for Python 3

YannickJadoul:
As discussed in issue #23, adding onto PR #24

@lelit Since AppVeyor is not yet building with Python 3 (open pull request #25), would it be easy for you to have a quick test if this branch works for you. (I'm assuming it should, because everything that's happening is a `flush()` call on an output stream.)


YannickJadoul:
Correction: I thought #25 was not working, but that was because of the unbuffered output error. So I have rebased my commit onto the commits from #25, which is why they now show up in this PR.

I guess that if this PR builds on Travis and AppVeyor, both PRs are working, and #25 could be merged first before merging this one.

lelit:
Yes, I did a quick test and it seems to work as expected, both on Travis CI and on Appveyor.

BTW, looking at the check failures above, I wonder why the Travis CI build starts a "Ruby" instance and the proceeds to install the Python interpreter...

YannickJadoul:
@lelit Thanks! Good to know it seems to be working.

To the Ruby remark: is Python being installed? I would expect that Python is just already installed on their base image, but python3's pip, etc isn't? I made a suggestion in #25 to add `language: Python` to the Travis CI config, though.

lelit:
I'm not sure to understand the questions, but AFAICT Travis supports Py3.x: 3.4 and 3.5 are already on board, while 3.6 is installed on demand.

YannickJadoul:
But if I am reading the log correctly, `pip3` isn't installed: `/usr/bin/python3: No module named pip`
As far as I know, Travis CI has multiple images, depending on the language. So if you then set that `language` key in `.travis.yml`, you probably get an exteded configuration.

lelit:
Ah, you meant in travis log of this PR... sorry: yes, when you specify the "right" language, you get a different image and/or initial setup, where `pip` is available.

lelit:
Was this PR actually merged? I'm still [hitting the problem](https://ci.appveyor.com/project/lelit/python-rapidjson/build/1.0.87), with latest 0.5.0

joerick:
This fix is in #25, so it’ll land when that gets merged. That still needs some CI fixes, I could cherry pick the Python 3 fixes out of there and onto the master branch, though 🤔

YannickJadoul:
Ooof, for a second I thought I had deleted this branch too early, when cleaning up my fork :-/
Or should this have been merged by itself? I can still restore the branch, apparently?

joerick:
@lelit These Python 3 fixes have now been released as 0.5.1. We've also now got CI running on Python 3 too, so we're unlikely to get more problems like this :)

joerick:
@YannickJadoul don't worry about keeping the branch here, these commits were merged into master with my PR #25

lelit:
Yeah, thank you!

PR #31, opened 2017-08-31
use os.makedirs to allow nested output directory

beniwohli:
As the title says. Reasoning for this it to e.g. have an output directory of `wheelhouse/${TRAVIS_TAG}`.

Not sure how to test this, any pointers are welcome :)


joerick:
I don't think a test is necessary. Cheers @beniwohli!

PR #33, opened 2017-09-07
Adding an overview of available options

YannickJadoul:
Added a short overview of the options cibuildwheel offers. (As proposed in #21.)

I wondered where to put it, at the start of the Options section or at the end of the Usage one. In the end I went for the latter, since after reading the 'quick start example', potential users might be wondering what is available to adapt cibuildwheel to their needs. This table then would them an idea of the kind of customizations that are provided, before actually diving into the more detailed documentation per configuration.


joerick:
Nice!

PR #34, opened 2017-09-11
Faster builds on Mac

joerick:
I noticed that cibuildwheel installs downloads Python and installs it even when it's already installed on Mac. That adds a lot of time to our CI, where cibuildwheel in run multiple times.


YannickJadoul:
Are we then sure to have the latest Python version, still? (i.e., 2.7.13, and not just 2.7.something)

joerick:
I think we'll still get 2.7.13, since the Travis VM doesn't preinstall any Python.org packages. But it's not a problem if not, because there's ABI compatibility between patch releases 😎

PR #35, opened 2017-09-12
Adding $PYBIN to path when calling pip wheel

YannickJadoul:
I found that when installing scripts through pip (such as `pip install cmake`, for example), these scripts are not in the `PATH` when `pip wheel` is called to build the wheel.

Since OS X and Windows do have this correctly (the path just being part of the `env` dictionary), this will also increase consistency, I think

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -62,11 +62,11 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
                 mkdir /tmp/delocated_wheel
 
                 if [ ! -z {before_build} ]; then
-                    PATH=$PYBIN:$PATH sh -c {before_build}
+                    PATH="$PYBIN:$PATH" sh -c {before_build}
                 fi
 
                 # Build that wheel
-                "$PYBIN/pip" wheel . -w /tmp/built_wheel --no-deps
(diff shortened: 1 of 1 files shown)


YannickJadoul:
(cfr. my failing build, not finding the `cmake` executable when installed through `pip`: https://travis-ci.org/YannickJadoul/Parselmouth/jobs/274290903)

YannickJadoul:
For the record, this small change indeed fixes the build: https://travis-ci.org/YannickJadoul/Parselmouth/builds/274720456

But of course, there would be other ways of achieving that, too

YannickJadoul:
Can't think of anything else to add to this. I don't think there are other lines in `linux.py` where this PATH should be added?

Oh, one thing: `$PYBIN` is added to `PATH` before `CIBW_ENVIRONMENT` in `macos.py`, while here it is now the opposite case. I don't know how much you value consistency, but otherwise we could still change that?

If not, go ahead and merge, as far as I'm concerned.

joerick:
> Oh, one thing: $PYBIN is added to PATH before CIBW_ENVIRONMENT in macos.py, while here it is now the opposite case. I don't know how much you value consistency, but otherwise we could still change that?

I agree this is a slight problem, it would be better to allow those CIBW_ENVIRONMENT calls to execute for every PYBIN, before the BEFORE_BUILD. But that's a separate problem, so I've merged this.

PR #36, opened 2017-09-13
Adding extra mount / to /host on docker invocation for linux wheels

YannickJadoul:
Under the assumption that https://travis-ci.org/YannickJadoul/Parselmouth/jobs/275233309 succeeds, a sillily small PR, mounting the root of the Travis CI/host file system as `/host` in the docker image for Linux builds. This allows to access directories outside of the project folder to be accessed inside of docker (e.g. the `ccache` dir that is saved/restored, in my specific use case), without big hacks (such as hard-linking that directory inside of the project root folder, or who knows what other tricks there would be).

If accepted, I still need to add some documentation in the README, perhaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ All being well, you should get wheels delivered to you in a few minutes.
 
 A more detailed description of the options, the allowed values, and some examples can be found in the [Options](#options) section.
 
+### Linux builds on Docker
+
+Linux wheels are built in the [`manylinux1` docker images](https://github.com/pypa/manylinux) to provide binary compatible wheels on linux, according to [PEP 513](https://www.python.org/dev/peps/pep-0513/). Because of this, when building with `cibuildwheel` on Linux, a few things should be taken into account:
+- Progams and libraries cannot be installed on the Travis CI Ubuntu host with `apt-get`, but can be installed inside of the Docker image using `yum` or manually. The same goes for environment variables that are potentially needed to customize the wheel building. `cibuildwheel` supports this by providing the `CIBW_ENVIRONMENT` and `CIBW_BEFORE_BUILD` options to setup the build environment inside the running Docker image. See [below](#options) for details on these options.
+- The project directory is mounted in the running Docker instance as `/project`, the output directory for the wheels as `/output`. In general, this handled transparently by `cibuildwheel`. For a more finegrained level of control however, the root of the host file system is mounted as `/host`, allowing for example to access shared files, caches, etc. on the host file system.
+
 
 Options
 -------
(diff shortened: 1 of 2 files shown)


joerick:
An amazingly simple solution to lots of problems! Yes this is a good idea. If you have time could you add some docs to README and I'll review before merging 👍

YannickJadoul:
Voilà.

I could not find a suitable section, so I created a short one about what makes Linux builds different from Windows & OS X. But I'm of course happy to change stuff if you think this is too lengthy/not at the right spot in the README/...

README.md
```diff
@@ -84,6 +84,12 @@ All being well, you should get wheels delivered to you in a few minutes.
...
 
+### Linux builds on Docker
+
+Linux wheels are built in the [`manylinux1` docker images](https://github.com/pypa/manylinux) to provide binary compatible wheels on linux, according to [PEP 513](https://www.python.org/dev/peps/pep-0513/). Because of this, when building with `cibuildwheel` on Linux, a few things should be taken into account:
+- Progams and libraries cannot be installed on the Travis CI Ubuntu host with `apt-get`, but can be installed inside of the Docker image using `yum` or manually. The same goes for environment variables that are potentially needed to customize the wheel building. `cibuildwheel` supports this by providing the `CIBW_ENVIRONMENT` and `CIBW_BEFORE_BUILD` options to setup the build environment inside the running Docker image. See [below](#options) for details on these options.
+- The project directory is mounted in the running Docker instance as `/project`, the output directory for the wheels as `/output`. In general, this handled transparently by `cibuildwheel`. For a more finegrained level of control however, the root of the host file system is mounted as `/host`, allowing for example to access shared files, caches, etc. on the host file system.
```
hroncok: this **is** handled

README.md
```diff
@@ -84,6 +84,12 @@ All being well, you should get wheels delivered to you in a few minutes.
 
 A more detailed description of the options, the allowed values, and some examples can be found in the [Options](#options) section.
 
+### Linux builds on Docker
+
+Linux wheels are built in the [`manylinux1` docker images](https://github.com/pypa/manylinux) to provide binary compatible wheels on linux, according to [PEP 513](https://www.python.org/dev/peps/pep-0513/). Because of this, when building with `cibuildwheel` on Linux, a few things should be taken into account:
```
hroncok: **L**inux.

Also, i would mention that this is CentOS 5.

joerick:
Niiiiice! Thanks @YannickJadoul (and @hroncok for the review). I'll merge this now.

Issue #40, opened 2017-09-20
Python 3.6 on macOS have some SSL problems


hroncok:
When using `setup_requires`, the 3.6 version of Python on macOS will blow up.

See [actual Travis CI log](https://travis-ci.org/joerick/cibuildwheel-autopypi-example/jobs/277539350) from https://github.com/joerick/cibuildwheel-autopypi-example/pull/4

This can be workarounded by manually pip installing the requirement in `CIBW_BEFORE_BUILD`. But given it only blows up with 3.6, I think it's a problem with Python installation.

joerick:
might be solved with `pip install certifi`?

joerick:
Any update on this @hroncok? Did you get it working using another method?

hroncok:
Sorry, I have it on my TODO, but new semester  have just started here and
everything is a bit crazy. I can get back to this in couple of weeks.

Dne 9. 10. 2017 20:29 napsal uživatel "Joe Rickerby" <
notifications@github.com>:

tgarc:
I've seen this issue too. To the best of my recollection, it was an issue with the fact that any packages from `setup_requires` are installed in some hackish way that doesn't always work nicely. I ran into a similar issue where installing cffi with setup_requires raised an error because it couldn't find the installed ffi library; again I had to use the pip install workaround.

I would defer this issue to the pip guys...this definitely isn't an issue particular to cibuildwheel.

hroncok:
A note: `setup_requires` don't use wheels, pip installing the dependencies before is a good optimization (especially when requiring stuff like Cython or Numpy). How this works inside is a mystery for me. Maybe the root of this issue is in setuptools.

joerick:
Thanks for the report @hroncok. The more I learn about `setup_requires`, the more I'm sure it shouldn't be used. Packages installed though it are installed in some arcane setuptools way that isn't well supported in a post-`pip` world.

In summary - don't use `setup_requires`. If you must use it, preinstall those dependencies using `CIBW_BEFORE_BUILD="pip install x y z"`.

hroncok:
That's basically what I have learned as well.

PR #41, opened 2017-09-20
Update macOS Python 3.6 to 3.6.2

hroncok:
This might possibly fix https://github.com/joerick/cibuildwheel/issues/40


hroncok:
We'll see in https://travis-ci.org/hroncok/cibuildwheel-autopypi-example/jobs/277760433

hroncok:
Doesn't help

PR #43, opened 2017-09-30
add bashlex to install_requires

tgarc:
bashlex was missing from the setup.py dependencies (which causes pip install to fail). Seems like testing should've caught this. I'm confused how this didn't come up in the travis tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,4 +34,7 @@ setup(
             'cibuildwheel = cibuildwheel.__main__:main',
         ],
     },
+    install_requires=[
+        'bashlex',
+    ],
 )


hroncok:
Actually, it is there: https://github.com/tgarc/cibuildwheel/blob/242f0019bdce7e60b2ec700e3023d2b6a0215442/setup.py#L12

YannickJadoul:
Yes, indeed: it already is there. And Travis also seems to agree on that, spitting out `SyntaxError: keyword argument repeated`.

What exactly was the immediate reason for this PR?

tgarc:
yeah. apparently I'm losing my mind.

I think I did a pip install --upgrade --no-deps cibuildwheel which caused pip to ignore the bashlex dependency.

Thanks

PR #47, opened 2017-12-20
Support for CIBW_SKIP platform specific variants

lelit:
To workaround a still mysterious problem building Py3.6 binary wheel on Travis CI for a C++ project I'm maintaining, I need a way to say "don't build Py36 wheel for GNU/Linux, but do that on other systems".

This adds `CIBW_SKIP_<platform>` variants, as done for other variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,9 @@ Example: `cp27-macosx_10_6_intel`  (don't build on Python 2 on Mac)
 Example: `cp27-win*`  (don't build on Python 2.7 on Windows)  
 Example: `cp34-* cp35-*`  (don't build on Python 3.4 or Python 3.5)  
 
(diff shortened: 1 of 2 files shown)


YannickJadoul:
I'm not per se against doing this, if only for consistency reasons, but what's wrong with `CIBW_SKIP=cp36-manylinux1*`, which includes both Python 3.6 as well as the platform?

lelit:
Doh, right, I completely missed that way of doing!

YannickJadoul:
One can argue that the `manylinux1` part is not the most intuitive, though, so maybe we could still integrate this PR. Any thoughts, @joerick?

lelit:
Ok, thank you.

Issue #48, opened 2017-12-25
pip: command not found


zfrenchee:
Hi, 

I just pushed the first commit in a long time to my project, and cibw ran into a couple issues. The first was, on osx: 

```
$ pip install cibuildwheel==0.6.0
/Users/travis/.travis/job_stages: line 57: pip: command not found
```

This might be related to this issue: https://github.com/travis-ci/travis-ci/issues/8829 and perhaps fixed with this comment: https://github.com/travis-ci/travis-ci/issues/8829#issuecomment-348322979

YannickJadoul:
Well, since this happens when you're installing `cibuildwheel`, it's indeed more of a problem with the Travis environment than with `cibuildwheel`.

That being said, I recently also had to fix my builds, and ended up with `python -m pip` or `python3 -m pip` (but according to that issue you mention, `pip2` or `pip3` should also do the job?)

zfrenchee:
@YannickJadoul Since cibw is so dependent on travis, an issue with travis is an issue with cibw, imo. Also, at the very least, the docs / readme will need updating in accordance with this change. 

I believe changing to pip2/pip3 will fix this issue, but I don't want to change the command to pip3 for the python2.7 wheel builds -- that just seems wrong. What would be ideal is `{pip} install cibw` but I'm not certain that will work. @joerick what do you think?

YannickJadoul:
Then main thing is that `cibuildwheel` can't really do anything about this, since the command you enter is the one that installs `cibuildwheel` (i.e., if `cibuildwheel` isn't installed, there's not a lot of code in this repository that can help to fix the problem, besides changing the example).

- `{pip}` is never going to work, because that's something `cibuildwheel` replaces once it's running, but we can't change that before it's installed.
- Running `pip3` for Python 2.7 builds is actually no problem either, since this will just be the outer Python interpreter. `cibuildwheel` will call the right version of Python interpreters while it's running.

YannickJadoul:
I have just created a PR (#49) to fix the 'Minimal example' in the README, but I'm quite afraid that's about the only thing we can do about this issue

zfrenchee:
@YannickJadoul changing `pip` to `pip3` did not work. 
```
The command "pip3 install cibuildwheel==0.6.0" exited with 127.
```
I'm now trying what you suggest, `python -m pip install...`

And no worries! That PR is what I was hoping for. Just want to make sure the provided example actually works.

zfrenchee:
[...]

PR #49, opened 2017-12-26
Fixing 'Minimal setup' in README to have `python -m pip` instead of `pip`

YannickJadoul:
Cfr. #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Usage
         - os: osx
 
     script:
-      - pip install cibuildwheel==0.6.0
+      - python -m pip install cibuildwheel==0.6.0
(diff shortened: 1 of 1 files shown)


kavvkon:
The above does not really work. I manage to make it work on both linux and osx with the following:
```
matrix:
  include:
    - sudo: required
      language: python
      services:
        - docker
    - os: osx
      language: generic
      before_install:
        - brew update
        - brew install python3
        - virtualenv env -p python3
        - source env/bin/activate

script:
  - pip install cibuildwheel==0.6.0
  - cibuildwheel --output-dir wheelhouse

```

joerick:
Now using `pip2` and `language: generic` on mac. see d16d81e

PR #57, opened 2018-03-31
macOS update

mayeut:
This PR does the following things:
- fix python3 test on macOS:
   brew python formula has been updated for pyhton3, c.f. https://travis-ci.org/mayeut/cibuildwheel/builds/360577716 macOS python3 errored build
- Update python versions used to build wheels (latest binaries)
- Use latest pip:
   With the new PyPI stuff on-going, pip is failing with required TLS 1.2, c.f. https://travis-ci.org/mayeut/cibuildwheel/builds/360577716 macOS python2 failing build
   The job has been started in the brownout window enforcing TLS 1.2, c.f. pypa/warehouse#3411
   pip 9.0.2 has been updated to support TLS 1.2 on macOS (where openSSL is too old) in pypa/pip#4454

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
         - "PYTHON=python3"
       before_install:
         - brew update
-        - brew install python3
+        - brew outdated python || brew upgrade python
 
 script:
   - |
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -18,6 +18,8 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.5/python-3.6.5-macosx10.6.pkg'),
     ]
+    get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
+    get_pip_script = '/tmp/get-pip.py'
 
     pkgs_output = subprocess.check_output(['pkgutil',  '--pkgs'])
     if sys.version_info[0] >= 3:
@@ -34,6 +36,9 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         return subprocess.check_call(args, env=env, cwd=cwd, shell=shell)
 
     abs_project_dir = os.path.abspath(project_dir)
+
+    # get latest pip once and for all
+    call(['curl', '-L', '-o', get_pip_script, get_pip_url])
 
     for config in python_configurations:
         if skip(config.identifier):
@@ -63,7 +68,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         call([python, '--version'], env=env)
 
         # install pip & wheel
-        call([python, '-m', 'ensurepip', '--upgrade'], env=env)
+        call([python, get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         call([pip, '--version'], env=env)
         call([pip, 'install', 'wheel'], env=env)
         call([pip, 'install', 'delocate'], env=env)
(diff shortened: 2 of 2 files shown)


YannickJadoul:
Thanks for this pull request! Looks pretty good to me, but I've got two minor questions:

- Can you explain why you are downloading and calling the `get-pip.py` script? Don't newly installed python versions already have a `pip`? So we could just `pip install -U pip`? (As suggested by pip itself in your failed build: https://travis-ci.org/mayeut/cibuildwheel/jobs/360577719#L148). Or is there a problem with that that I'm missing?
- Closely linked to this: would you mind having a look at the [minimal example in the `README.md` file](https://github.com/joerick/cibuildwheel/blob/master/README.md#minimal-setup). Brew on OS X is also being used there (I believe?), so we should make sure it still works. (Yeah, we should probably turn that into an actual test, but ... faster/easier said than done.)

Apart from that, I'm happy to merge this if @joerick and @tgarc don't have any remarks? I guess given the severity of the problem, it might be nice to have a patch/release for this reasonably soon?

mayeut:
@YannickJadoul,

> Can you explain why you are downloading and calling the get-pip.py script? Don't newly installed python versions already have a pip? So we could just pip install -U pip?

`pip` will fail to upgrade itself for the same reason. It lacks TLS 1.2 support. Using the `get-pip.py` script is the only way to install a working `pip`.

> Closely linked to what you did in this PR: would you mind having a look at the minimal example in the README.md file. Brew on OS X is also being used there (I believe?), so we should make sure it still works.

Though the minimal example does not exactly install `cibuildwheel` in the same way as in the `.travis.yml` file, it's using python2 from Brew which is built against a Brew OpenSSL library that supports TLS 1.2.  This differs from the official python packages used by `cibuildwheel` which are linked against the outdated macOS OpenSSL library.

YannickJadoul:
@mayeut Aaargh, that's quite a silly situation. Either way; then this seems to be the best solution indeed. Thanks for looking into it!

@joerick, should we then merge this soon?

cibuildwheel/macos.py
```diff
@@ -63,7 +68,7 @@ def call(args, env=None, cwd=None, shell=False):
         call([python, '--version'], env=env)
 
         # install pip & wheel
-        call([python, '-m', 'ensurepip', '--upgrade'], env=env)
+        call([python, get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
```
joerick: any particular reason for `--no-setuptools` and `--no-wheel`?

.travis.yml
```diff
@@ -32,7 +32,7 @@ matrix:
         - "PYTHON=python3"
       before_install:
         - brew update
-        - brew install python3
+        - brew outdated python || brew upgrade python
```
joerick: `brew outdated` returns with success if python isn't installed. Perhaps this line should be `brew install python || brew upgrade python`?

mayeut: python is installed on macOS travis-ci images.
Otherwise, I can change this by
```
brew install python || brew outdated python || brew upgrade python
```

mayeut: Sorry should be 
```
brew install python && ( brew outdated python || brew upgrade python )
```
as install returns success even if already installed.

cibuildwheel/macos.py
mayeut: I'd rather get the script only install pip and let the remaining as it was before.
wheel is installed shortly after.
This is also what travis-ci/dpl is doing for deployment so I guess I just did the same.

.travis.yml
mayeut: well, I just did a test and it seems the behavior of brew install is not really consistent.
brew install python exits with 0 on my laptop (python is already installed). It exits with error code on travis-ci.  
https://discourse.brew.sh/t/what-exit-code-is-brew-install-supposed-to-return-when-formula-is-already-installed/1607

joerick: Oh that's cool, I didn't know about brew bundle. But seems a little heavy-handed for this. Let's stick with your original for the moment.

cibuildwheel/macos.py
joerick: 👍

joerick:
Released as 0.7.1. thanks all!

PR #60, opened 2018-04-08
Removing replacement of {python} and {pip} in CIBW_BEFORE_BUILD

YannickJadoul:
Checking how the CI builds go wrong if we always use `python` and `pip` on OS X instead of specifying `2` or `3`. Cfr. #59.


YannickJadoul:
The tests on this seem to be passing, so I am assuming it is working, now. The one trick I had to pull were these two `os.symlink` lines in `macos.py`, since the Python 3 installations do not provide `python` and `pip`, but only `python3` and `pip3`.

A few more things to consider:
- I pushed the symlinking into the if-statement preventing the installation of Python to be repeated between different runs of `cibuildwheel`. Another option would be to independently check if there is already a `python` and `pip` executable in the installation folder (and symlink if these files are not there). Not sure which one is better, but this alternative has the advantage of not failing when the CI has already made this symlink before `cibuildwheel` is run...
- This PR or PR #59 are probably mutually exclusive. Though the one thing we could consider from #59 - if we choose to keep this one - is whether we also want to replace `{project}` in `CIBW_BEFORE_BUILD`?
- Merging and releasing this PR will be backwards incompatible with some existing `cibuildwheel` configurations, since `{python}` and `{pip}` are not exactly valid executable names.

joerick:
> This PR or PR #59 are probably mutually exclusive. Though the one thing we could consider from #59 - if we choose to keep this one - is whether we also want to replace {project} in CIBW_BEFORE_BUILD?

I'd say PR #59 is still valid, and this builds on that since we want to keep backward compatibility. Let me know if that makes sense, happy to help with this!

YannickJadoul:
> So a test for the active Python/pip version would be cool

In principle, we already have this because `python` and `pip` are being called directly from within `macos.py`. Consequently, if there is a mismatch, this results in an error during the installation/tests. But yeah, maybe an extra, explicit test would be a good idea.

> I think we should keep supporting the prepare_command substitutions until 1.0.0 (just replacing with python/pip, since that now works

Fair enough. Seems like a plan, I'll see if I can add/revert that, still. Is there an easy way to mark this deprecated, such that 1) users will know, looking at the output, and 2) we don't forget when releasing 1.0?

> On Mac, maybe it would be neater to do the python/pip symlinks in a different directory (on PATH before installation_bin_path), so we're not modifying the Python installation?

Yeah, I thought about that, but then I could not really see the problem with adding a two symlinks. Plus, I reckoned putting a new `bin/` dir somewhere would complicate things more than this semi-hack.


But I'll push this branch to the the main repo now, so you can see/test for yourself!

joerick:
Cool! I'll have a play with this now :)

YannickJadoul:
Is there a way to change this PR to the new branch? Otherwise I'd have to merge your changes into my branch before you can merge this PR :-D

joerick:
Not sure... can you do so using the [Edit] button at the top of the page?

YannickJadoul:
No, I can only update the title and the branch I want to merge into. But I probably gave you, as owner, the rights to make changes to my branch, so maybe it's better to work on that one?

joerick:
yeah, I think you're right, I can push to your branch. Okay let's go with that!

YannickJadoul:
> I'd say PR #59 is still valid, and this builds on that since we want to keep backward compatibility. Let me know if that makes sense, happy to help with this!

Yeah, sure, let's do that. The sooner PR #59 can be merged, the better, then, since we'll probably need to rebase this one.

joerick:
> In principle, we already have this because python and pip are being called directly from within macos.py. Consequently, if there is a mismatch, this results in an error during the installation/tests. 

I agree, I can't even see a way to test this to be honest!

joerick:
Cool, so I've merged in master with the previous PR applied, I made a couple stylistic changes and added deprecation warnings when `{python}` and `{pip}` are used.

Let me know what you think!

cibuildwheel/__main__.py
```diff
@@ -164,7 +164,27 @@ def print_preamble(platform, build_options):
...
     print('\nHere we go!\n')
 
+def detect_warnings(platform, build_options):
+    warnings = []
+
+    # warn about deprecated {python} and {pip}
+    for option_name in ['test_command', 'before_build']:
+        option_value = build_options.get(option_name)
+
+        if option_value:
```
YannickJadoul: For the sake of being slightly shorter and non-nested: `build_options.get(option_name, '')` and remove this if statement?

cibuildwheel/linux.py
```diff
@@ -103,10 +103,10 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             pybin_paths=' '.join(c.path+'/bin' for c in platform_configs),
             test_requires=' '.join(test_requires),
             test_command=shlex_quote(
-                prepare_command(test_command, python='python', pip='pip', project='/project') if test_command else ''
+                prepare_command(test_command, project='/project') if test_command else ''
```
YannickJadoul: Maybe it's possible to squeeze on a single line now, if you care about such things ;-) If not, eh...

cibuildwheel/macos.py
```diff
@@ -47,31 +47,41 @@ def call(args, env=None, cwd=None, shell=False):
 
         # if this version of python isn't installed, get it from python.org and install
         python_package_identifier = 'org.python.Python.PythonFramework-%s' % config.version
+        installation_bin_path = '/Library/Frameworks/Python.framework/Versions/{}/bin'.format(config.version)
```
YannickJadoul: This can now be moved down a little bit (as I'd first done, before moving the symlinks into the if statement of the installation), to just before the `if config.version[0] == '3'`. It feels slightly more relevant and in place there.

cibuildwheel/macos.py
```diff
@@ -47,31 +47,41 @@ def call(args, env=None, cwd=None, shell=False):
...
         if python_package_identifier not in installed_system_packages:
             # download the pkg
             call(['curl', '-L', '-o', '/tmp/Python.pkg', config.url])
             # install
             call(['sudo', 'installer', '-pkg', '/tmp/Python.pkg', '-target', '/'])
 
+        # Python bin folders on Mac don't symlink python3 to python, so we do that
+        # so `python` and `pip` always point to the active configuration.
+        if os.path.exists('/tmp/cibw_bin'):
+            shutil.rmtree('/tmp/cibw_bin')
```
YannickJadoul: I know it's just symlinking, but this feels slightly like a waste, removing and creating this folder over and over again?

cibuildwheel/macos.py
```diff
@@ -47,31 +47,41 @@ def call(args, env=None, cwd=None, shell=False):
...
+        os.makedirs('/tmp/cibw_bin')
+
+        if config.version[0] == '3':
```
YannickJadoul: If we do this in our own directory anyway, we could even just do this for Python 2 as well (i.e., `'python' + config.version[0]`). But I've no clue if that's better or worse, and more robust against further changes or not. So more of a thought/consideration than of a review remark.

cibuildwheel/util.py
```diff
@@ -1,16 +1,15 @@
...
+def prepare_command(command, project):
     '''
-    Preprocesses a command by expanding variables like {python} or {pip}.
+    Preprocesses a command by expanding variables like {project}.
 
-    For example, used for the before_build option, where the user would
-    like to run a command like `python setup.py test`. If the command should run on
-    Python 3, the user could write `{python} setup.py test`. This command would expand
-    it out to python2 or python3 as appropriate.
+    For example, used in the test_command option, to specify the path to the
```
YannickJadoul: Oops, I'd forgotten that :-D Well spotted!

cibuildwheel/util.py
```diff
@@ -1,16 +1,15 @@
...
+    tests directory.
     '''
-    return command.format(python=python, pip=pip, project=project)
+    return command.format(python='python', pip='pip', project=project)
```
YannickJadoul: Niiice :-)

cibuildwheel/macos.py
```diff
@@ -47,31 +47,41 @@ def call(args, env=None, cwd=None, shell=False):
...
+        installation_bin_path = '/Library/Frameworks/Python.framework/Versions/{}/bin'.format(config.version)
         if python_package_identifier not in installed_system_packages:
             # download the pkg
             call(['curl', '-L', '-o', '/tmp/Python.pkg', config.url])
             # install
             call(['sudo', 'installer', '-pkg', '/tmp/Python.pkg', '-target', '/'])
 
+        # Python bin folders on Mac don't symlink python3 to python, so we do that
+        # so `python` and `pip` always point to the active configuration.
+        if os.path.exists('/tmp/cibw_bin'):
```
YannickJadoul: Is it worth turning this little `if exists: rmtree; mkdirs` snippet into a function? (Is it used on other operating systems as well?)

joerick: It's used on windows as well, but I don't mind repeating it, since it's clear and concise enough I think

cibuildwheel/macos.py
joerick: computers are fast :)

cibuildwheel/__main__.py
joerick: good idea

cibuildwheel/macos.py
YannickJadoul: Heh, fair enough. Not really concerned with that, but more with the fact that this might suggest for a better alternative :-) But fair enough, I can't immediately come up with one...

cibuildwheel/__main__.py
YannickJadoul: Aaaaaah, dang. Well, then the only thing to stop nesting is to combine the if statements, but ... bleh, not worth a commit :-(

PR #61, opened 2018-04-08
Adding an to make the `pip wheel` invocating more or less verbose

YannickJadoul:
Recently ran into the problem where I wanted to see the actual compilation output from within `pip wheel`. Instead of throwing away that temporary commit, I thought to extend it, turn it into a full and documented option, and submit a PR, in case it would be considered useful. First version, so feel free to clean up stuff or tell me what to clean up. Maybe there's also other places where we want to influence the verbosity of the executed commands? Also, I couldn't decide on `VERBOSE_BUILD` vs. `BUILD_VERBOSITY`, and on whether the negative values (resulting in `-q`s instead of `-v`) would be useful to include or not really.


YannickJadoul:
Next to this, I cannot really think of a quick-and-easy test for this since we're not intercepting the output of the commands (except for adding the options in a test and see that `cibuildwheel` at least does't crash on it)

YannickJadoul:
Not completely happy with how it looks, but this won't break anything, so I guess it can be merged quite easily.
I've still cleaned up a minor thing, but if that works, I'll happily merge it :-)

PR #62, opened 2018-04-20
windows: upgrade pip with 'python -m pip'

anthrotype:
this fixes an "ImportError: cannot import name main" when self-upgrading pip to latest version 10: https://github.com/pypa/pip/issues/5240#issuecomment-382989420

Before applying this patch, I was getting this error:
https://ci.appveyor.com/project/trufont/uharfbuzz/build/1.0.4#L91

After applying this patch (and pip installing from my cibuildwheel fork), the pip self-upgrade works and the build succeeds:
https://ci.appveyor.com/project/trufont/uharfbuzz/build/1.0.7

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -72,7 +72,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
 
         # prepare the Python environment
-        shell(['pip', 'install', '--disable-pip-version-check', '--user', '--upgrade', 'pip'],
+        shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'],
               env=env)
         shell(['pip', 'install', 'wheel'], env=env)
 


YannickJadoul:
I am slightly annoyed that this kind of upgrade is somehow failing, but ... yeah, if this small change is what it takes to get it working, let's merge it?

@joerick Could you have a look and run the Bitrise CI again? It says "aborted" here, but I don't have access to check out the details. Then again, since this is a Windows-only change, maybe we can just merge without paying attention to Bitrise.

lelit:
Thank you Cosimo!

joerick:
Hello guys, sorry I'm suuuper busy with work at the moment so I don't have time to investigate this further. Happy to merge now as this is an immediate problem.

Questions (for post-merge?) are-

- Should we use `python -m pip` for all subsequent invocations of `pip`?
- Perhaps using the `get-pip.py` script (as we've started doing on macOS) would be a more robust way to get pip up-to-date?

p.s. don't worry about the bitrise build - not sure why it's aborted but it's really so I can get test-runs on mac when the Travis Mac build queue is backed up. Green Travis + Green Appveyor = 👍

anthrotype:
> Should we use python -m pip for all subsequent invocations of pip?

not necessarily. The only reason we do `python -m pip install --upgrade pip` instead of `pip install --upgrade pip` is that on windows the pip executable cannot remove itself while it's running (the file is locked by the OS).

> Perhaps using the get-pip.py script would be more robust

I don't think so. That's for bootstrapping in a python environment which doens't have already a pip installed, or doesn't have ensurepip.

joerick:
Alright, cool let's go with this! Thanks all

joerick:
Released as 0.7.2

PR #66, opened 2018-05-02
Give ownership of the wheel files to the current user instead of root on Linux

mayeut:
Give ownership of the wheel files to the current user instead of root on Linux, this also allows to cleanup the `.travis.yml` file a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,42 +5,28 @@ matrix:
     # Linux Python 2
     - sudo: required
       language: python
-      services:
-        - docker
-      env:
-        - "PYTHON=python2"
+      python: 2.7
+      services: docker
+      env: PYTHON=python
 
     # Linux Python 3
     - sudo: required
       language: python
-      services:
-        - docker
-      env:
-        - "PYTHON=python3"
-      before_install:
-        - sudo apt-get -qq update
-        - sudo apt-get install -y python3-pip
+      python: 3.4
+      services: docker
+      env: PYTHON=python
 
     # macOS Python 2
     - os: osx
-      env:
-        - "PYTHON=python2"
+      env: PYTHON=python2
 
     # macOS Python 3
     - os: osx
-      env:
-        - "PYTHON=python3"
+      env: PYTHON=python3
       before_install:
         - brew update
         - brew outdated python || brew upgrade python
 
-script:
-  - |
-    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      $PYTHON -m pip install -r requirements-dev.txt
-      $PYTHON ./bin/run_tests.py
-    else
-      # linux test requires root to clean up the wheelhouse (docker runs as root)
-      sudo $PYTHON -m pip install -r requirements-dev.txt
-      sudo $PYTHON ./bin/run_tests.py
-    fi
+install: $PYTHON -m pip install -r requirements-dev.txt
+
+script: $PYTHON ./bin/run_tests.py
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -95,6 +95,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
 
                 # we're all done here; move it to output
                 mv "$delocated_wheel" /output
+                chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"
             done
         '''.format(
             package_name=package_name,
@@ -108,6 +109,8 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             ),
             build_verbosity_flag=' '.join(get_build_verbosity_extra_flags(build_verbosity)),
             environment_exports='\n'.join(environment.as_shell_commands()),
+            uid=os.getuid(),
+            gid=os.getgid(),
         )
 
         docker_process = subprocess.Popen([


mayeut:
Failure on travis-ci is linked to the inability to pull the dockcross image in a timely manner: https://travis-ci.org/joerick/cibuildwheel/jobs/374170512#L2560
A restart shall correct this (travis-ci build is OK on my fork: https://travis-ci.org/mayeut/cibuildwheel/builds/374156382)

No idea for bitrise as I don't have access

YannickJadoul:
Nevermind bitrise (only @joerick has access there, but maybe it would be nice to temporarily disable?), but I've restarted the failed Travis build

YannickJadoul:
Furthermore:

> Python 3.3.x has reached end-of-life. This is its final release. It is a security-fix source-only release.
>
> Python 3.3.0 was released on 2012-09-29 and has been in security-fix-only mode since 2014-03-08. Per Python Development policy, all support for the 3.3 series of releases ended on 2017-09-29, five years after the initial release. This release, Python 3.3.7, was the final release for the 3.3 series.

 (cfr. https://www.python.org/downloads/release/python-337/)

So should we maybe just drop 3.3 support overall?

mayeut:
Dropping 3.3 support overall is to be considered yes. However, this PR only aims at dropping support on linux because builds are failing right now. Windows Python 3.3 builds are not broken (yet).

YannickJadoul:
Wait, because these are actually three PRs in one, if I see that correctly: Python 3.3 for manylinux, the wheel ownership, and simplifying the configuration for Travis?

I'm quite happy with the content, but is there a part that's urgent to merge? If not, I'll take out those changes. Otherwise, we can just wait for @joerick to review things as well.

.travis.yml
```diff
@@ -5,21 +5,20 @@ matrix:
     # Linux Python 2
     - sudo: required
       language: python
+      python: "2.7"
```
YannickJadoul: "2.7" -> 2.7?

.travis.yml
```diff
@@ -5,21 +5,20 @@ matrix:
...
       services:
         - docker
       env:
-        - "PYTHON=python2"
+        - "PYTHON=python"
 
     # Linux Python 3
     - sudo: required
       language: python
+      python: "3.4"
```
YannickJadoul: Why 3.4, and not the latest version? Or just 3, then? (All in all, the different minor versions of Python will be tested inside the `cibuildwheel` tests anyway)

.travis.yml
```diff
@@ -5,21 +5,20 @@ matrix:
...
       services:
         - docker
       env:
-        - "PYTHON=python3"
-      before_install:
-        - sudo apt-get -qq update
-        - sudo apt-get install -y python3-pip
+        - "PYTHON=python"
```
YannickJadoul: If distinction between `python2` and `python3` is not needed, we can just get rid of this environment variable

mayeut:
The only commit relevant to dropping python 3.3 on linux and reenable passing builds is e262877
You can cherry-pick only this commit (or I can separate commits in different PR).

.travis.yml
mayeut: Well,
I copy/pasted this from travis-ci doc https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions
If it has no impact I'm happy to change it

.travis.yml
mayeut: According to travis-ci docs, https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions, a major.minor shall be provided
This is the python version running cibuildwheel driver, not the python used to build wheels (so, cibuildwheel itself is not tested with every minor python versions). If support is dropped for 3.3 for building wheels, I found somewhat logical not to use it either to run cibuildwheel and thus, use the next supported version which is python 3.4

.travis.yml
mayeut: This is still needed for macOS which uses `python2` and `python3` (might not show in the diff)

mayeut:
@YannickJadoul, I opened #67 for the only urgent commit. This one now builds on top of the commit in #67 and is not that important

.travis.yml
YannickJadoul: Sorry, should've looked better!

YannickJadoul:
Thanks a lot! #67 is merged, so you could rebase this PR, if you prefer (and drop e6fa70a, I suppose)

mayeut:
@YannickJadoul, PR rebased.

cibuildwheel/linux.py
```diff
@@ -95,6 +95,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
 
                 # we're all done here; move it to output
                 mv "$delocated_wheel" /output
+                chown {uid}:{gid} /output/$(basename $delocated_wheel)
```
YannickJadoul: Since all variable expansions before are also quoted, can we still change this to `"/output/$(basename "$delocated_wheel")"` ?

YannickJadoul:
OK, given that #71 depends on this, and that this is a small change anyway, I'm happy to merge this (after that one change, if you agree, @mayeut).

mayeut:
@YannickJadoul, change requested has been committed

PR #67, opened 2018-05-03
Drop python 3.3 support on Linux

mayeut:
Python 3.3 support has been dropped in pypa/manylinux@ccb88109d9fe1076cd1d8e359761644cb3010eb5


YannickJadoul:
Perfect, thanks a lot!

Given the size of the change, let's merge this. And then @joerick can have a look and say about the non-urgent changes when he has time. Though we probably still need him to do the actual release...

Issue #63, opened 2018-04-24
OSX build fails (TLS version error)


TheBB:
Hi,

I tried following the instructions [here](https://github.com/joerick/cibuildwheel#minimal-setup), which work fine for Linux, but seem to fail on OSX. Example: https://travis-ci.org/sintefmath/Splipy/jobs/370590848

This seems related to https://github.com/joerick/cibuildwheel/pull/57 which claims to be fixed, so I am sure I'm just doing something wrong here.

.travis.yml file is

```yaml
language: python

matrix:
  include:
    - sudo: required
      services:
        - docker
      env:
        - PIP=pip
        - CIBW_SKIP="cp27-* cp33-*"
    - os: osx
      language: generic
      env:
        - PIP=pip2
        - CIBW_SKIP="cp27-* cp33-*"

script:
  - $PIP install cibuildwheel==0.7.1
  - cibuildwheel --output-dir wheelhouse
```

YannickJadoul:
I do not really see what's going wrong here. Could you maybe try running it again, to see if it wasn't a weird non-reproducible error? - Probably not, but fingers crossed.

Otherwise, I'll have a look at that minimal example, because I have the feeling that it could/should be made simpler anyway...

mayeut:
This seems to be a setuptools issue rather than pip issue which was corrected with #57 
c.f. pypa/setuptools#1320

You can try to install your dependencies using pip before building the wheel (using CIBW_BEFORE_BUILD environment variable)

YannickJadoul:
@TheBB Does this then mean the issue is fixed, given @mayeut's comment?

TheBB:
Sorry, I wasn't able to test it yet. I'll close this for now assuming the fix is good. Thanks everyone.

YannickJadoul:
Thanks! Feel free to re-open if necessary.

joerick:
btw @TheBB, `CIBW_BEFORE_BUILD={pip} install .` is quite common - setuptools installing the `setup_requires` dependencies is pretty unreliable these days.

mayeut:
setuptools is not updated by cibuildwheel and the version installed with python 3.4 on macOS is quite old... Shall cibuildwheel install an up-to-date version of setuptools ? (manylinux images are up-to-date, don't know about windows on AppVeyor)

YannickJadoul:
Hmmm, maybe we should do that. After all, we're also updating `pip` and installing `wheel`...

[...]

Issue #58, opened 2018-03-31
macOS travis-ci build fails


mayeut:
With enforced TLS 1.2 connection, pip fails to collect dependencies on macOS.

c.f. #57

joerick:
Fixed in #57

PR #69, opened 2018-05-03
macOS/Windows: Update setuptools

mayeut:
Update setuptools on macOS/Windows
Not required on linux since the manylinux image is updated frequently

It might help with the issue seen in #63

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -70,6 +70,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         # install pip & wheel
         call([python, get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         call([pip, '--version'], env=env)
+        call([pip, 'install', '--upgrade', 'setuptools'], env=env)
         call([pip, 'install', 'wheel'], env=env)
         call([pip, 'install', 'delocate'], env=env)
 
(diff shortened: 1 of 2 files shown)


mayeut:
travis-ci build failing on linux. c.f. #67

YannickJadoul:
#67 is merged, so builds should work after rebasing.

But which one should be merged? This one and #68 seem to be addressing the same issue, or did I fail to notice anything?

mayeut:
@YannickJadoul 

This PR address your comment in https://github.com/joerick/cibuildwheel/issues/63#issuecomment-386249609 but unfortunately, as shown in #70, It won't help for the issue in #63  (I thought it might by misreading changelog of python 3.4 & 3.5 where openSSL is shipped with the installer, but the 32bit only 10.5 installer, not the 10.6+ 32/64 bits installer)

=> This PR won't fix anything, however, macOS & Windows now have a behavior that more closely match the one from manylinux images which are updated frequently.

The only PR that can help #70 is the #68 which updates 2.7 installer from 2.7.14 to 2.7.15 (but that will only fix 2.7 build, not 3.4 or 3.5 builds)

YannickJadoul:
@mayeut Sorry, I think I was confusing this with the other one from some time ago, #58 and #57.

PR #73, opened 2018-06-16
Add support for Python 3.7

lelit:
This enables building binary wheels for Python 3.7, now at its rc1 release.

It is included in the latest manylinux1 Docker image, see https://github.com/pypa/manylinux/pull/196.

Accordingly to https://www.appveyor.com/docs/build-environment/#python it is not yet available on Appveyor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Space-separated list of builds to skip. Each build has an identifier like `cp27-
 
 The format is `python_tag-platform_tag`. The tags are as defined in [PEP 0425](https://www.python.org/dev/peps/pep-0425/#details).
 
-Python tags look like `cp27` `cp34` `cp35` `cp36`
+Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37`
 
 Platform tags look like `macosx_10_6_intel` `manylinux1_x86_64` `manylinux1_i686` `win32` `win_amd64`
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ What does it do?
 | Python 3.4 | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Python 3.5 | ✅ | ✅ | ✅ | ✅ | ✅ |
 | Python 3.6 | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Python 3.7 | ✅ | ✅ | ✅ |    |    |
 
 - Builds manylinux, macOS and Windows (32 and 64bit) wheels using Travis CI and Appveyor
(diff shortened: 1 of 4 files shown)


lelit:
It fails, because the dockcross/manylinux1 images do not carry py37 yet.

joerick:
awesome work @lelit! any idea when the dockcross image will be published?

lelit:
No, unfortunately they pushed an update just before the standard manylinux1 gained support for 3.7, and from https://hub.docker.com/r/dockcross/manylinux-x64/tags/ it's not clear how often that happens.

YannickJadoul:
Nice! :-)
Is this urgent to include (given that AppVeyor still doesn't have it either), or could we wait 1 or 2 more weeks for the full release (cfr. https://www.python.org/dev/peps/pep-0537/)?

hroncok:
> Is this urgent to include (given that AppVeyor still doesn't have it either), or could we wait 1 or 2 more weeks for the full release

Please, release this. Getting 3.7 builds building at rc1 stage means we can find more 3.7 bugs before the actual release.

YannickJadoul:
> Please, release this.

Fair enough. Let's just temporarily disable that one test, then?

lelit:
> Fair enough. Let's just temporarily disable that one test, then?

Would be it any better if we just ignore/skip unavailable versions within a
given image? In that way, there would be less pressure on cibuildwheel to
follow whatever Docker image one is using (I mean for example the cases when
"imagexyz dropped Python 3.4"...).

YannickJadoul:
Myeah, but wouldn't you want to know by having that as an error and manually disabling that version for that build, instead of silently ignoring what could be a potential oversight?

lelit:
Possibly yes... I was just thinking aloud (I do not use alternative manylinuxXX images myself): the question is about whether it should be **always** an hard error or if it could be downgraded (by default, or with an option) to a warning message instead.

YannickJadoul:
Yes, no, sorry, didn't mean to shut that suggestion down completely, but just my 2 cents on that. Since `CIBW_SKIP` (and e.g. `cp37-manylinux*`) is so easy, I thought there's maybe something to say for "explicit is better than implicit"?

joerick:
Yeah, let's just disable that test on Python 3.7 for now then :)

joerick:
The 06_docker_images test is really just there to check that the CIBW_MANYLINUX1_XXX_IMAGE options are honored, it's up to the user to create a valid config.

joerick:
Released as 0.9.0!

hroncok:
Thank You!

lelit:
Great, many thanks!

PR #76, opened 2018-06-30
Build Windows wheels for Python 3.7 (fixes #75)

### LINKED ISSUE #75: Windows wheels are not built for python 3.7
On Linux wheels are correctly built for all supported python versions. Windows however stops at python 3.6.

Looking at windows.py there seem to be no references to python 3.7

jlaine:


joerick:
Thanks jlaine, and yannick for tracking down the Appveyor thread over in #75.

https://github.com/appveyor/ci/issues/2475

Let's hold on this until their image gets updated.

In the meantime I might improve that error message you're getting - probably we should check that a directory exists before adding to PATH, to avoid the fallback to different versions.

jlaine:
Oh. I just gratuitously rebased my branch, I saw your commit checking for the right python and hadn't realized it wasn't on master.

joerick:
haha, whoops. I can cherrypick it back onto here, don't worry

YannickJadoul:
@joerick So it's probably not worth to have `cibuildwheel` install the Python versions that aren't installed on AppVeyor (similarly to how we do it on macOS) ?
I do agree on that (because I expect the extra amount of code to not be worth it), but just noting/checking, because we might run into similar problems in the future, depending on AppVeyor's update speed...

joerick:
We're green! thanks @jlaine

joerick:
@YannickJadoul we could consider it if they're more late in the future. In this case it would only buy us two weeks before we switch back to the preinstalled version – it's probably not worth it, unless people are really desperate.

YannickJadoul:
@joerick Yes, agreed. But for now we can just keep track of the issues, and if this keeps on happening, maybe it's worth to add the extra code.

YannickJadoul:
Thanks, @jlaine, for notifying and making the PR!

Issue #39, opened 2017-09-19
Multiple CIBW_TEST_COMMAND commands not supported on macOS


hroncok:
I have the following:

      - CIBW_TEST_COMMAND='cd {project}; python -m pytest -v test'

And on Linux, it works, but on macOS, it's broken, because:

```pycon
>>> shlex.split('cd foo; python -m pytest -v test')
['cd', 'foo;', 'python', '-m', 'pytest', '-v', 'test']
```

And

```console
$ cd '/Users/travis/build/admesh/python-admesh;' python -m pytest -v test
/usr/bin/cd: line 4: cd: /Users/travis/build/admesh/python-admesh;: No such file or directory
```

I will now fix my tests not to assume running from the projects directory and will workaround this, but there might be more use cases, where multiple CIBW_TEST_COMMAND commands are desired.

joerick:
Agreed, this should run in a shell, similar to CIBW_BEFORE_BUILD. Should be a simple fix, PR welcome :)

hroncok:
Once I make my test pass and my macOS wheels to build, I'll have a look at this.

PR #81, opened 2018-07-04
macOS: run test_command in shell

madig:
Fixes https://github.com/joerick/cibuildwheel/issues/39.

Let's see if this works.


madig:
Hm. I could change 02_test's `"CIBW_TEST_COMMAND": "cd nosetests {project}/test"` to `"CIBW_TEST_COMMAND": "cd {project}/test && nosetests"`, but that would bomb on AppVeyor.

joerick:
I've added a basic test for shells, but only on Mac and Linux. But it seems that option doesn't have platform-specific variants! I'll add those too.

joerick:
Thanks @madig!

PR #83, opened 2018-07-07
Add pybase64 to the list of working examples

mayeut:
Thanks for the great work !

@joerick, `pybase64` uses `pypi deploy plugin` on `travis` which is not documented in `joerick/cibuildwheel-autopypi-example`. `python-admesh` also uses the plugin but does not build for `macOS` using `python3` which requires a trick when deploying with `pypi deploy plugin`
I can create a PR on  `cibuildwheel-autopypi-example` if you're interested in that.


joerick:
Thanks @mayeut ! Could you re-do this without removing trailing whitespace? Markdown uses two trailing spaces to force a line-break, I think this will break some formatting.

PR #84, opened 2018-08-02
Applying unbuffered output to all platforms instead of only Windows

YannickJadoul:
I have started noticing that the output from `cibuildwheel` on Travis CI is not interleaved well with the output of the subprocesses (just like on Windows/AppVeyor, see #23, #24, #26) in my build: see e.g. https://travis-ci.org/YannickJadoul/Parselmouth/jobs/410930945 (Weirdly enough, this problem does not seem to show up on the CI tests of `cibuildwheel` itself?)

By extending this fix for Windows to all platforms, the problem seems to be fixed: https://travis-ci.org/YannickJadoul/Parselmouth/jobs/411213753

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -4,7 +4,7 @@ import argparse, os, subprocess, sys, textwrap
 import cibuildwheel
 import cibuildwheel.linux, cibuildwheel.windows, cibuildwheel.macos
 from cibuildwheel.environment import parse_environment, EnvironmentParseError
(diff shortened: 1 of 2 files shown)


YannickJadoul:
Can we merge this, or is there something that still needs to be done?

joerick:
I think we're good to merge! Thanks for figuring it out.

YannickJadoul:
Nice, thanks :-)
Nothing urgent, but it's nice to clean up and to have it in some future release.

PR #86, opened 2018-08-19
Update README.md

miguelsousa:
Add horizontal rules between option descriptions; they make it easier to understand what text belongs to each option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,8 @@ optional arguments:
 
 Most of the config is via environment variables. These go into `.travis.yml` and `appveyor.yml` nicely.
 
+***
+
 | Environment variable: `CIBW_PLATFORM` | Command line argument: `--platform`
 | --- | ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -258,7 +274,6 @@ Example: `nose==1.3.7 moto==0.4.31`
 Platform-specific variants also available:
 `CIBW_TEST_REQUIRES_MACOS` | `CIBW_TEST_REQUIRES_WINDOWS` | `CIBW_TEST_REQUIRES_LINUX`
 
---
```
YannickJadoul: If this is removed, maybe still add a horizontal line or make the "Example YML syntax" as `###` title, since it is a part of the "Options" section, a not of this specific environment variable.

miguelsousa: done

YannickJadoul:
Thanks! It is just the bit easier to read, like this :-)

PR #88, opened 2018-08-21
Improve error reporting during package name fetching

### LINKED ISSUE #87: 'wheel' missing from setup.py install_requires list?
While trying to use a vanilla cibuildwheel configuration with AppVeyor, I kept running into this error,
```
cibuildwheel: Failed to get name of the package. Command was ['c:\\python27\\python.exe', '.\\setup.py', '--name']
```

It was only after running that command (`python .\\setup.py --name`) ahead of the `cibuildwheel --output-dir wheelhouse` that I finally got a clue of what was going on,
```
setup requires that the Python package 'wheel' be installed. Try the command 'pip install wheel'.
```

To resolve this problem I've added `'wheel'` to [setup.py `install_requires` list](https://github.com/joerick/cibuildwheel/blob/master/setup.py#L12) in our fork.

I think that two changes are necessary:
1. On [line 111 of `__main__.py`](https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/__main__.py#L111), print the **`output`** of [CalledProcessError](https://docs.python.org/2/library/subprocess.html#subprocess.CalledProcessError), and exit with the value of **`returncode`**.
2. Add `'wheel'` to [setup.py `install_requires` list](https://github.com/joerick/cibuildwheel/blob/master/setup.py#L12)

Do you want a PR?

miguelsousa:
Fixes #87

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -108,9 +108,10 @@ def main():
             print('cibuildwheel: Could not find setup.py at root of project', file=sys.stderr)
             exit(2)
         else:
+            print(err.output)
             print('cibuildwheel: Failed to get name of the package. Command was %s' % err.cmd,
                   file=sys.stderr)
-            exit(2)
+            exit(err.returncode)
 
     if package_name == '' or package_name == 'UNKNOWN':
         print('cibuildwheel: Invalid package name "%s". Check your setup.py' % package_name,


joerick:
Nice one. Thank you!

YannickJadoul:
Should/could we make a test that this actually works? Because the [Python docs](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments) say that `CalledProcessError.output` contains `Output of the child process if it was captured by run() or check_output(). Otherwise, None.`, and I don't see the output being captured?

miguelsousa:
@YannickJadoul I tested these changes with our repo and got the expected results. See https://ci.appveyor.com/project/adobe-type-tools/afdko/build/1.0.1115
```
cibuildwheel --output-dir wheelhouse
setup requires that the Python package 'wheel' be installed. Try the command 'pip install wheel'.

cibuildwheel: Failed to get name of the package. Command was ['c:\\python27\\python.exe', '.\\setup.py', '--name']
Command exited with code 1
```
The first message and the return code came from https://github.com/adobe-type-tools/afdko/blob/da379f5e143b471faaa370c6494471b4118ba408/setup.py#L29

I'm all in favor of adding test cases along with any code changes, but in this case I'm not sure how to craft one.

Issue #90, opened 2018-08-28
CircleCI support


altendky:
I'm setting up [Twisted](https://twistedmatrix.com/) with `cibuildwheel` over in twisted/twisted#1051.  I haven't really tested anything but I got Windows wheels building on AppVeyor and Linux wheels on Travis so far.  Next step? CircleCI for OSX wheels...  Before I dive in, any comments on any other efforts to do this?  Interest in trying to do some of the work?  Suggestions how I get started?  I honestly haven't done anything quite yet but I like to raise the topic publicly first in case there's advice or opinions worth sharing.

Cheers,
-kyle

P.S. Thanks so much for cibuildwheel, it certainly made this all go a lot quicker so far.

joerick:
Thanks Kyle! I'll reply to the PR.

altendky:
More work can certainly be done but this CircleCI is working.

PR #91, opened 2018-08-29
Add CircleCI support

altendky:
joerick/cibuildwheel#90

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -131,9 +131,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             '--env',
             'CIBUILDWHEEL',
             '--name', container_name,
-            '--rm',
             '-i',
-            '-v', '%s:/output' % os.path.abspath(output_dir),
             '-v', '/:/host',
             docker_image,
             '/bin/bash',
@@ -164,21 +162,32 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
 
         time.sleep(10)
 
-        if host == Host.Circle:
-            command = [
-                'docker',
-                'cp',
-                './.',
-                '{}:/project'.format(container_name),
-            ]
-            print('docker command: {}'.format(command))
-            subprocess.check_call(command)
+        command = [
+            'docker',
+            'cp',
+            './.',
+            '{}:/project'.format(container_name),
+        ]
+        print('docker command: {}'.format(command))
+        subprocess.check_call(command)
 
         try:
             docker_process.communicate(bash_script)
         except KeyboardInterrupt:
             docker_process.kill()
             docker_process.wait()
+
+        command = [
+            'docker',
+            'cp',
+            './.',
+            '{}:/output'.format(container_name),
+            os.path.abspath(output_dir),
+        ]
+        print('docker command: {}'.format(command))
+        subprocess.check_call(command)
+
+        # TODO: dropped -rm above so cleanup here
 
         if docker_process.returncode != 0:
             exit(1)
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 import os, subprocess, sys, time, uuid
 from collections import namedtuple
 from .util import prepare_command, get_build_verbosity_extra_flags
-from .environment import Host
 
 import monotonic
 
@@ -12,7 +11,7 @@ except ImportError:
     from pipes import quote as shlex_quote
 
 
-def build(project_dir, package_name, output_dir, test_command, test_requires, before_build, build_verbosity, skip, environment, manylinux1_images, host):
+def build(project_dir, package_name, output_dir, test_command, test_requires, before_build, build_verbosity, skip, environment, manylinux1_images):
     try:
         subprocess.check_call(['docker', '--version'])
     except:
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -180,7 +180,6 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         command = [
             'docker',
             'cp',
-            './.',
             '{}:/output'.format(container_name),
             os.path.abspath(output_dir),
         ]
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -181,7 +181,7 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         command = [
             'docker',
             'cp',
-            '{}:/output'.format(container_name),
+            '{}:/output/.'.format(container_name),
             os.path.abspath(output_dir),
         ]
         print('docker command: {}'.format(command))
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -127,7 +127,9 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
 
         command = [
             'docker',
-            'exec',
+            'run',
+            '--env',
+            'CIBUILDWHEEL',
             '--name', container_name,
             '-d',
             '-i',
@@ -151,10 +153,11 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
 
         command = [
             'docker',
-            'run',
+            'exec',
             '--env',
             'CIBUILDWHEEL',
             container_name,
+            '/bin/sh',
         ]
         print('docker command: {}'.format(command))
         docker_process = subprocess.Popen(command, stdin=subprocess.PIPE, universal_newlines=True)
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -14,7 +14,8 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
         subprocess.check_call(['docker', '--version'])
     except:
         print('cibuildwheel: Docker not found. Docker is required to run Linux builds. '
-              'If you\'re building on Travis CI, add `services: [docker]` to your .travis.yml.',
+              'If you\'re building on Travis CI, add `services: [docker]` to your .travis.yml.'
+              'If you\'re building on Circle CI in Linux, add a `setup_remote_docker` step to your .circleci/config.yml',
               file=sys.stderr)
         exit(2)
 
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -131,15 +131,20 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             '--env',
             'CIBUILDWHEEL',
             '--name', container_name,
-            '-d',
+            '--rm',
             '-i',
             '-v', '%s:/project' % os.path.abspath(project_dir),
             '-v', '%s:/output' % os.path.abspath(output_dir),
             '-v', '/:/host',
             docker_image,
+            '/bin/bash',
         ]
         print('docker command: {}'.format(command))
-        subprocess.check_call(command)
+        docker_process = subprocess.Popen(
+            command,
+            stdin=subprocess.PIPE,
+            universal_newlines=True,
+        )
 
         if host == Host.Circle:
             command = [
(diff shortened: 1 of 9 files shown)


altendky:
I can't say I've tested this yet but `--platform macos` did get me some wheels.

https://circleci.com/gh/twisted/twisted/1832#artifacts/containers/0

altendky:
I suppose I should probably add a CircleCI config to match the existing AppVeyor and Travis configs...

altendky:
Alrighty, so I decided to at least temporarily add a Linux wheel build on Circle since they offer artifact storage (unlike Travis).  I had to add a [`setup_remote_docker`](https://circleci.com/docs/2.0/building-docker-images/) (https://github.com/twisted/twisted/pull/1051/commits/034b148f5af1b7da7a2318339a12c648c3963580) step.  After that it seems that the files in the host current directory are not available in the guest current directory.  You can see in the log that I did print out the current directory prior to calling `cibuildwheel` and that `setup.py` is there.

https://circleci.com/gh/twisted/twisted/1853?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
+ /opt/python/cp27-cp27m/bin/pip wheel . -w /tmp/built_wheel --no-deps
Directory '.' is not installable. File 'setup.py' not found.
```

I'm trying to print docker logs etc to see why it's not mounting properly or such.  I'll update if I find something helpful.

joerick:
Sorry Kyle, I replied to this via email on my phone a few days ago and it didn't come through!

Well done on the progress with the Mac build, looks great. We'd also need a (super minimal) sample config example, and perhaps the ability to run our tests on CircleCI to catch any platform-related regressions.

As for the linux build, it might be a little more complicated from the looks of things? I found the following note from their docs:

> It is not possible to mount a folder from your job space into a container in Remote Docker (and vice versa). You may use the docker cp command to transfer files between these two environments. For example, to start a container in Remote Docker using a config file from your source code:  
> https://circleci.com/docs/2.0/building-docker-images/#mounting-folders

Shared mounts is a part of how cibuildwheel works, so might be a bit more tricky. If you can find a solution, that would be great, but I'd also happily accept this PR for the Mac parts only.

p.s. really chuffed that twisted is finding the library useful! :)

altendky:
Thanks for finding that in the docs.  Makes it pretty straightforward what is needed.  I'll take a pass at implementing that when I get a chance.

joerick:
I've enabled CircleCI for this project, so you should be able to try running tests using that.

altendky:
Good deal, thanks.

altendky:
It appears that OSX is not free on CircleCI.  I'll keep going with the Linux tests though where the issue is.

https://circleci.com/gh/organizations/joerick/settings#osx-pricing

joerick:
I've written to them asking if they can enable Mac builds, so presumably that will go through in the near couple hours

joerick:
> Is there some reason that the OS platform isn't detected directly via sys.platform()?

Yes, it's to prevent accidentally running on a development machine - the Windows implementation relies on paths specific to the Appveyor device, and the Mac implementation will install software globally, which isn't desirable. So to really run cibuildwheel outside of CI you have to specify it.

joerick:
Also, CircleCI just got in touch - macOS builds are enabled on this repo :)

altendky:
Would you consider having that as a separate item to be more explicit?  I'm not sure if it would be good to use but they do all seem to define `CI`, though I was originally thinking more like detecting platform via `sys.platform()`, detecting host via env vars, and letting you force a run on an unrecognized host via `--force-without-host` or whatever.

I don't like the name host for this but it's what I've got now.  Happy to change to something better.  `CiHost` maybe?

Thanks for the osx builds.  I'll plan to get that going as well.

altendky:
Also, would you rather the copy in/out be an exception?  Or just make it the normal operation so that all (Linux anyways) hosts run the same?

joerick:
> Would you consider having that as a separate item to be more explicit?

Maybe, if makes the user experience more elegant, and doesn't break backward compat. I'm quite used to doing `--platform linux` to run test builds on my Mac (since that works on docker as well) so that should still work. 

> Also, would you rather the copy in/out be an exception?

It depends on what the code ends up looking like, but I think I'd rather just have one technique, so long as performance doesn't take a big hit with all the copying back-and-forth.

altendky:
The `/host` mount isn't explicitly used by `cibuildwheel` but is provided for use by developers.  Shall we just document CircleCI as an exception to that for now and leave it in?

altendky:
Hmm...  presently it only copies _into_ the container's `/project` so any projects depending on changes written there being accessible after would be broken.  Breaking change with documentation?  Or copy `/project` back out as well as `/output`?

altendky:
@joerick, alrighty, it seems to be 'working'.  I tidied a little bit anyways.

* Need answers to questions in above two comments.
* Need to add documentation and examples as you commented, but I usually wait for that until the functionality is agreed upon.
* Pending the Twisted PR being accepted I guess there's a list of 'projects using `cibuildwheel`' that I would need to update.
* There are some more diagnostic prints.  I tend to leave stuff like that in but...  if you don't want it I'll take it out.  Perhaps another ticket is needed (not volunteering at present :] ) to use `logging` (or some logger) with the ability to set verbosity.

I updated my [Twisted PR](https://github.com/twisted/twisted/pull/1051#issuecomment-418147553) to use this latest commit and we've got a bunch of wheels.  :]

altendky:
Oh, and sorry for the shotgun commits and the resulting CI clogs...  that's a bad habit of mine.

cibuildwheel/linux.py
```diff
@@ -115,25 +125,64 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
...
+                'cp',
+                '{}/.'.format(os.path.abspath(project_dir)),
+                '{}:/project'.format(container_name),
+            ]
+            print('docker command: {}'.format(command))
+            if subprocess.call(command) == 0:
+                break
+
+            if monotonic.monotonic() - start > 60:
+                break
```
joerick: A comment about why this retry loop is required would be nice :) and perhaps an informative error on timeout

joerick:
Thanks @altendky! Good job on getting this working. 

About your questions above-

> The /host mount isn't explicitly used by cibuildwheel but is provided for use by developers. Shall we just document CircleCI as an exception to that for now and leave it in?

Let's just leave it in with documented exception, as you say. 

> Hmm... presently it only copies into the container's /project so any projects depending on changes written there being accessible after would be broken

I'm not sure this will be a problem for many people. I'd be happy to leave as-is for now, and if anyone pops up in the issues we can look at it again. 

Overall, happy with the approach ✅

You can leave the diagnostic logs that were useful in there :)

altendky:
Can you grab a Circle status badge from the project settings?  Looks like you've been using the SVG option for the others (which makes me :] ).
https://circleci.com/docs/2.0/status-badges/

altendky:
Humm, just saw this.

https://github.com/joerick/cibuildwheel/blame/7c6b331e1555880f2d0b769c5f04fbcfae22b41e/README.md#L354

> Since `cibuildwheel` runs the wheel through delocate or auditwheel, it will automatically bundle library dependencies. This is similar to static linking, so it might have some licence implications. Check the license for any code you're pulling in to make sure that's allowed.

This seems likely not what would be wanted for a wheel you put on PyPi?  Better suited to something you use to do a standalone distribution.  Well, better is a judgement call, but it wouldn't be expected anyways.  I am not familiar with those tools.  Would it be easy to make this optional?  (separate PR presumably)

joerick:
SVG badge here-

```
[![CircleCI](https://circleci.com/gh/joerick/cibuildwheel.svg?style=svg)](https://circleci.com/gh/joerick/cibuildwheel)
```

And...

> Since cibuildwheel runs the wheel through delocate or auditwheel, it will automatically bundle library dependencies. This is similar to static linking, so it might have some licence implications. Check the license for any code you're pulling in to make sure that's allowed.

This might be a subject for another issue! The above note might be a bit misleading. delocate and auditwheel will bundle dynamic libraries that are linked as part of compilation of the C extension - things like `libgfortran` or `libffi`. It doesn't include python package dependencies specified in setup.py.

(The goal of auditwheel and delocate is to let users `pip install X` and have all the software required to use that library - the alternative is to require the user `brew install` this or `apt-get install` that)

Would you still be interested in an option to toggle that?

YannickJadoul:
> This seems likely not what would be wanted for a wheel you put on PyPi?

Isn't that the whole point of building wheels? To be independent of compilation and installation on the user's setup? For example, if `auditwheel` hasn't run, it doesn't match the `manylinux1` PEP standard, and shouldn't be uploaded as `manylinux1` wheel to PyPI.

altendky:
I did misunderstand that statement and thought it meant it would include other Python packages.  I'm much less concerned now.  (And yes, that would certainly have been a separate issue) Thanks for the clarification.

joerick:
This looking great! Thanks for all your work Kyle!

joerick:
Merged. I'll try to get a release out on Friday/the weekend.

cibuildwheel/linux.py
```diff
@@ -115,25 +129,72 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
...
+            stdin=subprocess.PIPE,
+            universal_newlines=True,
+        )
+
+        # It will take a bit for docker to get the container up and
+        # running.  Keep trying to copy in the project directory until
+        # it succeeds.  There is a timeout to avoid spinning forever.
+
+        timeout = 180
+        start = monotonic.monotonic()
```
YannickJadoul: Do we really need this `monotonic` dependency? I can see it's maybe good practice to not trust your clock, but if it's just to timeout Docker, wouldn't `time` just do? Especially since `cibuildwheel`is installed again and again on the CI services, it seems nice to limit the amount of dependencies.

Otherwise, if necessary to use `monotonic`, let's document why, so no one will be asking that same question again.

cibuildwheel/linux.py
```diff
@@ -115,25 +129,72 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
...
+        command = [
+            'docker',
+            'cp',
+            '{}:/output/.'.format(container_name),
+            os.path.abspath(output_dir),
+        ]
+        print('docker command: {}'.format(command))
+        subprocess.check_call(command)
+
+        command = [
```
YannickJadoul: Would it be nice to have these different docker calls in separate functions, to keep the overview?

YannickJadoul: (Or maybe collapse some of them to single lines?)

cibuildwheel/linux.py
altendky: It's proper to avoid affects of system time getting changed. It's unlikely to matter in practice. I'm surprised the added dependency got accepted.  :]  If dropping the dependency I would use monotonic in py3 and time in py2 in suppose.  We could probably vendor it as well... But I expect we don't get there.

cibuildwheel/linux.py
altendky: Separate functions? Maybe.  For formatting, that's just what I do when it doesn't fit on one line.  Less judgement, more consistency, less diff noise when you change one thing.

cibuildwheel/linux.py
YannickJadoul: Myeah, but we're mainly running this for a short time on a protected image (where no one's playing with the time), anyway, on the CI servers, so I don't know if this precaution is worth it?

If time does change:
- Either the build fails immediately, if time moved forward and docker wasn't ready. Restarting the job will work.
- Or the build itself will stall for a bit, until a longer timeout is reached (or until the main build times out).

I guess this is up to @joerick? But I'd vote to get rid of it. I've seen Travis CI fail more times because of random errors (apt install failing, or simply no output, or ..) than because of `cibuilwheel`, so as long as nothing bad, undetected happens (such as corrupt wheels), I'm not sure about the dependency.

YannickJadoul: (So even using `time` overall would be fine with me, to keep the Python code equal for 2 and 3?)

joerick: monotonic:

download

I'm okay with the dependency. 

Also the `monotonic` code is sufficiently hairy that I'd rather not have to deal with it :P

cibuildwheel/linux.py
altendky: @YannickJadoul the PR is closed...  or I probably would have done some of this.  :]  Good ideas.

cibuildwheel/linux.py
altendky: @YannickJadoul yeah, I'm not strongly attached to monotonic in this case.  Though it generally is a pet peeve of mine.  I won't be particularly bothered if you talk @joerick out of it.  Mostly I just started with what I considered best practice and then @joerick went and accepted it without giving you a chance to review.  :]

YannickJadoul: Eh, it's still @joerick's call. I don't really mind, but I thought it might be worth explicitly making the decision.

cibuildwheel/linux.py
joerick: Ah, I may have gotten a bit trigger happy in merging! Let's continue discussion in #96.

Issue #93, opened 2018-09-02
Feature request: provide a way to build a wheel based on a PyPI release


astrofrog:
It would be nice to have a way to build wheels based on an existing source release on PyPI without having to manually download the tar file, expand it, etc. This could look like:

```
cibuildwheel --from-pypi=mypackage
```

It would be nice to be able to optionally specify a version (when building wheels for old versions of packages).

Another related feature that would be nice is the ability to also skip the building of wheels when a wheel with the same name exists on PyPI. This would then allow a workflow of having a dedicated repository with a bunch of ``cibuildwheel`` commands for different packages, and have it run e.g. as a daily cron job, building new wheels whenever necessary.

joerick:
Interesting! I guess this is for running cibuildwheel on something other than Travis, Appveyor, something like Jenkins?

Just something that popped into my head - how does it work when the source tarball is uploaded before the wheels are? Do the installs in the meantime require source-builds, or could the version be marked as prerelease until the wheels are available?

astrofrog:
@joerick - actually I still would like to use this on Travis and AppVeyor. I'd just rather keep the repo separately from the actual package repo - because we have 4-5 packages we need to build and it'd be nicer to just centralize the building.

When the source tarball is uploaded before the wheels (which is common for our workflow at the moment), the source version is used automatically by pip if wheels are not available.

astrofrog:
@joerick @YannickJadoul - just for information, I've built a package called [autowheel](https://github.com/astrofrog/autowheel) that is a wrapper around cibuildwheel, and makes it easy to automatically build any missing wheels from PyPI releases for one or more packages. I'm not sure if you are interested in having something like that be part of cibuildwheel or if it's best to keep it separate, but I just wanted to give you a heads-up in case you are interested. I haven't actually released it at this point and am just using it for personal projects, but I do plan to release it at some point (unless there is interest in making it part of cibuildwheel).

astrofrog:
@joerick - yes, in my case I'll have a repo that just has this config and runs on a cron job and builds wheels for a number of different packages.

joerick:
Sweet. It's great that you just built it, wrapping cibuildwheel to use it the way you wanted! I like the YAML config too. 

I'll add a link from the README. Also, let me know if there's something we can do to make it easier to wrap!

PR #94, opened 2018-09-02
Adding CIBW_FROM_PYPI option

YannickJadoul:
First step towards addressing #93.

A nice extra is that I expect this will also work for a value `my-pypi-package==0.42.1`, building (or downloading, if the wheel already exists) that specific release.

Things to do, if this approach works:
- [ ] Does this need to be an environment variable option (as I've made it, for the moment), or the suggested `--from-pypi=...` argument?
- [ ] Document this feature in `README.md` (depending on the actual form the option will take).
- [ ] Not sure what's the best way to test this without actually using a project that's already been uploaded to PyPI. @astrofrog, maybe you could give it a shot, using this branch to build your project?
- [ ] The list of arguments to the different platform functions is getting quite long. What about introducing a `collections.namedtuple` (e.g., `BuildOptions`) to nicely wrap all the options and only pass one argument to `linux.build`/`macos.build`/`windows.build`?

EDIT: Sorry about initially pushing the commit to the master. Something went wrong, there.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -85,6 +85,7 @@ def main():
     build_verbosity = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
     skip_config = os.environ.get('CIBW_SKIP', '')
     environment_config = get_option_from_environment('CIBW_ENVIRONMENT', platform=platform, default='')
+    from_pypi = os.environ.get('CIBW_FROM_PYPI', '')
 
     try:
         build_verbosity = min(3, max(-3, int(build_verbosity)))
@@ -136,6 +137,7 @@ def main():
         build_verbosity=build_verbosity,
         skip=skip,
         environment=environment,
+        from_pypi=from_pypi,
     )
 
     if platform == 'linux':
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -15,7 +15,7 @@ class DockerRunTimeoutError(Exception):
     pass
 
 
-def build(project_dir, package_name, output_dir, test_command, test_requires, before_build, build_verbosity, skip, environment, manylinux1_images):
+def build(project_dir, package_name, output_dir, test_command, test_requires, before_build, build_verbosity, skip, environment, manylinux1_images, from_pypi):
     try:
         subprocess.check_call(['docker', '--version'])
     except:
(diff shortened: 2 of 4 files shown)


YannickJadoul:
@astrofrog Up to you to argue, here. I don't really mind either way. I just saw an easy way of implementing the requested feature in this PR.

astrofrog:
> moving off the pip wheel style of building wheels, to the python setup.py bdist_wheel method

Note that as far as I can tell, using ``setup.py`` to do various things is going to be deprecated in the long term in favor of using ``pip`` - see for example some of the replies here:

https://github.com/pypa/setuptools/issues/1156

In particular the sentence in one of the replies *Also, running setup.py directly will (after a long period of deprecation) be transitioned to implementation-defined behavior*

So I'm not sure it makes sense to transition to ``setup.py`` from ``pip`.

> Added to that, I'm not sure I totally understand the workflow when running in Travis CI or Appveyor - since those tools are designed to run as a result of a git change, not manually or reacting to a change in PyPI 🤔

I do something similar for conda packages - basically I have Travis and AppVeyor running as cron jobs, so that they can watch for changes in e.g. PyPI rather than react to changes in the git repo.

In any case, I'm fine with just doing the downloading from PyPI myself - this was just an idea in case it was useful for others too :)

joerick:
Cool, thanks for the info! I didn't know that pip will eventually inherit those commands. 

How do you run Travis/Appveyor as cron jobs? Just push empty commits to them every day or something?

If we were to add something like this to cibuildwheel (which could be useful), I'd like it to be implemented separately from the `pip wheel <pypi_package>` feature. So an explicit `cibuildwheel --from-pypi my-project` - then on getting that command, cibuildwheel would download the tarball, expand it into a temp dir and then run as normal on that. That would keep the feature localised to one bit of the code base. Just trying to keep things consistent.

(By the way, do we even need `package_name` any more? Maybe I'll file an issue to remove it.)

astrofrog:
@joerick - Travis and AppVeyor both have options in the settings for setting up cron jobs, e.g. https://docs.travis-ci.com/user/cron-jobs/

I'll have a think about how ``cibuildwheel --from-pypi my-project`` could be implemented.

joerick:
Oh, interesting! I'm starting to get it :)

YannickJadoul:
So, should we close this particular PR, then?

joerick:
Yes, I think so. I agree on the potential usefulness, just not this implementation. If you're happy, lets close the PR, but keep the issue to track the feature request.

YannickJadoul:
Nope, no problem. Just saw the issue, thought of this approach, and wanted to test/propose it :-)

PR #95, opened 2018-09-05
overwrite existing wheel file with same name

anthrotype:
In https://github.com/adobe-type-tools/afdko/pull/595 we attempt to build a universal `py2.py3` wheel
thus the filename of the wheel is the same as the previous ones, and it causes a `shutil.Error`
https://travis-ci.org/adobe-type-tools/afdko/jobs/424691105#L1460

Our current workaround is to just remove the previous existing file before moving the new one (they'll be identical anyway, or at least they *should*).

Ideally, it would be nice if one could build a wheel once, and run tests on multiple python versions, but that this would involve many more changes.


YannickJadoul:
Hmmm, apparently, if you specify the full destination file name, it will overwrite an existing file? https://stackoverflow.com/questions/31813504/move-and-replace-if-same-file-name-already-existed-in-python

This might be a slightly shorter solution, then (leaving everything else to the OS)?

```python
shutil.move(delocated_wheel, os.path.join(output_dir, os.path.basename(delocated_wheel)))
```

joerick:
I'm cool with this, pending that style change @YannickJadoul  suggested

anthrotype:
That would only work on Unix, not on Windows. Shutil.move uses the os.rename semantics

> If the destination already exists but is not a directory, it may be overwritten depending on os.rename() semantics.

https://github.com/python/cpython/blob/d700f97b627989d41cd4629dc02969f9a6b56d2f/Lib/shutil.py#L668-L669

From https://docs.python.org/3/library/os.html#os.rename

> On Windows, if dst already exists, OSError will be raised even if it is a file.

anthrotype:
I can squash the commits i you like

PR #96, opened 2018-09-08
Linux / Docker tweaks

joerick:
Hey @altendky , @YannickJadoul 

Was looking at the linux build script again today and saw a few modifications that I'd like to make - mostly around reducing that timeout, and the error handling.

Let me know if this looks okay to you!

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -148,53 +148,60 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             stdin=subprocess.PIPE,
             universal_newlines=True,
         )
+        try:
+            # Copy the project into the Docker container
+            #
+            # It will take a bit for docker to get the container up and
+            # running.  Keep trying to copy in the project directory until
+            # it succeeds.  There is a timeout to avoid spinning forever.
+            timeout = 60
+            start = monotonic.monotonic()
+            while True:
+                time.sleep(1)
+                command = [
+                    'docker',
+                    'cp',
+                    '{}/.'.format(os.path.abspath(project_dir)),
+                    '{}:/project'.format(container_name),
+                ]
+                print('docker command: {}'.format(command))
+                if subprocess.call(command) == 0:
+                    break
 
-        # It will take a bit for docker to get the container up and
-        # running.  Keep trying to copy in the project directory until
-        # it succeeds.  There is a timeout to avoid spinning forever.
+                if monotonic.monotonic() - start > timeout:
+                    raise DockerRunTimeoutError(
+                        'Unable to successfully copy project directory'
+                        ' within {} seconds'.format(timeout)
+                    )
 
-        timeout = 180
-        start = monotonic.monotonic()
-        while True:
-            time.sleep(1)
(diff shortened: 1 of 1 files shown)


cibuildwheel/linux.py
```diff
@@ -148,53 +158,60 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
...
-        start = monotonic.monotonic()
-        while True:
-            time.sleep(1)
+        try:
+            # Copy the project into the Docker container
+            #
+            # It will take a bit for docker to get the container up and
+            # running.  Keep trying to copy in the project directory until
+            # it succeeds.  There is a timeout to avoid spinning forever.
+            timeout = 30
```
joerick: Think we can get away with 30 now the image has already been pulled @altendky? The few builds I watched, it took about 4 seconds to start, so 30 seconds seemed reasonable.

altendky: Sure?  It's a timeout to failure so...  I mean, this should all be configurable but that might make sense as a separate PR to add configurability of a bunch of stuff (I haven't looked enough to have much in mind, but there's usually lots that might sometimes need to be different).

YannickJadoul: Sorry to keep bringing up different suggestions that might have been considered before, but if I look at the link posted in the original PR (https://circleci.com/docs/2.0/building-docker-images/#mounting-folders), there seems to be a way to run different docker steps without the timeout?

I.e., `create`, `cp`, `run` (vs. the `run`, <wait>, `cp`, <send-script>) ? Did you try this? Would you guys object against me trying this out in a PR and seeing what happens? (It's not that I really mind the timeout, but there's so much code dragging away the attention from the actual building steps.)

joerick: @YannickJadoul That's a good idea - I hadn't realised that was possible! If we can remove the timeout that would be ace. I'd be interested in your take on it regardless. Feel free to push modifications to this branch or start a new one :)

altendky: I think... I went the route I did because it didn't require the separate cleanup step that got added here.  I was able to just pass the `--rm` to `run`.  So I think you either have a check/timeout or a separate cleanup step.  I'm not strongly attached to 'my way', but that's the explanation.  And in case it's not obvious, I barely know Docker...

altendky: @YannickJadoul, I just didn't know offhand how to be sure to totally cleanup.  I suspect it isn't that hard and I'll agree that'd likely be tidier (though it ought to become optional at some point `:]`).

YannickJadoul: @altendky Yeah, you had to figure out everything by yourself, and didn't have the advantage of hindsight, of course :D

YannickJadoul: Things seem to be working, btw, at #97, but I'm very happy to hear what you have to say and to know what I am forgetting ;-)

joerick:
Closed in favour of #97

YannickJadoul:
If there were other small things and cleanups in this PR, do push them to the #97 branch! :-)

Issue #98, opened 2018-09-10
Add CIBW_BUILD as correlary to CIBW_SKIP (and cli option)


altendky:
Presently the build list  is effectively everything supported.  It'd be nice to be able to easily request a single wheel build to fit into CI schemes with a build per Python version.  This is coming up in my Twisted PR's as I try to set it up to run all tests against the built wheels.  Travis doesn't save artifacts or even files between jobs (without using external storage accounts) so I'm shooting for building in the individual job.

I would expect the built items to be effectively the `set(CIBW_BUILD) - set(CIBW_SKIP)`.

Of course there we'll be some overhead to spinning up docker more times.  We'll see what that is after I get my PR working.  twisted/twisted#1067

I might work on this soon, I might not.  We'll see...  :]

altendky:
@YannickJadoul if you want to take it on, go for it.  :]  Though if there's interest in subcommands it might be worth switching over to click instead first.  I personally use click for all my CLI's unless I'm trying to make a bootstrap sort of script.

astrofrog:
I'd really like to see this feature too!

joerick:
PRs welcome! :D

YannickJadoul:
PR's there!

> Though if there's interest in subcommands it might be worth switching over to click instead first.

I don't know what the actual plan is for subcommands? So for now I have just added this on the same level as `CIBW_SKIP`. But it was rather easy to fit into the current design (I'm still surprised by how easy it is to make these small changes to the code base!), so the PR shouldn't stop any further developments.

altendky:
Thanks all. Released in 0.10.0.

PR #101, opened 2018-09-16
Adding CIBW_BUILD option

YannickJadoul:
Cfr. #98, this is adding CIBW_BUILD to positively select a set of build targets (rather than deselecting everything else).

I'm not completely happy with the naming of `BuildSelector` and `selection` (replacing `BuildSkipper` and the `skip` argument to the different platforms), but I can't think of anything better, right now.

Comments, remarks, and identification of bugs are very welcome, @altendky, @astrofrog, and @joerick!


altendky:
Thanks. Quick look over the code mostly makes sense. I'll give it a shot over in twisted/twisted#1067.

README.md
```diff
@@ -213,15 +216,16 @@ Platform tags look like `macosx_10_6_intel` `manylinux1_x86_64` `manylinux1_i686
...
-- Skip building on Python 2.7 on the Mac: `cp27-macosx_10_6_intel`  
-- Skip building on Python 2.7 on all platforms: `cp27-*`  
-- Skip Python 2.7 on Windows: `cp27-win*`  
-- Skip Python 2.7 on 32bit Windows: `cp27-win32`  
-- Skip Python 3.4 and Python 3.5: `cp34-* cp35-*`
-- Skip Python 3.6 on Linux: `cp36-manylinux*`
-- Only build on Python 3.6: `cp27-* cp34-* cp35-*`
-
-***
+- Only build on Python 3.6: `CIBW_BUILD`:`cp36-*`
```
joerick: ohhhh CIBW_BUILD can do wildcards too! 🤯that's nice

README.md
```diff
@@ -213,15 +216,16 @@ Platform tags look like `macosx_10_6_intel` `manylinux1_x86_64` `manylinux1_i686
...
+- Skip building on Python 2.7 on the Mac: `CIBW_SKIP`:`cp27-macosx_10_6_intel`
+- Skip building on Python 2.7 on all platforms: `CIBW_SKIP`:`cp27-*`
+- Skip Python 2.7 on Windows: `CIBW_SKIP`:`cp27-win*`
+- Skip Python 2.7 on 32-bit Windows: `CIBW_SKIP`:`cp27-win32`
+- Skip Python 3.4 and Python 3.5: `CIBW_SKIP`:`cp34-* cp35-*`
+- Skip Python 3.6 on Linux: `CIBW_SKIP`:`cp36-manylinux*`
+- Only build on Python 3 and skip 32-bit builds: `CIBW_BUILD`:`cp3?-*` and `CIBW_SKIP`:`*-win32 *-manylinux1_i686`
```
joerick: nice :)

cibuildwheel/__main__.py
```diff
@@ -99,7 +99,7 @@ def main():
         traceback.print_exc(None, sys.stderr)
         exit(2)
 
-    skip = BuildSkipper(skip_config)
+    selection = BuildSelector(build_config, skip_config)
```
joerick: BuildSelector seems like a reasonable name to me. `selection` is a little confusing though? Maybe just call it `build_selector`? open to other ideas though

cibuildwheel/util.py
```diff
@@ -21,15 +21,18 @@ def get_build_verbosity_extra_flags(level):
...
+class BuildSelector(object):
+    def __init__(self, build_config, skip_config):
+        self.build_patterns = build_config.split()
+        self.skip_patterns = skip_config.split()
 
     def __call__(self, build_id):
-        return any(fnmatch(build_id, pattern) for pattern in self.patterns)
+        def match_any(patterns):
+            return any(fnmatch(build_id, pattern) for pattern in patterns)
+        return match_any(self.build_patterns) and not match_any(self.skip_patterns)
```
joerick: So the logic is 'match any of the build patterns, and none of the skip patterns.' Makes sense!

> (b1 or b2 or b3) and not (s1 or s2 or s3)  
> (b1 or b2 or b3) and not s1 and not s2 and not s3

test/04_build_skip/environment.json
```diff
@@ -0,0 +1,4 @@
+{
+    "CIBW_BUILD": "cp3?-*",
+    "CIBW_SKIP": "cp33-*"
```
joerick: Python 3.3 is about to be removed from cibuildwheel - maybe better to change this to Python 3.4

README.md
YannickJadoul: Your original code, just duplicated ;-)

cibuildwheel/util.py
YannickJadoul: Credit to @altendky for coming up with those semantics. But indeed, I've been thinking a little bit more, and this gives a really intuitive interpretation (there's not even really a 'first `CIBW_BUILD` is applied, and then `CIBW_SKIP`' kind of sequential semantics)

cibuildwheel/__main__.py
YannickJadoul: Myeah, I liked how `skip(...)` read, but couldn't really reproduce that

YannickJadoul: The other option is to keep this as `BuildSkipper`, btw, and just add the extra argument of `build_config`? Depending on your point of view, this class is still determining what builds to skip.

joerick: BuildSkipper feels like it relates too closely with CIBW_SKIP. I like it the way it is, I think, just the `selection(identifier)` is a bit too vague.

altendky:
https://travis-ci.org/twisted/twisted/builds/429309631?utm_source=github_status&utm_medium=notification

It's a pretty minimal 'test' but I think it did what I wanted.

test/04_build_skip/environment.json
YannickJadoul: Changed the folder name anyway, and added the line. So to almost any diff, this file is (almost) completely changed anyway, so might as well clean this up? :-/

YannickJadoul:
Tests are failing because on OSX, there's now only 3 wheels being produced for the `04_build_skip` test (i.e., 3.5, 3.6, 3.7, because `CIBW_BUILD` selects only 3.<something> and `CIBW_SKIP` skips 3.4, and because 3.3 is already not supported anymore on macOS). I'm seeing if I can find a solution.

YannickJadoul:
> I'm seeing if I can find a solution.

Pfff, I wanted to add a file with the list of files to be produced by a test, but then I need to start detecting the platform, as well.

I could split the test again into one for `SKIP` and one for `BUILD`, I suppose, but ... meh :-/ Any suggestions on making this test and its assert work are welcome.

joerick:
> Any suggestions on making this test and its assert work are welcome.

Hmm. We've talked in the past of the need to update how these tests work with a .py file for each script. Maybe we create an issue to track that refactoring, and for now we can reduce that magic number to 3 (there's nothing special about minimum 4 wheels per platform).

altendky:
Do we need more review here?  Trying it out in twisted/twisted#1067.

altendky:
Yeah, I'm still happy with it from what I've checked.  @joerick, do you see anything else we need to address?

joerick:
Released in 0.10.0! Thanks both!

Issue #106, opened 2018-10-01
OSX: open_for_csv not found in wheel.util (on delocate-list)


ysig:
Hello,

although my package [builds normally on Linux under Travis](https://travis-ci.org/ysig/GraKeL/builds/435300832), [when running it on OSX](https://travis-ci.org/ysig/GraKeL/jobs/435300837) I get the strange error:

>  (most recent call last):
>   File "/Library/Frameworks/Python.framework/Versions/2.7/bin/delocate-listdeps", line 12, in <module>
>     from delocate import tree_libs, wheel_libs, __version__
>   File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/delocate/__init__.py", line 3, in <module>
>     from .delocating import delocate_path, delocate_wheel, patch_wheel
>   File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/delocate/delocating.py", line 18, in <module>
>     from .wheeltools import rewrite_record, InWheel
>   File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/delocate/wheeltools.py", line 14, in <module>
>     from wheel.util import urlsafe_b64encode, open_for_csv, native
> ImportError: cannot import name open_for_csv

for which I couldn't find a reference online.
Is it because a library is downgraded on OSX (as another one get's upgraded)?
Maybe `delocate` should be upgraded or `wheel`, although I don't have a clue :)
I am using the [default osx image of travis](https://docs.travis-ci.com/user/reference/osx/).

Thanks for the support and this amazingly helpful package!

saimn:
It seems related to the recent release of wheel (0.32), which removed `open_for_csv`:
https://github.com/pypa/wheel/commit/186ca2c82a7bdcc87b0a5584528507ad404ba790#diff-7dc328f716d014fd1a19755f7d89c4e0

YannickJadoul:
Does the latest version of `delocate` (with the latest version of `wheel`) still work on a local macOS machine?

YannickJadoul:
Update: the latest version on delocate's master seems to be working. I quickly forked the repo and started the tests on Travis: see https://travis-ci.org/YannickJadoul/delocate/builds/435534535.

But the commit that actually contains the fix (https://github.com/matthew-brett/delocate/commit/68355ee937826e0e5db4c75833f165db670e6ba4) seems to not have been released yet? :-(

[...]

PR #107, opened 2018-10-01
Pinning version of wheel to install to 0.31.1 on macOS

YannickJadoul:
See discussion in issue #106

It will depend on when the delocate dependency releases a new version whether we want this merged or not. But for now, this fix can be used by `pip` installing `git+https://github.com/joerick/cibuildwheel.git@wheel-0.31.1` or `https://github.com/joerick/cibuildwheel/archive/wheel-0.31.1.zip`.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -54,6 +54,12 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             mkdir /output
             cd /project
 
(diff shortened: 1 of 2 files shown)


YannickJadoul:
(Sorry for the branch on this repo instead of on my own fork, btw, but I thought that while we're discussing how to progress, it would be nice if people can already pip install a branch from the original repo)

PR #108, opened 2018-10-01
Fixing "AttributeError: 'Popen' object has no attribute 'args'" on Python 2.7 for Linux builds

YannickJadoul:
Silly thing, but Python 2.7's `subprocess.Popen` object does not have an `args` attribute, and so you now get an extra exception instead of a nice overview of the executed command, when the subprocess fails.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -122,15 +122,15 @@ def build(project_dir, package_name, output_dir, test_command, test_requires, be
             if stdin_str is None:
                 subprocess.check_call(['docker'] + command)
             else:
-                process = subprocess.Popen(['docker'] + command,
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Builds are failing because of #106, but héy, at least the error message is now correct! :-D

```
subprocess.CalledProcessError: Command '['/home/circleci/project/venv/bin/python', '-m', 'cibuildwheel', '/home/circleci/project/test/06_docker_images']' returned non-zero exit status 1
```

MuellerSeb:
Could you provide a release including this, like 0.10.1? At the moment I have to install cibuildwheel with

    $PIP install https://github.com/joerick/cibuildwheel/archive/master.zip

to get everything running.

PR #109, opened 2018-10-18
Add apriltags2-ethz library to readme

safijari:
Adding a link to my library that is using cibuildwheel to build linux wheels. Not currently deploying to PyPI but will soon :).

Thank you for this awesome tool.


YannickJadoul:
Merged. Thanks!

PR #110, opened 2018-10-26
Update Python from 3.6.6, 3.7.0 to 3.6.7, 3.7.1 on macOS

mayeut:
many bug fixes including CVE.
c.f. https://docs.python.org/3.6/whatsnew/changelog.html#python-3-6-7-final
c.f. https://docs.python.org/3.7/whatsnew/changelog.html#python-3-7-1-final


YannickJadoul:
Thanks a lot for noticing the updates! This should be good, but it seems something's wrong with the Python 3/macOS build on Travis CI.

Are these new versions also available for Linux and Windows, then?

YannickJadoul:
> Are these new versions also available for Windows, then?

Nevermind. I forgot that these were preinstalled and thus updated by AppVeyor...

YannickJadoul:
@mayeut If you want, have a look at #112, which should fix the build issue, making this PR mergeable :-)

mayeut:
I had a similar issue with python3 on travis-ci a while back. I did not have time to investigate until now (I switched to python2). Your fix in #112 seems good.

YannickJadoul:
@mayeut Yeah, I'm expecting the same issue on my next push to my project, so it was worth figuring out either way...

Let's get this merged, then :-)

PR #111, opened 2018-10-26
Update openssl patch from 1.0.2o to 1.0.2p on macOS

mayeut:
Fixes 2 low security CVE. c.f. https://www.openssl.org/news/vulnerabilities-1.0.2.html

PR #112, opened 2018-10-27
Fixing Travic CI build with HomeBrew upgrade of gcc

YannickJadoul:


YannickJadoul:
@joerick & @mayeut Could you have a look at this? The current build is broken (see #110 or #111), because of something in the combination between Travis CI and HomeBrew (see https://github.com/travis-ci/travis-ci/issues/8826), resulting in an error when upgrading `python`:

```
==> Pouring gcc-8.2.0.high_sierra.bottle.1.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink include/c++
Target /usr/local/include/c++
already exists. You may want to remove it:
  rm '/usr/local/include/c++'
To force the link and overwrite all conflicting files:
  brew link --overwrite gcc
To list all files that would be deleted:
  brew link --overwrite --dry-run gcc
Possible conflicting files are:
/usr/local/include/c++ -> /usr/local/Caskroom/oclint/0.13.1,17.4.0/oclint-0.13.1/include/c++
```

This PR fixes the issue by removing oclint (as suggested in the Travis CI issue).

The other option would be to either remove the link before, or to add `|| brew link --overwrite gcc` to the end of the upgrading line. If you prefer that one, I can adapt the PR to that.

I guess we should merge this soon enough, because any PR made will now result in a build failure?

PR #113, opened 2018-11-05
Refactoring: remove package_name

### LINKED ISSUE #102: Refactoring: remove package_name
I noticed while looking at the code today that the `package_name` property isn't actually used anywhere. We should probably remove it, unless anyone has particular designs for it?

Also, a few of the test cases have stuff like 

```
if sys.argv[-1] != '--name':
```

which could be removed at the same time

xanderyzwich:
Removals made per comments in the issue
Closes #102

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -109,8 +109,6 @@ def main():
         project_setup_py = os.path.join(project_dir, 'setup.py')
         name_output = subprocess.check_output([sys.executable, project_setup_py, '--name'],
                                               universal_newlines=True)
-        # the last line of output is the name
-        package_name = name_output.strip().splitlines()[-1]
     except subprocess.CalledProcessError as err:
         if not os.path.exists(project_setup_py):
             print('cibuildwheel: Could not find setup.py at root of project', file=sys.stderr)
@@ -121,14 +119,8 @@ def main():
                   file=sys.stderr)
             exit(err.returncode)
 
-    if package_name == '' or package_name == 'UNKNOWN':
-        print('cibuildwheel: Invalid package name "%s". Check your setup.py' % package_name,
-              file=sys.stderr)
-        exit(2)
-
     build_options = dict(
         project_dir=project_dir,
-        package_name=package_name,
         output_dir=output_dir,
         test_command=test_command,
         test_requires=test_requires,
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -105,6 +105,26 @@ def main():
     # This needs to be passed on to the docker container in linux.py
     os.environ['CIBUILDWHEEL'] = '1'
 
+    try:
+        project_setup_py = os.path.join(project_dir, 'setup.py')
+        name_output = subprocess.check_output([sys.executable, project_setup_py, '--name'],
+                                              universal_newlines=True)
+        # the last line of output is the name
+        package_name = name_output.strip().splitlines()[-1]
+    except subprocess.CalledProcessError as err:
+        if not os.path.exists(project_setup_py):
+            print('cibuildwheel: Could not find setup.py at root of project', file=sys.stderr)
+            exit(2)
+        else:
+            print(err.output)
+            print('cibuildwheel: Failed to get name of the package. Command was %s' % err.cmd,
+                  file=sys.stderr)
+            exit(err.returncode)
+     if package_name == '' or package_name == 'UNKNOWN':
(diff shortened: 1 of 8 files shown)


cibuildwheel/__main__.py
```diff
@@ -107,10 +107,8 @@ def main():
 
     try:
```
YannickJadoul: I believe this whole try-except block can go. The only reason for having it was to get the project's name, but since we're removing it, I don't believe it's still necessary. And it will probably solve the build failure on the CI services.

cibuildwheel/linux.py
```diff
@@ -136,7 +135,6 @@ def run_docker(command, stdin_str=None):
         try:
             run_docker(['create',
                         '--env', 'CIBUILDWHEEL',
-                        '--name', container_name,
```
YannickJadoul: This has nothing to do with the `package_name` variable, so I wouldn't remove it

test/06_docker_images/setup.py
```diff
@@ -2,12 +2,6 @@
 
 from setuptools import setup, Extension
 
-if sys.argv[-1] != '--name':
```
YannickJadoul: Same here as above.

test/03_before_build/setup.py
```diff
@@ -1,25 +1,6 @@
 from setuptools import setup, Extension
 import sys, os
 
-if sys.argv[-1] != '--name':
```
YannickJadoul: We'd still like to keep the tests, here, actually. We just don't need to guard it with the `if sys.argv[-1] != '--name':`. So please keep the body of the if-statement without the condition (but dedented, to be valid Python)

test/04_build_skip/setup.py
```diff
@@ -1,13 +1,6 @@
 from setuptools import setup, Extension
 import sys
 
-if sys.argv[-1] != '--name':
```
YannickJadoul: Same here. We dó want these tests, but since you've removed the call with the `--name` argument, the condition and if-statement can go, keeping the body.

test/05_environment/setup.py
```diff
@@ -1,22 +1,6 @@
 from setuptools import setup, Extension
 import sys, os
 
-if sys.argv[-1] != '--name':
```
YannickJadoul: Same here as the two other test `setup.py` changes above.

cibuildwheel/__main__.py
```diff
@@ -105,30 +105,8 @@ def main():
...
     os.environ['CIBUILDWHEEL'] = '1'
 
-    try:
-        project_setup_py = os.path.join(project_dir, 'setup.py')
-        name_output = subprocess.check_output([sys.executable, project_setup_py, '--name'],
-                                              universal_newlines=True)
-        # the last line of output is the name
-        package_name = name_output.strip().splitlines()[-1]
-    except subprocess.CalledProcessError as err:
-        if not os.path.exists(project_setup_py):
```
YannickJadoul: I'm thinking. While the whole of the try-except statement is not useful anymore, maybe this `if not os.path.exists(project_setup_py): ...` statement is useful? Failing early and giving the user a readable error message might be a good idea here, in case the project directory argument is wrong and doesn't contain a setup.py.

So I would suggest to actually keep those three lines and keep the rest deleted? @joerick, any thoughts?

(Sorry for changing my mind/not being more clear in the previous review :-/ )

YannickJadoul:
Thanks a lot, @xanderyzwich!

I have a few more remarks to clear up things that should (hopefully) also help fix the final build failures on Travis CI and CircleCI :-)

cibuildwheel/__main__.py
joerick: yeah, i'd agree, the test for setup.py is still useful, since there's no point in going to the trouble to e.g. downloading the docker images if there's no setup.py. So it would be good to keep that check.

YannickJadoul:
Yey, getting close :-)
Let's have a quick look to see why the tests are now failing.

soekah:
No

YannickJadoul:
> No

@soekah What's a 'No'?

YannickJadoul:
Due to an unrelated error, the Python 3 test on macOS now fails. However, I've just fixed this in commit 481b1cb060b23388d20d904c86789e6042b88cd4

Apart from that, all changes have been discussed with @joerick. So since the final diff is quite concise and since I'm happy tests are now succeeding, I'm going to merge this! :-D

Thanks, @xanderyzwich!

joerick:
thanks, both!

PR #114, opened 2018-11-29
Sorting build options dict items when printing preamble

YannickJadoul:
Silly thing, but I was very confused while looking at the tests, seeing that the build options are were always in a different order (for example [here](https://ci.appveyor.com/project/joerick/cibuildwheel/builds/20661534/job/3iec2iin8gcwfr87#L1176) vs. [here](https://ci.appveyor.com/project/joerick/cibuildwheel/builds/20661534/job/3iec2iin8gcwfr87#L1609)) and this really annoys me, somehow.

The quickest solution I could think of was sorting the `build_options.items()` when printing. If this is not worth it or if there's a better order, do please tell me.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -180,7 +180,7 @@ def print_preamble(platform, build_options):
 
     print('Build options:')
     print('  platform: %r' % platform)
-    for option, value in build_options.items():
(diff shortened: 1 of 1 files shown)


joerick:
👍 looks good

PR #118, opened 2019-01-05
Update openssl patch from 1.0.2p to 1.0.2q on macOS

mayeut:
Fixes 2 low security CVE. c.f. https://www.openssl.org/news/vulnerabilities-1.0.2.html


joerick:
Thanks @mayeut for maintaining this patch!

Issue #122, opened 2019-01-23
MacOS travis-ci build stalls


josh146:
I noticed today that our MacOS travis-CI build using cibuildwheel has started stalling at the following point of the cibuildwheel setup:

```bash
+ pip install --upgrade setuptools
Collecting setuptools
  Downloading https://files.pythonhosted.org/packages/37/06/754589caf971b0d2d48f151c2586f62902d93dc908e2fd9b9b9f6aa3c9dd/setuptools-40.6.3-py2.py3-none-any.whl (573kB)
Installing collected packages: setuptools
  Found existing installation: setuptools 28.8.0
    Uninstalling setuptools-28.8.0:

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

This hasn't affected our Windows/Linux builds, and no changes to our devops pipeline has occurred (and even then, only superficial Python commits in our codebase were committed). 

This issue happens no matter how many times we restart the build, and seems odd - this step is usually instantaneous on previous Mac cibuildwheel builds.

Since this is a command that is called by cibuildwheel, has there been a recent change that breaks this step?

YannickJadoul:
I can reproduce this on my build as well: https://travis-ci.org/YannickJadoul/Parselmouth/jobs/483834720, but I have no clue what's going on.
Let's run a branch with verbosity flags for this command? Unless anyone has better suggestions?

YannickJadoul:
Weird, Travis builds of cibuildwheel fail in the exact same way, and `-vvv` does not provide any more insight: https://travis-ci.org/YannickJadoul/cibuildwheel/jobs/483882146#L3949

YannickJadoul:
The problem seems to be with Python 3.4, though, because the Python 2.7 does work. Is this the same for you, @josh146?

YannickJadoul:
Installing `pip<19` seems to work, so I'm guessing it's a problem with pip and Python 3.4: https://github.com/YannickJadoul/cibuildwheel/commit/f876231da4adeca96636411da067fc234733a6f5 and https://travis-ci.org/YannickJadoul/cibuildwheel/builds/483890015

Not sure what we can do about it though. This fix seems very ad-hoc and ugly?

This seems to be the same/a similar issue: https://github.com/pypa/pip/issues/6169

[...]

PR #123, opened 2019-01-30
Running 'pip install --upgrade setuptools' with 'sudo' on macOS, fixing #122

YannickJadoul:
Seems the fix to `pip` to #122 might still take some time to get released, and is not guaranteed to completely fix our problem. Adding `--user` to all `pip install` commands on macOS should be a harmless (?) fix to get `cibuildwheel` working again, though.

The other option would be to add `sudo` to the `pip install --upgrade setuptools` call. If you prefer that, I can change this PR.


joerick:
Great! This ready to go @YannickJadoul ?

PR #126, opened 2019-02-22
WIP: Support Windows on Azure Pipelines

astrofrog:
Currently cibuildwheel doesn't work out of the box on Windows in Azure Pipelines - mainly because the Python executables aren't there by default. By adding:

```yaml

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '2.7'
      architecture: 'x86'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '2.7'
      architecture: 'x64'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.4'
      architecture: 'x86'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.4'
      architecture: 'x64'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.5'
      architecture: 'x86'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.5'
      architecture: 'x64'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.6'
      architecture: 'x86'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.6'
      architecture: 'x64'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.7'
      architecture: 'x86'

  - task: UsePythonVersion@0
    inputs:
      versionSpec: '3.7'
      architecture: 'x64'
```

to the config, you can set up all the required Python versions on a single machine. However, they are at a different location so this PR will try and detect if we are on Azure and adjust the paths.

However, this isn't quite enough - I'm getting errors about the fact the the Microsoft Visual C++ compilers are missing when trying to build packages with C extensions. I'm currently stuck at that point, so if anyone has any ideas, let me know! :)

**Update:** I can get Python 2.7 wheels to build by adding the following to the Azure config:

```
  - script: choco install vcpython27 -f -y
    displayName: Install Visual C
```

This installs the Microsoft Visual C++ 9.0 compiler required by Python 2.7. In addition, Python 3.5, 3.6, and 3.7 work fine because it seems that the 14.0 compiler is available by default. However, I can't get Python 3.4 to work.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -25,17 +25,45 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
         return subprocess.check_call(' '.join(args), env=env, cwd=cwd)
 
     PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'path'])
+
+    # At this point, we need to check if we are running on Azure, because if
+    # so Python is not located in the usual place. We recognize Azure by
+    # checking for a C:\hostedtoolcache directory - there aren't any nice
+    # environment variables we can use as on some other CI frameworks.
+
+    if os.path.exists('C:\hostedtoolcache'):
+
+        # We can't hard-code the paths because on Azure, we don't know which
+        # bugfix release of Python we are getting so we need to check which
+        # ones exist. We just use the first one that is found since there should
+        # only be one.
+
+        def python_path(version, arch):
+            major, minor = version.split('.')[:2]
+            suffix = 'x86' if arch == '32' else 'x64'
+            path = glob("C:\\hostedtoolcache\\windows\\Python\\" + version.replace('x', '*') + "\\" + suffix)[0]
+            return path
+
+    else:
+
+        def python_path(version, arch):
+            major, minor = version.split('.')[:2]
+            path = 'C:\\Python' + major + minor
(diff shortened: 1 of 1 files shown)


YannickJadoul:
@astrofrog If it's any consolation, Python 3.4 is almost not supported anymore. Numpy already [dropped support in 1.16.0](https://docs.scipy.org/doc/numpy/release.html#numpy-1-16-0-release-notes), and will reach end-of-life status in less than a month: https://devguide.python.org/#status-of-python-branches.

So maybe it's not necessary to get that fixed? ;-)

astrofrog:
That's true! Just in case though, here's an issue I opened for the azure python task: https://github.com/Microsoft/azure-pipelines-tasks/issues/9674

joerick:
Oh, before I forget, we should also add a config to run our tests on Azure, too, so we can be sure it's still working once we announce support.

astrofrog:
Ok so just an update, it seems that it'll be difficult to support Python 3.4 so maybe indeed worth just dropping support for that (or never having it on Azure): https://github.com/Microsoft/azure-pipelines-tasks/issues/9674#issuecomment-468815193

I'll look into the other issues.

joerick:
Closed in favour of #132

Issue #127, opened 2019-03-10
OSX Permission error when installing dependencies


djhoese:
See https://travis-ci.org/pytroll/aggdraw/jobs/504187990#L106 for an example. Using cibuildwheel 0.10.1

```
$ source ci/travis-install.sh
Collecting cibuildwheel
  Downloading https://files.pythonhosted.org/packages/dd/db/8c652a74782377db4ba40e8868f6084a431d8cbe65670d4d7e1b55e37e95/cibuildwheel-0.10.1.tar.gz
Collecting bashlex (from cibuildwheel)
  Downloading https://files.pythonhosted.org/packages/a2/c6/4f647fce8138b9b7cf948ee65ed1cef958f75a8c81903d801bbbcc6ec741/bashlex-0.13.tar.gz (60kB)
    100% |████████████████████████████████| 61kB 11.0MB/s 
Collecting enum34 (from bashlex->cibuildwheel)
  Downloading https://files.pythonhosted.org/packages/c5/db/e56e6b4bbac7c4a06de1c50de6fe1ef3810018ae11732a50f15f62c7d050/enum34-1.1.6-py2-none-any.whl
Building wheels for collected packages: cibuildwheel, bashlex
  Running setup.py bdist_wheel for cibuildwheel ... done
  Stored in directory: /Users/travis/Library/Caches/pip/wheels/81/1d/e5/035783bf1e755a03aa30ed46471afc8ab00d4963d917cac82b
  Running setup.py bdist_wheel for bashlex ... done
  Stored in directory: /Users/travis/Library/Caches/pip/wheels/5c/1c/b5/fcccabef06e01d6b958cbfec6995c755e8168c02c91e6b226c
Successfully built cibuildwheel bashlex
Installing collected packages: enum34, bashlex, cibuildwheel
Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/LICENSE'
Consider using the `--user` option or check the permissions.
```

Any ideas?

joerick:
I think some changes to Pip mean this has to be `sudo pip install` now.  Let me know if that fixes your issue and I can update the docs. 

Does that should right @yannickjadoul?

YannickJadoul:
@djhoese @joerick Adding `sudo` could indeed be a solution, but I'm not entirely sure what's wrong.

[...]

PR #129, opened 2019-03-10
Revert temporary fix of #122

### LINKED ISSUE #122: MacOS travis-ci build stalls

YannickJadoul:
While I was looking at #127, I realized the temporary fix to #122 (which turned out to be a similar but unrelated issue) is not necessary anymore now that `pip` has been patched, fixing the issue.

If @joerick would decide to release a new version soon to include #128, then this might as well be part of it, too?

This PR would (finally!) close #122.

And, oh, yes: I've based it on #128, to get the tests green, so I'd suggest to first merge #128, and then I could rebase this PR?

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -86,8 +86,7 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
         # install pip & wheel
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         call(['pip', '--version'], env=env)
-        # sudo required, because the removal of the old version of setuptools might cause problems with newer pip versions (see issue #122)
-        call(['sudo', 'pip', 'install', '--upgrade', 'setuptools'], env=env)
+        call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
         call(['pip', 'install', 'wheel'], env=env)
         call(['pip', 'install', 'delocate'], env=env)
 


joerick:
Nice one. No need to rebase, I'll squash and merge, once the tests are green. 👍

YannickJadoul:
Oh, yeah, for the record, and future debugging and decisions: i realized that if Homebrew is correctly installed (and it seems it is, on Travis CI), permissions allow to install write to the Homebrew prefix without `sudo` and we should normally never be forced to use `sudo` (except as workarounds around other problems).

PR #137, opened 2019-04-22
Custom test checks alternate

joerick:
Building on #136 - main differences are 

- environment.json is rolled into each test script
- the checks are more specific to the exact wheels produced
- tests are run using pytest for nice output when the assertions fail


YannickJadoul:
Cool! Using pytest is a great improvement indeed!

I'm still not 100% convinced of getting rid of `environment.json` and making the `cibuildwheel` call to be added manually per test. I liked the test script doing all that and enforcing thhat framework. But that's also up to taste, ofc, and maybe I'm overlooking tests we'd like to write?

I'll have a fuller look once I'm behind my laptop screen, but do close #136 if you are happy with this PR :-)

test/01_basic/cibuildwheel_test.py
```diff
@@ -0,0 +1,50 @@
+import subprocess, sys, os
+from glob import glob
+project_dir = os.path.dirname(__file__)
+sys.path.append(os.path.join(os.path.dirname(project_dir), 'shared'))
```
YannickJadoul: Could we set this in the test driver, since it's (by definition) shared by all tests anyway? That's one less line to copy when adding a new test/one less line to change in every test file if there's a bug.

Adding (or appending to) the `PYTHONPATH` environment variable could work, here?

test/03_before_build/cibuildwheel_test.py
```diff
@@ -0,0 +1,22 @@
...
+sys.path.append(os.path.join(os.path.dirname(project_dir), 'shared'))
+
+import utils
+
+def test():
+    # set up the environment
+    env = os.environ.copy()
+    # write python version information to a temporary file, this is checked
+    # in setup.py
+    env['CIBW_BEFORE_BUILD'] = "python -c \"import sys; open('/tmp/pythonversion.txt', 'w').write(sys.version)\" && python -c \"import sys; open('/tmp/pythonexecutable.txt', 'w').write(sys.executable)\""
```
YannickJadoul: `'''python -c "import sys; open('/tmp/pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('/tmp/pythonexecutable.txt', 'w').write(sys.executable)"'''` saves some backslashes

test/03_before_build/version.txt
```diff
@@ -1,2 +0,0 @@
-3.6.0 (default, Feb  7 2017, 23:55:32)
```
YannickJadoul: What was this file doing? I don't see this in the new tests, anymore?

test/06_docker_images/cibuildwheel_test.py
```diff
@@ -0,0 +1,24 @@
+import subprocess, sys, os, pytest
+from glob import glob
+project_dir = os.path.dirname(__file__)
+sys.path.append(os.path.join(os.path.dirname(project_dir), 'shared'))
+import utils
+
+def test():
+    if not sys.platform.startswith('linux'):
```
YannickJadoul: `utils.platform` again?

test/01_basic/cibuildwheel_test.py
joerick: Hm, yes I agree. PYTHONPATH is a good idea! I'll look at changing that.

test/03_before_build/cibuildwheel_test.py
joerick: nice.

test/03_before_build/version.txt
joerick: I don't know what it was doing there! It must have been committed by mistake 😳

test/06_docker_images/cibuildwheel_test.py
joerick: yes good point

joerick:
thanks for the review, all good points.

> I'm still not 100% convinced of getting rid of environment.json and making the cibuildwheel call to be added manually per test. I liked the test script doing all that and enforcing thhat framework. But that's also up to taste, ofc, and maybe I'm overlooking tests we'd like to write?

On the test framework enforcing a methodology - I think it's a good point, but I'm not sure it's worth the complexity of splitting the test setup and check across two different files - much nicer to have cause and effect right there in one file IMO. But I agree it's a taste thing - I also wasn't keen on the 'environment as a JSON file' thing - never felt all that pleasant to edit JSON files 🤷‍♂️

test/01_basic/cibuildwheel_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+import utils
+
+def test():
+    # build the wheels
+    subprocess.check_call([sys.executable, '-m', 'cibuildwheel', project_dir]) 
+    
+    # check that every wheel is produced
+    if utils.platform == 'linux':
+        assert glob('wheelhouse/*-cp27-cp27m-manylinux1_x86_64.whl')
```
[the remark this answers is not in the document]
joerick: > What about adding all filenames to utils (possibly in a function that takes project and version parameters, and then generates the filename),

ooh.... i'm maybe... 50/50 on this. I like the explicitness of how it's currently written, it's very clear what's being looked for and the logic is fully within the test. But a set comparison might be neater and we could run it for every test, not just these two. I might try an implementation and see how it feels

bin/run_test.py
```diff
@@ -1,31 +1,27 @@
...
+    # set up an environment that gives access to the test utils
     env = os.environ.copy()
-    project_env = {str(k): str(v) for k, v in project_env.items()} # unicode not allowed in env
-    env.update(project_env)
-    print('Building %s with environment %s' % (test_project, project_env))
-    subprocess.check_call([sys.executable, '-m', 'cibuildwheel', test_project], env=env)
-    wheels = glob('wheelhouse/*.whl')
-    print('%s built successfully. %i wheels built.' % (test_project, len(wheels)))
+    env.update({
+        'PYTHONPATH': test_utils_dir,
```
YannickJadoul: Maybe, just to be sure:
```
pythonpath = test_utils_dir.get('PYTHONPATH', '')
if pythonpath:
    pythonpath += os.pathsep
env.update({
    'PYTHONPATH': pythonpath + test_utils_dir
})
```

(Or something like that; I'm not completely happy with that code)

test/01_basic/cibuildwheel_test.py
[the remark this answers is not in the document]
YannickJadoul: Fair enough; just an idea. Indeed, if this works for the basic test and the tests on skipping, that should be enough, and the other tests can focus on something else. I was also just thinking of getting rid of the `--print-build-identifiers` dependency in the test, since there could be errors in there as well?

test/01_basic/cibuildwheel_test.py
```diff
@@ -0,0 +1,49 @@
+import subprocess, sys, os
+from glob import glob
+import utils
+
+def test():
+    project_dir = os.path.dirname(__file__)
+
+    # build the wheels
+    subprocess.check_call([sys.executable, '-m', 'cibuildwheel', project_dir])
```
YannickJadoul: Minor idea: what about moving this into `utils` as well? `utils.call_cibuildwheel()` or so? This gets rid of `import subprocess, sys, os` as well, and then the test has even more focus on the things we're checking.

YannickJadoul: (OK, I now see you'd need an `env` parameter, or `add_env`, with a dictionary of things to update)

test/06_docker_images/cibuildwheel_test.py
```diff
@@ -0,0 +1,26 @@
+import subprocess, sys, os, pytest
```
YannickJadoul: Now I'm really nitpicking, so just ignore if you want, but wouldn't the filename `test.py` be a lot shorter and less work to type?

YannickJadoul:
Great! I've closed #136, because I think this updated version clearly has advantages, so let's continue on this :-)

test/01_basic/cibuildwheel_test.py
joerick: yes! I was just now working on something very similar....

image

test/06_docker_images/cibuildwheel_test.py
joerick: Projects like 02_test have a test _folder_ that contains unit tests for the package, so I wanted to have a longer name that wouldn't be confused for tests that are specific to the package.

bin/run_test.py
joerick: good point. It's a surprisingly annoying bit of code to write, I settled on

```python
    if 'PYTHONPATH' in env:
        env['PYTHONPATH'] += os.pathsep + test_utils_dir
    else:
        env['PYTHONPATH'] = test_utils_dir
```

YannickJadoul: Anything's good, there, indeed :-) I was writing a oneliner, to then think that that's not the kind of code to commit to a project ;-)

test/01_basic/cibuildwheel_test.py
YannickJadoul: Yesss, great :-)

For me this takes away the burden that `cibuildwheel` isn't called from `run_test.py`, if it's just a single line to write anywa.

test/06_docker_images/cibuildwheel_test.py
YannickJadoul: I hadn't thought of that indeed. I just thought `test_cibuildwheel.py` was so long, and curious why you didn't go for short

joerick:
This might now be ready, if it looks good to you @YannickJadoul ?

YannickJadoul:
@joerick Sorry for the delay, but as far as I can see, it's great! :-) And it's not very vital functionality for releases, so if we want to tweak further or find inconveniences, we can always update the test framework?

I'll rebase #135 onto this, once it gets merged

PR #142, opened 2019-05-29
Specify cibuildwheel=0.11 in Azure example 

kngwyu:
Thanks for your great work :slightly_smiling_face: 

Looks `0.10.1` doesn't work on Azure.


joerick:
Hey, thanks for this! Looks like the bumpversion tool got out of sync when I added the extra configs. I'll have to check that.

PR #150, opened 2019-07-06
Document that requirements.txt can be used in CIBW_TEST_REQUIRES

jbarlow83:


YannickJadoul:
Two things:
- In general, is this intentional, or is this something that might break in later versions?
- In particular, wouldn't it be more appropriate to insert this into the example values of `CIBW_TEST_REQUIRES` (e.g., https://github.com/joerick/cibuildwheel/pull/150/files#diff-04c6e90faac2675aa89e2176d2eec7d8R410) rather than into the `.yml` example?

jbarlow83:
I have a feeling it was not intentional, but I do think this is useful behavior that should be retained. The arguments are just forwarded to `pip install` so it's harder to suppress this than retain it.

One already has to repeat dependencies half a dozen times in different formats for a widely distributed project (`requirements.txt`, `setup.py`, .spec/.rpm/.deb control files for Linux, Homebrew scripts...); the ability to reuse any of them is greatly appreciated.

YannickJadoul:
Yes, agreed! :-) Just wanted to make sure this was meant to be supported or just something that happened to work.

It seems to be somewhere in between. Because currently, this will stop working when the requirements file has a space in its name. I think changing [this `.split()`](https://github.com/joerick/cibuildwheel/blob/ab86b8cd93d38786bcf59d279936c98ae1f4d23a/cibuildwheel/__main__.py#L93) to `shlex.split()` and [this `' '.join()`](https://github.com/joerick/cibuildwheel/blob/ab86b8cd93d38786bcf59d279936c98ae1f4d23a/cibuildwheel/linux.py#L109) to `' '.join(map(shlex.quote, test_requires))` should probably fix that, though. Do you want to add this to your PR, or do you prefer that I create a new one for that?

README.md
```diff
@@ -426,6 +426,11 @@ Platform-specific variants also available:
     CIBW_TEST_REQUIRES: nose
     CIBW_TEST_COMMAND: "nosetests {project}\\tests"
 </code></pre></td>
+<td><i>example using pytest requirements.txt</i><pre><code>environment:
```
YannickJadoul: Move to the "Environment variable: `CIBW_TEST_REQUIRES`" section a few lines above, rather than adding it here as an example of "YML syntax"?

joerick:
This is clever, but I'm not sure I agree with how this changes the meaning of the option `CIBW_TEST_REQUIRES` - it really puts the pip shell syntax front-and-center. I'd be happier with this as an undocumented 'feature' - but feel free to use it yourself!

YannickJadoul:
@joerick You can also use this kind of `-r sub_requirements.txt` in a `requirements.txt` file. So in that way, I don't think it breaks the abstraction too much.

> It seems to be somewhere in between. Because currently, this will stop working when the requirements file has a space in its name. I think changing this .split() to shlex.split() and this ' '.join() to ' '.join(map(shlex.quote, test_requires)) should probably fix that, though. Do you want to add this to your PR, or do you prefer that I create a new one for that?

Should we still make a PR for this, though? That'd be slightly more robust at any rate?

PR #152, opened 2019-07-23
Remove example of using yum install

jbarlow83:
Since CentOS 5 is 2.5 years old past of end of life, it's not recommend to use its binaries as they may have unpatched vulnerabilities. It's better to build any dependencies not on the whitelist from source.

The manylinux2010 image should be fine in this respect.


joerick:
Thanks, but this'll be fine once we get manylinux2010 in!

PR #155, opened 2019-08-14
Adding support for building manylinux2010 wheels - alternative

YannickJadoul:
This is a bit of a (cheeky/daring?) alternative proposal for supporting `manylinux2010` and following versions ([`manylinux2014` has also been approved](https://www.python.org/dev/peps/pep-0599/), recently).

Advantages:
- The new options `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE` provide a nice generalization and should be future-proof. Rather than adding new environment variables every time a new `manylinux` standard is released, these new options encompass both the need to switch between different `manylinux` version as well as use custom images.
- We also don't need to keep on adding new Python-platform combinations to the `python_configurations` list.
- "Building for `manylinux`" and "choosing the version of `manylinux`/using a custom image" are made independent, now.

Disadvantages:
- Two breaking changes: `CIBW_MANYLINUX1_X86_64_IMAGE` and `CIBW_MANYLINUX1_I686_IMAGE` are not used anymore, and the `*-manylinux1-x86_64` and `*-manylinux1-i686` selectors in `CIBW_BUILD` and `CIBW_SKIP` will not work anymore. I think we can check for these, and warn the user (or even just fail) when an existing configuration still uses these, though.

The other option I see would to be just add a bunch of entries to `python_configurations`, associate a `manylinux` version with each of them, and create a `CIBW_MANYLINUX_VERSION`. I feel this would create a lot of repetition, though (`CIBW_MANYLINUX*-X86_64/I686` environment variables, and `cp??-manylinux*_x86_64/i686` build tags.

That being said, I'm not convinced this proposal is the perfect solution either, so I'd be grateful for some feedback (@joerick, @jbarlow83, @mayeut ?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,8 +221,8 @@ All being well, you should get wheels delivered to you in a few minutes.
 | **Build parameters** | `CIBW_BUILD_VERBOSITY` | Increase or decrease the output of `pip wheel` |
 | **Build environment** | `CIBW_ENVIRONMENT` | Set environment variables needed during the build |
 |   | `CIBW_BEFORE_BUILD` | Execute a shell command preparing each wheel's build |
-|   | `CIBW_MANYLINUX1_X86_64_IMAGE` | Specify an alternative manylinx1 x86_64 docker image |
-|   | `CIBW_MANYLINUX1_I686_IMAGE` | Specify an alternative manylinux1 i686 docker image |
+|   | `CIBW_MANYLINUX_X86_64_IMAGE` | Specify an alternative manylinux x86_64 docker image |
+|   | `CIBW_MANYLINUX_I686_IMAGE` | Specify an alternative manylinux i686 docker image |
 | **Tests** | `CIBW_TEST_COMMAND` | Execute a shell command to test all built wheels |
 |   | `CIBW_TEST_REQUIRES` | Install Python dependencies before running the tests |
 |   | `CIBW_TEST_EXTRAS` | Install Python dependencies before running the tests using ``extras_require``|
@@ -231,10 +231,10 @@ A more detailed description of the options, the allowed values, and some example
 
 ### Linux builds on Docker
 
-Linux wheels are built in the [`manylinux1` docker images](https://github.com/pypa/manylinux) to provide binary compatible wheels on Linux, according to [PEP 513](https://www.python.org/dev/peps/pep-0513/). Because of this, when building with `cibuildwheel` on Linux, a few things should be taken into account:
+Linux wheels are built in the [`manylinux` docker images](https://github.com/pypa/manylinux) to provide binary compatible wheels on Linux, according to [PEP 571](https://www.python.org/dev/peps/pep-0571/). Because of this, when building with `cibuildwheel` on Linux, a few things should be taken into account:
 - Programs and libraries cannot be installed on the Travis CI Ubuntu host with `apt-get`, but can be installed inside of the Docker image using `yum` or manually. The same goes for environment variables that are potentially needed to customize the wheel building. `cibuildwheel` supports this by providing the `CIBW_ENVIRONMENT` and `CIBW_BEFORE_BUILD` options to setup the build environment inside the running Docker image. See [below](#options) for details on these options.
 - The project directory is mounted in the running Docker instance as `/project`, the output directory for the wheels as `/output`. In general, this is handled transparently by `cibuildwheel`. For a more finegrained level of control however, the root of the host file system is mounted as `/host`, allowing for example to access shared files, caches, etc. on the host file system.  Note that this is not available on CircleCI due to their Docker policies.
-- Alternative dockers images can be specified with the `CIBW_MANYLINUX1_X86_64_IMAGE` and `CIBW_MANYLINUX1_I686_IMAGE` options to allow for a custom, preconfigured build environment for the Linux builds. See [below](#options) for more details.
+- Alternative dockers images can be specified with the `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE` options to allow for a custom, preconfigured build environment for the Linux builds. See [below](#options) for more details.
 
 
 Options
@@ -291,17 +291,19 @@ For `linux` you need Docker running, on Mac or Linux. For `macos`, you need a Ma
 
 Optional.
 
-Space-separated list of builds to build and skip. Each build has an identifier like `cp27-manylinux1_x86_64` or `cp34-macosx_10_6_intel` - you can list specific ones to build and `cibuildwheel` will only build those, and/or list ones to skip and `cibuildwheel` won't try to build them.
+Space-separated list of builds to build and skip. Each build has an identifier like `cp27-manylinux_x86_64` or `cp34-macosx_10_6_intel` - you can list specific ones to build and `cibuildwheel` will only build those, and/or list ones to skip and `cibuildwheel` won't try to build them.
 
 When both options are specified, both conditions are applied and only builds with a tag that matches `CIBW_BUILD` and does not match `CIBW_SKIP` will be built.
 
-The format is `python_tag-platform_tag`. The tags are as defined in [PEP 0425](https://www.python.org/dev/peps/pep-0425/#details).
+The format is `python_tag-platform_tag`. The tags are similar but not identical to the ones defined in [PEP 425](https://www.python.org/dev/peps/pep-0425/#details).
 
 Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37`
 
-Platform tags look like `macosx_10_6_intel` `manylinux1_x86_64` `manylinux1_i686` `win32` `win_amd64`
+Platform tags look like `macosx_10_6_intel` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
 
-You can also use shell-style globbing syntax (as per `fnmatch`) 
+You can also use shell-style globbing syntax (as per `fnmatch`).
+
+The list of supported and currently selected build identifiers can be retrieved by passing the `--print-build-identifiers` flag to `cibuildwheel`.
 
 Examples:
 - Only build on Python 3.6: `CIBW_BUILD`:`cp36-*`
@@ -311,7 +313,7 @@ Examples:
 - Skip Python 2.7 on 32-bit Windows: `CIBW_SKIP`:`cp27-win32`
 - Skip Python 3.4 and Python 3.5: `CIBW_SKIP`:`cp34-* cp35-*`
 - Skip Python 3.6 on Linux: `CIBW_SKIP`:`cp36-manylinux*`
-- Only build on Python 3 and skip 32-bit builds: `CIBW_BUILD`:`cp3?-*` and `CIBW_SKIP`:`*-win32 *-manylinux1_i686`
+- Only build on Python 3 and skip 32-bit builds: `CIBW_BUILD`:`cp3?-*` and `CIBW_SKIP`:`*-win32 *-manylinux_i686`
 
 ***
 
@@ -370,15 +372,20 @@ Platform-specific variants also available:
 
 ***
 
-| Environment variables: `CIBW_MANYLINUX1_X86_64_IMAGE` and `CIBW_MANYLINUX1_I686_IMAGE`
+| Environment variables: `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE`
 | ---
 
 Optional.
 
-An alternative docker image to be used for building [`manylinux1`](https://github.com/pypa/manylinux) wheels. `cibuildwheel` will then pull these instead of the official images, [`quay.io/pypa/manylinux1_x86_64`](https://quay.io/pypa/manylinux1_i686) and [`quay.io/pypa/manylinux1_i686`](https://quay.io/pypa/manylinux1_i686).
+An alternative Docker image to be used for building [`manylinux`](https://github.com/pypa/manylinux) wheels. `cibuildwheel` will then pull these instead of the default images, [`quay.io/pypa/manylinux2010_x86_64`](https://quay.io/pypa/manylinux2010_x86_64) and [`quay.io/pypa/manylinux2010_i686`](https://quay.io/pypa/manylinux2010_i686).
 
-Beware to specify a valid docker image that can be used the same as the official, default docker images: all necessary Python and pip versions need to be present in `/opt/python/`, and the `auditwheel` tool needs to be present for `cibuildwheel` to work. Apart from that, the architecture and relevant shared system libraries need to be manylinux1-compatible in order to produce valid `manylinux1` wheels (see https://github.com/pypa/manylinux and [PEP 513](https://www.python.org/dev/peps/pep-0513/) for more details).
+The value of this option can either be set to `manylinux1` or `manylinux2010` to use the [official `manylinux` images](https://github.com/pypa/manylinux), or any other valid Docker image name.
 
+Beware to specify a valid Docker image that can be used in the same way as the official, default Docker images: all necessary Python and pip versions need to be present in `/opt/python/`, and the `auditwheel` tool needs to be present for `cibuildwheel` to work. Apart from that, the architecture and relevant shared system libraries need to be manylinux1- or manylinux2010-compatible in order to produce valid `manylinux1`/`manylinux2010` wheels (see https://github.com/pypa/manylinux, [PEP 513](https://www.python.org/dev/peps/pep-0513/), and [PEP 571](https://www.python.org/dev/peps/pep-0571/) for more details).
+
+Note that `auditwheel` detects the version of the `manylinux` standard in the Docker image through the `AUDITWHEEL_PLAT` environment variable, as `cibuildwheel` has no way of detecting the correct `--plat` command line argument to pass to `auditwheel` for a custom image. If a Docker image does not correctly set this `AUDITWHEEL_PLAT` environment variable, the `CIBW_ENVIRONMENT` option can be used to do so (e.g., `CIBW_ENVIRONMENT="manylinux2010_$(uname -m)"`).
+
+Example: `manylinux1`  
 Example: `dockcross/manylinux-x64`  
 Example: `dockcross/manylinux-x86`
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -146,7 +146,7 @@ def main():
         # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
 
         manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', 'manylinux2010')
-        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
+        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', 'manylinux1')
 
         default_manylinux_images_x86_64 = {'manylinux1': 'quay.io/pypa/manylinux1_x86_64',
                                            'manylinux2010': 'quay.io/pypa/manylinux2010_x86_64'}
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -141,11 +141,19 @@ def main():
     )
 
     if platform == 'linux':
-        manylinux1_x86_64_image = os.environ.get('CIBW_MANYLINUX1_X86_64_IMAGE', None)
-        manylinux1_i686_image = os.environ.get('CIBW_MANYLINUX1_I686_IMAGE', None)
+        # TODO Check CIBW_MANYLINUX1_X86_64_IMAGE and CIBW_MANYLINUX1_I686_IMAGE environment variables and error if they exist
+        # TODO Check CIBW_BUILD and CIBW_SKIP for "manylinux1" and error if present
+        # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
+        manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', '2010')
+        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
+
+        default_manylinux_images_x86_64 = {'1': 'quay.io/pypa/manylinux1_x86_64',
+                                           '2010': 'quay.io/pypa/manylinux2010_x86_64'}
+        default_manylinux_images_i686 = {'1': 'quay.io/pypa/manylinux1_i686'}
 
         build_options.update(
-            manylinux1_images={'x86_64': manylinux1_x86_64_image, 'i686': manylinux1_i686_image},
+            manylinux_images={'x86_64': default_manylinux_images_x86_64.get(manylinux_x86_64_image) or manylinux_x86_64_image,
+                              'i686': default_manylinux_images_i686.get(manylinux_i686_image) or manylinux_i686_image},
         )
     elif platform == 'macos':
         pass
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -54,6 +54,8 @@ def main():
 
     args = parser.parse_args()
 
+    detect_obsolete_options()
+
     if args.platform != 'auto':
         platform = args.platform
     else:
@@ -141,11 +143,7 @@ def main():
     )
 
     if platform == 'linux':
-        # TODO Check CIBW_MANYLINUX1_X86_64_IMAGE and CIBW_MANYLINUX1_I686_IMAGE environment variables and error if they exist
-        # TODO Check CIBW_BUILD and CIBW_SKIP for "manylinux1" and error if present
         # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
-        manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', '2010')
-        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
 
         manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', 'manylinux2010')
         manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
@@ -179,6 +177,26 @@ def main():
         cibuildwheel.macos.build(**build_options)
     else:
         raise Exception('Unsupported platform')
+
+
+def detect_obsolete_options():
+    # Check the old 'MANYLINUX1_*_IMAGE' options
+    for (deprecated, alternative) in [('CIBW_MANYLINUX1_X86_64_IMAGE', 'CIBW_MANYLINUX_X86_64_IMAGE'),
+                                      ('CIBW_MANYLINUX1_I686_IMAGE', 'CIBW_MANYLINUX_I686_IMAGE')]:
+        if deprecated in os.environ:
+            print("'{}' has been deprecated, and will be removed in a future release. Use the option '{}' instead.".format(deprecated, alternative))
+            if alternative not in os.environ:
+                print("Using value of option '{}' as replacement for '{}'".format(deprecated, alternative))
+                os.environ[alternative] = os.environ[deprecated]
+            else:
+                print("Option '{}' is not empty. Please unset '{}'".format(alternative, deprecated))
+                exit(2)
+
+    # Check for 'manylinux1' in the 'CIBW_BUILD' and 'CIBW_SKIP' options
+    for deprecated in ['CIBW_BUILD', 'CIBW_SKIP']:
+        if deprecated in os.environ and 'manylinux1' in os.environ[deprecated]:
+            print("Build identifiers with 'manylinux1' been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
+            os.environ[deprecated] = os.environ[deprecated].replace('manylinux1', 'manylinux')
 
 
 def print_preamble(platform, build_options):
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -147,9 +147,12 @@ def main():
         manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', '2010')
         manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
 
-        default_manylinux_images_x86_64 = {'1': 'quay.io/pypa/manylinux1_x86_64',
-                                           '2010': 'quay.io/pypa/manylinux2010_x86_64'}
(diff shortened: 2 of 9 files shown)


YannickJadoul:
*Cross-referencing #135*

YannickJadoul:
Grmbl. Of course tests are failing.

Reason: the name of the created wheels depends on the code, rather than on the `cibuildwheel` configuration... :-|

jbarlow83:
manylinux2014 supports a lot of new platforms, so avoiding the Python configurations is the way to go. I don't think the breaking changes are a big deal - @joerick's instructions to pin the version of cibuildwheel used in configuration will help here.

Taking a step back a bit, I think concern about producing manylinux2010 wheels when we could produce manylinux1 is a bit overblown. manylinux1 is technically "manylinux2007"; we're worried about 12 year old vs 9 year old systems, both already rare in production. What would be rarer still is a system from that era in production that can install manylinuxX wheels. As far as I can tell pip couldn't install wheels until 2013, and I don't think it could install manylinux1 wheels until 2015. I suspect it would be very difficult to get a 2007 or 2010 distribution to the point where it could install manylinux wheels, and it may be well easier to build necessary packages from source. For example, Ubuntu 14.04 can't install wheels without PPAs. The biggest practical concern by far is that older versions of pip (<18) that can't install manylinux2010.

joerick:
Okay. Fantastic! This is much better in my eyes! Here we have the default behaviour of building using the 2010 image (which will sometimes produce manylinux1 as a nice side effect for backward-compat), but users still have the freedom to build using manylinux1 image (a more advanced use-case imo) if they like. And we didn't need to add any more options to do it :+1:

Three things are occurring to me-
- do we need to add an option to assert manylinux1 was produced when building on 2010? (Actually, maybe an explicit section of the docs that talks about manylinux1, and shows an assert done in the bash shell after the cibuildwheel run would be better)
- the CIBW_BUILD/CIBW_SKIP identifiers are really diverging from pip's now. We should include a list of them in the docs.
- breaking backward compat on the option. Yes, this should be okay, as @jbarlow83 says, I've always tried to encourage version pinning in cibuildwheel (always a good idea for repeatable builds), so it shouldn't be too big a problem. Maybe there's something clever we can do to rewrite the option? `if 'manylinux1' in cibw_skip: cibw_skip.replace('manylinux1', 'manylinux'); print_warning()`? 

But thank you @YannickJadoul for continuing to push this forward!

YannickJadoul:
@jbarlow83 Thanks for that! I hadn't actually realized the chronology of the manylinux1 being so much older than the PEP and its implementation in pip. So I agree this is a good enough default! 

> The biggest practical concern by far is that older versions of pip (<18) that can't install manylinux2010.

Indeed. And the main problem is probably that older `pip` versions cannot even detect the problem and warn the user to upgrade pip to install the wheels. Luckily, we could hopefully expect the average Linux user to be recently up-to-date or at least tech-savvy enough when maintaining an old system?

YannickJadoul:
@joerick Great, good to hear you like this approach. Let's continue on this track then and hope I haven't overlooked any future problems :-)

> do we need to add an option to assert manylinux1 was produced when building on 2010?

I don't know about this one. Is this going to be a common scenario, that users will want `manylinux1` wheels even though they're building on `manylinux2010`? @jbarlow83 just argued that manylinux2010 is probably widely-compatible enough, by now. Then again, adding something to the docs is easy enough, and can only help to clarify further.

> the CIBW_BUILD/CIBW_SKIP identifiers are really diverging from pip's now. We should include a list of them in the docs.

Yes, this also worried me for a second, but then I realized this was already the case before for some other identifiers (I can't remember which ones, though) But, for example, the ABI tags `cpXXm` are already not present either).

> Maybe there's something clever we can do to rewrite the option? `if 'manylinux1' in cibw_skip: cibw_skip.replace('manylinux1', 'manylinux'); print_warning()`?

I like this idea! This will probably fix it, most of the time :-)
On a second thought, it might also be safer to just fail and have the user fix the options. As you said: if a user didn't want updates, the version should have been pinned.

One more thing: what do we do with the multiple wheels coming from cibuildwheel? I agree with [the argument that every `manylinux1` wheel is also a `manylinux2010` wheel](https://github.com/pypa/manylinux/issues/179#issuecomment-520523730), but how confusing is that for the user? It makes our tests quite annoying, especially because we now don't have a 32-bit version, so we still don't get the same wheels as when building on `manylinux1` images before.

I'll try to get this a bit closer to finished tonight. Now that we almost know how we want things, it shouldn't be too hard to finish!

joerick:
> One more thing: what do we do with the multiple wheels coming from cibuildwheel?

Offhand opinion- I think we just hand them both back to the user. They can upload both if they like. manylinux2010 wheels give the impression of a modern, secure build toolchain. I'd prefer to install a 2010 over a 1 given the choice!

YannickJadoul:
> They can upload both if they like. manylinux2010 wheels give the impression of a modern, secure build toolchain. I'd prefer to install a 2010 over a 1 given the choice!

Even though they are both exactly the same, except for name and metadata? :-D But true, that's also why I'm in doubt on what the best thing to do is...

jbarlow83:
I'd think it's probably better for the packagers to provide all the wheels they can and let `pip` worry about what to install.

YannickJadoul:
Finally, progress! Tests are succeeding, for the first time :-)

Sorry for all the force-push spam. Can I actually test this locally without polluting my own system too much, @joerick?

joerick:
Great! I should get a chance to review in the next day or two
> Sorry for all the force-push spam. Can I actually test this locally without polluting my own system too much, @joerick?
> 
Yes, if you have Docker running locally you can run individual tests with `bin/dev_run_test test/00_test_name`

YannickJadoul:
> Great! I should get a chance to review in the next day or two

I/we still need to:
- Handle old `CIBW_MANYLINUX1_..._IMAGE` options and `manylinux1` tags
- Add tests
- Add documentation

But I'm a bit stuck on (cleanly) getting the first one done.
I'll add the test from #135 later or tomorrow, and I'll make a list of things notes?

> Yes, if you have Docker running locally you can run individual tests with `bin/dev_run_test test/00_test_name`

Ah yes, now that it's Linux-related, I can just do that, ofc. The problem was before on macOS or Windows PRs :-D

YannickJadoul:
OK, ready for a first review, @joerick :-) And maybe @jbarlow83, @mayeut, if you two are interested? I'd love some more feedback on this.

A few concerns/things I'm not sure how to handle:
- I'm not a bash expert, so there's probably a better way to handle the multiple wheels in the bash array. I just put a for loop around `chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"`, for now, but any suggestions are welcome. Maybe this should become Python as well, at some point?
- The `AUDITWHEEL_PLAT` solution is less clean than I envisioned, mainly because we only have a single `CIBW_ENVIRONMENT` variable, but there are two different `--platform` options for `auditwheel` (`auditwheel*_x86_64` and `auditwheel*_i686`), so you can't just say `CIBW_ENVIRONMENT="AUDITWHEEL_PLAT=manylinux1"`. See `test/06_docker_images`.
    - The dockcross images do not have `AUDITWHEEL_PLAT` set by default, that's why we need to set them manually. They also contain `auditwheel` 2.0, and `AUDITWHEEL_PLAT` is only supported in 2.1, so that's why the update is there.
- Should we maybe turn these `'1'` and `'2010'` strings for default options into `'manylinux1'` and `'manylinux2010'`. More writing but also more clarity? I couldn't make up my mind.
- In `__main__.py`, I couldn't make up my mind how to organize the warnings/replacement of deprectated `manylinux1` tags and options. There's the nice `detect_warnings` function, but by then it's too late to still adapt the new `build_options` (since it's already printed in the preamble, for example)? Maybe you have a great idea on the organization, there, @joerick?

cibuildwheel/__main__.py
```diff
@@ -136,11 +136,19 @@ def main():
     )
 
     if platform == 'linux':
-        manylinux1_x86_64_image = os.environ.get('CIBW_MANYLINUX1_X86_64_IMAGE', None)
-        manylinux1_i686_image = os.environ.get('CIBW_MANYLINUX1_I686_IMAGE', None)
+        # TODO Check CIBW_MANYLINUX1_X86_64_IMAGE and CIBW_MANYLINUX1_I686_IMAGE environment variables and error if they exist
+        # TODO Check CIBW_BUILD and CIBW_SKIP for "manylinux1" and error if present
+        # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
```
joerick: I agree - there need to be breaking changes in this version, so let's detect obsolete options and error here, directing people to the relevant part of the changelog.

cibuildwheel/__main__.py
```diff
@@ -136,11 +136,19 @@ def main():
...
+        manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', '2010')
+        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
+
+        default_manylinux_images_x86_64 = {'1': 'quay.io/pypa/manylinux1_x86_64',
+                                           '2010': 'quay.io/pypa/manylinux2010_x86_64'}
+        default_manylinux_images_i686 = {'1': 'quay.io/pypa/manylinux1_i686'}
```
joerick: `CIBW_MANYLINUX_X86_64_IMAGE=1` looks like you're enabling a boolean option, or something numeric, but it's an identifier, so let's use `CIBW_MANYLINUX_X86_64_IMAGE=manylinux1` (and `manylinux2010`).

cibuildwheel/linux.py
```diff
@@ -79,14 +81,14 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
...
                 else
-                    auditwheel repair "$built_wheel" -w /tmp/delocated_wheel
+                    auditwheel repair "$built_wheel" -w /tmp/delocated_wheels
                 fi
-                delocated_wheel=(/tmp/delocated_wheel/*.whl)
+                delocated_wheels=(/tmp/delocated_wheels/*.whl)
 
                 # Install the wheel we just built
-                "$PYBIN/pip" install "$delocated_wheel"
+                "$PYBIN/pip" install "${{delocated_wheels[0]}}"
```
joerick: arg, this is tricksy... do we need to predefine a precedence here? Although, if auditwheel produces two wheels, they're the same. If that's true, we just need to comment it here, something like `# if auditwheel produced two wheels, it's because its output conforms to multiple manylinux standards. They're functionally the same, so it doesn't matter which one we run the tests on.`

test/08_manylinux2010_only/cibuildwheel_test.py
```diff
@@ -0,0 +1,20 @@
+import os, pytest
+import utils
+
+def test():
+    project_dir = os.path.dirname(__file__)
+
+    if utils.platform != 'linux':
+        pytest.skip('the docker test is only relevant to the linux build')
```
joerick: should probably be `this test is only relevant to the linux build`

test/shared/utils.py
```diff
@@ -42,25 +42,39 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
     )
 
 
-def expected_wheels(package_name, package_version):
+def expected_wheels(package_name, package_version, manylinux_versions={'1_x86_64', '2010_x86_64'}):
```
joerick: This `manylinux_versions` param... why is the default x86_64 only? I'd say the default would be to expect all wheels. 

I'm wondering if perhaps a nicer signature would be like-
`def expected_wheels(package_name, package_version, i686=True, x86_64=True, manylinux1=True)`

Then tests can deselect the parts they don't need. (Just a suggestion!)

cibuildwheel/linux.py
```diff
@@ -101,8 +103,8 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
                 fi
 
                 # we're all done here; move it to output
-                mv "$delocated_wheel" /output
-                chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"
+                mv "${{delocated_wheels[@]}}" /output
+                for delocated_wheel in "${{delocated_wheels[@]}}"; do chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"; done
```
joerick: I think `chown` will accept multiple files, but with the basename in there, the loop is probably the right approach. Although perhaps we could just do the chown before moving to output, then the basename wouldn't be needed.

YannickJadoul:
Great! I'll start implementing and completing the changes!

> I don't totally understand the AUDITWHEEL_PLAT note, because I only see that referenced in the dockcross test - seems fine to me, if it's a quirk of this old image?

Yeah, I guess so. I just thought: if someone uses a different image than the standard manylinux images, we can't detect what to pass to `auditwheel` as `--plat` option, but it's easy to use `CIBW_ENVIRONMENT="AUDITWHEEL_PLAT=manylinux2010"` (for example). However, that will not work, because `AUDITWHEEL_PLAT` also needs the `_x86_64` and  `_i686` suffixes, so a simple `CIBW_ENVIRONMENT` addition will never really work.

> > In **main**.py, I couldn't make up my mind how to organize the warnings/replacement of deprectated manylinux1 tags and options.
> 
> I suppose the `detect_warnings` approach isn't right since we don't even want to read the old obsolete options anymore. So maybe a `detect_obsolete_options` function that accesses os.environ, prints errors and exits the program? That could come before options parsing.

Oh, you don't want to try and check/use the old options (cfr. your `if 'manylinux1' in cibw_skip: cibw_skip.replace('manylinux1', 'manylinux'); print_warning()` idea) ?

cibuildwheel/__main__.py
YannickJadoul: Yes, ok, good point. Something started feeling lazy about using `'1'` and `'2010'`, so I agree!

cibuildwheel/linux.py
YannickJadoul: That's what I was thinking indeed.
They should be the same. Only the `pip` detection/installation is influenced by this, I guess. But we're testing on the same platform as we're building, so in this regard, it's not a good test anyway.

test/shared/utils.py
YannickJadoul: Interesting, myeah. I wasn't fully happy with the signature either, but:

> why is the default x86_64 only?

Because currently there is only a `manylinux2010_i686` standard, but not an image :-( See [this reply](https://github.com/pypa/manylinux/issues/179#issuecomment-520523082). So, since the general consensus seems to be to move away from `manylinux1`, I made `manylinux2010_x86_64` the default image for 64-bit, and *nothing* the default image for 32-bit.
Happy to change opinion, though, if you disagree. Or maybe we should make a PR with `manylinux2010_i686` to make our code cleaner? :-P

> I'm wondering if perhaps a nicer signature would be like-
`def expected_wheels(package_name, package_version, i686=True, x86_64=True, manylinux1=True)`

Problem is that I was trying to be future-proof as well. A PEP on `manylinux2014` has already been approved, I believe, so I was hoping to limit the places where we'd need to copy-paste-replace2010by2014. But your signature is nicer, so I'd like to find a way in between!

cibuildwheel/linux.py
YannickJadoul: Yeah, it's quite annoying.

I thought there was maybe a reason you first `mv`ed, then `chown`ed, though? Something that there's a chance you can't move the files anymore if they're not yours (although I guess we are root in the docker image). But then there was no reason for this order?

Anyway, no strong opinion here, so I can put whatever you prefer.

YannickJadoul:
One more thing I just thought of. Do we want to allow multiple manylinux images? E.g., `CIBW_MANYLINUX_X86_64_IMAGE="manylinux2010 dockcross/manylinux1-x64"`.

I'm wondering if there might be reasons to build for multiple targets? For example: https://github.com/pypa/manylinux/issues/179#issuecomment-524203791

joerick:
> One more thing I just thought of. Do we want to allow multiple manylinux images? E.g., CIBW_MANYLINUX_X86_64_IMAGE="manylinux2010 dockcross/manylinux1-x64".

Curious how this might play into the build identifier system that cibuildwheel uses - if both those images produce a manylinux1 wheel, which one wins? perhaps the auditwheel `--plat` option resolves that? (question - does specifying `--plat` mean that only that wheel is produced?)

I'm concerned that maybe having such an option as well as the CIBW_BUILD/CIBW_SKIP would get confusing... at least I'm a little confused right now!

YannickJadoul:
> I'm concerned that maybe having such an option as well as the CIBW_BUILD/CIBW_SKIP would get confusing... at least I'm a little confused right now!

True, yes. The way I was approaching it now was that `CIBW_BUILD`/`CIBW_SKIP` determine if you want to build CPython wheels for `manylinux` (later we will hopefully get PyPy!), and `CIBW_MANYLINUX_..._IMAGE` determines which image(s?) you are building on.

But maybe those are worries for later?

I'm just a bit concerned to not break backwards-compatibility again for `manylinux2014` or PyPy (or whatever else might come). So before merging or releasing, it might be a good idea to see how easy it is to add PyPy into this framework?

YannickJadoul:
Back to you, @joerick. Getting closer to 'done'?

I'm not entirely happy with the `detect_obsolete_options`, but it'll do, I guess. Should we add tests for this, somehow? Or is a thorough code review good enough for the handling of obsolete options?

I still need to:
- Update the docs (well, README) on the new options and usage.
- Fix some feedback, after we've reached a conclusion on what to do (see replies).
- Try how easy it would be to add PyPy manylinux, into this new design (?)

cibuildwheel/__main__.py
```diff
@@ -136,11 +138,18 @@ def main():
     )
 
     if platform == 'linux':
-        manylinux1_x86_64_image = os.environ.get('CIBW_MANYLINUX1_X86_64_IMAGE', None)
-        manylinux1_i686_image = os.environ.get('CIBW_MANYLINUX1_I686_IMAGE', None)
+        # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
+
+        manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', 'manylinux2010')
+        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', '')
```
joerick: hmmm.... not sure how I feel about this... this means that CIBW_BUILD can specify a i686 wheel, but it won't get built because there's no image specified. I think that's a bit confusing. CIBW_BUILD should specify which wheels are built.

I'd prefer this to specify the old manylinux1, and not allow this option to be empty.

cibuildwheel/linux.py
```diff
@@ -42,11 +42,13 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
...
     platforms = [
-        ('manylinux1_x86_64', manylinux1_images.get('x86_64') or 'quay.io/pypa/manylinux1_x86_64'),
-        ('manylinux1_i686', manylinux1_images.get('i686') or 'quay.io/pypa/manylinux1_i686'),
+        ('manylinux_x86_64', manylinux_images['x86_64']),
+        ('manylinux_i686', manylinux_images['i686']),
     ]
 
     for platform_tag, docker_image in platforms:
+        if not docker_image:
+            continue
```
joerick: See above comment on the empty `manylinux_i686_image` setting. Builds should be skipped using CIBW_SKIP, not with empty image settings.

cibuildwheel/__main__.py
```diff
@@ -136,11 +138,18 @@ def main():
     )
 
     if platform == 'linux':
-        manylinux1_x86_64_image = os.environ.get('CIBW_MANYLINUX1_X86_64_IMAGE', None)
-        manylinux1_i686_image = os.environ.get('CIBW_MANYLINUX1_I686_IMAGE', None)
+        # TODO Add documentation on CIBW_ENVIRONMENT and possibility of AUDITWHEEL_PLAT if not defined by custom manylinux image
```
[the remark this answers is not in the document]
YannickJadoul: Yes. But I was waiting for that until we had decided on the everything, before writing docs ;-)

cibuildwheel/__main__.py
YannickJadoul: That's a good point. I hadn't realized that, actually. :-/
The problem is that there was a strong suggestion in pypa/manylinux#179 to discourage people from building on the `manylinux1` image, given the potential security issues in the old CentOS 5 image. And if not now, at some point `manylinux1` will be deprecated (and a newer 32-bit docker image might not become available).

Any thoughts on that, @joerick? Should we make `CIBW_SKIP="manylinux-i686"` the default? That doesn't feel too right either, somehow :-/

jbarlow83: On my library I decided to skip 32-bit Linux because <1% of Linux machines are i686. I figured it was not worth the energy and CO2 to have a server build binaries no one will ever use. Two years ago, I decided if just one person asked for them, I would make it available. 

Last year, someone opened an issue that amounted to failing to compile 32-bit from source because there was no manylinux1 wheel. I explained this, and the person said they selected a 32-bit virtual machine by mistake. They closed the issue. To date, no one has ever wanted i686. This library has on the order of 10k users including some enterprise ones.

I think x64 only is a reasonable default for Linux. (32-bit Windows is another matter entirely - there are still lots of 32-bit users.) But maybe it should be an explicit opt-in?

joerick: Thanks for the perspective @jbarlow83! I wasn't aware of how rare that variant was. I might try to get some stats out of PyPI to back that up. 

As a mechanism to deselect manylinux-i686 by default, it should be something that doesn't surprise the user. The trouble with a default on CIBW_SKIP is that  setting that value (e.g. to skip a win32 build) would accidentally enable i686. 

I'm wondering about a default value for CIBW_BUILD - `*-manylinux_x86_64 *-win32 *-macosx_10_6_intel ...`, since I'm guessing CIBW_BUILD is less commonly used than CIBW_SKIP, and when it is used, it's specifically targeting something.

joerick:
I've done a little research on the i686 question - whether to deselect building manylinux-i686 by default. I've been using the [BigQuery](https://console.cloud.google.com/bigquery) interface to get some stats... it's kinda fun 

Here's the query I've come up with:

```sql
SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(file.filename, r"([^-]*)\.whl$") as platform_tag
WHERE
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY platform_tag
ORDER BY num_downloads DESC
```

Row | num_downloads | platform_tag |  
-- | -- | -- | --
1 | 2288918055 | any |  
2 | 706319759 | null [source tarball] |  
3 | 266871186 | manylinux1_x86_64 |  
4 | 17764422 | win_amd64 |  
5 | 10038858 | win32 |  
6 | 4955484 | manylinux1_i686 |  
7 | 3541978 | macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64 |  
8 | 3040174 | macosx_10_6_intel |  
9 | 2110448 | manylinux2010_x86_64 |  
10 | 2070123 | macosx_10_9_x86_64

I'm guessing these stats are dominated by build-boxes/CI systems, but we see that manylinux_x86_64 by far the biggest wheel. Here, i686 comes in at around 2% of the largest platform. It appears quite high up this list, but there are a large range of macos tags - if they were added up, macos would be higher.

Given the surprising dominance of manylinux1_x86_64 here, I also wanted to try my own package - pyinstrument, which is a developer tool, not a library:

```sql
SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(file.filename, r"([^-]*)\.whl$") as platform_tag
WHERE
   file.project = 'pyinstrument-cext' AND
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY platform_tag
ORDER BY num_downloads DESC
```

Row | num_downloads | platform_tag |  
-- | -- | -- | --
1 | 13456 | manylinux1_x86_64 |  
2 | 2056 | win_amd64 |  
3 | 1867 | win32 |  
4 | 1734 | macosx_10_6_intel |  
5 | 1618 | manylinux1_i686 |  
6 | 637 | null

In this example, i686 comes in at 12% of the largest. This was a surprising result for me! Especially how much windows and how little macOS.

But, in light of these numbers, I don't think we can drop i686 yet.

cibuildwheel/__main__.py
joerick: See https://github.com/joerick/cibuildwheel/pull/155#issuecomment-529647114 - I don't think we can drop i686 yet!

YannickJadoul:
@joerick Thanks for those numbers! I always got discouraged when trying to figure out how to get them :-)

Now that we have these, I'm thinking we should maybe go back to the discussion at pypa/manylinux#179, and particularly @takluyver's comments there (https://github.com/pypa/manylinux/issues/179#issuecomment-520523082).

There is a considerate amount of concern over the security of wheels build on the `manylinux1` image, so if we need to provide the 32-bit wheel, I would argue we would need the `manylinux2010_i686` image to be the default?

YannickJadoul:
(Rebased to server merge conflicts after #161)

joerick:
> I would argue we would need the manylinux2010_i686 image to be the default

My feeling is that we continue to build i686 on manylinux1, since for now, that's the best option for i686 linux wheels.

YannickJadoul:
OK, I'm not entirely convinced yet, but your final call. I'll make the changes, and push them (hopefully) soon :-)

YannickJadoul:
OK, there we go. If this builds and tests pass, and if everyone's happy, I'll add some docs and try adding PyPy in a different PR. And then we should be done!!

joerick:
All looks great to me! Let's get some docs in and call it done :)))

cibuildwheel/__main__.py
```diff
@@ -141,11 +143,18 @@ def main():
...
+
+        manylinux_x86_64_image = os.environ.get('CIBW_MANYLINUX_X86_64_IMAGE', 'manylinux2010')
+        manylinux_i686_image = os.environ.get('CIBW_MANYLINUX_I686_IMAGE', 'manylinux1')
+
+        default_manylinux_images_x86_64 = {'manylinux1': 'quay.io/pypa/manylinux1_x86_64',
+                                           'manylinux2010': 'quay.io/pypa/manylinux2010_x86_64'}
+        default_manylinux_images_i686 = {'manylinux1': 'quay.io/pypa/manylinux1_i686'}
```
mayeut: A manylinux2010_i686 now exists.

YannickJadoul: Done!

YannickJadoul:
(Sorry, @mayeut; I wanted to get rid of the red icon showing "Changes requested", but I have no clue how to do so, so maybe I've been spamming your inbox with all kinds of GitHub mails, now :smile: )

YannickJadoul:
If these CI builds succeed (and they should), I think this PR could (finally) be finished :-)
Sorry it took so long! In the end, writing the docs went quicker than I thought it would go.

@joerick Do you want to first merge this, or first merge your new docs? Because one of the two PRs will need to transfer the new/changed docs from the README to the new-style docs :-/

README.md
```diff
@@ -370,15 +370,20 @@ Platform-specific variants also available:
...
 ***
 
-| Environment variables: `CIBW_MANYLINUX1_X86_64_IMAGE` and `CIBW_MANYLINUX1_I686_IMAGE`
+| Environment variables: `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE`
 | ---
 
 Optional.
 
-An alternative docker image to be used for building [`manylinux1`](https://github.com/pypa/manylinux) wheels. `cibuildwheel` will then pull these instead of the official images, [`quay.io/pypa/manylinux1_x86_64`](https://quay.io/pypa/manylinux1_i686) and [`quay.io/pypa/manylinux1_i686`](https://quay.io/pypa/manylinux1_i686).
+An alternative Docker image to be used for building [`manylinux`](https://github.com/pypa/manylinux) wheels. `cibuildwheel` will then pull these instead of the default images, [`quay.io/pypa/manylinux2010_x86_64`](https://quay.io/pypa/manylinux2010_i686) and [`quay.io/pypa/manylinux2010_i686`](https://quay.io/pypa/manylinux2010_i686).
```
mayeut: link for `manylinux2010_x86_64` refers to `manylinux2010_i686`: 
```[`quay.io/pypa/manylinux2010_x86_64`](https://quay.io/pypa/manylinux2010_i686)``` should be ```[`quay.io/pypa/manylinux2010_x86_64`](https://quay.io/pypa/manylinux2010_x86_64)```

Comment (YannickJadoul) [reply README.md]: Argh, good catch!

File: README.md [hunk 12]
```diff
@@ -221,8 +221,8 @@ All being well, you should get wheels delivered to you in a few minutes.
 | **Build parameters** | `CIBW_BUILD_VERBOSITY` | Increase or decrease the output of `pip wheel` |
 | **Build environment** | `CIBW_ENVIRONMENT` | Set environment variables needed during the build |
 |   | `CIBW_BEFORE_BUILD` | Execute a shell command preparing each wheel's build |
-|   | `CIBW_MANYLINUX1_X86_64_IMAGE` | Specify an alternative manylinx1 x86_64 docker image |
-|   | `CIBW_MANYLINUX1_I686_IMAGE` | Specify an alternative manylinux1 i686 docker image |
+|   | `CIBW_MANYLINUX_X86_64_IMAGE` | Specify an alternative manylinx x86_64 docker image |
```
merwok:
```suggestion
|   | `CIBW_MANYLINUX_X86_64_IMAGE` | Specify an alternative manylinux x86_64 docker image |
```

YannickJadoul: Arggh, yes. Thanks, well spotted :)

README.md
```diff
@@ -291,15 +291,15 @@ For `linux` you need Docker running, on Mac or Linux. For `macos`, you need a Ma
 
 Optional.
 
-Space-separated list of builds to build and skip. Each build has an identifier like `cp27-manylinux1_x86_64` or `cp34-macosx_10_6_intel` - you can list specific ones to build and `cibuildwheel` will only build those, and/or list ones to skip and `cibuildwheel` won't try to build them.
+Space-separated list of builds to build and skip. Each build has an identifier like `cp27-manylinux_x86_64` or `cp34-macosx_10_6_intel` - you can list specific ones to build and `cibuildwheel` will only build those, and/or list ones to skip and `cibuildwheel` won't try to build them.
 
 When both options are specified, both conditions are applied and only builds with a tag that matches `CIBW_BUILD` and does not match `CIBW_SKIP` will be built.
 
 The format is `python_tag-platform_tag`. The tags are as defined in [PEP 0425](https://www.python.org/dev/peps/pep-0425/#details).
```
```suggestion
The format is `python_tag-platform_tag`.
```

Our tweak from `manylinux2010` to `manylinux` means we've deviated from this spec, so let's remove the link.

YannickJadoul: Ah, yes :-/ That's the downside of this PR, of course :-(

Maybe I'll still add something on using `--print-build-identifiers` to get the list.

YannickJadoul: I tried to rephrase and keep the reference (since it still seems relevant for users of `cibuildwheel` to know about this PEP).

Just a suggestions. If you don't agree and want to completely get rid of the sentence, do just push that change before you merge! :)

joerick:
AAAAAAAaaaaand it's merged! Congrats @YannickJadoul !

YannickJadoul:
WOOHOO! This deserves something of a party! :)

Once this gets released, we can let inform pypa/manylinux#179 as well :)

YannickJadoul:
Thanks @joerick, @mayeut, @jbarlow83, and everyone else who gave input, as well!

Issue #144, opened 2019-06-10
Windows support in Travis CI


cher-nov:
FYI
https://blog.travis-ci.com/2018-10-11-windows-early-release
https://travis-ci.community/t/python-support-on-windows/241

And yes, your project is simply amazing, keep up the good work!

hannesmuehleisen:
FWIW, I managed to get `cibuildwheel` to work on Windows Travis, here the relevant snippet of my `.travis.yml`

````yaml
language: cpp

before_install:
  - eval "${MATRIX_EVAL}"

env:
  global:
    - CIBW_BUILD='cp37-*'
    - CIBW_TEST_REQUIRES='pytest'
    - CIBW_BEFORE_BUILD='pip install "numpy>=1.16" && pip install "pandas>=0.24" && pip install "pytest>=4.3" && pip install cmake==3.13.3'
    - CIBW_TEST_COMMAND='python -m pytest {project}/tests'

matrix:
  include:

    - os: windows
      name: Python Package
      language: cpp
      filter_secrets: false

      before_install:
        - choco install python3 --version 3.7.3 --sidebyside -y --forcex86 --force --params "/InstallDir:C:\Python37"
        - choco install python3 --version 3.7.3 --sidebyside -y --force --params "/InstallDir:C:\Python37-x64"

        - C:/Python37-x64/python.exe -m pip install --upgrade pip
        - C:/Python37-x64/python.exe -m pip install "cibuildwheel==0.10.2"

      script:
        - cd tools/pythonpkg
        - C:/Python37-x64/python.exe  setup.py sdist
        - mkdir duckdb_tarball && tar xvf dist/duckdb-*.tar.gz --strip-components=1 -C duckdb_tarball
        - C:/Python37-x64/python.exe -m cibuildwheel --platform windows --output-dir wheelhouse duckdb_tarball
        - cd ../..
````

cher-nov:
@hannesmuehleisen, I also thought that this was possible, but in any case, some changes are still needed from the side of the **cibuildwheel**, such as the following.
https://github.com/joerick/cibuildwheel/blob/ab86b8cd93d38786bcf59d279936c98ae1f4d23a/cibuildwheel/__main__.py#L66-L67

Czaki:
#160 is merged. So you may close this issue.

cher-nov:
@Czaki https://github.com/joerick/cibuildwheel#usage

[...]

PR #156, opened 2019-08-15
upgrade macos version for support c++11

Czaki:
macos below 10.6 do not support c++11 features. Next python version of pkg is 10.9. To not break backward compatibility I suggest to manually force 10.9 version

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -201,10 +201,10 @@ def detect_obsolete_options():
     # Check for 'manylinux1' in the 'CIBW_BUILD' and 'CIBW_SKIP' options
     for deprecated in ['CIBW_BUILD', 'CIBW_SKIP']:
         if deprecated in os.environ and 'manylinux1' in os.environ[deprecated]:
-            print("Build identifiers with 'manylinux1' been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
+            print("Build identifiers with 'manylinux1' have been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
             os.environ[deprecated] = os.environ[deprecated].replace('manylinux1', 'manylinux')
         if deprecated in os.environ and ("macosx_10_6" in os.environ[deprecated] or "macosx_10_9" in os.environ[deprecated]):
-            print("Build identifiers with 'macosx_10_6' or 'macosx_10_9' been deprecated. Replacing with 'macosx' in the option '{}'".format(deprecated))
+            print("Build identifiers with 'macosx_10_6' or 'macosx_10_9' have been deprecated. Replacing all occurences with 'macosx' in the option '{}'".format(deprecated))
             os.environ[deprecated] = os.environ[deprecated].replace('macosx_10_6', 'macosx').replace('macosx_10_9', 'macosx')
 
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -203,6 +203,9 @@ def detect_obsolete_options():
         if deprecated in os.environ and 'manylinux1' in os.environ[deprecated]:
             print("Build identifiers with 'manylinux1' been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
             os.environ[deprecated] = os.environ[deprecated].replace('manylinux1', 'manylinux')
+        if deprecated in os.environ and ("macosx_10_6" in os.environ[deprecated] or "macosx_10_9" in os.environ[deprecated]):
+            print("Build identifiers with 'macosx_10_6' or 'macosx_10_9' been deprecated. Replacing with 'macosx' in the option '{}'".format(deprecated))
+            os.environ[deprecated] = os.environ[deprecated].replace('macosx_10_6', 'macosx').replace('macosx_10_9', 'macosx')
 
 
 def print_preamble(platform, build_options):
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -14,11 +14,11 @@ from .util import prepare_command, get_build_verbosity_extra_flags
 def get_python_configurations(build_selector):
     PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
     python_configurations = [
-        PythonConfiguration(version='2.7', identifier='cp27-macosx_10_6_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
-        PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
-        PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.6.pkg'),
-        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.8.1/python-3.8.1-macosx10.9.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
+        PythonConfiguration(version='3.5', identifier='cp35-macosx_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.6', identifier='cp36-macosx_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
+        PythonConfiguration(version='3.7', identifier='cp37-macosx_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg')
     ]
 
     # skip builds as required
@@ -96,7 +96,9 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env, cwd="/tmp")
         assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
-        call(['pip', 'install', '--upgrade', 'setuptools', 'wheel', 'delocate'], env=env)
+        call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
+        call(['pip', 'install', 'git+https://github.com/Czaki/wheel.git@1a907d36eb4921efa0c9e7dce6b570aa5f84d7f5'], env=env)
+        call(['pip', 'install', 'delocate'], env=env)
 
         # run the before_build command
         if before_build:
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -97,7 +97,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'git+https://github.com/Czaki/wheel.git@0dd9f3e66893517ec4c9521db523ea3a899a5148'], env=env)
+        call(['pip', 'install', 'git+https://github.com/pypa/wheel.git@882650d0fd2fadd75f4ea7c5d6b62f15f81b398a'], env=env)
         call(['pip', 'install', 'delocate'], env=env)
 
         # run the before_build command
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -96,9 +96,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env, cwd="/tmp")
         assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
-        call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'git+https://github.com/pypa/wheel.git@882650d0fd2fadd75f4ea7c5d6b62f15f81b398a'], env=env)
-        call(['pip', 'install', 'delocate'], env=env)
+        call(['pip', 'install', '--upgrade', 'setuptools', 'wheel', 'delocate'], env=env)
 
         # run the before_build command
         if before_build:
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -97,7 +97,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'git+https://github.com/Czaki/wheel.git@1a907d36eb4921efa0c9e7dce6b570aa5f84d7f5'], env=env)
+        call(['pip', 'install', 'git+https://github.com/Czaki/wheel.git@0dd9f3e66893517ec4c9521db523ea3a899a5148'], env=env)
         call(['pip', 'install', 'delocate'], env=env)
 
         # run the before_build command
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -18,7 +18,7 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.5', identifier='cp35-macosx_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
-        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg')
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
     ]
 
     # skip builds as required
--- a/test/10_cpp_standards/cibuildwheel_test.py
+++ b/test/10_cpp_standards/cibuildwheel_test.py
@@ -6,53 +6,55 @@ import pytest
 import utils
 
 
+project_dir = os.path.dirname(__file__)
+
 def test_cpp11(tmp_path):
-    add_env = {"CIBW_SKIP": "cp27-win*", "CIBW_ENVIRONMENT": "STANDARD=11"}
-    # VC for python 2.7 do not support modern standards
-    if utils.platform == "macos":
-        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
-    project_dir = os.path.dirname(__file__)
-    # this test checks if c++11 standard is supported.
+    # This test checks that the C++11 standard is supported
+
+    add_env = {'CIBW_SKIP': 'cp27-win*', 'CIBW_ENVIRONMENT': 'STANDARD=11'}
+    # VC++ for Python 2.7 does not support modern standards
+    if utils.platform == 'macos':
+        add_env['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
-    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0',
-                       macosx_deployment_target="10.9") if "cp27-cp27m-win" not in x]
+    expected_wheels = [x for x in utils.expected_wheels(
+                            'spam', '0.1.0', macosx_deployment_target='10.9')
+                       if 'cp27-cp27m-win' not in x]
     assert set(actual_wheels) == set(expected_wheels)
 
 
 def test_cpp14():
-    add_env = {"CIBW_SKIP": "cp27-win* cp35-win*", "CIBW_ENVIRONMENT": "STANDARD=14"}
-    # VC for python 2.7 do not support modern standards
-    # manylinux1 docker image do not support compilers with standards newer than c++11
-    # python 3.4 and 3.5 are compiled with MSVC 10. which not support c++14
-    if utils.platform == "macos":
-        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
-    project_dir = os.path.dirname(__file__)
-    # this test checks if c++14 standard is supported.
+    # This test checks that the C++14 standard is supported
+
+    add_env = {'CIBW_SKIP': 'cp27-win* cp35-win*', 'CIBW_ENVIRONMENT': 'STANDARD=14'}
+    # VC++ for Python 2.7 does not support modern standards
+    # The manylinux1 docker image does not have a compiler which supports C++11
+    # Python 3.4 and 3.5 are compiled with MSVC 10, which does not support C++14
+    if utils.platform == 'macos':
+        add_env['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
     expected_wheels = [x for x in utils.expected_wheels(
-                       'spam', '0.1.0', macosx_deployment_target="10.9")
-                       if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
+                            'spam', '0.1.0', macosx_deployment_target='10.9')
+                       if 'cp27-cp27m-win' not in x and 'cp35-cp35m-win' not in x]
     assert set(actual_wheels) == set(expected_wheels)
+
 
 def test_cpp17():
-    # python 2.7 use `register` keyword which is forbidden in c++17 standard 
-    # manylinux1 docker image do not support compilers with standards newer than c++11
-    # python 3.4 and 3.5 are compiled with MSVC 10. which not support c++17
-    if os.environ.get("APPVEYOR_BUILD_WORKER_IMAGE", "") == "Visual Studio 2015":
-        pytest.skip("Visual Studio 2015 does not support c++17")
+    # This test checks that the C++17 standard is supported
 
-    add_env = {"CIBW_SKIP": "cp27-win* cp35-win*", "CIBW_ENVIRONMENT": "STANDARD=17"}
-    if utils.platform == "macos":
-        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.13"
+    # Python 2.7 uses the `register` keyword which is forbidden in the C++17 standard 
+    # The manylinux1 docker image does not have a compiler which supports C++11
+    # Python 3.4 and 3.5 are compiled with MSVC 10, which does not support C++17
+    if os.environ.get('APPVEYOR_BUILD_WORKER_IMAGE', '') == 'Visual Studio 2015':
+        pytest.skip('Visual Studio 2015 does not support C++17')
 
-    project_dir = os.path.dirname(__file__)
-    # this test checks if c++17 standard is supported.
+    add_env = {'CIBW_SKIP': 'cp27-win* cp35-win*', 'CIBW_ENVIRONMENT': 'STANDARD=17'}
+    if utils.platform == 'macos':
+        add_env['MACOSX_DEPLOYMENT_TARGET'] = '10.13'
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
-    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0',
-                       macosx_deployment_target="10.13") 
-                       if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
+    expected_wheels = [x for x in utils.expected_wheels(
+                            'spam', '0.1.0', macosx_deployment_target='10.13')
+                       if 'cp27-cp27m-win' not in x and 'cp35-cp35m-win' not in x]
     assert set(actual_wheels) == set(expected_wheels)
-
--- a/test/10_cpp_standards/cibuildwheel_test.py
+++ b/test/10_cpp_standards/cibuildwheel_test.py
@@ -15,7 +15,8 @@ def test_cpp11(tmp_path):
     # this test checks if c++11 standard is supported.
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
-    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.9") if "cp27-cp27m-win" not in x]
+    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0',
+                       macosx_deployment_target="10.9") if "cp27-cp27m-win" not in x]
     assert set(actual_wheels) == set(expected_wheels)
 
 
@@ -30,7 +31,8 @@ def test_cpp14():
     # this test checks if c++14 standard is supported.
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
-    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.9")
+    expected_wheels = [x for x in utils.expected_wheels(
+                       'spam', '0.1.0', macosx_deployment_target="10.9")
                        if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
     assert set(actual_wheels) == set(expected_wheels)
 
@@ -49,7 +51,8 @@ def test_cpp17():
     # this test checks if c++17 standard is supported.
 
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
-    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.13") 
+    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0',
+                       macosx_deployment_target="10.13") 
                        if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
     assert set(actual_wheels) == set(expected_wheels)
 
--- a/test/10_cpp_standards/cibuildwheel_test.py
+++ b/test/10_cpp_standards/cibuildwheel_test.py
@@ -1,0 +1,55 @@
+import os
+import sys 
+
+import pytest
+
+import utils
+
+
+def test_cpp11(tmp_path):
+    add_env = {"CIBW_SKIP": "cp27-win*", "CIBW_ENVIRONMENT": "STANDARD=11"}
+    # VC for python 2.7 do not support modern standards
+    if utils.platform == "macos":
+        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
+    project_dir = os.path.dirname(__file__)
+    # this test checks if c++11 standard is supported.
+
+    actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
+    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.9") if "cp27-cp27m-win" not in x]
+    assert set(actual_wheels) == set(expected_wheels)
+
+
+def test_cpp14():
+    add_env = {"CIBW_SKIP": "cp27-win* cp35-win*", "CIBW_ENVIRONMENT": "STANDARD=14"}
+    # VC for python 2.7 do not support modern standards
+    # manylinux1 docker image do not support compilers with standards newer than c++11
+    # python 3.4 and 3.5 are compiled with MSVC 10. which not support c++14
+    if utils.platform == "macos":
+        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
+    project_dir = os.path.dirname(__file__)
+    # this test checks if c++14 standard is supported.
+
+    actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
+    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.9")
+                       if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
+    assert set(actual_wheels) == set(expected_wheels)
+
+def test_cpp17():
+    # python 2.7 use `register` keyword which is forbidden in c++17 standard 
+    # manylinux1 docker image do not support compilers with standards newer than c++11
+    # python 3.4 and 3.5 are compiled with MSVC 10. which not support c++17
+    if os.environ.get("APPVEYOR_BUILD_WORKER_IMAGE", "") == "Visual Studio 2015":
+        pytest.skip("Visual Studio 2015 does not support c++17")
+
+    add_env = {"CIBW_SKIP": "cp27-win* cp35-win*", "CIBW_ENVIRONMENT": "STANDARD=17"}
+    if utils.platform == "macos":
+        add_env["MACOSX_DEPLOYMENT_TARGET"] = "10.13"
+
+    project_dir = os.path.dirname(__file__)
+    # this test checks if c++17 standard is supported.
+
+    actual_wheels = utils.cibuildwheel_run(project_dir, add_env=add_env)
+    expected_wheels = [x for x in utils.expected_wheels('spam', '0.1.0', "10.13") 
+                       if "cp27-cp27m-win" not in x and "cp35-cp35m-win" not in x]
+    assert set(actual_wheels) == set(expected_wheels)
+
--- a/test/10_cpp_standards/setup.py
+++ b/test/10_cpp_standards/setup.py
@@ -6,7 +6,7 @@ standard = os.environ["STANDARD"]
 
 language_standard = "/std:c++" + standard if platform.system() == "Windows" else "-std=c++" + standard
 
-extra_compile_args=[language_standard, "-DSTANDARD=" + standard]
+extra_compile_args = [language_standard, "-DSTANDARD=" + standard]
 
 if standard == "17":
     if platform.system() == "Windows":
--- a/test/10_cpp_standards/setup.py
+++ b/test/10_cpp_standards/setup.py
@@ -1,0 +1,21 @@
+import os, sys
+from setuptools import setup, Extension
+import platform
+
+standard = os.environ["STANDARD"]
+
+language_standard = "/std:c++" + standard if platform.system() == "Windows" else "-std=c++" + standard
+
+extra_compile_args=[language_standard, "-DSTANDARD=" + standard]
+
+if standard == "17":
+    if platform.system() == "Windows":
+        extra_compile_args.append("/wd5033")
+    else:
+        extra_compile_args.append("-Wno-register")
+
+setup(
+    name="spam",
+    ext_modules=[Extension('spam', sources=['spam.cpp'], language="c++", extra_compile_args=extra_compile_args)],
+    version="0.1.0",
+)
--- a/test/shared/utils.py
+++ b/test/shared/utils.py
@@ -66,7 +66,8 @@ def cibuildwheel_run(project_path, env=None, add_env=None, output_dir=None):
     return wheels
 
 
-def expected_wheels(package_name, package_version, manylinux_versions=['manylinux1', 'manylinux2010']):
+def expected_wheels(package_name, package_version, manylinux_versions=['manylinux1', 'manylinux2010'],
+                    macosx_deployment_target=None):
     '''
     Returns a list of expected wheels from a run of cibuildwheel.
     '''
@@ -83,20 +84,30 @@ def expected_wheels(package_name, package_version, manylinux_versions=['manylinu
                 platform_tags.append('{manylinux_version}_{architecture}'.format(
                     manylinux_version=manylinux_version, architecture=architecture
                 ))
+            
         def get_platform_tags(python_abi_tag):
             return platform_tags
+        
     elif platform == 'windows':
+
         def get_platform_tags(python_abi_tag):
             return ['win32', 'win_amd64']
+        
     elif platform == 'macos':
+        if macosx_deployment_target is not None:
+            tag = macosx_deployment_target.replace(".", "_")
+            tag1 = macosx_deployment_target.replace(".", "_")
+        else:
+            tag = os.environ.get("MACOSX_DEPLOYMENT_TARGET", "10_6").replace(".", "_")
+            tag1 = os.environ.get("MACOSX_DEPLOYMENT_TARGET", "10_9").replace(".", "_")
+        
         def get_platform_tags(python_abi_tag):
             if python_abi_tag == 'cp38-cp38':
-                return ['macosx_10_9_x86_64']
+                return ['macosx_' + tag1 + '_x86_64']
             else:
-                return ['macosx_10_6_intel']
+                return ['macosx_' + tag + '_intel']
     else:
         raise Exception('unsupported platform')
-
     templates = []
     for python_abi_tag in python_abi_tags:
         for platform_tag in get_platform_tags(python_abi_tag):
--- a/test/shared/utils.py
+++ b/test/shared/utils.py
@@ -53,6 +53,8 @@ def cibuildwheel_run(project_path, env=None, add_env=None, output_dir=None):
     '''
     if env is None:
         env = os.environ.copy()
+        # If present in the host environment, remove the MACOSX_DEPLOYMENT_TARGET for consistency
+        env.pop('MACOSX_DEPLOYMENT_TARGET', None)
 
     if add_env is not None:
         env.update(add_env)
@@ -94,20 +96,15 @@ def expected_wheels(package_name, package_version, manylinux_versions=['manylinu
             return ['win32', 'win_amd64']
         
     elif platform == 'macos':
-        if macosx_deployment_target is not None:
-            tag = macosx_deployment_target.replace(".", "_")
-            tag1 = macosx_deployment_target.replace(".", "_")
-        else:
-            tag = os.environ.get("MACOSX_DEPLOYMENT_TARGET", "10_6").replace(".", "_")
-            tag1 = os.environ.get("MACOSX_DEPLOYMENT_TARGET", "10_9").replace(".", "_")
         
         def get_platform_tags(python_abi_tag):
             if python_abi_tag == 'cp38-cp38':
-                return ['macosx_' + tag1 + '_x86_64']
+                return ['macosx_' + (macosx_deployment_target or "10.9").replace(".", "_") + '_x86_64']
             else:
-                return ['macosx_' + tag + '_intel']
+                return ['macosx_' + (macosx_deployment_target or "10.6").replace(".", "_") + '_intel']
     else:
         raise Exception('unsupported platform')
+
     templates = []
     for python_abi_tag in python_abi_tags:
         for platform_tag in get_platform_tags(python_abi_tag):
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -2,30 +2,32 @@
 title: Modern C++ standards
 ---
 
-Building Python wheels with modern C++ (C++11) requires a few tricks.
+Building Python wheels with modern C++ standards (C++11 and later) requires a few tricks.
 
-Creating wheel for python 2.7 which need c++17 needs special configuration because python 2.7 header files using `register` keyword. 
-In c++17 this keyword is reserved ((see)[https://en.cppreference.com/w/cpp/keyword/register])
-It is possible to allow usage of `register` using proper flag `-Wno-register` for gcc/clang and `/wd5033` for MSVC. 
 
-## Linux
-When using default `manylinux1` image it is possible to use only c++11 and earlier standards. 
-This is about how you can make C++14 wheels in `manylinux1`, through some tricks https://github.com/pypa/manylinux/issues/118
+## Python 2.7 and C++17
 
-`manylinux2010` supports all C++ standards (up to c++17). 
+The Python 2.7 header files use the `register` keyword, which is [reserved and unused from C+17 onwards](https://en.cppreference.com/w/cpp/keyword/register). Compiling a wheel for Python 2.7 with the C++17 standard is still possible to allow usage of `register` using proper flag `-Wno-register` for gcc/clang and `/wd5033` for MSVC. 
 
-## MacOS
+## manylinux1 and C++14
+The default `manylinux1` image (based on CentOS 5) contains a version of GCC and libstdc++ that only supports C++11 and earlier standards. There are however ways to compile wheels with the C++14 standard (and later): https://github.com/pypa/manylinux/issues/118
 
-To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.9"`.
+`manylinux2010` and `manylinux2014` are newer and support all C++ standards (up to C++17).
 
-To get C++17 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.13"` or `"10.14"`. (`"10.13"` supports C++17 partially, e.g. the filesystem header is in experimental: `#include <filesystem>` -> `#include <experimental/filesystem>`)
+## macOS and deployment target versions
 
-For more details see https://en.cppreference.com/w/cpp/compiler_support and https://xcodereleases.com/
-(Xcode 10 needs MacOs 10.13 and Xcode 11 needs MacOS 10.14)
+OS X/macOS allows you to specify a so-called "deployment target" version that will ensure backwards compatibility with older versions of macOS. One way to do this is by setting the `MACOSX_DEPLOYMENT_TARGET` environment variable. If not set, Python will set this variable to the version the Python distribution itself was compiled on (10.6 or 10.9, for the python.org packages), when creating the wheel.
 
-## Windows
+However, to enable modern C++ standards, the deploment target needs to be set high enough (since older OS X/macOS versions did not have the necessary modern C++ standard library).
 
-Visual C++ for Python 2.7 does not support modern standards of C++.
-When building on Appveyor, you'll need to use either the 'Visual Studio 2017' or 'Visual Studio 2019' image. Note that Python 2.7 isn't supported on these images - you should skip it using `CIBW_SKIP=cp27-win*`.
+To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.9"`.
 
-There is option for workaround this limitation. PyBind project has in documentation example how to compile python 2.7 extension with newer compiler https://github.com/pybind/python_example
+To get C++17 support, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
+
+For more details see https://en.cppreference.com/w/cpp/compiler_support and https://xcodereleases.com/: Xcode 10 needs macOS 10.13 and Xcode 11 needs macOS 10.14.
+
+## Windows and Python 2.7
+
+Visual C++ for Python 2.7 does not support modern standards of C++. When building on Appveyor, you will need to either use the "Visual Studio 2017" or "Visual Studio 2019" image, but Python 2.7 is not supported on these images - skip it by setting `CIBW_SKIP=cp27-win*`.
+
+There is an optional workaround for this, though: the pybind11 project argues and shows that it is [possible to compile Python 2.7 extension with a newer compiler](https://pybind11.readthedocs.io/en/stable/faq.html#working-with-ancient-visual-studio-2008-builds-on-windows) and has an example project showing how to do this: https://github.com/pybind/python_example. The main catch is that a user might need to install a newer "Microsoft Visual C++ Redistributable", since the newer C++ standard libraries are not included by default with the Python 2.7 installation.
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -1,0 +1,31 @@
+---
+title: Modern C++ standards
+---
+
+Building Python wheels with modern C++ (C++11) requires a few tricks.
+
+Creating wheel for python 2.7 which need c++17 needs special configuration because python 2.7 header files using `register` keyword. 
+In c++17 this keyword is reserved ((see)[https://en.cppreference.com/w/cpp/keyword/register])
+It is possible to allow usage of `register` using proper flag `-Wno-register` for gcc/clang and `/wd5033` for MSVC. 
+
+## Linux
+When using default `manylinux1` image it is possible to use only c++11 and earlier standards. 
+This is about how you can make C++14 wheels in `manylinux1`, through some tricks https://github.com/pypa/manylinux/issues/118
+
+`manylinux2010` supports all C++ standards (up to c++17). 
+
+## MacOS
+
+To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.9"`.
+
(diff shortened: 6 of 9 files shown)


YannickJadoul:
@Czaki How is this the case? I've been building wheels using C++17 with the current version of cibuildwheel. Have I been doing something wrong? Could you explain in slightly more detail what the problem is?

Czaki:
When I try to setup ci build on azure-pipelines. When building wheels for macos I got this error:
```
In file included from package/PartSeg/utils/multiscale_opening/mso_bind.cpp:645:
  package/PartSeg/utils/multiscale_opening/mso.h:1:10: fatal error: 'array' file not found
  #include <array>
           ^~~~~~~
  1 warning and 1 error generated.
  error: command 'gcc' failed with exit status 1
```
Here is pipeline: https://dev.azure.com/PartSeg/PartSeg/_build/results?buildId=74

Here is source code: https://github.com/4DNucleome/PartSeg/tree/cibuildwheel_problem

When I upgrade to 10.9 version everything works. Maybe you have some suggestion how workaround this problem in other way.

YannickJadoul:
Hmmm, interesting. I haven't worked with azure pipelines yet, so I don't know. But I'm intrigued how a different Python version solves this problem, so I'd like to have a look indeed!

Czaki:
This and similar questions: 
https://stackoverflow.com/questions/14494513/how-do-i-get-back-c0x-c11-support-for-mac-os-x-10-6-deployment-using-xcode-4
suggest that the problem is not in python, but in this that wheel is build against 10.6 sdk which not support c++11.  And changing to build against newer macos sdk fix this. Maybe upgrading also python build is not important. I think that most important is changing identifier. (`identifier='cp36-macosx_10_9_intel'`) etc.

YannickJadoul:
Thanks! Maybe I should check the validity of my macOS wheels (built on TravisCI, btw), as well, then!

Czaki:
It may depend of macos version on virtual machine. Maybe I move part of build to travis, but I prefer to have all three system on one service. In my repository I have configured travis for windows test also. Maybe it can be base for `cibuildwheel` windows build for travis?

YannickJadoul:
> It may depend of macos version on virtual machine. Maybe I move part of build to travis, but I prefer to have all three system on one service.

No, I understand. You're completely right: this should work on azure as well! (It's already been 8 years since C++11, c'mon, this should not be an issue!) I'd just love to understand this issue and implications for old builds/wheels, and be able to give more advise in the README :-)

> In my repository I have configured travis for windows test also. Maybe it can be base for cibuildwheel windows build for travis?

That would be great! There's already an issue open on this (#144), but I guess no one has taken the time to implement it, yet, since we're still trying to fix the `manylinux2010` support (#135 and #155).
But any kind of help is welcome, be it a full PR or just a comment on #144 with a link to your configuration.

YannickJadoul:
Update: I tried installing and importing my own library (which C++17) on a Mac OS X 10.6 installation, and I'm getting the following error:

Screenshot from 2019-08-28 15-27-22

I'm guessing the reason is that Apple was still using libstdc++ back then? I'm curious to give it another try, but so, by default, 10.6 is not really supported anyway, it seems. Although that maybe just using C, things might work? I'll try to give that a go.

At any rate, I also realized how old this platform has become (2009-2011), and apparently it is officially not supported anymore since 2014. I also couldn't install Python 3.7.4 (not sure why, but I got an error), but had to go with 3.7.0. And while installing, I also noticed that numpy's latest version (1.17.1) does not provide a 10.6 wheel anymore (but 1.17.0 still did).

So, maybe it's just time to drop support for 10.6, or make 10.9 the default and provide 10.6 only through an explicit option. @joerick, @mayeut, any thoughts?

@Czaki, I need to still look at what's actually causing the compilation error in your case, but thanks for bringing this to our attention! Are you happy working with your own branch, for the moment, or should we strive to get a short-term solution as well?

Czaki:
I can work with my branch, but I would like to not need port future changes to my repository. 

I think that putting libstdc++ inside wheel is risky strategy.

YannickJadoul:
> I can work with my branch, but I would like to not need port future changes to my repository.

No, sure, this needs to get solved. But then we don't need a temporary solution too soon.

> I think that putting libstdc++ inside wheel is risky strategy.

Yeah, probably. Though I am already statically linking `libstdc++` to my `manylinux1` wheels, since C++14 is not supported on there. So I'm just curious what would happen...

Czaki:
Thanks. 
I`m not sure but maybe solve statistical linking on Linux should be connected with manylinux2010. 

How I can got old MacOs images for testing purpose? I'm not very familiar with envirornment.

YannickJadoul:
> How I can got old MacOs images for testing purpose? I'm not very familiar with envirornment.

We still had some old Mac around at the lab, that I could use (and apparently Apple still sells images). Since OS X is only allowed to run on Apple hardware, ... which is amazingly annoying, but apparently Apple thinks that's more important than allowing developers to easily target their platform.

joerick:
> So, maybe it's just time to drop support for 10.6, or make 10.9 the default and provide 10.6 only through an explicit option.

I tend to agree, but I wasn't sure... Using my [favourite new toy](https://packaging.python.org/guides/analyzing-pypi-package-downloads/) I crunched some numbers on the OS version of PyPI downloads-

```sql
SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(details.distro.version, r"^(\d+.\d+)") as system_version
WHERE
  details.system.name = 'Darwin' AND
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY system_version
ORDER BY num_downloads DESC
```

Row | num_downloads | system_version |  
-- | -- | -- | --
1 | 36722694 | 10.14 |  
2 | 15164382 | 10.13 |  
3 | 1840143 | 10.12 |  
4 | 588444 | 10.11 |  
5 | 412806 | 10.15 |  
6 | 167678 | 10.10 |  
7 | 38686 | 10.9 |  
8 | 25143 | null |  
9 | 3957 | 10.6 |  
10 | 2077 | 12.4 |  
11 | 2041 | 10.7 |  
12 | 1896 | 10.16 |  
13 | 1738 | 10.8 |  
14 | 326 | 10.5

We see that pre-10.9 versions of macOS make up.... less than 0.02% of the downloaders :)

So, based on this, I'd be happy just to drop 10.6 altogether. If we get a couple separate requests for an option to add it back, we could add an option for 10.6 then.... but I suspect we won't!

joerick:
If @YannickJadoul is happy to go with this, @Czaki could you remove the references to the 10.6 pkgs from macos.py?

@YannickJadoul since this includes build identifier changes, it would be great to use the same CIBW_BUILD/CIBW_SKIP migration stuff that we used in the manylinux1->manylinux transition. So let's plan to land this after the manylinux2010 PR is merged.

Czaki:
I fail with all test and there is no option for upgrade python 3.4 and 3.5 to macosx10.9. There is no duch builds on https://www.python.org/downloads/mac-osx/. I do not know what now. There are also needs for small fix in readme.  Do this?

joerick:
ah, that changes things a bit... python3.5 is still used quite a bit on mac.

```sql
SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(details.python, r"^(\d+\.\d+)") as python_version
WHERE
  details.system.name = 'Darwin' AND
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY python_version
ORDER BY num_downloads DESC
```

image

Regardless, I think that your current solution makes the most sense - use the 10.9 version if it's available.

It does make the build identifiers a little tricky, but really this just highlights our need to have a big list of them available in the readme. That's a docs concern, I can have a look at that later. As for the build identifier migration warnings, we could add a warning on the string `*-macosx_10_6_intel` - suggest that it is changed to `*-macosx_*`.

Happy for this to merge @YannickJadoul ?

YannickJadoul:
@joerick Hmmm, I'm not sure. I'm a bit annoyed that I still don't fully understand what's going on here. Because, as I mentioned before, I can perfectly build wheels with C++17 with the current cibuildwheel. (Though they don't seem to work, because libc++ is not available on 10.6, but that's a different problem.)

I was still planning to research further, if I get the time (but there's a few conferences going on this month). However, if you're happy merging this, go ahead, and I'll catch up later and make a new PR if I figure out what's going on :-)

Czaki:
@YannickJadoul The problem with building is not targeting for macos 10.6, but wrong environment. It looks that on azure (i do not have experience from other ci for this case) when targeting build for macos 10.6 it have only access to this level environment. So in this case two fixes are possible. Upgrade build target to have access to c++11 headers (and latest) or modify environment with installing this dependencies in it. I have only access to one macos device and cannot use it for such experiments. My experience with this system is to low to write it without many tests.

YannickJadoul:
@Czaki Yeah, I can see the environment of the CI matters, but I'm very baffled how the installation of just Python inside of `cibuildwheel` affects the compiler/environment :-(
Do these other versions of Python (for 10.9) then come with new C++ stdlib headers?

Czaki:
Maybe it is connected with compilers. If I good understand it is need to pass minimum sdk version to it. And basing on it it decides which headers are available? 
https://stackoverflow.com/questions/25352389/what-is-the-difference-between-macosx-deployment-target-and-mmacosx-version-min/25362535

YannickJadoul:
Oh, right, that's an interesting point!

Czaki:
pypa/wheel#312

Czaki:
I think that I found the solution. The MACOSX_DEPLOYMENT_TARGET was the key. But there is bug in wheel package pypa/wheel#312 and it not bump platform tag when compile against newer platform. I will write the patch and create pull request, but it may take a time. This is explanation of @YannickJadoul case of creating c++17 wheel.  

Now I use bypass by overwriting platform tag manually. 

I also add test for support of c++ standards. (#165)

This changes are also tested on windows on travis from #160.

Czaki:
On AppVeyor there is a problem. One can choose image with Visual C++ Compiler for Python 2.7 or newer MSVC version 
obraz

I do not know how to solve this problem. Create some array? Skip test base on "APPVEYOR_BUILD_WORKER_IMAGE" ?

YannickJadoul:
@Czaki Thanks for that! So if I understand correctly, the problem is related to `distutils` setting compiler flags/options? Or did I get that wrong?
That would explain why things 'just work' for me, since I'm only using `setup.py` to run CMake, whih then determines its own options. (But that would probably means my wheels are not built in a compatible way? :-( )

Something different: could you rebase your PR onto the master? Because you've merged in these changes from the master, there's a bunch of code in this PR that's unrelated and it's hard to get an overview of the relevant changes.

Czaki:
I think that error is not in distutils, because it return information about current python, but bug is in wheel package which do not check environment and only copy information about python platform tag.

I wrote solution that check MACOSX_DEPLOYMENT_TARGET. But it need to be set outside. And I do not know how check what version is chosen by cmake. Maybe variable CMAKE_OSX_DEPLOYMENT_TARGET will help. I do not have any code for reproducing this use case.  

> Something different: could you rebase your PR onto the master? Because you've merged in these changes from the master, there's a bunch of code in this PR that's unrelated and it's hard to get an overview of the relevant changes.

You thnk to remove previous commits? Or remove tests of c++ standards compatibility? Or split it and move c++ standard tests to another commit (I think that they are part of this problem because it tests the solution proposed here)

YannickJadoul:
> I think that error is not in distutils, because it return information about current python, but bug is in wheel package which do not check environment and only copy information about python platform tag.

Right, I see. Thanks for the extra info!

> You thnk to remove previous commits? Or remove tests of c++ standards compatibility? Or split it and move c++ standard tests to another commit (I think that they are part of this problem because it tests the solution proposed here)

No, no need to split, I think. There's just a bunch of stuff from the `CIBW_TEST_EXTRAS` #161 PR that was merged recently. I think if you just update the master reference and run `git rebase master` and push that branch again, we should get rid of those unrelated changes.

Czaki:
Done. What with AppVeyor problem (tests fails on it) as mentioned in previous comments.

YannickJadoul:
Thanks. Something weird is going on with the history of this PR, still, but at least the changelog on GitHub is clear now :-)
This should make it easier to have a look at why this AppVeyor build can be failing!

Czaki:
AppVeyor is falling because modern standards of c++ are supported from MSVC15 and more modern (for c++17 there is need for MSVC17). 

But from image "Visual Studio 2017" the "Visual C++ Compiler for Python 2.7" was removed. 
So there is no option for test against c++17 and python2.7 (different test) on one image. Posibilites are:  

- skip test for c++17 on AppVeyor
- add next entry to AppVeyor and modify all test to skip python 2.7 test if run on "Visual Studio 2017" or "Visual Studio 2019" image
- find way to install "Visual C++ Compiler for Python 2.7" compiler on 2017 and 2019 images.

YannickJadoul:
> But from image "Visual Studio 2017" the "Visual C++ Compiler for Python 2.7" was removed.

Weirdly enough, I again manage to build C++17 wheels for Windows without this "Visual C++ Compiler for Python 2.7". I still don't fully understand what's going on there, with that ancient compiler, and the pybind11 docs say the following: 

> The official Windows distributions of Python are compiled using truly ancient versions of Visual Studio that lack good C++11 support. Some users implicitly assume that it would be impossible to load a plugin built with Visual Studio 2015 into a Python distribution that was compiled using Visual Studio 2009. However, no such issue exists: it’s perfectly legitimate to interface DLLs that are built with different compilers and/or C libraries.

See https://pybind11.readthedocs.io/en/stable/faq.html#working-with-ancient-visual-studio-2009-builds-on-windows

Besides, you can already not build C++11 or C++14 in with this "Visual C++ Compiler for Python 2.7", right? Or am I getting something wrong, there?

README.md
```diff
@@ -236,6 +236,25 @@ Linux wheels are built in the [`manylinux1` docker images](https://github.com/py
...
+-------------------
+Python 2.7 use `register` keyword which is prohibited in c++17 standard. 
+
+Linux
+=====
+For `manulinux1` images only c++11 standards is supported. c++14 and c++17 will be supported with manylinux2010 image
+
+MacOS
+=====
+To get c++11 and c++14 upport need to set `MACOSX_DEPLOYMENT_TARGET` variable to `10.9` value
```
hardbyte:
```suggestion
To get c++11 and c++14 support need to set `MACOSX_DEPLOYMENT_TARGET` variable to `10.9` value
```

Do newer macos versions work as well? Or just `10.9`?

Czaki: 10.9 is minimal version of macos on which package will work.

Czaki:
@YannickJadoul 
The problem is with Python 2.7. When I run test on image without "Visual C++ Compiler for Python 2.7" I get this error ([log](https://ci.appveyor.com/project/joerick/cibuildwheel/builds/27426401/job/2t97tkhr2a5h1kw6)):

```
building 'spam' extension
  error: Microsoft Visual C++ 9.0 is required. Get it from http://aka.ms/vcpython27
  ----------------------------------------
```

Maybe it is possible to build it with new python version. On python [wiki](https://wiki.python.org/moin/WindowsCompilers) you can see the information that official python releases are compiled with MSVC 9.0. 
Here is example repo with some steps needed to build extension with old python versions. https://github.com/pybind/python_example

I do not understand this.

YannickJadoul:
@Czaki Hmmm, interesting, so somehow those environment variable can force `setuptools` to use another compiler than the "Visual C++ Compiler for Python 2.7"? That's making this already a bit more clear on why it can sometimes work.

What I don't really get though, is that your tests do work for C++11 and C++14? Because this VC++ 9.0 is really old (Visual Studio 2008, I believe) and should not support C++11 features, either? So are the tests then still somehow using a more modern compiler?

I'm not entirely sure how we could check this; is there a way of getting verbose output, or getting the compile commands? You'd probably need to at least add `-s` to the test commands to get the output of the succeeding tests, but that's not going to be enough.

Czaki:
@YannickJadoul The problem is with c++17 test. If on AppVeyor I use windows image which have "Microsoft Visual C++ 9.0" compiler then the compiler used by other version of python is "Visual Studio 2015". So test on c++17 fails. When I use "Visual Studio 2017" image then all test on python 2.7 fails. 

In test on c++ standards I ignore `cp27-win*`

YannickJadoul:
> If on AppVeyor I use windows image which have "Microsoft Visual C++ 9.0" compiler then the compiler used by other version of python is "Visual Studio 2015".

Oooh, you're skipping 2.7, when building the C++11/14/17 test?

Czaki:
On windows yes.

joerick:
> In test on c++ standards I ignore cp27-win*

I think that's fine - it's documented that 2.7 and modern C++ don't work together. @Czaki - I tweaked your docs a little, can you let me know if my changes are accurate as you understand it?

mayeut:
The 33% download of cp27 is quite scary given that 2.7 is EOL in January 2020.

cibuildwheel/macos.py
```diff
@@ -81,6 +81,8 @@ def call(args, env=None, cwd=None, shell=False):
             env['PATH'],
         ])
         env = environment.as_dictionary(prev_environment=env)
+        if "MACOSX_DEPLOYMENT_TARGET" not in env:
+            env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
```
mayeut: If targeting macOS 10.9, then I see 2 options:
- keep building using 10.6 installer of python (32/64 bits)
- use the 10.9 installer of python when available (64 bits), dropping 32 bits support

Any idea what shall be done ? (Don't know if the analysis can be done to differentiate downloads triggered by 32/64 bits python)

YannickJadoul: That's a good point, actually. So we're losing 32-bit compatibility?
But then, if I interpret [this table](https://en.wikipedia.org/wiki/MacOS#Hardware_compatibility) correctly, all versions after 10.6 were 64-bit only?

mayeut: With the PR in its current state, we don't loose 32-bit compatibility. It's also still possible to target 10.6 by defining `MACOSX_DEPLOYMENT_TARGET` to `10.6`.

10.6 installers do install a dual 32/64 bits python and, when building using these, wheels will be 32/64 bits. If switching to 10.9 installers then, wheels will only be 64 bits (and tags will need to be updated from `intel` to `x86_64`). As you mentioned, 10.7+ is always 64 bits (but can run 32bits software up to 10.14, 10.15 will drop 32 bits support). The PR in this state still builds 32/64 bits wheels (`intel` wheels) and is still the most compatible with everything out there. My question is : shall 32 bits support be dropped as well or not ? (by using 10.9 installers).

Czaki:
I think it is better to keep 10.6. Python 3.5 do not have pure 64 bits instalator. So it will create inconsistence. At this moment i do not see profit of changing to 10.9 installers.

I think that python identifier on macos should be modified. To not contain macos version information.

cibuildwheel/macos.py
Czaki: There are few problems. Python 3.5 do not have 10.9 installer. So we cannot change installer in this case which introduce incompatibility. Maybe wait with changing installers to drop of Python3.5 support. 

What will be profit of updating versions?

Czaki:
I'm working on pull request to wheel package which will patch the platform tag. But to do this properly I need to learn how to parse Match-O headers. This may take a time.

README.md
```diff
@@ -236,25 +236,26 @@ Linux wheels are built in the [`manylinux1` docker images](https://github.com/py
...
+#### MacOS
+
+To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `10.9`.
+
+To get C++17 support need to set `MACOSX_DEPLOYMENT_TARGET` variable to `10.13` or `10.14`. (`10.13` supports C++17 partially, e.g. filesystem header is in experimental: `#include <filesystem>` -> `#include <experimental/filesystem>`)
+
+#### Windows
+
+Visual C++ for Python 2.7 do not support modern standards of C++.
+When building on Appveyor, you'll need to use the 'Visual Studio 2019' image. Note that Python 2.7 isn't supported on this image - you should skip it using `CIBW_SKIP=cp27-win*`.
```
Czaki: For c++14 and c++17 both images should work: 'Visual Studio 2017' or 'Visual Studio 2019'

README.md
```diff
@@ -236,25 +236,26 @@ Linux wheels are built in the [`manylinux1` docker images](https://github.com/py
...
 
+#### MacOS
+
+To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `10.9`.
+
+To get C++17 support need to set `MACOSX_DEPLOYMENT_TARGET` variable to `10.13` or `10.14`. (`10.13` supports C++17 partially, e.g. filesystem header is in experimental: `#include <filesystem>` -> `#include <experimental/filesystem>`)
+
+#### Windows
+
+Visual C++ for Python 2.7 do not support modern standards of C++.
```
Czaki: this configuration do not support. Here is example how get this possibility, but I do not understand it yet: https://github.com/pybind/python_example

README.md
```diff
@@ -236,25 +236,26 @@ Linux wheels are built in the [`manylinux1` docker images](https://github.com/py
...
+Python 2.7 wheels cannot use C++17 at all, because Python 2.7 uses the `register` keyword, which is prohibited in C++17 standard.
 
-To get c++17 support need to set `MACOSX_DEPLOYMENT_TARGET` variable to `10.13` or `10.14` value. `10.13` supports c++17 partialy, eg. filesystem header is in experimental: `#include<filesystem> -> #include<experimental/filesystem>`
+#### Linux
 
-Windows
-=======
-Visual C++ for Python 2.7 do not support modern standards of c++.  
-Also MSVC 10.0 
+For `manylinux1` images, only C++11 standards is supported. C++14 and C++17 will be supported with manylinux2010 image (coming soon!).
```
Czaki: change manylinux2010 to `manylinux2010`

Czaki:
I create pull request to wheel package wheel/pull#314

cibuildwheel/macos.py
mayeut: > What will be profit of updating versions?

Mostly:
- reducing build time
- reducing test time (well, since 32bits is probably not tested in most cases - at least all working examples, including mine-, probably not applicable)
- reducing package size

This being said, waiting for python 3.5 support to be dropped might be a good idea for consistency across python versions.

Czaki: Maybe add some MacOSX specific variable to allow user set other python installer?

README.md
joerick:
```suggestion
When building on Appveyor, you'll need to use either the 'Visual Studio 2017' or 'Visual Studio 2019' image. Note that Python 2.7 isn't supported on these images - you should skip it using `CIBW_SKIP=cp27-win*`.
```

Czaki:
Waiting on wheel upgrade I would like to mention that macos 10.15 drop support of 32-bits application. Some programs like Photoshop have 32 parts. So many people can stuck on 10.14.

Czaki:
@YannickJadoul My pull request to wheel was accepted. This solve some problems but need next changes. And I'm not sure if continue this pull request or create next which will not have so many commit and strange changes in history. 

What do you think?

YannickJadoul:
@Czaki Hmmm, good point. I guess that's mostly up to @joerick. But if you're asking me, I'd say go for whatever you prefer. But if you worry about history, we can always squash the pull request into a single commit, when merging it.

Do note that meanwhile, #155 has been merged, so you might need to change a few things in the readme?

joerick:
Good job getting the wheel PR merged @Czaki ! I guess that'll have to wait for a minor release before it's ready to use inside cibuildwheel?

>  I'm not sure if continue this pull request or create next which will not have so many commit and strange changes in history.

Like @YannickJadoul said, whatever you prefer. You're right that the history is getting a little crazy, but I can 'squash & merge' in GitHub when I merge.

Czaki:
I modify existing code for changes in wheel package. Current code on MacOS install wheel from repo instead of  pypi. So it is not ready for production. 
I also remove system version from macosx platform tag. 
I`m not sure if there is need to add something which will check and update old platform tag or pr #179 will be accepted in some form and user will update config.

YannickJadoul:
Umpf, so many changes, suddenly. Is there a possibility to rebase onto the master instead of merging the master into your PR, @Czaki?

Czaki:
@YannickJadoul I have hope that I fixed.

YannickJadoul:
Seems easier to read, indeed. Thanks! :)

cibuildwheel/macos.py
```diff
@@ -91,7 +91,7 @@ def call(args, env=None, cwd=None, shell=False):
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'wheel'], env=env)
+        call(['pip', 'install', 'git+https://github.com/pypa/wheel.git@master'], env=env)
```
joerick: Installing from master may not be a good idea, probably better to wait for a release or pin to a specific commit in the mean time

```suggestion
        call(['pip', 'install', 'git+https://github.com/pypa/wheel.git@a7510b6510ab3f1768de7eb75d7a6d06303bee24'], env=env)
```

cibuildwheel/macos.py
```diff
@@ -14,11 +14,11 @@
...
-        PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
+        PythonConfiguration(version='3.4', identifier='cp34-macosx_intel', url='https://www.python.org/ftp/python/3.4.4/python-3.4.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.5', identifier='cp35-macosx_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.6', identifier='cp36-macosx_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
+        PythonConfiguration(version='3.7', identifier='cp37-macosx_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
```
joerick: I just want to check I understand this - it's now the case that `wheel` will actually pull the MACOS_DEPLOYMENT_TARGET from the dylibs themselves? So the cibuildwheel identifier is no longer in control of whether this is 10_6 or 10_9 or whatever.

Is that right?

README.md
```diff
@@ -314,15 +334,15 @@ The format is `python_tag-platform_tag`. The tags are similar but not identical
 
 Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37`
 
-Platform tags look like `macosx_10_6_intel` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
```
joerick: Could we do one of those identifier migrations like how @YannickJadoul did the `manylinux1`->`manylinux` warnings?

test/shared/utils.py
```diff
@@ -88,12 +88,16 @@ def expected_wheels(package_name, package_version):
...
-            '{package_name}-{package_version}-cp27-cp27m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp34-cp34m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp35-cp35m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp36-cp36m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp37-cp37m-macosx_10_6_intel.whl',
+            '{package_name}-{package_version}-cp27-cp27m-macosx_' + tag + '_intel.whl',
+            '{package_name}-{package_version}-cp34-cp34m-macosx_' + tag + '_intel.whl',
+            '{package_name}-{package_version}-cp35-cp35m-macosx_' + tag + '_intel.whl',
+            '{package_name}-{package_version}-cp36-cp36m-macosx_' + tag + '_intel.whl',
+            '{package_name}-{package_version}-cp37-cp37m-macosx_' + tag + '_intel.whl',
```
joerick: We'll need to add Python 3.8 to this list now, too :)

cibuildwheel/macos.py
Czaki: yes

README.md
Czaki: added

Czaki: added

cibuildwheel/__main__.py
```diff
@@ -198,6 +198,8 @@ def detect_obsolete_options():
         if deprecated in os.environ and 'manylinux1' in os.environ[deprecated]:
             print("Build identifiers with 'manylinux1' been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
             os.environ[deprecated] = os.environ[deprecated].replace('manylinux1', 'manylinux')
+        if deprecated in os.environ and ("macosx_10_6" in os.environ[deprecated] or "macosx_10_9" in os.environ[deprecated]):
+            os.environ[deprecated] = os.environ[deprecated].replace('macosx_10_6', 'macosx').replace('macosx_10_9', 'macosx')
```
joerick:
```suggestion
            print("Build identifiers with 'macosx_10_6' or 'macosx_10_9' been deprecated. Replacing with 'macosx' in the option '{}'".format(deprecated))
            os.environ[deprecated] = os.environ[deprecated].replace('macosx_10_6', 'macosx').replace('macosx_10_9', 'macosx')
```

README.md
```diff
@@ -248,6 +248,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
 
+### Modern C++ standards
+
+Building Python wheels with modern C++ (C++11) requires a few tricks.
+
+Python 2.7 wheels cannot use C++17 at all, because Python 2.7 uses the `register` keyword, which is prohibited in C++17 standard.
+
+#### Linux
+
+`manylinux2010` support all c++ standards. If you would like to build `manylinux1` images then your project is limited to c++11 and earlier.
```
nsoranzo:
```suggestion
`manylinux2010` supports all C++ standards. If you would like to build `manylinux1` images, then your project is limited to C++11 and earlier. 
```

README.md
```diff
@@ -248,6 +248,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
+
+#### Linux
+
+`manylinux2010` support all c++ standards. If you would like to build `manylinux1` images then your project is limited to c++11 and earlier. 
+
+#### MacOS
+
+To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.9"`.
+
+To get C++17 support need to set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.13"` or `"10.14"`. (`"10.13"` supports C++17 partially, e.g. filesystem header is in experimental: `#include <filesystem>` -> `#include <experimental/filesystem>`)
```
nsoranzo:
```suggestion
To get C++17 support, set `MACOSX_DEPLOYMENT_TARGET` variable to `"10.13"` or `"10.14"`. (`"10.13"` supports C++17 partially, e.g. the filesystem header is in experimental: `#include <filesystem>` -> `#include <experimental/filesystem>`)
```

README.md
nsoranzo:
```suggestion
Visual C++ for Python 2.7 does not support modern standards of C++.
```

cibuildwheel/macos.py
```diff
@@ -14,10 +14,11 @@
 def get_python_configurations(build_selector):
     PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
     python_configurations = [
-        PythonConfiguration(version='2.7', identifier='cp27-macosx_10_6_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
-        PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
-        PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
+        PythonConfiguration(version='3.4', identifier='cp34-macosx_intel', url='https://www.python.org/ftp/python/3.4.4/python-3.4.4-macosx10.6.pkg'),
```
nsoranzo: Python 3.4 support has just been dropped in #168.

```suggestion
```

test/shared/utils.py
```diff
@@ -82,11 +82,16 @@ def expected_wheels(package_name, package_version):
...
+            tag = "10_6"
+        else:
+            tag = os.environ["MACOSX_DEPLOYMENT_TARGET"].replace(".", "_")
         templates = [
-            '{package_name}-{package_version}-cp27-cp27m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp35-cp35m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp36-cp36m-macosx_10_6_intel.whl',
-            '{package_name}-{package_version}-cp37-cp37m-macosx_10_6_intel.whl',
+            '{package_name}-{package_version}-cp27-cp27m-macosx_' + tag + '_intel.whl',
+            '{package_name}-{package_version}-cp34-cp34m-macosx_' + tag + '_intel.whl',
```
nsoranzo:
```suggestion
```

cibuildwheel/macos.py
```diff
@@ -90,7 +91,7 @@ def call(args, env=None, cwd=None, shell=False):
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'wheel'], env=env)
+        call(['pip', 'install', 'git+https://github.com/Czaki/wheel.git@cbaa8f623115a14fe0a03f73f0b898eb84e3abfd'], env=env)
```
Czaki: PR pypa/wheel/pull/317

YannickJadoul:
Can we still hold this a bit before merging? Or maybe give me a deadline to review this in detail?
I'd really like to understand this and read up on it, but I haven't had the time or courage yet to get into it :-/

Czaki:
I think that it will be best to wait with this merge since new release of `wheel` package is created. 
I use not released version to shoretn gap between `wheel` release and merge to `cibuildwheel`.

joerick:
From looking at https://github.com/pypa/wheel/pull/317 , I agree that it will be best to wait for that to be resolved and for an official wheel release, which will give you time to review @yannickjadoul.

Czaki:
@YannickJadoul @joerick If there is decision to create both `manylinux1` and `manylinux2010` then maybe create  both `intel` and `x86_64` wheels to easier allow user choose their needs? 

It will increase time build for whole, but gave user bigger chose (and maybe smaller wheels)

YannickJadoul:
> If there is decision to create both manylinux1 and manylinux2010

We have not made that decision. The user chooses on which image to build (default `manylinux2010`), but as I explained in #187, sometimes a wheel built on the `manylinux2010` image is also `manylinux1`-compatible. We dón't build the same wheel twice; it's just `auditwheel` that returns the same wheel with 2 different tags.

Czaki:
I think a little what can user do if he write x86_64 code and would like to produce only 64-bits wheel.

test/shared/utils.py
```diff
@@ -88,12 +98,18 @@ def expected_wheels(package_name, package_version):
             '{package_name}-{package_version}-cp38-cp38-win_amd64.whl',
         ]
     elif platform == 'macos':
+        if "MACOSX_DEPLOYMENT_TARGET" not in env:
+            tag = "10_6"
+            tag1 = "10_9"
+        else:
+            tag = env["MACOSX_DEPLOYMENT_TARGET"].replace(".", "_")
+            tag1 = env["MACOSX_DEPLOYMENT_TARGET"].replace(".", "_")
```
joerick: I guess we should only be using major/minor here, ignoring patch

joerick:
> I think a little what can user do if he write x86_64 code and would like to produce only 64-bits wheel.

This is an interesting thought... I was also thinking it might be nice with the build identifiers to use just `macosx_...` rather than `macosx_..._intel` and `macosx_..._x86_64` for Python 3.8.

In which case, cibuildwheel could build universal 32/64 where possible by default, but build only 64 bit with an option `CIBW_MACOS_64BIT_ONLY`...?

@Czaki do you know how you get pip/setuptools to build a 64bit only wheel on macos?

Czaki:
I take look to build_ext code and not seen easy option. It looks like it need subclass of `build_ext` and then change compiler parameters (remove `-arch i386` from arguments). Changing used python installation looks easier (add additional architectures for cross compilation looks easier).

zackw:
I am working on a package that needs C++11 for some of its code and I was just pondering how to get the platform tags right on OSX.  What I have now produces wheels tagged `macosx_10_6_intel` but with code inside that will fail to load on <10.9.  Please let me know if I can help with this PR, e.g. by testing stuff.

test/shared/utils.py
joerick: Oops sorry, this is a test. Nevermind !

Czaki:
@zackw All tests of this PR are welcome. Current revision of this tests bases on changes in wheel package which try to calculate macos version base on library files headers. I try to write best solution, but not all things are documented well. 

I do not know how log we need to wait on next release of wheel package.

YannickJadoul:
@zackw How did you get C++11 working, then, with the current version of `cibuildwheel`? (OK, it fools `delocate` into thinking it's a macOS 10.6 wheel, OK, but I'd rather say that this is a problem with `decloate` than with `cibuildwheel`?)

Czaki:
@YannickJadoul the only thing which is needed to have working c++11 on macosx with current `cibuildwheel` is to set `MACOSX_DEPLOYMENT_TARGET` to `"10.09"`or above. Then wheel build finish with success. 

All other changes in this PR are documentation, tests for c++ standards, and upgrade of platform tag.

zackw:
@YannickJadoul @Czaki It was a bit more involved than that for me.  I had to force everything to be built with `-std=c++11 -stdlib=libc++ -mmacosx-version-min=10.9`, both in `extra_compile_args` to my extension, and in the build scripts for the third-party C++ library that I'm wrapping.  I tried just setting `MACOSX_DEPLOYMENT_TARGET=10.9` at the `CIBW_ENVIRONMENT` level, but distutils overwrote that when it invoked the C++ compiler.

YannickJadoul:
@zackw Aaaah, thanks! Things are starting to make sense, now. I'm calling CMake from my `setup.py` (maybe not the cleanest, but at least the easiest), and apparently CMake is handling that for me: https://cmake.org/cmake/help/v3.15/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html

So thát's why I never ran into serious problems with this!

Czaki:
@zackw `MACOSX_DEPLOYMENT_TARGET` should be set as general environment variable (you can see [here](https://github.com/Czaki/tsp_spanning/blob/master/azure-pipelines.yml). Then it working.  For safety reason compiler flags which indicates C++ standard need to be added. Not all compilers on Linux use modern standards as default.

zackw:
Nope. Tried that. Didn’t work. Distutils reset it to 10.6.

Czaki:
@zackw Is repository with this code public? If not could you prepare small project which will reproduce this problem?

zackw:
@Czaki Repo is https://bitbucket.org/elwoz/cyexiv2/ ; you probably want to read both ci/azure-build.py and setup.py.  You can see build logs at https://dev.azure.com/zackw0744/cyexiv2/_build (not all builds have OSX enabled).  Everything I've said so far was observed with the stock cibuildwheel; I haven't done any tests with your patches applied.

Czaki:
@zackw Setting variable is still enough, but it need to be set in proper place. I create copy of your repo and try i two way. I link result bellow. Maybe you have some suggestion to documentation? (If I good understand distutuils do not allow both CXXFLAGS and extra_compile_args)


https://github.com/Czaki/cyexiv2/commit/e3bd0582010d4dfc8b1c9f60ca74d895473e858c
https://dev.azure.com/bokota/open%20source%20contrib/_build/results?buildId=16

and

https://github.com/Czaki/cyexiv2/commit/0efc8ea209000ebce848845a658da9374362567e
https://dev.azure.com/bokota/open%20source%20contrib/_build/results?buildId=20

Czaki:
@zackw The other part is that you do not compile all with cmake. So even set flags for cmake is not enough. There is lib build from setup.py  were compiler is caled by `build_ext`.

cibuildwheel/macos.py
mayeut: c.f. #202 / #203 for my point on 32-bit testing

README.md
```diff
@@ -242,6 +242,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
 - The project directory is mounted in the running Docker instance as `/project`, the output directory for the wheels as `/output`. In general, this is handled transparently by `cibuildwheel`. For a more finegrained level of control however, the root of the host file system is mounted as `/host`, allowing for example to access shared files, caches, etc. on the host file system.  Note that this is not available on CircleCI due to their Docker policies.
 - Alternative dockers images can be specified with the `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE` options to allow for a custom, preconfigured build environment for the Linux builds. See [below](#options) for more details.
 
+### Modern C++ standards
```
YannickJadoul: I'm a little bit in doubt on where we should add this, and in which amount of detail. I agree modern C++ versions are important, but `cibuildwheel`is independent of the language/compilation that generates/compiles the C extensions.

So in that case, it might be better to create a page with FAQ on different programming languages. And rather than explaining everything ourselves, we should link to the relevant resources so that users can find out more.

appveyor.yml
```diff
@@ -1,3 +1,5 @@
+image: Visual Studio 2015
```
YannickJadoul: Is this necessary?

This should be coordinated with @mayeut, who's been distributing the different tests between CI providers (cfr. #201, for example)

cibuildwheel/__main__.py
```diff
@@ -198,6 +198,9 @@ def detect_obsolete_options():
         if deprecated in os.environ and 'manylinux1' in os.environ[deprecated]:
             print("Build identifiers with 'manylinux1' been deprecated. Replacing all occurences of 'manylinux1' by 'manylinux' in the option '{}'".format(deprecated))
```
YannickJadoul: `"Build identifiers with 'manylinux1' have been deprecated."` Not sure how we've overlooked that typo before.

cibuildwheel/__main__.py
```diff
@@ -198,6 +198,9 @@ def detect_obsolete_options():
...
             os.environ[deprecated] = os.environ[deprecated].replace('manylinux1', 'manylinux')
+        if deprecated in os.environ and ("macosx_10_6" in os.environ[deprecated] or "macosx_10_9" in os.environ[deprecated]):
+            print("Build identifiers with 'macosx_10_6' or 'macosx_10_9' been deprecated. Replacing with 'macosx' in the option '{}'".format(deprecated))
```
YannickJadoul: Same here, `"... have been deprecated."`

YannickJadoul: And maybe also `"Replacing all occurences with 'macosx' ..."`?

cibuildwheel/macos.py
```diff
@@ -14,11 +14,11 @@
...
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
-        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
+        PythonConfiguration(version='3.5', identifier='cp35-macosx_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.6', identifier='cp36-macosx_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
+        PythonConfiguration(version='3.7', identifier='cp37-macosx_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg')
```
YannickJadoul: Ha, funny. So Python has not adopted the new "macOS" name yet, but is still using `macosx`? Are we sure this is not going to change too soon?

test/10_cpp_standards/cpp11/setup.py
```diff
@@ -0,0 +1,9 @@
+import os, sys
```
YannickJadoul: Can we somehow merge all of these three standards into one `setup.py` and one `spam.cpp` file?

We could access an new environment variable `CIBW_TEST_CPP_STANDARD` being `11`, `14`, `17` (or `20`, soon :-) ), and forward this one to `extra_compile_args`.

We could then also pass that as preprocessor variable to `define_macros` and have a single `spam.cpp` file with some preprocessor `#if`? Ór, we could use `__cplusplus` to detect the C++ standard used in the compilation (https://github.com/pybind/pybind11/blob/deb3cb238a9f299d7c57f810165e90a1b14b3e6f/include/pybind11/detail/common.h#L30-L46)

YannickJadoul: Just an idea, though. It would reduce the amount of code we need to maintain, but I don't know how clean the result will look.

test/10_cpp_standards/cibuildwheel_test.py
```diff
@@ -0,0 +1,58 @@
+import os
+import sys 
+
+import pytest
+
+import utils
+
+
+def test_cpp11(tmp_path):
```
YannickJadoul: Why do we need this `tmp_path`?

test/shared/utils.py
```diff
@@ -36,17 +40,23 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
...
     subprocess.check_call(
-        [sys.executable, '-m', 'cibuildwheel', project_path],
-        env=env,
+        [sys.executable, '-m', 'cibuildwheel', "--output-dir", output_dir, project_path],
+        env=env
     )
 
 
-def expected_wheels(package_name, package_version):
+def expected_wheels(package_name, package_version, add_env=None):
```
YannickJadoul: Pffff, isn't there a better way? Can't we somehow pass `**kwargs` and just `tag_version = kwargs.get('macosx_tag_version')` and `'{package_name}-{package_version}-cp27-cp27m-macosx_' + (tag_version or '10_6') + '_intel.whl'`, or something like that?

I don't feel like `expected_wheels` should deal with environment variables.

test/10_cpp_standards/cibuildwheel_test.py
Czaki: to store result of different cibuildwheel call in different directories. It allows check if result is correct. Before this only first test check content  of wheelhouse directory, so there is no collision.

test/10_cpp_standards/cpp11/setup.py
Czaki: This will reduce amount of code, but increase complexity, so may make harder to debug. I'm not sure what will be better.

appveyor.yml
Czaki: Problem is that newer version of AppVeyor image do not have python 2.7 development tool. So most of test will fail. 
without this pin the tests may start failing for no apparent reason.

cibuildwheel/macos.py
Czaki: You see that this platform tag is used in installer name. And it is default name given by setuptools.

README.md
```diff
@@ -242,6 +242,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
+
+Building Python wheels with modern C++ (C++11) requires a few tricks.
+
+Python 2.7 wheels cannot use C++17 at all, because Python 2.7 uses the `register` keyword, which is prohibited in C++17 standard.
```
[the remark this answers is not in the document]
Czaki: I think that all technical details should be moved to documentation and only linked from readme.

README.md
```diff
@@ -242,6 +242,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
+
+#### Linux
+
+`manylinux2010` supports all C++ standards. If you would like to build `manylinux1` images, then your project is limited to C++11 and earlier.
```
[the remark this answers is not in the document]
Czaki: I see, but I write this text before manylinux2010 is merged. In few days I will rebase it on current master and move most of changes to docs.

Czaki:
@YannickJadoul 

> So if I understand correctly, all the actual work will be done by the wheel PR? And all this PR does is add and update the build ids, tests, and docs?

Yes

> PS: Will PyPI allow you to upload wheels with newer macOS versions, like '{package_name}-{package_version}-cp37-cp37m-macosx_10_14_intel.whl' for example?

I will check in few days. 

I do not know how to estimate time needed for next release of `wheel` with accepted changes. 

Upgrading this PR to next revision of master is time consuming. If there are next big changes planned which can conflict with this changes (with assumption that I will move most technical information to docs)?

YannickJadoul:
> If there are next big changes planned which can conflict with this changes

Not that I know of.

> Upgrading this PR to next revision of master is time consuming.

Maybe I have/could make some time, this weekend. Would you like me to try and rebase, or would you prefer to do it yourself?

Czaki:
> Maybe I have/could make some time, this weekend. Would you like me to try and rebase, or would you prefer to do it yourself?

If you have time. Working with text is my weakest side.

test/10_cpp_standards/cibuildwheel_test.py
YannickJadoul: But we don't need this in other tests. See for example `02_...`.

Should we then adapt the other tests, or is this change you made not really necessary because `pip wheel` will already build in a separate directory)?

test/10_cpp_standards/cpp11/setup.py
YannickJadoul: I'm not sure it will be so much more complex. Maybe we can give it a try and see?

YannickJadoul: (I can try, if you are OK with that?)

appveyor.yml
YannickJadoul: That might actually be good to know if the default version starts failing?
But let's see if @mayeut has any input here.

README.md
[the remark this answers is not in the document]
YannickJadoul: Yes, definitely. But this comment is about the content, not about how to merge this into the new docs.

README.md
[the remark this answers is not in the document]
YannickJadoul: This has got nothing to do with `manylinux2010`. This is about how you can make C++11/C++14 wheels in `manylinux2014`, through some tricks.

cibuildwheel/macos.py
YannickJadoul: I know. But there is no plan to change this soon, is there?

YannickJadoul: I know. But there is no plan to change this soon, is there?

test/10_cpp_standards/cibuildwheel_test.py
Czaki: I see that package name from `01_test` and `02_test` use same name and same version of package. 
When I run test on my personal computer I see that, when I use different names I got wheels for all names (you may see that i name projects `spam11`, `spam14` and `spam17`). 

So it looks like `02_test` may pass if it silently fail, because `01_test` already produce wheels in `wheelhouse`. 

I think that `cibuildwheel` should use `tmp_path` fixture or clean `wheelhouse` before test

test/10_cpp_standards/cpp11/setup.py
Czaki: control with environment variables?

cibuildwheel/macos.py
mayeut: I did not come across anything suggesting this. IMHO, I think it would be a nightmare to change this and Python/CPython/PyPA will probably stick with macosx (at least as long as possible). The amount of libraries failing with such a change would be huge and carefully communicated during BETA/RC phases.

> PS: Will PyPI allow you to upload wheels with newer macOS versions, like '{package_name}-{package_version}-cp37-cp37m-macosx_10_14_intel.whl' for example?

Short answer , yes. Long answer:
https://github.com/pypa/warehouse/blob/a37ea8d9e06361688a498fbcb83015c58fd9150d/warehouse/forklift/legacy.py#L116-L138

README.md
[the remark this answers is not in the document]
Czaki: I do not understand. I can build c++11 on `manylinux1`. `manylinux2014` is still not supported by `cibuildwheel` and why its need special tricks to use these standards.

test/10_cpp_standards/cpp11/setup.py
Czaki: it looks simple, so I push this change.

Czaki:
I do rebase and simplify C++ standards, but do some chaos in README (i do not wand to lost some information).
 
It need to be cleaned.

appveyor.yml
mayeut: It's in fact already set to `Visual Studio 2015` in master after #204 was merged.
An image specification was needed for windows / linux matrix builds. (as far as I understood the docs: https://www.appveyor.com/docs/appveyor-yml/)

Czaki: Deal with this will be strange. I all test add check of build system and ban python2.7 if condition. Brr.

README.md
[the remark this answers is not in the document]
zackw: `-Dregister=` or equivalent (that is, defining `register` as early as possible as a macro which expands to nothing) should also be an effective workaround for this unfortunate change in C++17.

test/10_cpp_standards/cibuildwheel_test.py
YannickJadoul: OK, yes, so we should make all the tests follow the same structure! But maybe that's a different PR, then.

cibuildwheel/macos.py
YannickJadoul: OK, great! Good to know, thanks :-)

test/10_cpp_standards/setup.py
```diff
@@ -0,0 +1,13 @@
+import os, sys
+from setuptools import setup, Extension
+import platform
+
+standard = os.environ["STANDARD"]
+
+language_standard = "/std:c++17" if platform.system() == "Windows" else "-std=c++" + standard
```
YannickJadoul: Doesn't Windows allow `/std:c++11` etc?

test/10_cpp_standards/cpp11/setup.py
YannickJadoul: Amazing, that's exactly what I was thinking of! Thanks for trying! -)

README.md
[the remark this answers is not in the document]
YannickJadoul: Good point! I think I was planning to do that, but then the compiler had a flag to just do so, so that was easier :-)

Bottom line: a lot of things are possible, but are not really supported in a straightforward way and need some hacks here and there. But I think we should be careful formulating what's not possible.

README.md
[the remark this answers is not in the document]
YannickJadoul: Sorry. that was a typo. I meant to say: "This is about how you can make C++11/C++14 wheels in `manylinux1`, through some tricks."

appveyor.yml
YannickJadoul: Ah right. But anyway, seems it's already fixed, then. :)

YannickJadoul:
Thanks, @Czaki! I'll see if I can make some time this weekend to look at the README and docs :-)

test/10_cpp_standards/setup.py
Czaki: Allow. My bad.

README.md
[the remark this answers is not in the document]
Czaki: I change code to test also c++17 against python 2.7. But not on windows.

Czaki:
I do few changes.  But I would mention that I working mainly on Linux and MacoOS. My work with windows is occasionally. So everything what I wrote about C++ standards compatibility on windows need to be checked.

PyBind11 have example how to force to compile old python with new MSVC  [here](https://github.com/pybind/python_example). But I do not found way to do it on CI (but I do not try hard). If there are here someone with better windows knowledge? Founding way to do this will simplity many things with future manage of windows configuration (also with pinning to given verion on AppVeyor)

YannickJadoul:
Nice! Thanks again, @Czaki. Once the docs are done, we should be getting close to finally being done with this all :-)

Looking at the past `wheel` releases (+- every 3 months), I wouldn't be surprised if your PR gets into a release soon, as well: https://github.com/pypa/wheel/releases

test/10_cpp_standards/cibuildwheel_test.py
mayeut: I'm working on adding a test for `manylinux2014` and I ran into the same issue. I'll try to make a PR that reworks a bit testing that should answer @Czaki remarks.

Czaki: I just finishing refactor of test. https://github.com/Czaki/cibuildwheel/tree/test_fix

test/shared/utils.py
Czaki: I do this with enviroment because I think aboutt fileterin result base on `CIBW_BUILD` and `CIBW_SKIP`, but not sure if it is proper approach.

YannickJadoul: Hmmm, I don't really like that. Íf it needs to be filtered there (and I'm not really convinced of that either), we should at least pass a `BuildSelector` object.

Czaki: Ok. I will wait with any changes in code till the decision about which test PR is accepted.

YannickJadoul: Sure, no rush!

YannickJadoul: (It's also a minor detail, compared to all the work you have done in this PR, ofc, but since there's still time to fix things, ...)

YannickJadoul:
@Czaki @zackw I'm having a similar problem, now, on #185: inside `cibuildwheel`, I'm setting `MACOSX_DEPLOYMENT_TARGET` to 10.13 (as temporary workaround; PyPy's setting this internally correctly, next release), but I get the following error:

`error: $MACOSX_DEPLOYMENT_TARGET mismatch: now "10.13" but "10.14" during configure`

I'm thinking this is the same error as @zackw had?
Any insights in what to do?

Czaki:
@YannickJadoul Similar. To late definition of MACOSX_DEPLOYMENT_TARGET and pypy is build against higher version. I add comment in this PR

YannickJadoul:
@Czaki Thanks a lot! I replied there, as well. See https://github.com/joerick/cibuildwheel/pull/185#discussion_r347433262, if @zackw is interested in following.

Czaki:
@YannickJadoul I move c++ standards description to docs.

YannickJadoul:
Sorry this took so long, @Czaki, but I've finally taken the time to add some more things to the documentation!

If taken the liberty to push to your branch; I hope this is OK with you? Please change/remove the commit if you're not happy with it, of course :-)

I have one more general remark/question: should we have a separate page for C++ standards? Or should we make a page for all kinds of build issues, also for other programming languages/build systems in the future? It's just that it feels a bit weirdly specific to have the modern C++ there. @joerick, any thoughts on this?

test/10_cpp_standards/cibuildwheel_test.py
YannickJadoul: So is this now OK, or do we need to still fix this to work with the refactoring of the tests we did some weeks ago?

test/shared/utils.py
```diff
@@ -111,12 +111,21 @@ def expected_wheels(package_name, package_version):
             '{package_name}-{package_version}-cp38-cp38-win_amd64.whl',
         ]
     elif platform == 'macos':
+        if macosx_deployment_target is not None:
+            tag = macosx_deployment_target.replace(".", "_")
+            tag1 = macosx_deployment_target.replace(".", "_")
+        elif "MACOSX_DEPLOYMENT_TARGET" in os.environ:
```
YannickJadoul: Why are we checking this? It would somehow feel a lot cleaner if tests don't check the environment but always make sure they get the same output?

Czaki: I do not use `"MACOSX_DEPLOYMENT_TARGET` in test. I introduce this if to test not fail if `"MACOSX_DEPLOYMENT_TARGET` is defined in user environment. 
Other approach is to modify environment in test call to be sure  that this environment variable is not defined.

Czaki:
Travis fail because of get-pip.py download fail.

test/shared/utils.py
YannickJadoul: Hmmm, myeah, I see. I'm not entirely convinced, but it would be nice to be independent from the environment, no? I guess in that case, we'd add a line like `env.pop('MACOSX_DEPLOYMENT_TARGET', None)` in `run_cibuildwheel`?

YannickJadoul:
> Travis fail because of get-pip.py download fail.

Restarted. Seems to be going well now :-)

test/shared/utils.py
YannickJadoul: OK, I've tried this. Let's see what the CIs say :-)

Czaki: In your commit there is no removing `MACOSX_DEPLOYMENT_TARGET` from environment.  

the c++ standards tests are not problematic because `MACOSX_DEPLOYMENT_TARGET` is set inside. Problem is with all other tests.

YannickJadoul: @Czaki, I did, no?

https://github.com/joerick/cibuildwheel/pull/156/commits/fc67e3db947e3ba9186fe39d1010d47300101fbe#diff-8437219d14af4459b32737acc9521403R57

Czaki: My bad. I removed this commit.

YannickJadoul: But tests still fail? :-(

I don't have rights on CircleCI, I'm afraid. :-/ Maybe amend the commit without changes and force push again?

Czaki: It is canceled because I push next commit. 

I restart all tests with force push.

YannickJadoul: Great, let's see :-)

YannickJadoul: Do you agree with these changes, btw? Or do you consider the previous solution to be better?
I mainly just wanted to try this out, and see if it would look better; we can still revert it if you want.

Czaki: I agree. It is better, because it still check version forced by choosing python installer and reduce change of behavior tests between environments.

YannickJadoul: OK, great, then! :-) Then we mainly just have to wait for a new wheel release. Wondering if that's getting close; should be, I hope.

Czaki: I see one problem. If there will be a test that will run cibuildwheel in different way than using function from `utils` then this protection will not be apply.  But maybe it needs only mention in Readme?

YannickJadoul: I wouldn't put it in the README. This is more of a development thing, right?

I wouldn't really know where to mention this, actually. Maybe we should just solve the problem when/if it occurs in some PR?

Czaki: It think about this readme https://github.com/joerick/cibuildwheel/blob/master/test/README.md

other approach is to put something like this: 
```
@pytest.fixture(autouse=True)
def no_macosx_deployment_target(monkeypatch):
    """remove MACOSX_DEPLOYMENT_TARGET from environment"""
    if "MACOSX_DEPLOYMENT_TARGET" in os.environ:
        env = os.environ.copy()
        del env["MACOSX_DEPLOYMENT_TARGET"]
        monkeypatch.setattr(os, "environ", env)
```
in conftest.py 

But maybe keep in memory this edge case is the best solution.

YannickJadoul: Oh, I had no idea that was there!
Well, maybe. But I think normally, to make a new test, you copy the simple test `01_basic` anyway, and start working from there?

The one thing I was unsure about was if someone passes `env`. But then again, worst case, tests break and we'll fix something. It's still much better to have a test fail when there's no error, than to have them succeed when there is an error.

YannickJadoul: Actually, I do like your suggestion/solution with the fixture as well. But it's a bit more hidden away because it's in `conftest.py`

Czaki: Ok. I'm quite new in such projects and not sure how strong protection should be.  I'm ok with all proposition.

YannickJadoul:
Argh. Mind if I rebase and force-push, @Czaki?

Also, no new on a wheel release? :-(

Czaki:
I do not know. It is merged to wheel master. But yet no release.

YannickJadoul:
Good news: https://github.com/pypa/wheel/issues/328#issuecomment-577120524

Czaki:
@joerick @YannickJadoul 
Today is that day https://github.com/pypa/wheel/releases/tag/0.34.0

YannickJadoul:
HURRAAAAAAY!

MERGE! MERGE! MERGE!

Czaki:
Hmm. If I good remember threr was a try of merge cpp standards tests in one? Or not?

And if yes then it was somehow lost during some rebase? 
@YannickJadoul

joerick:
KAPOW! 👏 👏 👏  Congrats @czaki and @YannickJadoul ! 👏 👏 👏  Only 5 months in the making!

joerick:
Now to figure out our strategy for the next release! #220 is a must, #194, #185 and #185 would be great too, and we should get in #212 while we can :)

YannickJadoul:
Let's see what can be rebased :-)

PR #157, opened 2019-08-29
Add Twisted to working examples list

altendky:


altendky:
I don't know where to put this but you could add some examples if you want.  You have to do a bit of CI setup (create a build, hook to my repo, add some credentials) but between https://github.com/altendky/romp and cibuildwheel here, a full suite of good wheels is one command line away.  I still find it really slick what you've done here.  Thanks as always.

https://github.com/sqlalchemy/sqlalchemy/issues/4827

YannickJadoul:
@altendky `romp` looks really cool! I think we should add it to the example, or at least refer to it at the bottom of the README.

Or, maybe we can even use this in `cibuildwheel`. This 'quick build' command from #74 or something similar might really benefit from this?

altendky:
It certainly relates though it doesn't let you iterate the actual CI config.  Just files and a command.  I use it for my locking and for experiments especially when I am not working on a project with CI (like checking random things for people in #python).

joerick:
Seconded, Romp looks great @altendky ! I'm wondering if we could tie the two tools together into a great UX that allows people to iterate on config, or to build wheels on-demand without a CI. 

> doesn't let you iterate the actual CI config

Why is this? Because it doesn't allow setting environment variables?

altendky:
> > doesn't let you iterate the actual CI config
> 
> Why is this? Because it doesn't allow setting environment variables?

I mean that if you have an `azure-pipelines.yml` that that doesn't get used at all.  Romp provides it's own CI config that creates the build matrix and runs the requested `--command` in each (and reads in submitted files and collects artifacts and...).  So romp lets you 'run a command' not 'let the server run your CI config'.

joerick:
Ah, yes. Okay, cool! If, let's say, cibuildwheel could pull its config out of `pyproject.toml` or `setup.cfg`, this could become an interesting proposition...

altendky:
I may lose battle (that I'm not really fighting anyways) but I tend to think that a common directory for configs makes more sense than a common file.  There was just so much breakage as pip started doing that 517/518 stuff and just having even an pyproject.toml triggered it.  So, I'm not big on one common config, would prefer a per-tool config I think.  But yeah, not necessarily where things are going so I wouldn't put up a real fight or anything.

joerick:
(this is an aside, but) I totally agree with you, I've always been partial to an `/etc` directory in projects, where we could chuck all those pesky tool config files. Maybe one day! In the absence of such a thing, I think tool config in `pyproject.toml` is better than individual config files cluttering up the root of a project, but I guess there's no reason why cibuildwheel couldn't read a `cibuildwheel.toml` at the root if the community prefers it!

PR #160, opened 2019-09-04
Build Python wheels for windows on travis

Czaki:
This changes allow to build parts of windows wheels on travis-ci. 
Supported python version: 3.5+
For python 2.7 I do not found method to install  VCForPython27.msi. Try of installation with msiexec stuck without any action  https://travis-ci.com/Czaki/cibuildwheel/builds/125771121
Try of installation it from choco fail because bug in DotNet3.5 formula. https://travis-ci.com/Czaki/cibuildwheel/builds/125753492

For python 3.4 there is collision with python installed by some tool under strange path. And I do not know how to write script for search its path. The collision is only for 32 bits version, butt I decided to remove both. 

on the end of yaml file I put the lines which fails. Maybe someone with better windows knowledge find a way to solve it. 

If there will be decision to not fight for old releases then also is possible to use https://www.nuget.org/packages/python/ as python source (which is suggested in python docs). But it contains only python from version 3.5.


YannickJadoul:
@Czaki Amazing, thanks! Looks pretty good to me, as well.

I have one major suggestion: as far as I know, part of the idea is to abstract away the management of different versions of Python. So just like we download and install all the different pkg-files on macOS, it would be `cibuildwheels` philosophy to install the wheels inside of the `cibuildwheel` script (right, @joerick?).
Any idea how feasible that would be, @Czaki ?

Czaki:
I change code to install python inside cibuildwheel. Initial python still need to be installed manually.

YannickJadoul:
@Czaki Great, thank you very much! I'll have a close-up review later :-)

Czaki:
@YannickJadoul You can also look on https://github.com/Czaki/cibuildwheel/tree/windows_on_travis2 branch. There is try of better protection of collision with existing installation based on [https://www.python.org/dev/peps/pep-0514/](pep 514).

Problem with python 3.4 is that there exist this version installation in windows image. But hidden on very strange path and windows installers do not allow to add second copy. In future this tools may upgrade python version and generate problems. 

I do not found pep 514 entries for python 3.4, but maybe this protection will be usefull? If yes then I will clean the code. Or maybe You know how find existing python installation on windows.

joerick:
This is heading in a good direction! Agree that installing within pyinstrument is the right approach - the user shouldn't be concerned with that.

I have lots of smaller comments on the diff as it is now, but before that, would it be easy/possible to install Python directly from downloads from python.org rather than using `choco`? That would be preferable to me, since it reduces our dependency on other tools, which can be slow to publish packages, and they add potential security holes to the build process.

Czaki:
I have bad experience with direct calling `msiexec` because it blocks all process  without any error message, but I will try this in this week.

Czaki:
@joerick What do you think about usage of nuget for installing? It is recommended in python docs as method of installng python for CI.
https://docs.python.org/3/using/windows.html#windows-nuget

.travis.yml
```diff
@@ -4,29 +4,50 @@ matrix:
   include:
     # Linux Python 2
     - sudo: required
+      if: branch=master
```
YannickJadoul: Let's not forget about this when merging :-)

cibuildwheel/windows.py
```diff
@@ -31,26 +37,51 @@ def get_python_path(config):
...
 
+        
 
 def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'path'])
+
+    bace_choco_args = "--no-progress --force -y --allowmultiple --override".split() 
+    instal_args = "'/quiet  InstallAllUsers=1 TargetDir={}'"
+    
+    class PythonConfiguration(object):
```
YannickJadoul: Just a though: could we keep this as a simple `namedtuple`, and format/create the full args when we need them.

We could even just make a simple function `get_choco_args` that takes the `version` and `arch` and returns all the arguments? Then we don't need to change the current `PythonConfiguration`.

test/shared/utils.py
```diff
@@ -36,12 +37,12 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
     if add_env is not None:
         env.update(add_env)
 
-    subprocess.check_call(
+    p = subprocess.check_call(
```
YannickJadoul: What are these changes?

YannickJadoul:
@joerick @Czaki The current, very incomplete Travis CI docs for Windows seem to suggest using chocolatey: https://docs.travis-ci.com/user/reference/windows/

.travis.yml
```diff
@@ -4,29 +4,50 @@ matrix:
...
       env: PYTHON=python3
       before_install:
         - brew update
         - brew outdated python || brew upgrade python
 
+    - os: windows
+      language: shell
+      before_install:
+       - choco install python3 --version 3.5.4 --no-progress --force -y --allowmultiple --override --installargs "'/quiet  InstallAllUsers=1 TargetDir=C:\Python35-x64'"
```
YannickJadoul: Just for future reference, is there a reason we need all of these different arguments, here?
https://docs.travis-ci.com/user/languages/python/ seems to be suggesting to just `choco install python` (or maybe `python3`, I guess) and use the `python` executable afterwards.

I understand we need more flags/control for the different versions, but for the one that's just running `cibuildwheel` in the examples, it would be great to keep things simple.

Czaki: `--no-progress` means not show progress of download. In other cases the 4MB limit of log on travis may run out. 
`-y` accept all questions during installation.  
`-allowmultiple` allow multiple versions of package. Not sure if need here. Needed with next installations
`--override --installargs "'/quiet  InstallAllUsers=1 TargetDir=C:\Python35-x64'"` - set install directory

test/shared/utils.py
Czaki: I will remove this. Artefact after debug.

cibuildwheel/windows.py
```diff
@@ -31,26 +37,51 @@ def get_python_path(config):
...
+        PythonConfiguration(version='3.6.x', arch="32", identifier='cp36-win32', path='C:\Python36', choco_args="python3 --version 3.6.8 --x86"),
+        PythonConfiguration(version='3.6.x', arch="64", identifier='cp36-win_amd64', path='C:\Python36-x64', choco_args="python3 --version 3.6.8"),
+        PythonConfiguration(version='3.7.x', arch="32", identifier='cp37-win32', path='C:\Python37', choco_args="python3 --version 3.7.4 --x86"),
+        PythonConfiguration(version='3.7.x', arch="64", identifier='cp37-win_amd64', path='C:\Python37-x64', choco_args="python3 --version 3.7.4")
     ]
 
-    if IS_RUNNING_ON_AZURE:
+    if IS_RUNNING_ON_AZURE or IS_RUNNING_ON_TRAVIS:
         # Python 3.4 isn't supported on Azure.
+        # I meet problem with install python on travis.
```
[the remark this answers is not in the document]
Czaki: Exact problem is that python installer do not allow to install two instances of same python version. One of pre-installed program has own instance of python 3.4 x64. But it is installed under strange path. I try to write part of code for discovery installation [https://github.com/Czaki/cibuildwheel/commit/b0cefb63c75e17952a19dabe4ffe410b75af3254#diff-bf885b72fecfdf7485cab147f7957d5c](here) but all info which i found in internet is connected with pep 514, but python 3.4 do not satisfy it.  

should I describe it in comment inside code?

test/shared/utils.py
YannickJadoul: Oh, don't worry. Sorry to be too quick about it!

cibuildwheel/windows.py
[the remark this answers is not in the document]
YannickJadoul: Aha, I see. It's probably just fine to write a single line that this version clashes with an already installed version of Python 3.4 on Travis CI's Windows image. Thanks!

joerick:
@Czaki is this ready for review? Let me know when and I'll take a proper look at the code :)

Czaki:
Generally yes. 
In second PR there shown this link https://github.com/pybind/python_example that show some method which will allow for enable support of python 2.7 on travis. But I think that this change can be target of next PR with enhance of c++ modern standards support.

README.md
```diff
@@ -123,6 +127,16 @@ jobs:
       - cibuildwheel --output-dir wheelhouse
     ```
 
+  and matrix entry for windows
```
```suggestion
  To build on Windows too, add this matrix entry:
```

README.md
```diff
@@ -123,6 +127,16 @@ jobs:
...
+  and matrix entry for windows 
+  ```yaml
+    - os: windows
+      language: shell
+      before_install:
+       - choco install python3 --version 3.6.8 --no-progress -y
+      env:
+       - PATH=/c/Python36:/c/Python36/Scripts:$PATH
+  ```
+
```
```suggestion

  Note that building Windows Python 2.7 wheels on Travis is unsupported.
```

cibuildwheel/windows.py
```diff
@@ -77,10 +106,32 @@ def shell(args, env=None, cwd=None):
...
+        # instal nuget as best way for provide python
+        call(["choco", "install", "nuget.commandline"])
+        # get pip fo this installation which not have. 
+        get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
+        get_pip_script = 'C:\get-pip.py'
+        call(['curl', '-L', '-o', get_pip_script, get_pip_url])
+
+    python_configurations = get_python_configurations(build_selector)
     for config in python_configurations:
+        print(config, file=sys.stderr)
```
joerick: Can we remove this log?
```suggestion
```

cibuildwheel/windows.py
```diff
@@ -32,32 +44,49 @@ def get_python_path(config):
...
+    if configuration.exact_version is None:
+        return None
+    python_name = "python" if configuration.version[0] == '3' else "python2"
+    if configuration.arch == "32":
+        python_name = python_name + "x86"
+    return [python_name, "-Version", configuration.exact_version, "-OutputDirectory", "C:/python"]
+
 def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'path'])
+    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'path', "exact_version"])
```
joerick: Perhaps rename `exact_version` to `nuget_version`, since I don't think it's accurate in e.g. Appveyor

cibuildwheel/windows.py
```diff
@@ -77,10 +106,32 @@ def shell(args, env=None, cwd=None):
...
+
+    python_configurations = get_python_configurations(build_selector)
     for config in python_configurations:
+        print(config, file=sys.stderr)
         config_python_path = get_python_path(config)
+        if IS_RUNNING_ON_TRAVIS and config.exact_version is not None and not os.path.exists(config_python_path):
+            call(["nuget", "install"] + get_nuget_args(config))
+            if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
+                call([os.path.join(config_python_path, 'python.exe'), get_pip_script ])
+            print(config_python_path, file=sys.stderr)
```
joerick:
```suggestion
```

cibuildwheel/windows.py
```diff
@@ -22,6 +28,12 @@ def get_python_path(config):
             return glob(path_pattern)[0]
         except IndexError:
             raise Exception('Could not find a Python install at ' + path_pattern)
+    elif IS_RUNNING_ON_TRAVIS:
+        if config.version == "3.4.x":
+            return config.path
```
joerick: Better to return `None` here I think

cibuildwheel/windows.py
```diff
@@ -107,8 +158,12 @@ def shell(args, env=None, cwd=None):
         shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
 
         # prepare the Python environment
-        shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'],
+        if config.version == "3.4.x":
+            shell(['python', '-m', 'pip', 'install', 'pip==19.1.1'],
               env=env)
+        else:
+            shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'],
+                env=env)
```
joerick: Is this change still needed?

test/shared/utils.py
```diff
@@ -85,9 +86,12 @@ def expected_wheels(package_name, package_version):
     else:
         raise Exception('unsupported platform')
     
-    if IS_RUNNING_ON_AZURE:
+    if IS_WINDOWS_RUNNING_ON_AZURE or IS_WINDOWS_RUNNING_ON_TRAVIS:
         # Python 3.4 isn't supported on Azure.
         templates = [t for t in templates if '-cp34-' not in t]
+    if IS_WINDOWS_RUNNING_ON_TRAVIS:
+        # Python 2.7 and 3.4 isn't supported on Travis.
+        templates = [t for t in templates if '-cp27-' not in t and '-cp34-' not in t]
```
joerick: The 3.4 exclusion is already covered by the previous `if` statement

cibuildwheel/windows.py
```diff
@@ -77,10 +108,21 @@ def shell(args, env=None, cwd=None):
...
 
-    python_configurations = get_python_configurations(build_selector)
+    def call(args, env=None, cwd=None, shell=False):
+        # print the command executing for the logs
+        if shell:
+            print('+ %s' % args)
+        else:
+            print('+ ' + ' '.join(shlex_quote(a) for a in args))
+
+        return subprocess.check_call(args, env=env, cwd=cwd, shell=shell)
```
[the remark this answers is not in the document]
Czaki: No. My mistake. Why using different names in macos and windows file?

.travis.yml
```diff
@@ -27,6 +27,24 @@ matrix:
         - brew update
         - brew outdated python || brew upgrade python
 
+    - os: windows
+      language: shell
+      env: VS100COMNTOOLS=$VS140COMNTOOLS
```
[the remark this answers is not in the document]
Czaki: VS100COMNTOOLS=$VS140COMNTOOLS forces to use Visual studio 14 instead of Visual Studio 10. 

But if we do not try to support Python 3.4 it is not required.

.travis.yml
```diff
@@ -27,6 +27,24 @@ matrix:
...
 
 script: $PYTHON ./bin/run_tests.py
+
+
+# - choco install python3-x86_32 --version 3.4.3.20150501 --no-progress  -y --allowmultiple --override --installargs "'/qn /norestart ALLUSERS=1 TARGETDIR=C:\Python34'"
+# - choco install python3 --version 3.4.4.20180111 --no-progress --force -y --allowmultiple --override --installargs "'/quiet  InstallAllUsers=1 TargetDir=C:\Python34-x64'"
+# - choco install python2 --version 2.7.16 --no-progress --x86 -y --allowmultiple --override --installargs "'/quiet  InstallAllUsers=1 TargetDir=C:\Python27'"
+# - choco install python2 --version 2.7.16 --no-progress -y --allowmultiple --override --installargs "'/quiet  InstallAllUsers=1 TargetDir=C:\Python27-x64'"
+#       - wget -q http://download.microsoft.com/download/7/9/6/796EF2E4-801B-4FC4-AB28-B59FBF6D907B/VCForPython27.msi
+#       - msiexec /i VCForPython27.msi ALLUSERS=1 ACCEPT=YES /passive
```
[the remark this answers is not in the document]
Czaki: yes

cibuildwheel/windows.py
Czaki: This case allow someone, who really  python 3.4 to install it manually and use. The result of this function in rest of code is not checked and is treated as string. I think that it is better to left this or literally raise the exception than return `None`

cibuildwheel/windows.py
Czaki: This is not connected with travis but generally with python. From version 19.2 `pip` drop support of python 3.4. and I see errors in test when calling upgrade. on python 3.4

test/shared/utils.py
Czaki: I removed it from previous statement, because whole comment is in this if branch

cibuildwheel/windows.py
Czaki: Agree

cibuildwheel/windows.py
Czaki: yes

README.md
Czaki: ok

cibuildwheel/windows.py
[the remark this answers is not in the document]
Czaki: This is old version of text? Can you check current?

cibuildwheel/windows.py
joerick: Okay, fair enough. Btw, do you have a specific use case for Python 3.4? We're going to be removing it very soon from cibuildwheel since it hit end-of-life this year (#168)

cibuildwheel/windows.py
joerick: I see, thanks

cibuildwheel/windows.py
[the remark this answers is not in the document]
joerick: Err... I think because the `call(...)` functions run programs directly unless `shell=True` (similar to subprocess.call), but the `shell(...)` function always runs via a shell, because they're called using `cmd /C`

cibuildwheel/windows.py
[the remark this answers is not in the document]
joerick: looks good

test/01_basic/cibuildwheel_test.py
```diff
@@ -19,4 +19,5 @@ def test_build_identifiers():
     # identifiers
     expected_wheels = utils.expected_wheels('spam', '0.1.0')
     build_identifiers = utils.cibuildwheel_get_build_identifiers(project_dir)
+    print("bb", build_identifiers)
```
YannickJadoul: What's this, still?

test/shared/utils.py
```diff
@@ -38,10 +39,10 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
 
     subprocess.check_call(
         [sys.executable, '-m', 'cibuildwheel', project_path],
-        env=env,
+        env=env
     )
-
-
+ 
+
```
YannickJadoul: A small spurious change left?

cibuildwheel/windows.py
```diff
@@ -1,12 +1,18 @@
...
+try:
+    from shlex import quote as shlex_quote
+except ImportError:
+    from pipes import quote as shlex_quote
+
 from .util import prepare_command, get_build_verbosity_extra_flags
 
 
 IS_RUNNING_ON_AZURE = os.path.exists('C:\\hostedtoolcache')
+IS_RUNNING_ON_TRAVIS = os.environ.get('TRAVIS_OS_NAME') == 'windows'
```
YannickJadoul: Any reason not to use the constants from `utils.py`?

cibuildwheel/windows.py
Czaki: No. I only use python 3.6+.

cibuildwheel/windows.py
Czaki: Which utils? This from test?

YannickJadoul: Oh, yes, that's code from the tests. Nevermind, my bad. Sorry!

test/01_basic/cibuildwheel_test.py
Czaki: Sorry.

YannickJadoul: No problem. Very easily overlooked. I was just lucky to notice it ;-)

PR #161, opened 2019-09-07
Add support for specifying test-time dependencies via extras_require

astrofrog:
For a number of packages I work on, the test dependencies can be installed via an [extras_require](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) option called ``test``, e.g.:

```
pip install astropy[test]
```

This PR adds support for a new environment variable ``CIBW_TEST_EXTRAS`` that can be used to make sure the wheel is installed with e.g.:

```
pip install <wheel_file>[test]
```

This then means that we can avoid duplication of listing test-time dependencies in ``CIBW_TEST_REQUIRES`` if they are already defined in ``setup.py`` or ``setup.cfg``.

Open questions:

* Does the environment variable name make sense? I was originally thinking ``CIBW_TEST_EXTRAS_REQUIRE`` but that's a bit long.
* Another approach would be to allow some special syntax in ``CIBW_TEST_REQUIRES`` but that seems more error-prone.

Of course feel free to close if you think you'd prefer not to have this, but it is a pretty simple change that can greatly simplify configuration in some cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,7 @@ All being well, you should get wheels delivered to you in a few minutes.
 |   | `CIBW_MANYLINUX1_I686_IMAGE` | Specify an alternative manylinux1 i686 docker image |
 | **Tests** | `CIBW_TEST_COMMAND` | Execute a shell command to test all built wheels |
 |   | `CIBW_TEST_REQUIRES` | Install Python dependencies before running the tests |
+|   | `CIBW_TEST_EXTRAS` | Install Python dependencies before running the tests using ``extras_require``|
 
 A more detailed description of the options, the allowed values, and some examples can be found in the [Options](#options) section.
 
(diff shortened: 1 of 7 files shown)


YannickJadoul:
@astrofrog Looks great to me :-)

Minor remark, at some point I also tried to clean up trailing spaces, but @joerick asked me to revert those changes. I've forgotten the exact reason (and can't find the PR anymore), though, but it might be a good idea to keep the changes of this PR small, to get a better overview later, anyway.

astrofrog:
@YannickJadoul - thanks! I just rebased and removed the changes to the whitespace

YannickJadoul:
@astrofrog Great! As far as I'm concerned this is ready to go in, but let's give @joerick a chance to get a look and merge. He's the one that needs to craft a new release, anyway, before you can easily use it from PyPI :-)

joerick:
All looks good! Thanks especially for the docs update.

Issue #121, opened 2019-01-09
Tests should be run in a clean environment


astrofrog:
Currently tests are run in the same environment that ``BEFORE_BUILD`` was run in, but it would be good to try installing the wheel into a clean environment to make sure that all the correct dependencies are installed and don't rely on ``BEFORE_BUILD``.

joerick:
Good point! Did you get burned by this?

astrofrog:
Somewhat - I have to currently install a more recent version of numpy manually in the test phase (since I want to test with the most recent version, not the version used for the build). So there's a workaround, but ideally we should fix this in cibuildwheel.

YannickJadoul:
Then a `CIBW_BEFORE_TEST` would also be nice to have? (Maybe even defaulting to `CIBW_BEFORE_BUILD` for backwards compatibility?)

joerick:
Maybe CIBW_TEST_REQUIRES supports this already? Ask for a version that's newer than whatever your built with?

astrofrog:
@joerick - yes I can use ``CIBW_TEST_REQUIRES`` in my case but it's more that conceptually the test and build phases should be separate to make sure that the dependencies in the package are correctly defined. For example if the fast-histogram package didn't define the numpy depdendency in ``install_requires``, I wouldn't have noticed (so I'm just pre-empting for other packages in future).

joerick:
Yep, I agree on the broader point about test environments. 

I suppose the solution is to create a venv, install our new wheel in there and run the tests using that.

Aside: I've never been totally happy with the strategy of moving out of the project dir and running the tests from elsewhere to avoid picking up the module (and using the {project} keyword in the test command). If the isolated test env could improve that too somehow, that would be great.

illume:
tox might be an option for creating different environments with different pythons and running tests separately.

joerick:
Fixed in #164, release should be soon!

PR #162, opened 2019-09-07
Run tests in clean virtual environment

astrofrog:
This fixes https://github.com/joerick/cibuildwheel/issues/121 by making it so that once the wheels are built, the tests are run in an isolated environment. This ensures isolation from the build environment and guarantees that the package is correctly defining its build and test-time dependencies.

I originally attempted to solve this with direct calls to virtualenv, but ran into issues with some of the CI tests here, which suggested to me that this was not a robust solution. I therefore have now taken the approach of using tox. The process is as follows:

* Create an empty temporary directory
* Write a simple tox configuration to that directory that installs the wheels, the test dependencies, and runs the test command
* Run tox in that directory

The tox configuration whitelists all external commands since it's hard to know what non-Python commands people might be relying on. For now we also do not pass any environment variables into the test environment.

If there are concerns that this could break existing projects (it's always hard to account for all corner cases) this behavior could be made opt-in, with the previous behavior being the default. Another idea would be to make it opt-in initially, and then choose to make it the default later once it's been stress tested by some projects.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from textwrap import dedent
 import os, subprocess, sys, uuid
 from collections import namedtuple
 from .util import prepare_command, get_build_verbosity_extra_flags
@@ -51,6 +52,9 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         if not platform_configs:
             continue
 
+        test_requires_cmd = 'pip install ' + ' '.join(test_requires) if test_requires else '#'
+        test_cmd_prepared = "sh -c '{0}'".format(prepare_command(test_command, project='/project')) if test_command else '#'
+
         bash_script = '''
             set -o errexit
             set -o xtrace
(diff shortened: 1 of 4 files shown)


astrofrog:
There are some CI failures here related to the ``cp27-cp27mu`` Linux wheels, investigating.

astrofrog:
I've managed to fix most CI issues but there's still one failure on Travis, which I'll investigate shortly.

astrofrog:
This is turning out to be more complicated than I thought it would be, so I'll need to put this on pause for a while!

cibuildwheel/linux.py
```diff
@@ -85,40 +89,48 @@ def build(project_dir, output_dir, test_command, test_requires, before_build, bu
...
+            envlist = test
+            skipsdist = true
+
+            [testenv]
+            whitelist_externals = *
+            commands =
+                pip install $delocated_wheel
+                {test_requires_cmd}
+                {test_cmd_prepared}
+            EOL
```
joerick: I'm not familiar with tox, so bear with me - what is tox doing for us here? I was expecting to see some virtualenv commands, but I guess tox is handling that on our behalf using these options. Could it perhaps be clearer by doing running the virtualenv commands directly? 

(In my head it's just a `virtualenv $test_dir/env`, then a subshell that calls `source $test_dir/env/bin/activate` before the package-specific stuff.... that feels clearer to me, and requires less knowledge)

joerick: Apologies, I now see your note in the PR. I can't see the previous attempts using direct virtualenv calls in this PR, I'd be curious what issues you ran into.

astrofrog: @joerick - I still have a backup of the virtualenv version locally so can push it to a branch for you to look at shortly. Essentially the issue is that sometimes when the virtualenv was set up, ``python -m pip`` would still point to the system install even though ``python`` pointed to the virtualenv version, and I couldn't find a way around it.

joerick: @astrofrog please do - I might be able to help you on that!

astrofrog: @joerick - I just opened https://github.com/joerick/cibuildwheel/pull/164 which has my original attempt with virtualenv alone.

joerick:
Closing in favour of #164

PR #168, opened 2019-09-28
Drop python 3.4

mayeut:
Python 3.4 reached the end of its life on 16 March 2019.
It's about time to drop support for Python 3.4.

c.f. https://hugovk.github.io/drop-python/3.4/


joerick:
I agree. 

I'm wondering if a `v1.0` is brewing, with docs, manylinux2010, etc... maybe this can wait for that release!

Issue #170, opened 2019-10-15
Add support for Python 3.8


Legrandin:
Python 3.8 was released on [October 14th, 2019](https://docs.python.org/3/whatsnew/3.8.html).

Can it be added to `cibuildwheel`?

YannickJadoul:
Yessss, thanks for the reminder! Should be reasonably quick to implement, though it depends a bit on the Python versions AppVeyor and Azure are offering

YannickJadoul:
@Legrandin, see #172 to follow progress :)

So far, it seems that Azure Pipelines and AppVeyor do not support Python 3.8 on Windows, yet.

joerick:
Fixed in #180. Release should be within the next couple weeks...!

MuellerSeb:
> Fixed in #180. Release should be within the next couple weeks...!

Are there any news on when to release it?

YannickJadoul:
@MuellerSeb In the meanwhile, `pip3 install git+https://github.com/joerick/cibuildwheel.git` should work, though?

MuellerSeb:
> @MuellerSeb In the meanwhile, `pip install git+https://github.com/joerick/cibuildwheel.git` should work, though?

Since I want to use it in CI, I'll wait for the next official release, to stick to one version.

joerick:
I want to be sure that  #175 isn't a problem and merge #169 before release... so I can't give an exact date, but within a couple weeks :)

MuellerSeb:
I'm cool with that. Thanks for your great work!

YannickJadoul:
@joerick In principle, the docs don't need to be ready before release, right? Commits to `README` on the master will show up immediately on GitHub, and the readthedocs can build independently of a release?

PR #172, opened 2019-10-15
Adding Python 3.8

YannickJadoul:
Added a 3.8 or 38 for every 3.7 or 37 currently present in master...
Let's see what works; let's see what breaks!

Cfr. #170

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -19,7 +19,7 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
-        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_intel', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
     ]
 
     # skip builds as required
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -19,6 +19,7 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_intel', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
     ]
 
     # skip builds as required
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ What does it do?
 | Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
 
 > ¹ Not supported on Azure Pipelines 
 >
@@ -87,6 +88,8 @@ jobs:
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x64}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x86}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x64}}
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x86}}
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x64}}
     - script: choco install vcpython27 -f -y
       displayName: Install Visual C++ for Python 2.7
     - bash: |
@@ -312,7 +315,7 @@ When both options are specified, both conditions are applied and only builds wit
 
 The format is `python_tag-platform_tag`. The tags are similar but not identical to the ones defined in [PEP 425](https://www.python.org/dev/peps/pep-0425/#details).
 
-Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37`
+Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37` `cp38`
 
 Platform tags look like `macosx_10_6_intel` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
 
--- a/README.md
+++ b/README.md
@@ -88,8 +88,12 @@ jobs:
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x64}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x86}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x64}}
+    #- {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x86}}
+    #- {task: UsePythonVersion@0, inputs: {versionSpec: '3.8', architecture: x64}}
+    - powershell: |
+        curl -o install_python.ps1 https://raw.githubusercontent.com/matthew-brett/multibuild/11a389d78892cf90addac8f69433d5e22bfa422a/install_python.ps1
+        & { $env:PYTHON = 'C:\\Python38'; .\install_python.ps1; Copy-Item 'C:\\Python38' 'C:\\hostedtoolcache\\windows\\Python\\3.8.0\\x86' -Recurse }
+        & { $env:PYTHON = 'C:\\Python38-x64'; .\install_python.ps1; Copy-Item 'C:\\Python38-x64' 'C:\\hostedtoolcache\\windows\\Python\\3.8.0\\x64' -Recurse }
     - script: choco install vcpython27 -f -y
       displayName: Install Visual C++ for Python 2.7
     - bash: |
@@ -210,6 +214,12 @@ jobs:
 
     ```
     build_script:
+      - ps: |
+          $PYTHON = $env:PYTHON
+          curl -o install_python.ps1 https://raw.githubusercontent.com/matthew-brett/multibuild/11a389d78892cf90addac8f69433d5e22bfa422a/install_python.ps1
+          & { $env:PYTHON = 'C:\\Python38'; .\install_python.ps1 }
+          & { $env:PYTHON = 'C:\\Python38-x64'; .\install_python.ps1 }
+          $env:PYTHON = $PYTHON
       - pip install cibuildwheel==0.12.0
       - cibuildwheel --output-dir wheelhouse
     artifacts:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,12 @@ environment:
     - PYTHON: "C:\\Python35\\python.exe"
 
 build_script:
+  - ps: |
+      $PYTHON = $env:PYTHON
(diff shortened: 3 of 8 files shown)


cibuildwheel/macos.py
```diff
@@ -19,6 +19,7 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.4/python-3.7.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_intel', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
```
mayeut: identifier should be `cp38-macosx_10_9_x86_64`, only `x86_64` is supported with this installer.

YannickJadoul:
Ah, sorry to be too fast, then; I was just curious to see what would break :)

Thanks for the suggestions! Fixed those just now, but feel free to just push to this branch; I think you should be able to, as maintainer?

YannickJadoul:
Cfr. https://github.com/microsoft/azure-pipelines-image-generation/issues/1317 and https://github.com/appveyor/ci/issues/3142

YannickJadoul:
@joerick Just a thought, but this keeps on happening: for Windows, we depend on the CI images being updated, and we have no control over them. For macOS, we manually install the official Python releases; should we perhaps do something similar for Windows?

Czaki:
Maybe install python 3.8 with nuget? https://www.nuget.org/packages/python/3.8.0
Like in windows on travis?

YannickJadoul:
Pffff, I'm not completely convinced of that. It takes more time if there are already Python installations. And then there's choco and nuget and whatelse? Is there any widely-used, semi-default package manager on Windows?

But more importantly, we also don't use homebrew on macOS, but somehow go through the trouble of downloading and installing the official binaries from python.org. I don't fully remember, but there's somehow difference between brew's Python and the official Python version, or what was the reason? Do you perhaps know, @mayeut?
So how's that for Windows?

Czaki:
In python docuementation nuget is suggested way to install python for CI. It install reduced version of Python (without upgrade options etc) and without checking i there exist some installation of python in system. 

Maybe best option is to wait on introducing new python version by CI provider but add command line option to insert path to user installation. eg. CIBW_WINDOWS_PYTHON_38_PATH. which will be ignored if provider has own installation? 

Problem with homebrew is that it **compile** python against **current version** of host system. So you cannot create 10.9 compatibile wheel from 10.13 system. And downloading version from python.org allows this.

YannickJadoul:
> Problem with homebrew is that it compile python against current version of host system.

But so, that's not the case for the Windows binaries from nuget?

Czaki:
Form nuget you get binaries, not source code. So there is no compilation. 

Moving everything to nuget it force to drop python 3.4.

YannickJadoul:
> Form nuget you get binaries, not source code. So there is no compilation.

The same goes for brew, as far as I know?

YannickJadoul:
Possible temporary solution, since AppVeyor and Azure are taking their time: https://github.com/appveyor/ci/issues/3142#issuecomment-545950043

mayeut:
> I don't fully remember, but there's somehow difference between brew's Python and the official Python version, or what was the reason?

> Problem with homebrew is that it compile python against current version of host system. So you cannot create 10.9 compatibile wheel from 10.13 system. And downloading version from python.org allows this.

My 2 cents regarding `brew`, as @Czaki said, built for a specific system (binaries or built from sources depending on availability of binaries for the specific system), can't build 32/64 bit intel wheels, it's painfully slow (just look at ci and compare the time between python 2 (pre-installed) and python 3 (brew)), there are only 2 formulas, python 2 & 3, installing multiple python 3 versions might be painful. Anyway, that's not the topic here.

> In python docuementation nuget is suggested way to install python for CI. It install reduced version of Python (without upgrade options etc) and without checking i there exist some installation of python in system.

https://docs.python.org/3/using/windows.html
Why not use nuget only for unexisting python versions on CI system ?

AppVeyor comes with nuget preinstalled, so nothing to do for the user once cibuildwheel manages this. Travis CI already uses nuget and since choco is available in Azure (don't know about nuget), I don't see why it couldn't work.

joerick:
Regarding the use of package managers, where possible I'd prefer not to use them in cibuildwheel. There are practical reasons - see the discussion of `brew` above, but more generally, it's better to use python.org releases, because:

- if an extension is built against a package manager's version of Python, and doesn't work with python.org, the extension is at fault
- if an extension is built against python.org Python, and doesn't work with a package manager's Python, it's the package manager's fault.

However, in the Windows on Travis PR, it seems `nuget` was necessary, since we couldn't figure out a way to install a `msi` file on CI.

> Possible temporary solution, since AppVeyor and Azure are taking their time: [appveyor/ci#3142 (comment)](https://github.com/appveyor/ci/issues/3142#issuecomment-545950043)

This is a cool solution, for the latest versions of Python. Looking at Python's ftp archive, there are only `.exe` installers for Python 3.5+. So it would solve the problem for new versions, but we'd still need to rely on Appveyor/Azure's versions for the older ones.

Czaki:
The problem with windows installer from python.org is that they are looking if there exist other python  installation in the system and not install if found. So you cannot assume that you know the path to python executable. The installers from nuget do not have this property. 

Also windows installer are by default with graphical interface. 

When cibuildwheel drop python 3.4 support the pep 514 can help with determine python location.

YannickJadoul:
@mayeut, @joerick Thanks for the refresher on those package managers :-)

The Python docs do advertise `nuget` as lightweight version for use in CI configurations (as @Czaki noted), but I see your argument against package managers.

I'm just not satisfied with the fact that we seem to have 3 different solutions for all three different platforms (_AppVeyor_: pre-installed, _Azure Pipelines_: pre-installed in different locations, and with extra lines in the config; and _Travis CI_: nuget). And of all these three solutions, the `nuget` one seems to be the only one already working for this PR, while we're waiting for the pre-installed ones?

@Czaki We should also beware to not make things too complex/complicated either. Querying the Windows registry according to PEP 514 soon becomes a huge mess, I guess.

Czaki:
@YannickJadoul If you would like to use standard python installer for windows then you cannot assume that python is under path which you set as installation path. It may end because python is installed under other path. 

I meet this problem when I try different approach for using `cibuildwheel` on travis windows. Best way to found already installed python is PEP 514. But I found that using nuget produce much more clean, and readable code.

mayeut:
I've opened 3 PR as draft to investigate different solutions and see what works:
- [Always use nuget to install python on Windows](https://github.com/joerick/cibuildwheel/pull/180):
  - Pros: This works on all CI providers, consistency between all CI providers.
  - Cons: This slows down CI a bit (not that much) if python is already available. It's not an installer retrieved on python.org

- [Install python with nuget if not found](https://github.com/joerick/cibuildwheel/pull/182):
  - Pros: This works on all CI
  - Cons: no consistency between CI providers. It's not an installer retrieved on python.org

- [Install python with official installer if not found](https://github.com/joerick/cibuildwheel/pull/184):
  - Pros: Installer retrieved from python.org (python 3.5+)
  - Cons: no consistency between providers. Does not work on travis-ci as of now (probably hitting issues with an already installed python 3.6 as mentioned in @Czaki comment), probably require to implement PEP514 which is a con.

There's one not opened that would be, "Install python with official installer if an official python install is not found".
  - Pros: Consistency between all CI providers. It's an installer retrieved on python.org
  - Cons: Might slow down CI a bit on Azure/Travis-CI (Appveyor uses official installers). Requires to implement PEP514.

The pros & cons take into account @joerick comment about package managers. That being said, while I can understand this comment, we're not talking about `brew` here (which is not mentioned in python.org documentation in the Mac section). The nuget package seems to be officially supported. Maybe @zooba, release manager for Windows installers and manager of the nuget package, can shed some light on wether or not nuget shall be used in this use case and what differences should be expected (as they're not really fully explained in the python.org docs).

mayeut:
PS: IMHO, the [Install python with nuget if not found](https://github.com/joerick/cibuildwheel/pull/182) shall be merged/reviewed in this PR in order to get a working python 3.8 solution and defer the whole "consistency on windows", "package managers vs official installer" discussion (or at least specify a deadline not to let the discussion prevent the majority of users to get working python 3.8 users).

YannickJadoul:
Wow, great, @mayeut! Thanks for your investigation!

I mostly agree with you, I believe. Furthermore, if the nuget can be confirmed to be (quasi-)official releases supported by Python.org, and if the installation overhead is OK, then it could also help us get rid of the differences between the different Windows CIs?
Do we actually know where the Python versions pre-installed on AppVeyor and Azure originate?

> MHO, the Install python with nuget if not found shall be merged/reviewed in this PR in order to get a working python 3.8 solution

Also agreed. I've been thinking for the two last weeks that "It's just a new version of Python, soon Azure and AppVeyor will get new images", but ... it's quite disappointing.

mayeut:
> Furthermore, if the nuget can be confirmed to be (quasi-)official releases supported by Python.org, and if the installation overhead is OK, then it could also help us get rid of the differences between the different Windows CIs?

Agreed

> Do we actually know where the Python versions pre-installed on AppVeyor and Azure originate?

Appveyor uses official installers from python.org:  https://github.com/appveyor/build-images/blob/master/scripts/Windows/install_python.ps1
I tried to find the information for Azure (https://github.com/microsoft/azure-pipelines-image-generation) but I couldn't figure this out...

Czaki:
It looks like it is not from anaconda
> 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)]

But in this case I do not see where they install python..

YannickJadoul:
> But in this case I do not see where they install python..

It seems to be pre-installed, somehow, [downloaded from some image](https://github.com/microsoft/azure-pipelines-image-generation/blob/866b368747a064f22ecfb7061e9034e3e21c63d4/images/win/scripts/Installers/Download-ToolCache.ps1)?

mayeut:
> It seems to be pre-installed, somehow, downloaded from some image?

I found that but could not figure out where are the scripts that build that cache... Even if it's from an official python install, it won't be found by a PEP514 implementation.

mayeut:
The 4 PR drafts now include python 3.8 on all platforms, and updated readme from this PR. Ready for review.

YannickJadoul:
@mayeut Which one do you prefer? And what's tje suggestion? To merge one of those PRs and close this? Or to merge your additions into this PR?

mayeut:
I think this one can be closed since all changes have been merged back in the 4 other PRs.
It's up to @joerick to review/merge the one he thinks most fits cibuildwheel I guess.
My personal preference goes to consistency on all CI so either #180 (nuget on all CIs) or #186 (official install, using PEP514 tools)

joerick:
Hi both, thank you for this excellent work. I'm away this weekend but I'll try to find time in the next few days to take a proper look and make a decision. What would be your preference @yannickjadoul?

YannickJadoul:
@joerick As a quick but safe solution to get something working but also compile Python 3.8, I'd go with merging #186. The sooner we could merge that, the better.

Again, I've been thinking that Python 3.8 support should soon by in Azure and AppVeyor, but it seems to still be taking quite long. Even though there are rumours in the microsoft/azure-pipelines-image-generation#1317 discussion that it's rolling out, I haven't been successful in restarting the builds, over the last two days.

On the long run, it would good to figure out if the Nuget installations are official enough to be used by cibuildwheel and think about #180. Doing this might greatly simplify the code that's now using different paths on different Windows CI platforms. But that's still worth a discussion, as I do understand your point of view on non-Python.org versions. But meanwhile, #186 should be harmless to merge soon, I think.

@mayeut, do you agree, or did I miss something?

YannickJadoul:
Right, this should work now, I think.

@joerick, yet another suggestion. If you want to take your time deciding what to choose, this current PR pushes the workaround to the `appveyor.yml` and `azure-pipelines.yml`, using the solution suggested in https://github.com/appveyor/ci/issues/3142#issuecomment-545950043.

Advantage: No changes to the `windows.py` code are needed, and once AppVeyor and Azure catch up with the Python release, the workaround is not necessary anymore.
Disadvantage: Users need to add a few lines to their config, before calling cibuildwheel. (Oh yeah, and that very ugly hack on Azure, copying the new installation to the expected location.)

joerick:
Right, I've looked at all the PRs now! 

My feeling is that #180 is my favourite. That solution gives us consistency that we didn't previously have. I hadn't fully considered the provenance of the preinstalled versions we use in CI now. If we have nuget on Travis, Appveyor's version there, Azure doing something different, we're potentially opening ourselves up to a maintenance headache down the road. 

I was wondering on the performance hit, but it doesn't seem to make a measurable difference according to our own CI.

| CI times | Azure | Appveyor |
|---|---|---|
| #172 this PR | 17min | 20min |
| #180 |  17min | 20min |
| #186 |  17min  | 20min |

Then there is the question of non-Python.org installer. Well, it's a risk, but I think the reduced complexity is worth it. If it turns out that it creates compatibility problems, at least then we'll know. If that's the case, we can change our mind, revert #180 and adopt #186 instead.

So, any further comments before we move forward with #180?

mayeut:
> So, any further comments before we move forward with #180?

Drop python 3.4 #168 ?

YannickJadoul:
Agreed with both :)
Though it's slightly sad/ironic that now that @mayeut got a nice PEP 514, we're ditching it anyway :-D

mayeut:
> Though it's slightly sad/ironic that now that @mayeut got a nice PEP 514, we're ditching it anyway :-

No worries, the goal with the 5 PRs was to get at least one selected to get Python 3.8 ASAP
I also do prefer #180 over #186 but, if at some point compatibility issues arise, as @YannickJadoul said, then #186 it's still an option (I'll keep the branch for reference).

YannickJadoul:
> but, if at some point compatibility issues arise

Perfect! Good to have a quick fallback at hand :)

joerick:
Closed, because this is included in #180

Issue #99, opened 2018-09-12
Refactoring the Linux build script


joerick:
Currently the linux build works by piping a bash script into /bin/bash inside the docker container.

A couple good ideas have come out of the discussion in #97, so I wanted to keep them in an open issue!

1. @altendky: maybe a good way would be a Python script with a CLI (probably small enough to comfortably do with stdlib and 2/3 compatibility) and we pass in the info via the command. No config file to write and copy, no template to fill out. Just copy the exact Python script over to the container and then docker run it with proper arguments.

2. @altendky: Or, maybe we just replace the string/file with python code calling docker run for each command?

altendky:
You could use `check_output()` as the general rule instead and could then read out variables and directory listings.  Just make sure of things like having `ls` output a file per line rather than splitting on spaces (hmm, can a file have a newline?).

https://docs.docker.com/engine/reference/run/#env-environment-variables

For setting env var you can (1) `['-e', 'deep=purple', 'do', 'something']` or (2) have the function take an additional dict for extra env vars or (3) `**kwargs` for env vars.  I think (2) uses Python syntax and is explicit and flexible so I'd tend towards that.

joerick:
The piped bash script was removed in #386 - replaced with a system similar to (2) above, but using a shell process and open pipes instead of `docker exec` for performance reasons. So this is fixed :)

PR #175, opened 2019-10-18
Make sure build fails when test command fails

GjjvdBurg:
First of all, thank you for creating this project!

I'm running the latest version (master) to test my package in a virtualenv and noticed that on Linux the return code of the test isn't used. This means that the build continues even though the test command fails. This PR fixes this by keeping the return code of the test command and making that the return code of the bash script.

Note that this is not needed on mac or windows because they use ``subprocess.check_call``.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -112,9 +112,9 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                         # Run the tests from a different directory
                         pushd $HOME
                         sh -c {test_command}
-                        test_retcode=$(( $test_retcode || $? ))
                         popd
                     )
+                    test_retcode=$(( $test_retcode || $? ))
 
                     # clean up
                     rm -rf "$venv_dir"
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -117,7 +117,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                         sh -c {test_command}
                         popd
                     )
-                    # exit if tests failed
+                    # exit if tests failed (needed for older bash versions)
                     if [ $? -ne 0 ]; then
                       exit 1;
                     fi
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -114,7 +114,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
                         # Run the tests from a different directory
                         pushd $HOME
-                        sh -o errexit -c {test_command}
+                        sh -c {test_command}
                         popd
                     )
                     # exit if tests failed
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -57,8 +57,6 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             mkdir /output
             cd /project
 
-            test_retcode=0
-
             {environment_exports}
 
             for PYBIN in {pybin_paths}; do
@@ -111,10 +109,13 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
                         # Run the tests from a different directory
                         pushd $HOME
-                        sh -c {test_command}
+                        sh -o errexit -c {test_command}
                         popd
                     )
-                    test_retcode=$(( $test_retcode || $? ))
+                    # exit if tests failed
+                    if [ $? -ne 0 ]; then
+                      exit 1;
+                    fi
 
                     # clean up
                     rm -rf "$venv_dir"
@@ -124,8 +125,6 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 mv "$delocated_wheel" /output
                 chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"
             done
-
-            exit $test_retcode
         '''.format(
             pybin_paths=' '.join(c.path+'/bin' for c in platform_configs),
             test_requires=' '.join(test_requires),
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -57,6 +57,8 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             mkdir /output
             cd /project
 
+            test_retcode=0
+
             {environment_exports}
 
             for PYBIN in {pybin_paths}; do
@@ -110,6 +112,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                         # Run the tests from a different directory
                         pushd $HOME
                         sh -c {test_command}
+                        test_retcode=$(( $test_retcode || $? ))
                         popd
                     )
 
@@ -121,6 +124,8 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 mv "$delocated_wheel" /output
                 chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"
             done
+
+            exit $test_retcode
         '''.format(
             pybin_paths=' '.join(c.path+'/bin' for c in platform_configs),
             test_requires=' '.join(test_requires),


joerick:
Hi! Thanks for the contribution. I'm a little confused why it's needed though, since our bash script uses `set -o errexit`. I was wondering if that wasn't passing to the subshell, so I tried this test:

```bash
joerick@joerick2 ~/Desktop> cat testsh.bash                                   1
#!/bin/bash

set -o errexit
set -o xtrace

echo "Starting"
true

(
  echo subshell 1
  (
    echo subshell 2
    false
    true
  )
  true
)

echo Complete
joerick@joerick2 ~/Desktop> bash testsh.bash 
+ echo Starting
Starting
+ true
+ echo subshell 1
subshell 1
+ echo subshell 2
subshell 2
+ false
```

In this test, the first failing command `false` stops the execution and exits the script with an error. Did you have an example (perhaps a build log) of that not happening?

GjjvdBurg:
Thanks for checking this out! You're right that your test scripts exits properly, so I'm now also confused why this didn't happen in my tests. 

Here is an example of a failing build: https://travis-ci.org/alan-turing-institute/CleverCSV/jobs/599248169 (around line 670). 

The ``mv "$delocated_wheel /output`` command still runs even though the test failed. The error itself is not really relevant, it can be fixed by removing the ``*.so`` files before running cibuildwheel.

GjjvdBurg:
Here is a build that prints out ``set -o`` when running the test command: https://travis-ci.org/GjjvdBurg/TestingTravis/builds/600010616 Line 309 shows that ``errexit`` is off in the subshell.

YannickJadoul:
Hmmm, so, can we then just add `set -o errexit` (and maybe `set -o xtrace` as well) to the subshell?

GjjvdBurg:
Yeah, that would be much cleaner than my solution. I'll make that change and update this PR when I have it.

GjjvdBurg:
The new commit simplifies the test and seems to work reliably. 

Bash 4.4 adds the ``inherit_errexit`` shell option, which would be good to use.  Unfortunately, the manylinux docker image comes with bash version 3.2.25. In that version the script given above by @joerick does not work as expected. It seems that ``errexit`` is a bit of an [issue](http://mywiki.wooledge.org/BashFAQ/105) in bash. [Here](https://fvue.nl/wiki/Bash:_Error_handling) is a more detailed overview of all the caveats.

I found that even when I place ``set -o errexit`` in every subshell, the top-level bash script doesn't exit if the subshell for the virtualenv returns a nonzero exit code. The simplest and most reliable solution was to check the exit code of that subshell explicitly using an if statement. Also, the ``sh -c {testcommand}`` has been changed to add ``-o errexit`` in case the user chains test commands using ``;`` instead of ``&&``.

YannickJadoul:
Great, thanks! 

Are we now just focusing on the `CIBW_TEST_COMMAND`, though? What happens if `pip install "$delocated_wheel"{test_extras}` would fail? Is there no way of putting `set -o errexit set -o xtrace`source "$venv_dir/bin/activate

GjjvdBurg:
Good question! Any failure in the subshell for the virtualenv will make the returncode of that subshell nonzero. The problem is only that that does not cause the toplevel shell to register an error. Here is an example where the ``pip install "$delocated_wheel"{test_extras}`` command fails due to a non-existing package: https://travis-ci.org/GjjvdBurg/TestingTravis/builds/600808779 As you can see the subshell exits at that point, but the if statement is reached (line 305) indicating that the top-level script didn't exit right away.

The links in my previous comment give some more details about the problems of ``errexit`` and subshells. Here's a demonstration that prints ``set -o`` after activating the virtualenv and in the test command, which shows that ``errexit`` is on in both, but despite that the if statement is reached (which it shouldn't, if ``errexit`` would be sufficient): https://travis-ci.org/GjjvdBurg/TestingTravis/builds/600813804

I agree that the if statement isn't a particularly clean solution, but it was the most robust approach I could find ([with some trial and error](https://github.com/GjjvdBurg/cibuildwheel/pull/1)).

YannickJadoul:
Ha, that's funny. But apparently this whole error-handling is quite fragile in bash, so fair enough. Then you have indeed found a very clean solution!

Thanks a lot for your investigation and extensive explanation! Always good to have this for future reference, if we forget the reasons for this change! :)

@joerick, I thought I've heard you mention long time ago you'd like to replace the bash script with proper Python? If we ever get to doing it, this can be added to the list of reasons why it's a maybe a good idea ;)

cibuildwheel/linux.py
```diff
@@ -109,9 +109,13 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
                         # Run the tests from a different directory
                         pushd $HOME
-                        sh -c {test_command}
+                        sh -o errexit -c {test_command}
```
joerick: I'm maybe 50/50 on adding the `-o errexit` here, because `cmd1 && cmd2` is well recognised, and I'm not sure how to recreate `cmd1; cmd2` when `-o errexit` is enabled.

But if we're changing the behaviour of this, we should change before_build and the Mac and Windows versions too, for consistency.

YannickJadoul:
> That test ensures that `CIBW_TEST_COMMAND=false` causes the build to fail... and it passes.

You're not alone to be confused, @joerick. I just tried this, thinking I knew it was going to fail, but ... nope

```shell
set -o errexit
set -o xtrace

echo A
(
        echo B
        sh -c "false"
        echo C
)
echo D
```

```shellsession
$ bash test.sh; echo $?
+ echo A
A
+ echo B
B
+ sh -c false
1
```

GjjvdBurg:
Thanks for your comments! I'll add a failing test that this PR fixes and get back to you soon.

joerick:
Hi @GjjvdBurg, just checking in, any progress on this?

GjjvdBurg:
Apologies for the delay, I'll try to figure it out this week.

cibuildwheel/linux.py
GjjvdBurg: I think I agree with you and will remove the ``-o errexit`` settting from this command, using `` && `` is indeed well recognized.

GjjvdBurg:
So, essentially, I can no longer reproduce my original problem with the latest version of cibuildwheel. It seems that upgrading the linux images to manylinux2010 has solved it, probably because manylinux2010 comes with an upgraded version of bash (3.2.25 -> 4.1.2).

Here is a build for a test package with the version of cibuildwheel before #155 was merged: 
https://travis-ci.org/GjjvdBurg/TestingTravis/builds/608870166#L184 (see [line 386](https://travis-ci.org/GjjvdBurg/TestingTravis/builds/608870166#L386), where failure is not caught).

And here is one with the version after manylinux was upgraded: https://travis-ci.org/GjjvdBurg/TestingTravis/builds/608871534#L184 (which correctly fails).

As further proof that the manylinux change solved it, [here is a version of cibuildwheel](https://travis-ci.org/GjjvdBurg/tempcibuildwheelfork/jobs/608863988) branched off from where #155 was merged and with the test from #177 added (in a temporary copy of the repo). With manylinux1 the failing test did indeed fail.

So there's no need to merge this PR anymore! The problem has gone away in another way and the test added in #177 ensures it can be detected in the future. Thanks for your time on this @joerick and @YannickJadoul!

YannickJadoul:
Huh, that's really weird! Thanks for further checking that.
We're still allowing `manylinux1`, so should we still try to merge this, though? I guess it can't hurt either, and it might solve discrepancies on sóme docker images?

@joerick Yet another reason to take some time for #99? Maybe after the next release, there's a bit more time to do these kinds of refactoring.

GjjvdBurg:
Yes, with manylinux1 this fix would still be needed and it can't hurt for the manylinux2010 images.

I've updated my branch with the changes requested above and have clarified the comment. I'll reopen the PR so you can decide on whether to merge or not.

joerick:
> Yes, with manylinux1 this fix would still be needed

We have a winner! I just ran a test with manylinux1 and indeed it's erroneously passing.

```
    def test_failing_test():
        '''Ensure a failing test causes cibuildwheel to error out and exit'''
        project_dir = os.path.dirname(__file__)
    
        with pytest.raises(subprocess.CalledProcessError):
>           utils.cibuildwheel_run(project_dir, add_env={
                'CIBW_TEST_COMMAND': 'false',
                # manylinux1 has a version of bash that's been shown to have
                'CIBW_MANYLINUX_I686_IMAGE': 'manylinux1',
                'CIBW_MANYLINUX_X86_64_IMAGE': 'manylinux1',
            })
E           Failed: DID NOT RAISE <class 'subprocess.CalledProcessError'>
```

Thank you all for raising this! Very happy we spotted this before a release.

YannickJadoul:
Thanks, @GjjvdBurg, for work and for holding on during the discussion!

GjjvdBurg:
Glad I could help!

PR #178, opened 2019-10-21
Fix CIBW_BUILD example

apcamargo:
Just a little fix to a `CIBW_BUILD` example. Using `*-manylinux_i686` won't skip `*-manylinux1_i686`.


YannickJadoul:
Auwtch, yes. It will, soon, once the master branch, supporting `manylinux2010`, gets released. But I guess currently the README is wrong indeed :-(

joerick:
Argh, yes. All the more reason to get the docs in - we can set those to track the latest release, not master.

PR #180, opened 2019-10-26
Adding Python 3.8, Always use nuget to install python on Windows

mayeut:
Just a draft to check if it works on all windows providers.
c.f. #172 for full discussion

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,10 +352,10 @@ You must set this variable to pass variables to Linux builds (since they execute
 
 You can use `$PATH` syntax to insert other variables, or the `$(pwd)` syntax to insert the output of other shell commands.
 
-Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`
-Example: `PATH=$PATH:/usr/local/bin`
-Example: `BUILD_TIME="$(date)"`
-Example: `PIP_EXTRA_INDEX_URL="https://pypi.myorg.com/simple"`
+Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`\
+Example: `PATH=$PATH:/usr/local/bin`\
+Example: `BUILD_TIME="$(date)"`\
+Example: `PIP_EXTRA_INDEX_URL="https://pypi.myorg.com/simple"`\
 
 Platform-specific variants also available:
 `CIBW_ENVIRONMENT_MACOS` | `CIBW_ENVIRONMENT_WINDOWS` | `CIBW_ENVIRONMENT_LINUX`
@@ -375,11 +375,11 @@ If dependencies are required to build your wheel (for example if you include a h
 
 The active Python binary can be accessed using `python`, and pip with `pip`; `cibuildwheel` makes sure the right version of Python and pip will be executed. `{project}` can be used as a placeholder for the absolute path to the project's root.
 
-Example: `pip install .`
-Example: `pip install pybind11`
+Example: `pip install .`\
+Example: `pip install pybind11`\
 Example: `yum install -y libffi-dev && pip install .`
 
-Platform-specific variants also available:
+Platform-specific variants also available:\
  `CIBW_BEFORE_BUILD_MACOS` | `CIBW_BEFORE_BUILD_WINDOWS` | `CIBW_BEFORE_BUILD_LINUX`
 
 ***
@@ -397,8 +397,8 @@ Beware to specify a valid Docker image that can be used in the same way as the o
 
 Note that `auditwheel` detects the version of the `manylinux` standard in the Docker image through the `AUDITWHEEL_PLAT` environment variable, as `cibuildwheel` has no way of detecting the correct `--plat` command line argument to pass to `auditwheel` for a custom image. If a Docker image does not correctly set this `AUDITWHEEL_PLAT` environment variable, the `CIBW_ENVIRONMENT` option can be used to do so (e.g., `CIBW_ENVIRONMENT="manylinux2010_$(uname -m)"`).
 
-Example: `manylinux1`
-Example: `dockcross/manylinux-x64`
+Example: `manylinux1`\
+Example: `dockcross/manylinux-x64`\
 Example: `dockcross/manylinux-x86`
 
 ***
@@ -426,7 +426,7 @@ Optional.
 
 Space-separated list of dependencies required for running the tests.
 
-Example: `pytest`
+Example: `pytest`\
 Example: `nose==1.3.7 moto==0.4.31`
 
 Platform-specific variants also available:
--- a/README.md
+++ b/README.md
@@ -12,13 +12,13 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
 What does it do?
 ----------------
 
-|   | macOS 10.6+ | manylinux i686 | manylinux x86_64 |  Windows 32bit | Windows 64bit |
-|---|---|---|---|---|---|
-| Python 2.7 | ✅ | ✅ | ✅ | ✅¹ | ✅¹ |
-| Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
-| Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
-| Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
-| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
+|   | macOS 10.6+ intel | macOS 10.9+ x86_64 | manylinux i686 | manylinux x86_64 |  Windows 32bit | Windows 64bit |
+|---|---|---|---|---|---|---|
+| Python 2.7 | ✅ |    | ✅ | ✅ | ✅¹ | ✅¹ |
+| Python 3.5 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
+| Python 3.6 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
+| Python 3.7 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 |    | ✅ | ✅ | ✅ | ✅  | ✅  |
 
 > ¹ Not supported on Travis
 
@@ -305,7 +305,7 @@ The format is `python_tag-platform_tag`. The tags are similar but not identical 
 
 Python tags look like `cp27` `cp35` `cp36` `cp37` `cp38`
 
-Platform tags look like `macosx_10_6_intel` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
+Platform tags look like `macosx_10_6_intel` `macosx_10_9_x86_64` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
 
 You can also use shell-style globbing syntax (as per `fnmatch`).
 
@@ -314,6 +314,7 @@ The list of supported and currently selected build identifiers can be retrieved 
 Examples:
 - Only build on Python 3.6: `CIBW_BUILD`:`cp36-*`
 - Skip building on Python 2.7 on the Mac: `CIBW_SKIP`:`cp27-macosx_10_6_intel`
+- Skip building on Python 3.8 on the Mac: `CIBW_SKIP`:`cp38-macosx_10_9_x86_64`
 - Skip building on Python 2.7 on all platforms: `CIBW_SKIP`:`cp27-*`
 - Skip Python 2.7 on Windows: `CIBW_SKIP`:`cp27-win*`
 - Skip Python 2.7 on 32-bit Windows: `CIBW_SKIP`:`cp27-win32`
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ What does it do?
 | Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
 
 > ¹ Not supported on Travis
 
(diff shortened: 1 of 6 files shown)


YannickJadoul:
Just before your summary in #172, I was trying to figure out if there's a way to force the official installers to work, but it's horrible; there's absolutely no error message. I also tried to run `py -0p`, as [it's supposed to give a list of installed Python installations](https://stackoverflow.com/questions/53312590/how-to-check-all-the-installed-python-versions-on-windows), but that [doesn't seem to give any useful information either](https://travis-ci.org/YannickJadoul/cibuildwheel/builds/605176514#L39).

I don't know if it will make a difference, but have you considered [pyenv-win](https://github.com/pyenv-win/pyenv-win)? It seems like it is also [downloading and installing official Python installers](https://github.com/pyenv-win/pyenv-win/blob/03918c29bc68769db4f5f0756943894199d2ef27/pyenv-win/libexec/pyenv-install.vbs#L45), and I have no clue if it somehow manages to get install an already installed version but it might be worth a try.

cibuildwheel/windows.py
```diff
@@ -106,21 +89,19 @@ def shell(args, env=None, cwd=None):
...
     built_wheel_dir = os.path.join(temp_dir, 'built_wheel')
 
-    if IS_RUNNING_ON_TRAVIS:
-        # instal nuget as best way for provide python
-        shell(["choco", "install", "nuget.commandline"])
-        # get pip fo this installation which not have. 
-        get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
-        get_pip_script = 'C:\\get-pip.py'
-        shell(['curl', '-L', '-o', get_pip_script, get_pip_url])
+    # instal nuget as best way for provide python
```
nsoranzo:
```suggestion
    # install nuget as best way to provide python
```

README.md
```diff
@@ -355,10 +349,10 @@ You must set this variable to pass variables to Linux builds (since they execute
...
 You can use `$PATH` syntax to insert other variables, or the `$(pwd)` syntax to insert the output of other shell commands.
 
-Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`  
-Example: `PATH=$PATH:/usr/local/bin`  
-Example: `BUILD_TIME="$(date)"`  
-Example: `PIP_EXTRA_INDEX_URL="https://pypi.myorg.com/simple"`  
+Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`
+Example: `PATH=$PATH:/usr/local/bin`
+Example: `BUILD_TIME="$(date)"`
+Example: `PIP_EXTRA_INDEX_URL="https://pypi.myorg.com/simple"`
```
YannickJadoul: Same set of remarks for spaces and line endings as for #186. (Sorry, but it feels silly to repeat myself/copy-past everything from there to here.)

cibuildwheel/windows.py
```diff
@@ -8,89 +8,72 @@
...
+        PythonConfiguration(version='2.7.x', arch="32", identifier='cp27-win32', nuget_version="2.7.17"),
+        PythonConfiguration(version='2.7.x', arch="64", identifier='cp27-win_amd64', nuget_version="2.7.17"),
+        PythonConfiguration(version='3.5.x', arch="32", identifier='cp35-win32', nuget_version="3.5.4"),
+        PythonConfiguration(version='3.5.x', arch="64", identifier='cp35-win_amd64', nuget_version="3.5.4"),
+        PythonConfiguration(version='3.6.x', arch="32", identifier='cp36-win32', nuget_version="3.6.8"),
+        PythonConfiguration(version='3.6.x', arch="64", identifier='cp36-win_amd64', nuget_version="3.6.8"),
+        PythonConfiguration(version='3.7.x', arch="32", identifier='cp37-win32', nuget_version="3.7.5"),
+        PythonConfiguration(version='3.7.x', arch="64", identifier='cp37-win_amd64', nuget_version="3.7.5"),
+        PythonConfiguration(version='3.8.x', arch="32", identifier='cp38-win32', nuget_version="3.8.0"),
+        PythonConfiguration(version='3.8.x', arch="64", identifier='cp38-win_amd64', nuget_version="3.8.0"),
```
YannickJadoul: We could potentially even make `version` be the `nuget_version`, or not?
Or keep it like in `macOS.py` and remove the `.x`; any reason to have it here?

cibuildwheel/windows.py
```diff
@@ -8,89 +8,72 @@
...
-        # See https://github.com/Microsoft/azure-pipelines-tasks/issues/9674
-        python_configurations = [c for c in python_configurations if c.version != '3.4.x']
-    
     if IS_RUNNING_ON_TRAVIS:
         # cannot install VCForPython27.msi which is needed for compiling C software
         # try with (and similar): msiexec /i VCForPython27.msi ALLUSERS=1 ACCEPT=YES /passive
-        # no easy and stable way fo installing python 3.4 
-        python_configurations = [c for c in python_configurations if c.version != '2.7.x' and c.version != '3.4.x']
+        # no easy and stable way fo installing python 3.4
+        python_configurations = [c for c in python_configurations if c.version != '2.7.x']
```
YannickJadoul: Is this still up to date? At least the comment about 3.4 isn't.
Why can't we get nuget 2.7 on Travis?

cibuildwheel/windows.py
```diff
@@ -106,21 +89,19 @@ def shell(args, env=None, cwd=None):
...
-        # get pip fo this installation which not have. 
-        get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
-        get_pip_script = 'C:\\get-pip.py'
-        shell(['curl', '-L', '-o', get_pip_script, get_pip_url])
+    # instal nuget as best way for provide python
+    nuget = 'C:\\nuget.exe'
+    download('https://dist.nuget.org/win-x86-commandline/latest/nuget.exe', nuget)
+    # get pip fo this installation which not have.
+    get_pip_script = 'C:\\get-pip.py'
+    download('https://bootstrap.pypa.io/get-pip.py', get_pip_script)
```
YannickJadoul: Just curious: do nuget versions not have `pip` installed?

cibuildwheel/windows.py
```diff
@@ -106,21 +89,19 @@ def shell(args, env=None, cwd=None):
...
 
     python_configurations = get_python_configurations(build_selector)
     for config in python_configurations:
         config_python_path = get_python_path(config)
-        if IS_RUNNING_ON_TRAVIS and config.nuget_version is not None and not os.path.exists(config_python_path):
-            shell(["nuget", "install"] + get_nuget_args(config))
-            if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
-                shell([os.path.join(config_python_path, 'python.exe'), get_pip_script ])
+        simple_shell([nuget, "install"] + get_nuget_args(config))
```
YannickJadoul: Do we need to do for every `config`?

README.md
mayeut: resolved same way as #186

cibuildwheel/windows.py
mayeut: argh, I forgot I had to drop python 3.4 on this PR.
The issue on travis with 2.7 is not install python but the tools needed to build native code for python 2.7

cibuildwheel/windows.py
mayeut: I've not redone the test, it's merely a rewrite of the travis-ci PR which was doing this.

mayeut: I'm not sure we hit the conditional statement lower in the code.

cibuildwheel/windows.py
mayeut: yes. we need to install all versions with nuget here. If already installed, nuget just does nothing.

mayeut:
Depends on #168 (sort of), drops python 3.4 on windows.

cibuildwheel/windows.py
mayeut: The version with `.x` is used in the code to set an environment variable. It's for the Appveyor build tool selection but reading the script, `x` could be anything. So yes, we could use the same field. I'll do that.

cibuildwheel/windows.py
YannickJadoul: Ah, so it might be good to leave merging this PR until after we drop 3.4?

On the 2.7: is it really necessary to keep worrying about these VS tools for 2.7? I'm building C++17 wheels for Python 2.7 (without the ancient compiler VS offers for 2.7). I've tested this and it does work. The only catch is that "Microsoft Visual C++ Redistributable for Visual Studio 2017" needs to be installed, then, so I [added a section about that to my docs](https://parselmouth.readthedocs.io/en/stable/installation.html#importerror-dll-load-failed-on-windows).

My point being: maybe we could still offer 2.7, but make people aware that the tools won't be installed?

cibuildwheel/windows.py
YannickJadoul: Sorry, I got confused looking for the line where `nuget` is installed. But I'm now realizing the executable you download probably doesn't need to be installed anymore but it already ready to be executed.

cibuildwheel/windows.py
YannickJadoul: Hmmm, two lines and a download. Not sure if it's worth testing whether we can remove the lines.

cibuildwheel/windows.py
mayeut: IMHO, this goes against the generic wheel principles (I think) which is, pip install should always work, no matter what, no need for further installs. My point is that if we want to offer python 2.7 support on travis, then it shall be an opt-in, not the default. This would require a specific PR.
Anyway, python 2.7 is EOL in less than 2 months. Shall we really do specific things for it ?

YannickJadoul: Hmmm, that's also true, yes. And there's still AppVeyor anyway.

mayeut: > Ah, so it might be good to leave merging this PR until after we drop 3.4?

yes, if we want to merge it after #186 (official vs nuget). My guess is that in a couple months (after python 2.7/3.4) have been dropped, the only difference between the 2 PRs will be what to install and that, with official installers, we'll still need pep514tools vendored-in. The code in `windows.py` will be almost the same, whatever the PR.

cibuildwheel/windows.py
mayeut: [Seems pip is not installed](https://travis-ci.org/mayeut/cibuildwheel/jobs/606729148?utm_medium=notification&utm_source=github_status)

> Hmmm, two lines and a download. Not sure if it's worth testing whether we can remove the lines.

Not sure I fully understand this comment. If it only refers to `get-pip` then no, we can't remove those (now tested).

YannickJadoul: Yeah, that's what I meant. That's kind of silly. Why do all these distributions not have `pip` installed?

README.md
```diff
@@ -309,7 +303,7 @@ When both options are specified, both conditions are applied and only builds wit
 
 The format is `python_tag-platform_tag`. The tags are similar but not identical to the ones defined in [PEP 425](https://www.python.org/dev/peps/pep-0425/#details).
 
-Python tags look like `cp27` `cp35` `cp36` `cp37`
+Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37` `cp38`
```
```suggestion
Python tags look like `cp27` `cp35` `cp36` `cp37` `cp38`
```

joerick:
This is looking good to me! Anything else you'd like to do before merging @mayeut ?

mayeut:
@joerick, it seemed it was not possible to merge the PR with the merge you did. Instead, I rebased on the new master. Looks good to me, let's wait for CI.
Matt$ diff -ruN -x .git cibuildwheel.bak cibuildwheel
diff -ruN -x .git cibuildwheel.bak/cibuildwheel/windows.py cibuildwheel/cibuildwheel/windows.py
--- cibuildwheel.bak/cibuildwheel/windows.py	2019-11-04 22:00:42.000000000 +0100
+++ cibuildwheel/cibuildwheel/windows.py	2019-11-04 22:07:06.000000000 +0100
@@ -9,9 +9,9 @@
     from pipes import quote as shlex_quote
 
 try:
-    from urllib2 import urlopen
-except ImportError:
     from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
 
diff -ruN -x .git cibuildwheel.bak/test/shared/utils.py cibuildwheel/test/shared/utils.py
--- cibuildwheel.bak/test/shared/utils.py	2019-11-04 22:00:42.000000000 +0100
+++ cibuildwheel/test/shared/utils.py	2019-11-04 22:08:34.000000000 +0100
@@ -99,7 +99,7 @@
         raise Exception('unsupported platform')
 
     if IS_WINDOWS_RUNNING_ON_TRAVIS:
-        # Python 2.7 and 3.4 isn't supported on Travis.
+        # Python 2.7 isn't supported on Travis.
         templates = [t for t in templates if '-cp27-' not in t]
 
     return [filename.format(package_name=package_name, package_version=package_version)
```

File: cibuildwheel/macos.py [hunk 8]
```diff
@@ -18,6 +18,7 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
         PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
```
Czaki: This inconsistency in python macos tag should be mentioned somehow in documentation (there is no x86_64 installer for python 3.5)

mayeut: I updated README, is that enough like this ?

Czaki: add it also to platform tag examples maybe (line 318)

mayeut: done

mayeut: also added line 308

joerick:
Those README updates are okay, I think it will be simpler once @Czaki's PR is merged anyway.

mayeut:
The only CI check remaining is AppVeyor which is rebuilding the 4 commits for the `README` change so it will take around 1h30 to get the all green CI status. (they are not building the commit they claim to be building anyway... otherwise, it would have already failed for the first 3 unexisting commits...)

mayeut:
@joerick, LGTM.

PR #183, opened 2019-10-26
Do not update brew python3

mayeut:
This doesn't seem necessary anymore and greatly speeds-up ci


YannickJadoul:
Let's merge this, so other PRs can benefit from this! It's just our own CI anyway, so we can always add this again if necessary.

YannickJadoul:
Thanks @mayeut!

PR #184, opened 2019-10-29
Adding Python 3.8, Install python with official installer if not found on windows

mayeut:
Draft to following discussion in #172

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,12 @@ matrix:
     - os: windows
       language: shell
       before_install:
-       - choco install python3 --version 3.6.8 --no-progress -y
+       - choco install nuget.commandline
+       - nuget install python -Version 3.7.5 -ExcludeVersion -OutputDirectory C:/python-cibuildwheel
       install:
-       - C:\\Python36\\python -m pip install -r requirements-dev.txt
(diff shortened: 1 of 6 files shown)


YannickJadoul:
Just before your summary in #172, I was trying to figure out if there's a way to force the official installers to work, but it's horrible; there's absolutely no error message. I also tried to run `py -0p`, as [it's supposed to give a list of installed Python installations](https://stackoverflow.com/questions/53312590/how-to-check-all-the-installed-python-versions-on-windows), but that [doesn't seem to give any useful information either](https://travis-ci.org/YannickJadoul/cibuildwheel/builds/605176514#L39).

I don't know if it will make a difference, but have you considered [pyenv-win](https://github.com/pyenv-win/pyenv-win)? It seems like it is also [downloading and installing official Python installers](https://github.com/pyenv-win/pyenv-win/blob/03918c29bc68769db4f5f0756943894199d2ef27/pyenv-win/libexec/pyenv-install.vbs#L45), and I have no clue if it somehow manages to get install an already installed version but it might be worth a try. (I was planning to do so, actually, but maybe you know your way around better, now.)

mayeut:
Now working on travis-ci

mayeut:
Closing in favor of #186

PR #186, opened 2019-11-01
Adding Python 3.8, Install python with official installer if an official install is not found on windows

mayeut:
Vendor-In pep514tools

Draft to following discussion in #172

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,11 @@ matrix:
     - os: windows
       language: shell
       before_install:
-       - choco install python3 --version 3.6.8 --no-progress -y
+       - choco install python3 --version 3.7.5 --no-progress -y
       install:
-       - C:\\Python36\\python -m pip install -r requirements-dev.txt
+       - C:\\Python37\\python -m pip install -r requirements-dev.txt
       script:
-       - C:\\Python36\\python ./bin/run_tests.py
+       - C:\\Python37\\python ./bin/run_tests.py
 
 install: $PYTHON -m pip install -r requirements-dev.txt
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ What does it do?
 | Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
 
 > ¹ Not supported on Azure Pipelines
 >
@@ -81,12 +82,6 @@ jobs:
   steps:
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x86}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.5', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.5', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x64}}
     - script: choco install vcpython27 -f -y
       displayName: Install Visual C++ for Python 2.7
     - bash: |
@@ -312,7 +307,7 @@ When both options are specified, both conditions are applied and only builds wit
 
 The format is `python_tag-platform_tag`. The tags are similar but not identical to the ones defined in [PEP 425](https://www.python.org/dev/peps/pep-0425/#details).
 
-Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37`
+Python tags look like `cp27` `cp34` `cp35` `cp36` `cp37` `cp38`
 
 Platform tags look like `macosx_10_6_intel` `manylinux_x86_64` `manylinux_i686` `win32` `win_amd64`
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ What does it do?
 | Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
 | Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
 
-> ¹ Not supported on Azure Pipelines 
+> ¹ Not supported on Azure Pipelines
 >
 > ² Not supported on Travis
 
@@ -58,7 +58,7 @@ Usage
 jobs:
 - job: linux
   pool: {vmImage: 'Ubuntu-16.04'}
-  steps: 
+  steps:
     - task: UsePythonVersion@0
     - bash: |
         python -m pip install --upgrade pip
@@ -68,7 +68,7 @@ jobs:
       inputs: {pathtoPublish: 'wheelhouse'}
 - job: macos
   pool: {vmImage: 'macOS-10.13'}
-  steps: 
+  steps:
     - task: UsePythonVersion@0
     - bash: |
         python -m pip install --upgrade pip
@@ -78,7 +78,7 @@ jobs:
       inputs: {pathtoPublish: 'wheelhouse'}
 - job: windows
   pool: {vmImage: 'vs2017-win2016'}
-  steps: 
+  steps:
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x86}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x64}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '3.5', architecture: x86}}
@@ -110,7 +110,7 @@ jobs:
 
     ```yaml
     language: python
-    
+
     matrix:
       include:
         - sudo: required
@@ -126,7 +126,7 @@ jobs:
       - cibuildwheel --output-dir wheelhouse
     ```
 
-  To build on Windows too, add this matrix entry: 
+  To build on Windows too, add this matrix entry:
   ```yaml
     - os: windows
       language: shell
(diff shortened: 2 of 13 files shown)


azure-pipelines.yml
```diff
@@ -1,31 +1,25 @@
...
-  steps: 
+  steps:
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x86}}
     - {task: UsePythonVersion@0, inputs: {versionSpec: '2.7', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.5', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.5', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x64}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x86}}
-    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x64}}
```
YannickJadoul: Hmmm, so this is now part of testing the new mechanism to install?
Maybe we could/should:
1. Add another job that tests the 2.7 installation? (Just setting `CIBW_BUILD` to `cp27` or so?)
2. Document that these versions are not listed so that the new "install-if-not-present" mechanism is tested?
3. (or 2.bis) Make sure it is clear/documented to the `cibuildwheel` users that want to use this file as an example configuration that it is recommended to not do it this way, but rather still add these versions, as it is faster?

cibuildwheel/_vendored/pep514tools/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
YannickJadoul: Why are we vendoring this? We're also listing and using `bashlex` as a dependency in `setup.py` so can't we do the same for `pep514tools`?

Great find, btw! This makes the whole use of PEP 514 doable! I was really dreading maintaining an implementation ourselves.

cibuildwheel/windows.py
```diff
@@ -1,91 +1,132 @@
...
 except ImportError:
     from pipes import quote as shlex_quote
 
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
+
 from .util import prepare_command, get_build_verbosity_extra_flags
+if hasattr(sys, 'getwindowsversion'):
```
YannickJadoul: Why? I'm not sure I understand why you're testing for this? Don't we already know we're on Windows? And if not, won't `pep514tools` just fail?

cibuildwheel/windows.py
```diff
@@ -1,91 +1,132 @@
...
+IS_RUNNING_ON_APPVEYOR = os.environ.get('APPVEYOR', 'false').lower() == 'true'
+
+
+@contextmanager
+def _mkdtemp():
+    path = tempfile.mkdtemp()
+    try:
+        yield path
+    finally:
+        shutil.rmtree(path)
```
YannickJadoul: There's probably no need to clean up, as we're on a CI image. But maybe it's cleaner, indeed?

We should remember to streamline this on macOS and Linux as well, then, this cleaning up? (in some other commit/PR)

cibuildwheel/windows.py
```diff
@@ -1,91 +1,132 @@
...
+        elif IS_RUNNING_ON_APPVEYOR:
+            # We're running on AppVeyor
+            major, minor = config.version.split('.')[:2]
+            python_path = 'C:\\Python{major}{minor}{arch}'.format(
+                major=major,
+                minor=minor,
+                arch = '-x64' if config.arch == '64' else ''
+            )
+            if os.path.exists(python_path):
+                return python_path
```
YannickJadoul: If we have this code anyway, isn't it easier/slighlty simpler code to just look for known locations, even on Python 3.5?

I mean:
```
if IS_RUNNING_ON_AZURE:
    # ...
elif IS_RUNNING_ON_APPVEYOR:
    # ...

if (int(major), int(minor)) >= (3, 5):
    # PEP 514

# ...
```

cibuildwheel/windows.py
```diff
@@ -1,91 +1,132 @@
...
+            print('+ Download ' + url + ' to ' + installer)
+            response = urlopen(url)
+            try:
+                with open(installer, 'wb') as file:
+                    file.write(response.read())
+            finally:
+                response.close()
+            print('+ Installing ' + installer + ' to ' + path)
+            args = ['cmd', '/E:ON', '/V:ON', '/C', 'start', '/wait', installer, '/quiet', 'TargetDir=' + path, 'Shortcuts=0', 'Include_launcher=0', 'InstallLauncherAllUsers=0']
+            #return subprocess.check_call(' '.join(args), env=env, cwd=cwd)
```
YannickJadoul: Supposed to be commented out, or forgotten to restore after debug?

README.md
```diff
@@ -355,10 +350,10 @@ You must set this variable to pass variables to Linux builds (since they execute
 
 You can use `$PATH` syntax to insert other variables, or the `$(pwd)` syntax to insert the output of other shell commands.
 
-Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`  
-Example: `PATH=$PATH:/usr/local/bin`  
-Example: `BUILD_TIME="$(date)"`  
-Example: `PIP_EXTRA_INDEX_URL="https://pypi.myorg.com/simple"`  
+Example: `CFLAGS="-g -Wall" CXXFLAGS="-Wall"`
```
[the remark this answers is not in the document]
YannickJadoul: See https://gist.github.com/shaunlebron/746476e6e7a4d698b373
(Which is also weird, because https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lines claims something else.)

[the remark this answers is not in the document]
mayeut: I now replaced trailing `  `  by `\` to show the visual intent clearly in the file and deal with editors which just strip trailing spaces (as do mine, but I could reconfigure for markdown).

cibuildwheel/windows.py
mayeut: It's cleaner indeed. Could be streamlined on other platforms indeed.

cibuildwheel/windows.py
mayeut: yeah, debug code. corrected.

.travis.yml
```diff
@@ -27,11 +27,11 @@ matrix:
     - os: windows
       language: shell
       before_install:
-       - choco install python3 --version 3.6.8 --no-progress -y
+       - choco install python3 --version 3.7.5 --no-progress -y
```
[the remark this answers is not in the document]
mayeut: No reason, it was part of the various tests.

cibuildwheel/_vendored/pep514tools/LICENSE
mayeut: there's no release of `pep514tools` and `pip install https://github.com/zooba/pep514tools/archive/master.zip` does not work on python 2. The easiest was to vendor this.

cibuildwheel/windows.py
mayeut: `windows.py` is imported by `__main__.py` unconditionnaly. Importing `pep514tools` fails with an import error if not protected because it uses modules only available on windows.

cibuildwheel/windows.py
mayeut: It's for the sake of consistency on all CI. Starting with python 3.5, always use `pep514tools` in order to find an official install. The one on Azure is not official (at least per pep514). I guess in less than a year, it won't matter anyway because Python 2.7 & 3.4 will be dropped thus leaving only a clean `pep514tools` method.

azure-pipelines.yml
mayeut: It's the only recommended way if you wan't python 2.7 support. Other python versions requires the installer to be run on Azure.

.travis.yml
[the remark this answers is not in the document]
YannickJadoul: Hmmm, I don't mind this single line, especially if it's related to the PR. Why not be a bit more modern :) But it was mainly that if there's a reason we need 3.7, it would be good to document.

README.md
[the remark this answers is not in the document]
YannickJadoul: Ah, yes, that might clear out future confusion as well :)
Still wondering if we should make that a separate commit/PR outside of this PR, though? It keeps the diff smaller. Then again, it's only in REAME, so it's maneagable, actually.

cibuildwheel/windows.py
YannickJadoul: OK, great!

cibuildwheel/windows.py
YannickJadoul: Hmmmm, ok, yes, why not, then.

cibuildwheel/_vendored/pep514tools/LICENSE
YannickJadoul: Ah, sorry, I hadn't tried it but I'd seen `pip install pep514tools` in the `pep514tools`'s `README`. But the project is not that active and there's no updates to expect, so we should be fine, actually. We could revisit this when dropping 2.7, though.

README.md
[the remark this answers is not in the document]
mayeut: I can make the change in a separate commit if this is clearer but it will be in the PR.

azure-pipelines.yml
YannickJadoul: Aha, yes, because we don't know the origin of the Python versions on 2.7?

In that case, the example configuration in `README` should be adapted to reflect this change, though.

README.md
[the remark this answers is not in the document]
YannickJadoul: Don't bother too much about it. I guess it's fine.

cibuildwheel/_vendored/pep514tools/LICENSE
mayeut: Yes, or if @zooba actually make a release (but I don't think he will given the project doesn't seem active)

azure-pipelines.yml
mayeut: The `README` is in sync with this change (or am I missing another line in the `README` where it's used ?)

cibuildwheel/_vendored/pep514tools/LICENSE
YannickJadoul: Yeah, no, even then the `setup.py` would still fail? Alternatively, we can make make a PR (or a fork) that works for 2.7, but the question is if that's really worth it :-/

azure-pipelines.yml
YannickJadoul: Apologies, I'd overlooked that! :-/

joerick:
Closed in favour of #180

PR #188, opened 2019-11-04
Readme: Fix the outline for Appveyor setup

paulmelnikow:
Hi! I wanted to say how impressed I was with this tool! I'm digging into wheel-building for the first time, and while it wasn't 100% smooth I manage to do it in the first sitting, which is very impressive for how tricky a problem this is! The documentation is excellent! Keep up the good work; hope I can continue to help in some small ways.

It seems like this bullet is intended to be part of the Appveyor setup item. Right now it's coming after it in the docs and it's a bit confusing, especially when using Azure DevOps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,8 +215,6 @@ jobs:
 
 </details>
 
-- Commit those files, enable building of your repo on Travis CI and AppVeyor, and push.
-
 All being well, you should get wheels delivered to you in a few minutes.
 
 > ⚠️ Got an error? Check the [checklist](#it-didnt-work) below.


joerick:
Thanks! I agree it's a little confusing, it's out of date. It might be best to remove that line completely, I think that's gonna be fixed when we move to [external docs](https://cibuildwheel.readthedocs.io/en/docs/setup/#appveyor-windows)

paulmelnikow:
Okeydoke, I removed it for now.

The new docs look great!

PR #194, opened 2019-11-05
Add support for GitHub Actions

### LINKED ISSUE #159: Support for GitHub actions
https://github.com/features/actions

(which is still in closed beta, but filing this anyhow since hopefully it's in the cards?).

mayeut:
Draft:
- [x] basic support (linux, macOS, windows 3.6, 3.7, 3.8)
- [x] python 2.7 on windows
- [x] python 3.5 on windows
- [x] documentation update

Closes #159

Here's a summary of python versions / CI providers used to test `cibuildwheel` after this PR:

|               |2.7                          |3.5           |3.6          |3.7       |3.8    |
|---------|------------------|----------|---------|-------|------|
|windows| AppVeyor               |TravisCI  |               |GitHub|Azure|
|linux       | TravisCI + CircleCI|TravisCI  |CircleCI  |TravisCI + GitHub |Azure|
|macOS  |                    CircleCI|               |TravisCI |CircleCI + GitHub |Azure|

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,29 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Test cibuildwheel on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+        python_version: ['3.7']
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      name: Install Python ${{ matrix.python_version }}
+      with:
+        python-version: ${{ matrix.python_version }}
+
+    - name: Install dependencies
+      run: |
+        python -m pip install -r requirements-dev.txt
+    - name: Install Visual C++ for Python 2.7
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        choco install vcpython27 -f -y
+    - name: Test cibuildwheel
+      run: |
+        python ./bin/run_tests.py
--- a/examples/github-minimal.yml
+++ b/examples/github-minimal.yml
@@ -1,0 +1,32 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Build wheel on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      name: Install Python
+      with:
+        python-version: '3.7'
+
+    - name: Install cibuildwheel
+      run: |
+        python -m pip install cibuildwheel==1.2.0
+    - name: Install Visual C++ for Python 2.7
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        choco install vcpython27 -f -y
+    - name: Build wheel
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+    - uses: actions/upload-artifact@v1
+      with:
+        name: wheels
+        path: ./wheelhouse
--- a/CI.md
+++ b/CI.md
@@ -1,10 +1,10 @@
 This is a summary of the Python versions and platforms covered by the different CI platforms:
 
-|          | 3.5              | 3.6              | 3.7                   | 3.8              |
-|----------|------------------|------------------|-----------------------|------------------|
-| Linux    | Travis CI        | CircleCI         | AppVeyor              | Azure Pipelines  |
-| macOS    | Azure Pipelines  | CircleCI         | Travis CI¹ / CircleCI | Azure Pipelines  |
-| Windows  | TravisCI         | Azure Pipelines  | AppVeyor              | Azure Pipelines  |
+|          | 3.5              | 3.6              | 3.7                                               | 3.8              |
+|----------|------------------|------------------|---------------------------------------------------|------------------|
+| Linux    | Travis CI        | CircleCI         | AppVeyor / GitHub Actions                         | Azure Pipelines  |
+| macOS    | Azure Pipelines  | CircleCI         | AppVeyor / Travis CI¹ / CircleCI / GitHub Actions | Azure Pipelines  |
+| Windows  | TravisCI         | Azure Pipelines  | AppVeyor / GitHub Actions                         | Azure Pipelines  |
 
 > ¹ Python version not really pinned, but dependent on the (default) version of image used.
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ cibuildwheel
 
 Python wheels are great. Building them across **Mac, Linux, Windows**, on **multiple versions of Python**, is not.
 
-`cibuildwheel` is here to help. `cibuildwheel` runs on your CI server - currently it supports Azure Pipelines, Travis CI, AppVeyor, and CircleCI - and it builds and tests your wheels across all of your platforms.
+`cibuildwheel` is here to help. `cibuildwheel` runs on your CI server - currently it supports Azure Pipelines, Travis CI, AppVeyor, GitHub Actions and CircleCI - and it builds and tests your wheels across all of your platforms.
 
 
 What does it do?
@@ -35,13 +35,14 @@ What does it do?
 Usage
 -----
 
-`cibuildwheel` currently works on **Travis CI**, **Azure Pipelines** and **AppVeyor** to build wheels for all three supported platforms (Linux, macOS, Windows). On **CircleCI** Linux and macOS wheels can be built.
+`cibuildwheel` currently works on **Travis CI**, **Azure Pipelines**, **AppVeyor** and **GitHub Actions** to build wheels for all three supported platforms (Linux, macOS, Windows). On **CircleCI** Linux and macOS wheels can be built.
 
 |                 | Linux | macOS | Windows |
 |-----------------|-------|-------|---------|
 | Azure Pipelines | ✅    | ✅    | ✅      |
 | Travis CI       | ✅    | ✅    | ✅      |
 | AppVeyor        | ✅    | ✅    | ✅      |
+| GitHub Actions  | ✅    | ✅    | ✅      |
 | CircleCI        | ✅    | ✅    |         |
 
 `cibuildwheel` is not intended to run on your development machine. Because it uses system Python from Python.org it will try to install packages globally - not what you expect from a build tool! Instead, isolated CI services like Travis CI, CircleCI, Azure Pipelines and AppVeyor are ideal.
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -77,12 +77,12 @@ def main():
     if args.platform != 'auto':
         platform = args.platform
     else:
-        ci = strtobool(os.environ.get('CI', 'false')) or 'BITRISE_BUILD_NUMBER' in os.environ or 'AZURE_HTTP_USER_AGENT' in os.environ
+        ci = strtobool(os.environ.get('CI', 'false')) or 'BITRISE_BUILD_NUMBER' in os.environ or 'AZURE_HTTP_USER_AGENT' in os.environ or 'GITHUB_WORKFLOW' in os.environ
         if not ci:
             print('cibuildwheel: Unable to detect platform. cibuildwheel should run on your CI server, '
-                  'Travis CI, AppVeyor, Azure Pipelines and CircleCI are supported. You can run on your '
-                  'development machine or other CI providers using the --platform argument. Check --help '
-                  'output for more information.',
+                  'Travis CI, AppVeyor, Azure Pipelines, GitHub Actions and CircleCI are supported. You '
+                  'can run on your development machine or other CI providers using the --platform argument. '
+                  'Check --help output for more information.',
                   file=sys.stderr)
             exit(2)
         if sys.platform.startswith('linux'):
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -65,6 +65,19 @@ AppVeyor will store the built wheels for you - you can access them from the proj
 
 For more info on this config file, check out the [docs](https://www.appveyor.com/docs/).
 
+# GitHub Actions [linux/mac/windows] {: #github-actions}
+
+Using GitHub Actions, you can build all three platforms on the same service. Create a `./github/workflows/build.yml` file in your repo.
+
+> build.yml
+```yaml
+{% include "../examples/github-minimal.yml" %}
+```
+
+Commit this file, enable building of your repo on GitHub Actions, and push.
+
+For more info on this file, check out the [docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions).
+
 > ⚠️ Got an error? Check the [FAQ](faq.md).
 
 <script>
--- a/unit_test/main_tests/main_platform_test.py
+++ b/unit_test/main_tests/main_platform_test.py
@@ -11,6 +11,7 @@ def test_unknown_platform_non_ci(monkeypatch, capsys):
     monkeypatch.delenv('CI', raising=False)
     monkeypatch.delenv('BITRISE_BUILD_NUMBER', raising=False)
     monkeypatch.delenv('AZURE_HTTP_USER_AGENT', raising=False)
+    monkeypatch.delenv('GITHUB_WORKFLOW', raising=False)
 
     with pytest.raises(SystemExit) as exit:
         main()


YannickJadoul:
Woohoo, great, @mayeut!

mayeut:
Python 3.5 on Windows works after patching distutils. I'm trying to identify if there was a regression in setuptools to signal that at setuptools rather than patching distutils.

mayeut:
macOS 10.14 support dropped in GitHub actions, this will require to build x86_64 only wheels. Waiting for #156 to land to see how to manage that.

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macOS-10.14]
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      name: Install Python ${{ matrix.python_version }}
+      with:
+        python-version: "3.7"
```
mayeut: c.f. #198

azure-pipelines.yml
```diff
@@ -19,6 +19,8 @@ jobs:
   pool: {vmImage: 'vs2017-win2016'}
   steps:
     - task: UsePythonVersion@0
+      inputs:
+        versionSpec: '3.7'
```
mayeut: c.f. #198

cibuildwheel/windows.py
```diff
@@ -132,6 +133,14 @@ def shell(args, env=None, cwd=None):
         simple_shell(['pip', '--version'], env=env)
         simple_shell(['pip', 'install', '--upgrade', 'setuptools', 'wheel'], env=env)
 
+        # setuptools does not setup the compiler correctly on GH Actions for python 3.5
+        # c.f. https://github.com/pypa/setuptools/issues/1851#issuecomment-537714885
+        # Let's patch setuptools with distutils discovery mechanism from 3.8.0
+        shutil.copyfile(
+            os.path.join(os.path.dirname(__file__), 'resources', 'msvc.py'),
+            os.path.join(config_python_path, 'Lib', 'site-packages', 'setuptools', 'msvc.py')
+        )
```
joerick: This is quite a funny hack! :) I guess it'll only last until this PR is merged/released? https://github.com/pypa/setuptools/pull/1904

mayeut: yes

mayeut: This is a big hack while waiting for upstream PR, I'm not keen on getting the hack baked into cibuildwheel.

mayeut:
> Just docs to go?

Still blocking are also #156, #202 and #220 since GH is using macOS 10.15

remorses:
Really appreciate your work guys, waiting for this pull request to finally be able to create wheels for my projects on github

mayeut:
The python 3.5 build started to fail again... Need to debug this some more.

myselfhimself:
Let me test this PR for macosx only on Github workflow here: https://github.com/dtschump/gmic-py/blob/master/.github/workflows/cpythonmacosbuild.yml (WIP)

myselfhimself:
I am unhappy with this branch vs. the regular setup.py's (our manylinux builds without any cibuildwheel worked well) native behaviour relying on xcode/gcc/g++....
I pretty much do not care about clang.. and cannot make it work because of an MACOSX_DEPLOYMENT_TARGET variable to tune or a  at some stage or '-stdlib=libc++' to add as an extra_compile_arg for the clang compiler.. Let me troubleshoot further..
Maybe delocate needs clang.. does it?
Here is an example failing build: https://github.com/dtschump/gmic-py/commit/9a5712d136fd81e42926a2c477cfb13f22f4a764/checks?check_suite_id=364999034

myselfhimself:
Hmm OK I fixed it somehow :)
Here is the working build:
https://github.com/dtschump/gmic-py/commit/1be036c03ea5b1bef85193e40ab986df6c658fdc/checks?check_suite_id=365014107

The corresponding github action file, which has hardcoded `MACOSX_DEPLOYMENT_TARGET=10.10` in several places (pip, before cibuildwheel, env) https://github.com/dtschump/gmic-py/commit/1be036c03ea5b1bef85193e40ab986df6c658fdc

adding  '-stdlib=libc++' clang flag within the extra_compile_arg as to setup.py did not impact the g++ based compilations (manylinux ones without cibuildwheel and handmade ubuntu/debian one)
Here is the working setup.py:
https://github.com/dtschump/gmic-py/commit/210b2eb5954d6102d8156c52e0d5112361ebbb0e

Some background info on the project that I am building: it is a Python binding to a C++11 graphics project

myselfhimself:
Sorry to drift a bit off-topic though, on Github actions, the existing clang through Xcode 1.11 does not want to understand `-fopenmp` which is used by many multi-process libraries. OpenMP was supposed to be included within Clang from version 4.8..

Here is the setup.py :
https://github.com/dtschump/gmic-py/commit/689de230cc3c7d24fd37f35d062786dafced564d

Here is the build log:
https://github.com/dtschump/gmic-py/commit/ad17f785a3b0167e230da2adffc628d224aba5d2/checks?check_suite_id=365184781

myselfhimself:
Just tried the fix here to get openmp working through the Xcode's clang:
https://iscinumpy.gitlab.io/post/omp-on-high-sierra/
And it did work!!! :+1: 
This boils down to this in a setup.py:
```python
if sys.platform == 'darwin':
    extra_compile_args += ['-Xpreprocessor -fopenmp']
```

myselfhimself:
Regarding this failing test:
https://github.com/dtschump/gmic-py/commit/ce4f54a54b5918fb5a0502f607e93ccc697d9340/checks?check_suite_id=365217158
The assumed working directory was `working-directory: ./cpython` (relative to the project rootdir) as listed in this GH recipe file: https://github.com/dtschump/gmic-py/commit/ce4f54a54b5918fb5a0502f607e93ccc697d9340?diff=unified#diff-13e025583f81712a1f3bbb1846c47bf5
in which directory lives tests: cpython/tests
I had ditched the `{project}` variable from some cibuildwheels documentation thinking it would be useless...

samuelcolvin:
I've tried this and I'm getting the same error as #244.

see [this](https://github.com/samuelcolvin/rtoml/pull/4/checks?check_run_id=384818038) build which used head of this PR.

samuelcolvin:
The reason this is currently failing on windows is that pip installed at `/c/hostedtoolcache/windows/Python/3.7.5/x64/Scripts/pip` (that's the output of `which pip`) not `C:/cibw/python\pythonx86.3.7.6\tools\Scripts`.

I have this working if I just comment out

```py
assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
```

mayeut:
@samuelcolvin,

`cibuildwheel` is right to check for `pip` where it installed `python` (i.e. `C:/cibw/python\pythonx86.3.7.6\tools\Scripts`)

The `/c/hostedtoolcache/windows/Python/3.7.5/x64/Scripts/pip` location is the one from the default `python` tool configured on the VM and is not the one `cibuildwheel` shall use.

The output in the linked failing build is really weird:
```
Successfully installed 'pythonx86 3.8.1' to C:\cibw\python
Executing nuget actions took 6.1 sec
+ where python
C:\hostedtoolcache\windows\Python\3.7.5\x64\python.exe
C:\hostedtoolcache\windows\Python\3.6.8\x64\python.exe
C:\ProgramData\Chocolatey\bin\python.exe
+ python --version
Python 3.7.5
+ python -c "import struct; print(struct.calcsize('P') * 8)"
+ python C:\cibw\get-pip.py
Collecting pip
  Downloading https://files.pythonhosted.org/packages/00/b6/9cfa56b4081ad13874b0c6f96af8ce16cfbc1cb06bedf8e9164ce5551ec1/pip-19.3.1-py2.py3-none-any.whl (1.4MB)
Installing collected packages: pip
  Found existing installation: pip 19.3.1
    Uninstalling pip-19.3.1:
      Successfully uninstalled pip-19.3.1
Successfully installed pip-19.3.1
  File "c:\hostedtoolcache\windows\python\3.7.5\x64\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\hostedtoolcache\windows\python\3.7.5\x64\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\hostedtoolcache\windows\Python\3.7.5\x64\Scripts\cibuildwheel.exe\__main__.py", line 7, in <module>
  File "c:\hostedtoolcache\windows\python\3.7.5\x64\lib\site-packages\cibuildwheel\__main__.py", line 180, in main
    cibuildwheel.windows.build(**build_options)
  File "c:\hostedtoolcache\windows\python\3.7.5\x64\lib\site-packages\cibuildwheel\windows.py", line 142, in build
    assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
AssertionError
```

This is the build log for python 3.8, but, as can be seen, `where python` returns:
```
C:\hostedtoolcache\windows\Python\3.7.5\x64\python.exe
C:\hostedtoolcache\windows\Python\3.6.8\x64\python.exe
C:\ProgramData\Chocolatey\bin\python.exe
```
which is missing the `python` that's been installed by `cibuildwheel`...
That probably means the issue is in setting up the environment for targeted python.
I can see that you have a specific environment: `ParsedEnvironment(['PATH="$HOME/.cargo/bin:$PATH"'])` which I guess is one source of investigation possible.

samuelcolvin:
The 3.8 build might have been installing python 3.7 in that build, I fixed it later.

But that doesn't explain why the 3.7 build was failing too.

mayeut:
@samuelcolvin, no matter what python you configure for GitHub Action,  `cibuildwheel` will install its own so "The 3.8 build might have been installing python 3.7 in that build, I fixed it later." is not a fix but rather shows there's something weird in the environment.
You could configure GitHub Action with python `3.5` and use `cibuildwheel` to build python `2.7, 3.5, 3.6, 3.7 & 3.8` in one pass.

samuelcolvin:
That's what I had assumed, just trying to help.

mayeut:
If you can printout the environment dict `env` [here](https://github.com/samuelcolvin/cibuildwheel/blob/c734bcae276eacc9ec2f664608230e73b49d1c4d/cibuildwheel/windows.py#L134) in your fork, I think that might help.

YannickJadoul:
> I can see that you have a specific environment: `ParsedEnvironment(['PATH="$HOME/.cargo/bin:$PATH"'])` which I guess is one source of investigation possible.

Windows' `PATH` entries are separated by `;` rather than `:`. Could this be the problem?

samuelcolvin:
I don't think so, but anyone is free to clone rtoml, make the changes and submit a PR which will try to build the binaries and invoke cibuildwheel.

YannickJadoul:
> I don't think so, but anyone is free to clone rtoml, make the changes and submit a PR which will try to build the binaries and invoke cibuildwheel.

If I find half an hour of spare time, I'll see if I can investigate.

YannickJadoul:
@samuelcolvin Do you still need someone to look at that problem of yours? I might have some time one of these days.

samuelcolvin:
> @samuelcolvin Do you still need someone to look at that problem of yours? I might have some time one of these days.

@YannickJadoul I'm still using my own fork of `cibuildwheel`, I don't know whether the problem i had is would still occur with HEAD of this PR, simplest approach would be to create a PR against https://github.com/samuelcolvin/rtoml and modify `.github/workflows/ci.yml` to use a different `cibuildwheel` and see what happens. Every push will run `cibuildwheel`.

YannickJadoul:
> > @samuelcolvin Do you still need someone to look at that problem of yours? I might have some time one of these days.
> 
> @YannickJadoul I'm still using my own fork of `cibuildwheel`, I don't know whether the problem i had is would still occur with HEAD of this PR, simplest approach would be to create a PR against https://github.com/samuelcolvin/rtoml and modify `.github/workflows/ci.yml` to use a different `cibuildwheel` and see what happens. Every push will run `cibuildwheel`.

Maybe I've done something wrong, but the PR didn't trigger your CI, it seems

cibuildwheel/resources/msvc.py
```diff
@@ -0,0 +1,1822 @@
...
+
+    def __init__(self, registry_info, vc_ver=None):
+        self.ri = registry_info
+        self.pi = self.ri.pi
+
+        self.known_vs_paths = self.find_programdata_vs_vers()
+
+        # Except for VS15+, VC version is aligned with VS version
+        self.vs_ver = self.vc_ver = (
+                vc_ver or self._find_latest_available_vs_ver())
```
nsoranzo:
```suggestion
            vc_ver or self._find_latest_available_vs_ver())
```

cibuildwheel/resources/msvc.py
```diff
@@ -0,0 +1,1822 @@
...
+
+        Return
+        ------
+        str
+            name
+        """
+        matching_dirs = (
+            dir_name
+            for dir_name in reversed(listdir(path))
+            if isdir(join(path, dir_name)) and
```
nsoranzo:
```suggestion
            if isdir(join(path, dir_name))
```

cibuildwheel/resources/msvc.py
```diff
@@ -0,0 +1,1822 @@
...
+            dir_name.startswith(prefix)
```
nsoranzo:
```suggestion
            and dir_name.startswith(prefix)
```

cibuildwheel/resources/msvc.py
```diff
@@ -0,0 +1,1822 @@
...
+        """
+        if self.vs_ver <= 10.0:
+            arch_subdir = self.pi.target_dir(hidex86=True, x64=True)
+            return [join(self.si.WindowsSdkDir, 'Lib%s' % arch_subdir)]
+
+        else:
+            arch_subdir = self.pi.target_dir(x64=True)
+            lib = join(self.si.WindowsSdkDir, 'lib')
+            libver = self._sdk_subdir
+            return [join(lib, '%sum%s' % (libver , arch_subdir))]
```
nsoranzo:
```suggestion
            return [join(lib, '%sum%s' % (libver, arch_subdir))]
```

cibuildwheel/resources/msvc.py
```diff
@@ -0,0 +1,1822 @@
...
+        if self.vs_ver >= 11.0:
+            libpath += [join(ref, r'CommonConfiguration\Neutral')]
+
+        if self.vs_ver >= 14.0:
+            libpath += [
+                ref,
+                join(self.si.WindowsSdkDir, 'UnionMetadata'),
+                join(ref, 'Windows.Foundation.UniversalApiContract', '1.0.0.0'),
+                join(ref, 'Windows.Foundation.FoundationContract', '1.0.0.0'),
+                join(ref,'Windows.Networking.Connectivity.WwanContract',
```
nsoranzo:
```suggestion
                join(ref, 'Windows.Networking.Connectivity.WwanContract',
```

cibuildwheel/resources/msvc.py
YannickJadoul: Thanks, @nsoranzo. I'm wondering, though: maybe we shouldn't try to fix vendored things, but just ignore `resources/`?

nsoranzo: Good point, I didn't realise that is was a case of vendoring.

joerick: `cibuildwheel/resources` is ignored now, I made that adjustment to `master`

https://github.com/joerick/cibuildwheel/blob/master/setup.cfg#L25

joerick: If @mayeut rebases/merges `master`, these lint errors should go away :)

mayeut: I will rebase tonight.
I did not apply the style changes to this file because I expect it to be included as-is in setuptools.
Anyway, I'm waiting for setuptools to include this change before opening this PR for review.
At this point, this file won't be included in cibuildwheel.

filips123:
@mayeut Is there any update? What needs to be done to properly support GitHub Actions?

joerick:
I meant to say @mayeut, I can see this is a popular feature, I'm happy to take over this PR and write some docs if you don't have time at the moment?

mayeut:
It was missing a patch in setuptools which has been released yesterday.
I updated the docs but the minimal example is still lacking an archive artifacts section.
@joerick, feel free to push on this branch to update the docs if needed.

cibuildwheel/windows.py
mayeut: This hack is not needed anymore, setuptools has been released with the patch.

YannickJadoul:
A quick reminder to check the table in `CI.md` and add the new CI configuration there. The file was since the start of this PR, and quite a few updates to our own CI have been done.

joerick:
Fantastic, thanks @mayeut! I'll take a good look soon.

examples/github-minimal.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Install cibuildwheel
+      run: |
+        python -m pip install cibuildwheel==1.2.0
+    - name: Install Visual C++ for Python 2.7
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        choco install vcpython27 -f -y
+    - name: Build wheel
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
```
joerick: It would be nice to keep the wheels as artifacts in the minimal config, I think

```suggestion
        python -m cibuildwheel --output-dir wheelhouse
    - uses: actions/upload-artifact@v1
       with:
         name: wheels
         path: ./wheelhouse
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  test:
+    name: Test cibuildwheel on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v1
```
jlaine: actions/checkout@v2 seems to be the current version, maybe go with that?

examples/github-minimal.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  test:
+    name: Build wheel on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v1
```
jlaine: same as above, maybe use v2?

examples/github-minimal.yml
jlaine: Definitely keep the wheels as artifacts, especially as uploading wheels to PyPI using `pypa/gh-action-pypi-publish` only works on Linux, so you have to:

- upload wheels for the various platforms as artifacts
- download all the wheels in a job running on Linux
- upload them to PyPI

NOTE: I'm not sure what happens if you give the same name to uploads from all platforms

.github/workflows/test.yml
henryiii: Note that version 2 doesn’t support submodules. Just an aside.

jlaine: OK, forget my comment then, v1 works :)

examples/github-minimal.yml
lithammer: >I'm not sure what happens if you give the same name to artifacts from all platforms

It's not a problem as long as the file names are unique.

jlaine:
One thing which might be worth testing with the provided example is the behaviour if cibuildwheel fails. I have had very confusing experiences with Github Actions on Windows: the default shell is powershell, and it seems exit with a success code despite one or more commands in a multi-line failing. I ended switching over to `bash` for all platforms to get proper failures on error.

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+    - uses: actions/setup-python@v1
+      name: Install Python ${{ matrix.python_version }}
```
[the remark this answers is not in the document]
mayeut: I'd rather not hardwire the version in the name since changing python version would require 2 changes instead of 1.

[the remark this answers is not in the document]
joerick: Oh... Maybe I'm misunderstanding - I don't see a matrix entry for `python_version`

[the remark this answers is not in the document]
mayeut: My mistake, you're absolutely right. Maybe just leaving it blank instead ? I hate changing things multiple times when it should be changed only once. I don't know if we can set it in the matrix with a single value (probably).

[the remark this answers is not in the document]
mayeut: I fixed this.

mayeut:
> Only other thing I wonder is- should this be the first CI service listed in setup.md, rather than the last? What's your impression of Github Actions, generally? I ask because Azure currently has that spot (because it was the first multi-platform service and has great build times) but I wonder if this might be simpler and have a better DX. Any thoughts @mayeut ?

Not sure it helps you decide the order but just my 2 cents on this:
I was wondering the same thing, let's see how it compares to Azure (mostly build times).
TravisCI now has more platform support but if you're not interested in s390x, aarch64, ppc64le, then I guess Azure/GitHub are faster (probably true especially on macOS where the waiting queue can grow quite fast, also Windows is still BETA at Travis).
I would not recommend AppVeyor today, even though it's now multipatform, because of build times with one "parallel" job.

henryiii:
A couple more cents: One nice feature of GHA is no special setup/permissions. Azure seems to always be a bit of a headache to get more than one collaborator on because it has a separate repo and permissions model. The performance is generally very similar (since they are pretty much the same platform in the backend). Azure is more powerful (manual build triggers, release pipelines).

joerick:
Thanks both. I'm leaning towards putting it at the top for the reasons you describe. 

This PR looks ready to me! Any more comments before we merge?

YannickJadoul:
Minor detail, but we could eliminate CircleCI running on Python 3.7 on macOS, since it's already running on Python 3.6 anyway.

joerick:
Thanks all! I'll try to get a release out with this very soon!

joerick:
Released as 1.3.0!

PR #199, opened 2019-11-09
Tests cibuildwheel across different python versions

### LINKED ISSUE #198: cibuildwheel CI tests failing on Azure for windows
`cibuildwheel` CI tests which are using the sample configuration in README are failing on Windows following Azure update to support python 3.8

Given the number of CI providers now tested, I guess we can try to test `cibuildwheel` on python 2.7, 3.5, 3.6, 3.7 and 3.8 without too much overhead on test time by dispatching the python versions running `cibuildwheel` across CI providers.

mayeut:
Fixes #198 

Here's a summary of python versions / CI providers used to test `cibuildwheel`:

|               |2.7                          |3.5           |3.6          |3.7|3.8    |
|---------|------------------|----------|---------|---|------|
|windows| AppVeyor               |AppVeyor|Travis CI|      |Azure|
|linux       | TravisCI + CircleCI|Travis CI  |CircleCI |      |Azure|
|macOS  |TravisCI + CircleCI|                  |TravisCI|CircleCI|Azure|

### DIFF
--- a/unit_test/environment_test.py
+++ b/unit_test/environment_test.py
@@ -1,3 +1,4 @@
+import os
 from cibuildwheel.environment import parse_environment
 
 
@@ -37,12 +38,15 @@ def test_inheritance():
 def test_shell_eval():
     environment_recipe = parse_environment('VAR="$(echo "a test" string)"')
 
+    env_copy = os.environ.copy()
+    env_copy.pop('VAR', None)
+
     environment_dict = environment_recipe.as_dictionary(
-        prev_environment={}
+        prev_environment=env_copy
     )
     environment_cmds = environment_recipe.as_shell_commands()
 
-    assert environment_dict == {'VAR': 'a test string'}
+    assert environment_dict['VAR'] == 'a test string'
     assert environment_cmds == ['export VAR="$(echo "a test" string)"']
 
 def test_shell_eval_and_env():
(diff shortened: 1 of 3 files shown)


mayeut:
I think we can safely remove macOS python 2.7 from travis-ci in order to reduce the time needed to pass tests in CI. Any thoughts @joerick, @YannickJadoul ?

unit_test/environment_test.py
```diff
@@ -37,12 +38,15 @@ def test_inheritance():
...
+    env_copy.pop('VAR', None)
+
     environment_dict = environment_recipe.as_dictionary(
-        prev_environment={}
+        prev_environment=env_copy
     )
     environment_cmds = environment_recipe.as_shell_commands()
 
-    assert environment_dict == {'VAR': 'a test string'}
+    assert environment_dict['VAR'] == 'a test string'
```
joerick: What's the thinking/reasoning behind this change?

joerick: ah, i see it fixes Python 3.8 on azure... what's your thinking why?

joerick: I checked the build log. a bug in python 3.8?

joerick:
I want to get master green again, so merging now.

unit_test/environment_test.py
mayeut: It's either a bug in python 3.8 on windows or a bug in the tests.
For sure, behavior has changed which would suggest a bug in python 3.8 however passing an empty environment to subprocess feels wrong to me so I updated the test to pass the full environment instead.

PR #200, opened 2019-11-09
Add check that failing tests fail on manylinux1 specifically

joerick:
Another test for #175 ... hopefully this fails :)


joerick:
The test failure is on Mac Travis, due to a temporary network failure. Otherwise, since merging master, this is green.

PR #185, opened 2019-10-30
Adding PyPy support

YannickJadoul:
There's probably still a lot to discuss on choosing defaults, option names, build tag names, ... but we can already give this a go and see if we can get the technical part working :)

Cfr. #143

PR #202, opened 2019-11-10
Run 32-bit tests on macOS

### LINKED ISSUE #203: 32-bit tests are not run on macOS when creating an intel wheel
Only 64-bit tests are run on macOS when creating a dual architecture `intel` wheel

mayeut:
Tests are only run in 64-bit mode on macOS while `intel` wheels are being produced.
This PR adds running tests in 32-bit mode.

There's an issue in [virtualenv](https://github.com/pypa/virtualenv/issues/1437) that prevents to get python2 working as well.

Fixes #203

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -76,6 +76,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
         if config.version[0] == '3':
             os.symlink(os.path.join(installation_bin_path, 'python3'), '/tmp/cibw_bin/python')
+            os.symlink(os.path.join(installation_bin_path, 'python3-32'), '/tmp/cibw_bin/python-32')
             os.symlink(os.path.join(installation_bin_path, 'python3-config'), '/tmp/cibw_bin/python-config')
             os.symlink(os.path.join(installation_bin_path, 'pip3'), '/tmp/cibw_bin/pip')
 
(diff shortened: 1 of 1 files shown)


mayeut:
I opened a PR at [virtualenv](https://github.com/pypa/virtualenv/pull/1438) to get things working smoothly.

mayeut:
It seems `virtualenv` is in the middle of a major rewrite so it won't be fixed for python 2 for at least 1-2 months.
I think providing a way to test what's built, even if only in python 3, it's better than nothing.
Opening for review.

cibuildwheel/macos.py
```diff
@@ -121,12 +122,23 @@ def call(args, env=None, cwd=None, shell=False):
...
         delocated_wheel = glob('/tmp/delocated_wheel/*.whl')[0]
 
+        test_drivers = []
         if test_command:
+            if 'intel' in config.identifier and config.version[0] == '3':
+                test_drivers = ['python', 'python-32']
+            else:
+                test_drivers = ['python']
+
+        for test_driver in test_drivers:
```
YannickJadoul: Hmmm, nicely compact, this way, but now I'm wondering if, logically/for readability, it wouldn't make sense to have all of this inside the `if test_command:`?

Czaki:
If there is an option for omit this test? Package may depend on 64 bits only library (compiled external). And there is no option for change python installer on 64-bits without fork.

cibuildwheel/macos.py
```diff
@@ -121,12 +122,23 @@ def call(args, env=None, cwd=None, shell=False):
...
             # there are no dependencies that were pulled in at build time.
-            call(['pip', 'install', 'virtualenv'], env=env)
+            if config.version[0] == '2':
+                call([test_driver, '-m', 'pip', 'install', 'virtualenv'], env=env)
             venv_dir = tempfile.mkdtemp()
-            call(['python', '-m', 'virtualenv', venv_dir], env=env)
+            if config.version[0] == '2':
+                call([test_driver, '-m', 'virtualenv', venv_dir], env=env)
+            else:
+                call([test_driver, '-m', 'venv', venv_dir], env=env)
```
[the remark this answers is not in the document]
mayeut: Exactly. I could now add a comment about that.

cibuildwheel/macos.py
```diff
@@ -73,6 +73,7 @@ def call(args, env=None, cwd=None, shell=False):
 
         if config.version[0] == '3':
             os.symlink(os.path.join(installation_bin_path, 'python3'), '/tmp/cibw_bin/python')
+            os.symlink(os.path.join(installation_bin_path, 'python3-32'), '/tmp/cibw_bin/python-32')
```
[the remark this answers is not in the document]
mayeut: yeah, but shouldn't be too hard to change one way or the other.

cibuildwheel/macos.py
mayeut: So just indent the for loop one level right then ?
You're probably right. I took a bad habit of finding tricks in order not to get too much multi-line statements with linters enforced rules (as the one coming in #212)

mayeut:
> If there is an option for omit this test? Package may depend on 64 bits only library (compiled external). And there is no option for change python installer on 64-bits without fork.

If a package depends on 64 bits only library, then it should not be deployed as an `intel` wheel (condition to trigger 32-bit tests). IMHO, if `cibuildwheel` fails to prevent such an upload, it's a major failure of `cibuildwheel`. I don't even know if such a wheel can be built... maybe you have more insight on that matter.

That being said, if such a wheel can be built, I think you're right in wanting to propose an  alternative but that would be another PR. Probably something in the like of having an option to target only x86_64 (even with `intel` installers like 3.5 where there are no alternative installers, maybe through a `delocate` option, might need a PR there).

Czaki:
@mayeut But this is current situation. I ask about this in dealocate repository https://github.com/matthew-brett/delocate/issues/56#issuecomment-549769943

Current revision of cibuildwheel do not allow to choose only 64 bits for macos. And I think that these need to be introduced before merge this PR. 

Also I remember that in #156 there shown suggestion to drop intel tag when drop python 3.5

mayeut:
> But this is current situation. I ask about this in dealocate repository matthew-brett/delocate#56 (comment)

Thanks for the reminder

> Current revision of cibuildwheel do not allow to choose only 64 bits for macos. And I think that these need to be introduced before merge this PR.

That I can agree with.

> Also I remember that in #156 there shown suggestion to drop intel tag when drop python 3.5

Right, but maybe we can do something in the meantime.

mayeut:
I'll check if I can add a quick test building a 64-bit only wheel.

cibuildwheel/macos.py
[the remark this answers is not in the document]
YannickJadoul: OK, fair enough; not a blocker at all. Rather an invitation to think about this part of the code in general, in #185. I'm already not happy about how I hacked that in #185, so maybe you have better solution up your sleeves :-)

cibuildwheel/macos.py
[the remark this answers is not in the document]
YannickJadoul: Yeah, it's getting harder and harder to keep track of all reasons why something happened, now the project is making such progress. Thanks!

cibuildwheel/macos.py
YannickJadoul: I don't know, either. Too much indentation is really annoying either, especially with line length limits (not a big fan there, but fair enough, there's a standard for some reason?), but in this case ... up to your judgment, but worth a mention.

Czaki: Extract whole test to another function?

YannickJadoul:
> @YannickJadoul Check your cmake build project if it is multiarch.

Interesting, yes, thanks! Any idea how to best test this? Just run `python-32` and try importing?

Czaki:
I see three options:
- Run whole all tests with python-32. 
- Run verbose compilation and check if there are two `-arch` in flags. 
- Run `otool -f` on all library files.

cibuildwheel/macos.py
```diff
@@ -122,39 +123,54 @@ def call(args, env=None, cwd=None, shell=False):
...
-            # (this ensures that Python runs the tests against the installed wheel
-            # and not the repo code)
-            test_command_prepared = prepare_command(test_command, project=abs_project_dir)
-            call(test_command_prepared, cwd=os.environ['HOME'], env=virtualenv_env, shell=True)
-
-            # clean up
-            shutil.rmtree(venv_dir)
+            # Because of https://github.com/pypa/virtualenv/issues/1437, we can only test
+            # 32-bit mode using python 3 embedded venv module.
+            if 'intel' in config.identifier and config.version[0] == '3':
```
Czaki: maybe change this to checking wheel file name. Then if there is method to "repair" wheel after construction it will still work.

mayeut:
I found a way to build `x86_64` wheel with python 3.5 `intel` installer. Will be tomorrow for the PR though, it's getting late.
Proposition: the other PR adds an option for targeting `x86_64` only on macOS, then this PR can always run tests when `intel` is found in the wheel name. My fear is that it will clash with @Czaki PR. Given that his `wheel` PR has been merged, maybe we can hope for something soon and merge those PRs after #156.

YannickJadoul:
@Czaki

> Run all tests with python-32.

Grrrrmbl, I imported it and get an error :-( Thanks for mentioning this!

Czaki:
@YannickJadoul you can participate in this discussion https://github.com/matthew-brett/delocate/issues/56
I think that in future release dealocate with deal with it.

YannickJadoul:
@Czaki Thanks. But it's maybe my own fault/problem, if I'm calling CMake. Btw, do not that I'm still calling `wheel` and `setup.py` and only then CMake (through a custom `build_ext` in `setup.py`). So the file and wheel names etc are nicely handled by `setup.py` :-)

Czaki:
@YannickJadoul Ia all libraries on which you depend is build by your cmake, or you have some external dependencies (if the folder `.lib` in wheel is empty? - it may be hidden)?

YannickJadoul:
@Czaki No, I don't think so. Everything is just included in the compilation.

mayeut:
blocked waiting for #220

Czaki:
What did you think to use `--require-archs` option in delocate?  https://github.com/matthew-brett/delocate/blob/3e2364bd4e92b9d4ca35e8ce46510925d38befc2/delocate/cmd/delocate_wheel.py#L40-L43

mayeut:
@Czaki, yes, I saw that and I'm going to add it to the PR on `intel` vs `x86_64` option

mayeut:
Closing in favor of #220 (only build `x86_64`)

PR #204, opened 2019-11-10
Rework platform detection to ignore CI provider.

### LINKED ISSUE #146: Linux support in AppVeyor
FYI
https://www.appveyor.com/blog/2018/03/06/appveyor-for-linux/
https://www.appveyor.com/blog/2018/05/15/appveyor-for-linux-is-generally-available/
https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/

mayeut:
CI provider is now mostly irrelevant (#180), rework platform detection with that in mind.
Run linux test in AppVeyor.

Fixes #146 

Here's a summary of python versions / CI providers used to test `cibuildwheel` after this PR:

|               |2.7                          |3.5           |3.6          |3.7       |3.8    |
|---------|------------------|----------|---------|-------|------|
|windows| AppVeyor               |TravisCI  |               |             |Azure|
|linux       | TravisCI + CircleCI|TravisCI  |CircleCI  |TravisCI|Azure|
|macOS  |                    CircleCI|               |TravisCI |CircleCI |Azure|

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -38,7 +38,7 @@ def main():
                               'script is going to automatically install MacPython on your system, '
                               'so don\'t run on your development machine. For "windows", you need to '
                               'run in Windows, and it will build and test for all versions of '
-                              'Python at C:\\PythonXX[-x64]. Default: auto.'))
+                              'Python. Default: auto.'))
     parser.add_argument('--output-dir',
                         default=os.environ.get('CIBW_OUTPUT_DIR', 'wheelhouse'),
                         help='Destination folder for the wheels.')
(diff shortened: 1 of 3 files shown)


mayeut:
> So maybe if platform is `auto` we'd still want a check for environment variables that indicate a CI environment?

I was wondering about that and went the simpler way, not the safest (i.e. manual mostly means we can run linux on macOS, and maybe windows). I can add a simple check for CI (at least the tested ones).

YannickJadoul:
Just thinking out loud, here, but didn't these also check whether the platform is supported by `cibuildwheel`? By which I mean, if the platform isn't supported, running the default invocation of `cibuildwheel` wouldn't work; so to some degree, it was a way of safeguarding users against using it on unsupported platforms?

mayeut:
> Just thinking out loud, here, but didn't these also check whether the platform is supported by cibuildwheel? By which I mean, if the platform isn't supported, running the default invocation of cibuildwheel wouldn't work; so to some degree, it was a way of safeguarding users against using it on unsupported platforms?

I think this was true before #180 was merged in. There's no more specific behavior in cibuildwheel except for skipping py27 on windows travis-ci.
Preventing users to do something with side-effects on their dev machine is one thing, this is now mostly true for macOS where installation of python is tampered with.
The documentation is clear enough on what CI providers are tested so I'm more in favor of letting people try cibuildwheel with other CI providers easily if they want to rather than being conservative. I don't think it will add much more support and might in fact lead to positive reports on other CIs.

YannickJadoul:
> I don't think it will add much more support and might in fact lead to positive reports on other CIs.

True, but that's what I'm actually afraid of: builds that seem to work on untested CI platforms but maybe have some specific problems that would come to light in our tests?

Then again, I think I do agree with your simplifications :) Just wanted to bring this up, in case we'd otherwise overlook it. (And maybe we should add a line to the docs about non-supported platforms and running the `cibuildwheel` tests? I'm not entirely sure; just a suggestion.)

cibuildwheel/__main__.py
```diff
@@ -59,35 +65,18 @@ def main():
...
-            elif os.environ['AGENT_OS'] == 'Windows_NT':
+            elif sys.platform == 'win32':
                 platform = 'windows'
-
         if platform is None:
             print('cibuildwheel: Unable to detect platform. cibuildwheel should run on your CI server, '
-                  'Travis CI, AppVeyor, and CircleCI are supported. You can run on your development '
-                  'machine using the --platform argument. Check --help output for more '
+                  'Travis CI, AppVeyor, Azure Pipelines and CircleCI are supported. You can run on your '
+                  'development machine using the --platform argument. Check --help output for more '
```
joerick: Maybe let's change this to "...development machine or other CI providers using..." to

joerick:
> True, but that's what I'm actually afraid of: builds that seem to work on untested CI platforms but maybe have some specific problems that would come to light in our tests?

I guess this is why it's good to get official support baked in to cibuildwheel. But these days, now that we use nuget for windows python (and manylinux and macpython from python.org), there's much less potential for differences between CI providers :)

I just left a little comment, but otherwise, I'm okay with this. Merge when you're happy @mayeut .

YannickJadoul:
@mayeut Should AppVeyor on Linux still be added to the docs, then?

mayeut:
I know I missed something... Obviously

YannickJadoul:
Well, I hadn't realized this PR achieved this, until I saw #146 being closed :-) Quite amazing that this now just works without anything else!

mayeut:
> Should AppVeyor on Linux still be added to the docs, then?

Merged #207

PR #209, opened 2019-11-12
Avoid script location warnings on Windows

YannickJadoul:
This should get rid of the "WARNING: The scripts [...] are installed in '...' which is not on PATH." warnings (e.g., https://travis-ci.org/joerick/cibuildwheel/jobs/610081892#L636).

Moreover, since macOS and Windows had different ordering of the steps setting up the build environment (e.g., `CIBW_ENVIRONMENT` before or after the installation of `pip`), I've reordered the steps on Windows. Is this the correct order?

Alternatively, we could just silence the warnings with `--no-warn-script-location`, but fixing this (and keeping the warnings when they are relevant!) seems like a better solution.

Besides that, when do we want to use `simple_shell` vs. `shell`, @mayeut?

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -92,9 +92,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
         assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
-        call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        call(['pip', 'install', 'wheel'], env=env)
-        call(['pip', 'install', 'delocate'], env=env)
+        call(['pip', 'install', '--upgrade', 'setuptools', 'wheel', 'delocate'], env=env)
 
         # setup dirs
         if os.path.exists('/tmp/built_wheel'):
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -63,6 +63,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 call(['sudo', 'tar', '-C', '/Library/Frameworks/Python.framework/Versions/%s/' % config.version, '-xmf', '/tmp/python-patch.tar.gz'])
 
         installation_bin_path = '/Library/Frameworks/Python.framework/Versions/{}/bin'.format(config.version)
+        assert os.path.exists(os.path.join(installation_bin_path, 'python3' if config.version[0] == '3' else 'python'))
 
         # Python bin folders on Mac don't symlink python3 to python, so we do that
         # so `python` and `pip` always point to the active configuration.
@@ -89,6 +90,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
         # install pip & wheel
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
+        assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
         call(['pip', 'install', 'wheel'], env=env)
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -118,9 +118,9 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         env = environment.as_dictionary(prev_environment=env)
 
         # for the logs - check we're running the right version of python
-        shell(['where', 'python'], env=env)
-        shell(['python', '--version'], env=env)
-        shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
+        simple_shell(['where', 'python'], env=env)
+        simple_shell(['python', '--version'], env=env)
+        simple_shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
 
         # make sure pip is installed
         if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
@@ -128,10 +128,10 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
 
         # prepare the Python environment
-        shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'], env=env)
-        shell(['pip', '--version'], env=env)
-        shell(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        shell(['pip', 'install', 'wheel'], env=env)
+        simple_shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'], env=env)
+        simple_shell(['pip', '--version'], env=env)
+        simple_shell(['pip', 'install', '--upgrade', 'setuptools'], env=env)
+        simple_shell(['pip', 'install', 'wheel'], env=env)
 
         # setup dirs
         if os.path.exists(built_wheel_dir):
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -124,14 +124,13 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
         # make sure pip is installed
         if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
-            simple_shell([os.path.join(config_python_path, 'python.exe'), get_pip_script], env=env)
+            simple_shell(['python', get_pip_script], env=env)
         assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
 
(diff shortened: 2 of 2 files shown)


YannickJadoul:
Hmmmm, the Ubuntu build on AppVeyor timed out; @joerick, the email you sent somehow didn't work to give me rights on AppVeyor. Anyhow, I'll just amend the commit without changing anything and force push again; that should start the CIs again.

YannickJadoul:
Seem that job is timing out again. I've no idea whether that's because of the PR or because of AppVeyor (though the former would be weird, especially since the other Windows builds do succeed).

joerick:
I've restarted it again, let's see...

YannickJadoul:
> the email you sent somehow didn't work to give me rights on AppVeyor

(Nope, somehow still not; I think it's because my main GitHub email is a different one than in the commit messages. Wondering how I can get that to you without putting my email plainly on GitHub without the risk of being scraped and spammed. But no worries; I can perfectly survive without :) )

cibuildwheel/windows.py
```diff
@@ -100,40 +100,44 @@ def shell(args, env=None, cwd=None):
...
             os.path.join(config_python_path, 'Scripts'),
             env['PATH']
         ])
+        # update env with results from CIBW_ENVIRONMENT
         env = environment.as_dictionary(prev_environment=env)
 
         # for the logs - check we're running the right version of python
+        shell(['where', 'python'], env=env)
         shell(['python', '--version'], env=env)
         shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
```
mayeut: These shall probably use the `simple_shell`

cibuildwheel/windows.py
```diff
@@ -100,40 +100,44 @@ def shell(args, env=None, cwd=None):
...
+        # make sure pip is installed
+        if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
+            simple_shell([os.path.join(config_python_path, 'python.exe'), get_pip_script], env=env)
+        assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
+
         # prepare the Python environment
         shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'], env=env)
+        shell(['pip', '--version'], env=env)
         shell(['pip', 'install', '--upgrade', 'setuptools'], env=env)
         shell(['pip', 'install', 'wheel'], env=env)
```
mayeut: These shall probably use the `simple_shell` as well.

cibuildwheel/windows.py
YannickJadoul: Thanks; that's what I thought, but they weren't before, so I wasn't sure :)

mayeut:
> Besides that, when do we want to use `simple_shell` vs. `shell`

`shell` has a different effect on AppVeyor only where it sets-up the variables to allow C extension to build. This means `shell` shall be used where a C extension might be built. I think this goes down to: `before_build`, `build`, `install` (both), and maybe the `test_command`. All other lines can use the `simple_shell`.

I don't know if the specific script is still needed.

cibuildwheel/macos.py
```diff
@@ -89,6 +90,7 @@ def call(args, env=None, cwd=None, shell=False):
 
         # install pip & wheel
         call(['python', get_pip_script, '--no-setuptools', '--no-wheel'], env=env)
+        assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools'], env=env)
         call(['pip', 'install', 'wheel'], env=env)
```
mayeut: It feels weird no to upgrade `wheel` (and thus not having a single line: `call(['pip', 'install', '--upgrade', 'setuptools', 'wheel'], env=env))`

cibuildwheel/windows.py
```diff
@@ -100,39 +100,43 @@ def shell(args, env=None, cwd=None):
...
+        assert os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe'))
 
         # prepare the Python environment
-        shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'], env=env)
-        shell(['pip', 'install', '--upgrade', 'setuptools'], env=env)
-        shell(['pip', 'install', 'wheel'], env=env)
+        simple_shell(['python', '-m', 'pip', 'install', '--upgrade', 'pip'], env=env)
+        simple_shell(['pip', '--version'], env=env)
+        simple_shell(['pip', 'install', '--upgrade', 'setuptools'], env=env)
+        simple_shell(['pip', 'install', 'wheel'], env=env)
```
mayeut: Same here, It feels weird no to upgrade `wheel` (and thus not having a single line: `call(['pip', 'install', '--upgrade', 'setuptools', 'wheel'], env=env))`

Any reason you know of ?

mayeut:
> Is this the correct order?

Seems right.

cibuildwheel/windows.py
YannickJadoul: Agreed; but I believe it's because wheel is not installed by default, so `--upgrade` wouldn't do anything.

mayeut: This seems to work perfectly on macOS at least:
```
(.venv) Matt$ pip uninstall wheel setuptools
Uninstalling wheel-0.33.6:
  Would remove:
    /Users/Matt/Dev/pypa/setuptools/.venv/bin/wheel
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/wheel-0.33.6.dist-info/*
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/wheel/*
Proceed (y/n)? y
  Successfully uninstalled wheel-0.33.6
Uninstalling setuptools-41.6.0:
  Would remove:
    /Users/Matt/Dev/pypa/setuptools/.venv/bin/easy_install
    /Users/Matt/Dev/pypa/setuptools/.venv/bin/easy_install-3.7
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/easy_install.py
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/pkg_resources/*
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/setuptools-41.6.0.dist-info/*
    /Users/Matt/Dev/pypa/setuptools/.venv/lib/python3.7/site-packages/setuptools/*
Proceed (y/n)? y
  Successfully uninstalled setuptools-41.6.0
(.venv) Matt$ pip install --upgrade wheel setuptools
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/00/83/b4a77d044e78ad1a45610eb88f745be2fd2c6d658f9798a15e384b7d57c9/wheel-0.33.6-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/d9/de/554b6310ac87c5b921bc45634b07b11394fe63bc4cb5176f5240addf18ab/setuptools-41.6.0-py2.py3-none-any.whl
Installing collected packages: wheel, setuptools
Successfully installed setuptools-41.6.0 wheel-0.33.6
```

YannickJadoul: Yeah, it should work; I have two explanations for this:
1) One was added afterwards.
2) @joerick somehow wanted to keep those two installations separate, for some reason?

Either way, I can change this now, and see if @joerick had a reason.

cibuildwheel/windows.py
```diff
@@ -100,39 +100,43 @@ def shell(args, env=None, cwd=None):
...
         # for the logs - check we're running the right version of python
-        shell(['python', '--version'], env=env)
-        shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
+        simple_shell(['where', 'python'], env=env)
+        simple_shell(['python', '--version'], env=env)
+        simple_shell(['python', '-c', '"import struct; print(struct.calcsize(\'P\') * 8)\"'], env=env)
+
+        # make sure pip is installed
+        if not os.path.exists(os.path.join(config_python_path, 'Scripts', 'pip.exe')):
+            simple_shell([os.path.join(config_python_path, 'python.exe'), get_pip_script], env=env)
```
YannickJadoul: @mayeut And here, any reason to keep the full path here, or should we shorten this to `'python'` now it's in `PATH`?

mayeut: This can definitely be shorten to `'python'`

YannickJadoul: Yeah, just wondering if there would be a good reason to keep it like this.

cibuildwheel/windows.py
joerick: Happy to have these both be --upgrade. Not sure why they're separate. Side tangent- I've always wanted to pin these in cibuildwheel, but it would create a maintenance burden to have to keep revving them on each release. Sure would be nice for repeatable builds though

mayeut:
Thanks @YannickJadoul, it will be much cleaner without all the warnings in the logs.

@joerick, do you want to review this some more or can it be merged ?

mayeut:
@YannickJadoul, if you don't have anything to add, can you merge this one ?

YannickJadoul:
Done, thanks, @mayeut & @joerick! :-)

PR #210, opened 2019-11-12
Use ensurepip instead of get-pip.py

YannickJadoul:
Just curious to see what would happen. It saves us a few lines and a download, so let's see...


YannickJadoul:
Heh, that's sad. NuGet's Python doesn't seem to have `ensurepip` ("No module named ensurepip") and on macOS, it seems `pip` is installed but maybe there's no script in the `bin` directory (not sure; the error's not very clear).

Too bad :-(

joerick:
Pretty sure it's python 3.4+, so not available on 2.7. Maybe that was the issue?

YannickJadoul:
Hmmm, maybe. I though I'd checked that :-/

Yeah, here it says "New in version 2.7.9.": https://docs.python.org/2.7/library/ensurepip.html
Also, now that I think further about it: on macOS, the error happened for Python 3.5, so after 2.7 was done. Presumably because there's only a `pip3` script instead of `pip` or so?

PR #211, opened 2019-11-13
Add `CIBW_REPAIR_WHEEL_COMMAND` env variable

### LINKED ISSUE #191: Hook to replace auditwheel/delocate/[windows equivalent that doesn't exist yet]
I would like a hook that would allow me to run a command of my own in place of the `auditwheel` or `delocate` invocation (on Linux or MacOS respectively), or at the point in the process where an equivalent tool _would_ be invoked on Windows, if one existed.

My primary reason for wanting this is that I need the equivalent tool that doesn't yet exist for Windows.  My package is simple enough that I can probably kludge together something that works well enough for me, but I need a way to make cibuildwheel run it.

A plausible interface would be: environment variables named `CIBW_FIXUP_WHEEL_${PLATFORM}` that take the name of a program.  This program will be invoked with two arguments: the absolute pathname of a wheel to process, and the absolute pathname to write the processed filename to.  On Linux the default is a wrapper for `auditwheel`, on MacOS the default is a wrapper for `delocate`, and on Windows it's `copy`.

nsoranzo:
to allow different options to auditwheel/delocate, alternative commands and a future Windows-equivalent.

Fix #191 .

Also fix some PEP-008 issues reported by flake8.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -30,7 +30,7 @@ def get_python_configurations(build_selector):
     return [c for c in python_configurations if build_selector(c.identifier)]
 
 
-def build(project_dir, output_dir, test_command, test_requires, test_extras, before_build, build_verbosity, build_selector, environment, manylinux_images):
+def build(project_dir, output_dir, test_command, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment, manylinux_images):
     try:
         subprocess.check_call(['docker', '--version'])
     except:
@@ -60,30 +60,28 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             {environment_exports}
 
             for PYBIN in {pybin_paths}; do
-                # Setup
-                rm -rf /tmp/built_wheel
-                rm -rf /tmp/delocated_wheels
-                mkdir /tmp/built_wheel
-                mkdir /tmp/delocated_wheels
-
                 if [ ! -z {before_build} ]; then
                     PATH="$PYBIN:$PATH" sh -c {before_build}
                 fi
 
-                # Build that wheel
+                # Build the wheel
+                rm -rf /tmp/built_wheel
+                mkdir /tmp/built_wheel
                 PATH="$PYBIN:$PATH" "$PYBIN/pip" wheel . -w /tmp/built_wheel --no-deps {build_verbosity_flag}
                 built_wheel=(/tmp/built_wheel/*.whl)
 
-                # Delocate the wheel
+                # repair the wheel
+                rm -rf /tmp/repaired_wheels
+                mkdir /tmp/repaired_wheels
                 # NOTE: 'built_wheel' here is a bash array of glob matches; "$built_wheel" returns
                 # the first element
-                if [[ "$built_wheel" == *none-any.whl ]]; then
-                    # pure python wheel - just copy
-                    mv "$built_wheel" /tmp/delocated_wheels
+                if [[ "$built_wheel" == *none-any.whl ]] || [ -z {repair_command} ]; then
+                    # pure Python wheel or empty repair command
+                    mv "$built_wheel" /tmp/repaired_wheels
                 else
-                    auditwheel repair "$built_wheel" -w /tmp/delocated_wheels
+                    built_wheel=$built_wheel sh -c {repair_command}
                 fi
-                delocated_wheels=(/tmp/delocated_wheels/*.whl)
+                repaired_wheels=(/tmp/repaired_wheels/*.whl)
 
                 if [ ! -z {test_command} ]; then
                     # Set up a virtual environment to install and test from, to make sure
@@ -105,7 +103,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                         # functionally the same, differing only in name, wheel metadata, and possibly include
                         # different external shared libraries. so it doesn't matter which one we run the tests on.
                         # Let's just pick the first one.
-                        pip install "${{delocated_wheels[0]}}"{test_extras}
+                        pip install "${{repaired_wheels[0]}}"{test_extras}
 
                         # Install any requirements to run the tests
                         if [ ! -z "{test_requires}" ]; then
@@ -127,8 +125,8 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 fi
 
                 # we're all done here; move it to output
-                mv "${{delocated_wheels[@]}}" /output
-                for delocated_wheel in "${{delocated_wheels[@]}}"; do chown {uid}:{gid} "/output/$(basename "$delocated_wheel")"; done
+                mv "${{repaired_wheels[@]}}" /output
+                for repaired_wheel in "${{repaired_wheels[@]}}"; do chown {uid}:{gid} "/output/$(basename "$repaired_wheel")"; done
             done
         '''.format(
             pybin_paths=' '.join(c.path+'/bin' for c in platform_configs),
@@ -141,6 +139,9 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 prepare_command(before_build, project='/project') if before_build else ''
             ),
             build_verbosity_flag=' '.join(get_build_verbosity_extra_flags(build_verbosity)),
+            repair_command=shlex_quote(
+                prepare_command(repair_command, wheel='"$built_wheel"', dest_dir='/tmp/repaired_wheels') if repair_command else ''
+            ),
             environment_exports='\n'.join(environment.as_shell_commands()),
             uid=os.getuid(),
             gid=os.getgid(),
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -79,7 +79,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                     # pure Python wheel or empty repair command
                     mv "$built_wheel" /tmp/repaired_wheels
                 else
-                    built_wheel=$built_wheel sh -c {repair_command}
+                    sh -c {repair_command} repair_command "$built_wheel"
                 fi
                 repaired_wheels=(/tmp/repaired_wheels/*.whl)
 
@@ -140,7 +140,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             ),
             build_verbosity_flag=' '.join(get_build_verbosity_extra_flags(build_verbosity)),
             repair_command=shlex_quote(
-                prepare_command(repair_command, wheel='"$built_wheel"', dest_dir='/tmp/repaired_wheels') if repair_command else ''
+                prepare_command(repair_command, wheel='"$1"', dest_dir='/tmp/repaired_wheels') if repair_command else ''
             ),
             environment_exports='\n'.join(environment.as_shell_commands()),
             uid=os.getuid(),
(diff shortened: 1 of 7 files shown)


joerick:
Hi @nsoranzo, thanks for submitting this! I preferred the naming `CIBW_REPAIR_WHEEL_${PLATFORM}`, as @mayeut suggested - would you mind changing the name to that?

Otherwise, I get that people have different tastes regarding style and pep8 might be a good place to be, but if we wanted to adopt that properly (and put tooling in place) that should be a separate PR. It's just tricky as a reviewer to review both style and functionality changes at the same time, and can cause lots of conflicts with other active PRs.

Would you mind removing those style changes from this PR? Feel free to send them in as a separate PR if you like.

nsoranzo:
@joerick Thanks for the fast feedback! I've moved the PEP-008 changes to a new PR #212.

Regarding the environment variable name, what about `CIBW_REPAIR_COMMAND` ? To me `CIBW_REPAIR_WHEEL` suggest a yes/no value. Also `_COMMAND` is already used for `CIBW_TEST_COMMAND`.

cibuildwheel/util.py
```diff
@@ -2,14 +2,14 @@
 import warnings
 
 
-def prepare_command(command, project):
+def prepare_command(command, replace_dict):
```
YannickJadoul: What about `**replace`? Then we can keep the old syntax.

Is there no way to check these format variables that are replaced, though? It would be nice to catch typos rather sooner than later.

cibuildwheel/windows.py
```diff
@@ -136,12 +131,27 @@ def shell(args, env=None, cwd=None):
...
         # run the before_build command
         if before_build:
-            before_build_prepared = prepare_command(before_build, project=abs_project_dir)
+            before_build_prepared = prepare_command(before_build, {'project': abs_project_dir})
             shell([before_build_prepared], env=env)
 
         # build the wheel
+        if os.path.exists(built_wheel_dir):
+            shutil.rmtree(built_wheel_dir)
+        os.makedirs(built_wheel_dir)
```
YannickJadoul: Why was this moved?

cibuildwheel/macos.py
```diff
@@ -94,32 +97,29 @@ def call(args, env=None, cwd=None, shell=False):
         call(['pip', 'install', 'wheel'], env=env)
         call(['pip', 'install', 'delocate'], env=env)
```
YannickJadoul: What do we do with this, then? It's (in principle) not always needed.

cibuildwheel/windows.py
```diff
@@ -136,12 +131,27 @@ def shell(args, env=None, cwd=None):
...
+        if os.path.exists(delocated_wheel_dir):
+            shutil.rmtree(delocated_wheel_dir)
+        os.makedirs(delocated_wheel_dir)
+        if built_wheel.endswith('none-any.whl') or not delocate_command:
+            # pure Python wheel or empty delocate command
+            shutil.move(built_wheel, delocated_wheel_dir)
+        else:
+            delocate_command_prepared = prepare_command(delocate_command, {'wheel': built_wheel, 'dest_dir': delocated_wheel_dir})
+            shell([delocate_command_prepared], env=env)
+        delocated_wheel = glob(os.path.join(delocated_wheel_dir, '*.whl'))[0]
```
YannickJadoul: Agreeing with @joerick here; the name "delocate" makes absolutely no sense in a Windows (or Linux) context.

docs/options.md
```diff
@@ -177,7 +177,9 @@ A shell command to run before building the wheel. This option allows you to run
 
 If dependencies are required to build your wheel (for example if you include a header from a Python module), set this to `pip install .`, and the dependencies will be installed automatically by pip. However, this means your package will be built twice - if your package takes a long time to build, you might wish to manually list the dependencies here instead.
 
-The active Python binary can be accessed using `python`, and pip with `pip`; `cibuildwheel` makes sure the right version of Python and pip will be executed. `{project}` can be used as a placeholder for the absolute path to the project's root.
+The active Python binary can be accessed using `python`, and pip with `pip`; `cibuildwheel` makes sure the right version of Python and pip will be executed. `{project}` can be used as a placeholder for the absolute path to the project's root and will be replaced by `cibuildwheel`.
```
YannickJadoul: What has changed here?

docs/options.md
```diff
@@ -177,7 +177,9 @@ A shell command to run before building the wheel. This option allows you to run
...
+
+On Linux and macOS, the command is run in a shell, so you can write things like `cmd1 && cmd2`.
```
YannickJadoul: And on Windows?

YannickJadoul:
> Regarding the environment variable name, what about `CIBW_REPAIR_COMMAND` ? To me `CIBW_REPAIR_WHEEL` suggest a yes/no value.

Agreed. But then to be nitpicky, make it `CIBW_REPAIR_WHEEL_COMMAND` (or `CIBW_WHEEL_REPAIR_COMMAND`?). "repair" could be many things, out of context, I'd say. And the use of this option should be rare enough to not care about a slightly longer name.

cibuildwheel/macos.py
nsoranzo: I'd keep it for simplicity, 99% of users will use this.

docs/options.md
nsoranzo: I've added ` and will be replaced by `cibuildwheel`` to the end of the line, like it was already specified for `CIBW_TEST_COMMAND`.

YannickJadoul: Aha, that's why :-)

docs/options.md
nsoranzo: To be honest I have no idea, I've just copy-pasted this from the `CIBW_TEST_COMMAND` documentation.

YannickJadoul: Hmmm, OK. I guess I was/am also just confused by changing this in the docs for CIBW_BEFORE_BUILD, while this PR is not touching that part of the code.

YannickJadoul:
> > Regarding the environment variable name, what about `CIBW_REPAIR_COMMAND` ? To me `CIBW_REPAIR_WHEEL` suggest a yes/no value.
> 
> Agreed. But then to be nitpicky, make it `CIBW_REPAIR_WHEEL_COMMAND` (or `CIBW_WHEEL_REPAIR_COMMAND`?). "repair" could be many things, out of context, I'd say. And the use of this option should be rare enough to not care about a slightly longer name.

Just noticed: it's also `CIBW_BEFORE_BUILD` and not `CIBW_BEFORE_BUILD_COMMAND`, though.

cibuildwheel/windows.py
nsoranzo: I prefer to have it close to the code that actually use the directory.

cibuildwheel/windows.py
nsoranzo: Replaced with `repair`/`repaired` everywhere.

cibuildwheel/util.py
nsoranzo: > What about `**replace`? Then we can keep the old syntax.

Replaced with the standard `**kwargs`.

cibuildwheel/windows.py
YannickJadoul: Still not 100% convinced that we don't need `wheel` in those names, but I'll leave that decision up to @joerick.

docs/options.md
```diff
@@ -237,7 +260,7 @@ CIBW_TEST_COMMAND: nosetests {project}/tests
 CIBW_TEST_COMMAND: nosetests {project}/tests
 ```
 
-Platform-specific variants also available:
```
YannickJadoul: Sorry, missed this before. Could you also revert these spaces at the end of the line? Maybe we should do the same in a different PR as @mayeut already did elsewhere, replacing them with \ instead of , but it's nice to keep focused changesets.

nsoranzo: Given I'm copy-pasting this for the new `CIBW_REPAIR_COMMAND` I opted to fix it for all 7 instances in this PR (using `<br/>`, looks like `\` does not work for documents built with `mkdocs`).

nsoranzo:
The Travis job failure is unrelated.

nsoranzo:
Rebased to fix a conflict.

docs/options.md
```diff
@@ -191,12 +193,33 @@ CIBW_BEFORE_BUILD: pip install pybind11
...
+
+A shell command to repair a built wheel by copying external library dependencies into the wheel tree and relinking them.
+The command is run on each built wheel (except for pure Python ones) before testing it.
+
+The following placeholders must be used inside the command and will be replaced by `cibuildwheel`:
+- `{wheel}` for the absolute path to the built wheel
+- `{dest_dir}` for the absolute path of the directory where to create the repaired wheel.
+
+On Linux and macOS, the command is run in a shell, so you can write things like `cmd1 && cmd2`.
+
```
~~~suggestion
#### Examples
```yaml
# don't repair mac wheels
CIBW_REPAIR_COMMAND_MACOS: "mv {wheel} {dest_dir}"

# pass the --lib-sdir flag to auditwheel on linux
CIBW_REPAIR_COMMAND_LINUX: "auditwheel repair --lib-sdir . -w {dest_dir} {wheel}"
```

~~~

nsoranzo: Should all the examples be moved after the "Platform-specific variants also available" bit?
It feels awkward to use the variants in examples before introducing them.

cibuildwheel/linux.py
```diff
@@ -60,30 +60,28 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
...
                 # the first element
-                if [[ "$built_wheel" == *none-any.whl ]]; then
-                    # pure python wheel - just copy
-                    mv "$built_wheel" /tmp/delocated_wheels
+                if [[ "$built_wheel" == *none-any.whl ]] || [ -z {repair_command} ]; then
+                    # pure Python wheel or empty repair command
+                    mv "$built_wheel" /tmp/repaired_wheels
                 else
-                    auditwheel repair "$built_wheel" -w /tmp/delocated_wheels
+                    built_wheel=$built_wheel sh -c {repair_command}
```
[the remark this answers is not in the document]
nsoranzo: Good idea! I had to use `sh -c {repair_command} repair_command "$built_wheel"` because the format is `sh -c command_string [command_name [argument ...]]`

docs/options.md
joerick: Yes, I think you're right. Better to move all those "Platform-specific variants also available" bits before Examples. 

Would you mind making the change? Then we can merge.

nsoranzo:
The AppVeyor failure is unrelated.

YannickJadoul:
Quick thing, still: I'd just made sure that all three platforms follow the same structure/do +- the same steps in the same order (cfr. #209). It's kind of nice to keep that, I think. So just for certainty, since I see some things moving around but I lack the overview: could you have a quick look whether that consistency between Windows & macOS (& Linux, to a certain degree) is still there?

nsoranzo:
@YannickJadoul Actually this PR slightly improves the consistency across architectures, I hope.
See e.g. https://github.com/joerick/cibuildwheel/pull/211/files#diff-b356ce3147eb040e0ec45382568d9b40R28-R33 , which replicates what was already done in windows.py .

YannickJadoul:
@nsoranzo OK, great, then! :-)

joerick:
Looking good! Thanks @nsoranzo !

PR #213, opened 2019-11-16
Change test to use tmp_path fixture and by pytest callable

Czaki:
This pull request contains four parts 

- change to use tmp_path fixture for separated build folder for each test (this allow for fix 02_test::test_failing_test assertion). Change `bin/run_tests.py` to not stop on first test fail. 
- allow to run each test separate by calling pytest ex. `pytest tests/02_test` 
- allow to run all tests by one pytest call ex. `pytest unit_test test`
  Pytest on test collection needs that test containing files has different names or be in different python modules. I change names. Other option, for save `cibuildwheel_test.py` name is to add `__init__.py` file in each test sub directory. 
- add test of pytest calling to azure test case and add installation of `pytest-azurepipelines` which integrate information about test passing in test tab. 

@YannickJadoul @mayeut


azure-pipelines.yml
```diff
@@ -7,8 +7,20 @@ jobs:
         versionSpec: '3.8'
     - bash: |
         python -m pip install -r requirements-dev.txt
+        python -m pip install pytest-azurepipelines
         python ./bin/run_tests.py
```
mayeut: Why not change to `pytest` invocation ?

azure-pipelines.yml
```diff
@@ -7,8 +7,20 @@ jobs:
...
+  pool: {vmImage: 'Ubuntu-18.04'}
+  steps:
+    - task: UsePythonVersion@0
+      inputs:
+        versionSpec: '3.8'
+    - bash: |
+        python -m pip install -r requirements-dev.txt
+        python -m pip install pytest-azurepipelines
+        pytest unit_test test
+
```
mayeut: Why did you add this whole section ? Just to show `pytest` direct command invocation ?

test/01_basic/cibuildwheel_basic_test.py
```diff
@@ -1,16 +1,20 @@
 import os
-import utils
+import sys 
+
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))
+
+from shared import utils
```
mayeut: I guess this changed is required by:
1. `pytest` direct invocation
2. the removal of `PYTHONPATH` fix in `bin/run_test.py` (which is probably removed because of the change for `pytest` didn't require `bin/run_test.py` to still fix the `PYTHONPATH `)

test/01_basic/cibuildwheel_basic_test.py
```diff
@@ -1,16 +1,20 @@
...
-def test():
+def test(tmp_path):
     # build the wheels
-    utils.cibuildwheel_run(project_dir)
+    utils.cibuildwheel_run(project_dir, output_dir=tmp_path)
 
     # check that the expected wheels are produced
     expected_wheels = utils.expected_wheels('spam', '0.1.0')
-    actual_wheels = os.listdir('wheelhouse')
+    actual_wheels = [x.name for x in tmp_path.iterdir()]
```
mayeut: why not keep `os.listdir ? I find this less readable at first glance.

test/02_test/cibuildwheel_test_test.py
```diff
@@ -1,55 +1,61 @@
...
             'CIBW_TEST_COMMAND': 'false',
             # manylinux1 has a version of bash that's been shown to have 
             # problems with this, so let's check that.
             'CIBW_MANYLINUX_I686_IMAGE': 'manylinux1',
             'CIBW_MANYLINUX_X86_64_IMAGE': 'manylinux1',
-        })
+        }, output_dir=tmp_path)
 
-    assert len(os.listdir('wheelhouse'))
+    assert len([x.name for x in tmp_path.iterdir()]) == 0
```
mayeut: Good catch for the bug in the test. I also came across this.
Still, I find `os.listdir` more readable

azure-pipelines.yml
Czaki: Current pipeline work is to call test by `python ./bin/run_tests.py`. When my changes are applied the whole content of bin directory can be removed, but from the other side its need change of workflow or all contributor.  So I implement this in conservative way, so show how to do this, but allow @joerick decide which approach is best for him.

test/01_basic/cibuildwheel_basic_test.py
Czaki: yes

test/01_basic/cibuildwheel_basic_test.py
Czaki: because of python 2.7. 
For me this is more elegant than convert tmp_path to string), 

```
>>> import os 
>>> from pathlib2 import Path
>>> p = Path(".")
>>> list(p.iterdir())
[PosixPath('setup.cfg'), PosixPath('bin'), PosixPath('.travis.yml'), PosixPath('cibuildwheel'), PosixPath('azure-pipelines.yml'), PosixPath('.git'), PosixPath('.vscode'), PosixPath('pytest.ini'), PosixPath('setup.py'), PosixPath('docs'), PosixPath('.circleci'), PosixPath('.readthedocs.yml'), PosixPath('.python-version'), PosixPath('requirements-dev.txt'), PosixPath('mkdocs.yml'), PosixPath('LICENSE'), PosixPath('unit_test'), PosixPath('README.md'), PosixPath('test'), PosixPath('CODE_OF_CONDUCT.md'), PosixPath('.gitignore'), PosixPath('.pytest_cache'), PosixPath('appveyor.yml'), PosixPath('cibuildwheel.egg-info')]
>>> os.listdir(p)
  File "<stdin>", line 1, in <module>
TypeError: coercing to Unicode: need string or buffer, PosixPath found
>>> os.listdir(str(p))
['setup.cfg', 'bin', '.travis.yml', 'cibuildwheel', 'azure-pipelines.yml', '.git', '.vscode', 'pytest.ini', 'setup.py', 'docs', '.circleci', '.readthedocs.yml', '.python-version', 'requirements-dev.txt', 'mkdocs.yml', 'LICENSE', 'unit_test', 'README.md', 'test', 'CODE_OF_CONDUCT.md', '.gitignore', '.pytest_cache', 'appveyor.yml', 'cibuildwheel.egg-info']
```

mayeut: It is indeed more elegant to get rid of the string conversion, but I was talking about readability.
I still think `os.listdir(str(tmp_path))` is more readable than `[x.name for x in tmp_path.iterdir()]`.

azure-pipelines.yml
YannickJadoul: If there's just the old and the new version in the diff, @joerick would also have noticed when reviewing the changes.

bin/run_tests.py
```diff
@@ -23,7 +23,12 @@
...
 
     run_test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'run_test.py')
+    test_failed = 0
     for project_path in test_projects:
-        subprocess.check_call([sys.executable, run_test_path, project_path])
+        if 0 != subprocess.call([sys.executable, run_test_path, project_path]):
+            test_failed += 1
 
+    if test_failed > 0:
+        print('%d of %d projects built failed.' % (test_failed, len(test_projects)))
```
YannickJadoul: 1) What are these changes doing?
2) Why old %-formatting rather than format strings?

pytest.ini
```diff
@@ -0,0 +1,3 @@
+# content of pytest.ini
```
YannickJadoul: norecursedirs = subtest

test/01_basic/cibuildwheel_basic_test.py
YannickJadoul: This `sys.path.append` should then be in `conftest.py`, I believe?

bin/run_tests.py
Czaki: 1. write how many test group failed. 
2. Because bellow there is line `print('%d projects built successfully.' % len(test_projects))` and I would like to use same style.

test/02_test/cibuildwheel_test_test.py
```diff
@@ -1,55 +1,61 @@
...
+import subprocess
+import os
+import sys 
+
 import pytest
-import utils
 
-def test():
-    project_dir = os.path.dirname(__file__)
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))
```
YannickJadoul: We cannot have to repeat this for every test. This is not simplifying things. If this is the only way, I prefer to keep using `run_tests.py`

YannickJadoul:
Thanks for making a PR on this, @Czaki! I've seen @mayeut also made one, but I haven't looked at that yet, so I can't compare yet. I will comment on that in a bit.

>     * change to use tmp_path fixture for separated build folder for each test (this allow for fix 02_test::test_failing_test assertion).

Nice, good to fix this for all tests :-)

> Change `bin/run_tests.py` to not stop on first test fail.

Why? Was this a problem?

>     * allow to run each test separate by calling pytest ex. `pytest tests/02_test`

That didn't work before? Couldn't you call `python bin/run_test.py tests/02_test`? What's the advantage of your changes?

>     * allow to run all tests by one pytest call ex. `pytest unit_test test`

Well, again, you could do that before with `python bin/run_tests.py`. And unit tests are arguably different, so I don't think we need to be able to call them all in 1 go (for example, the unit tests need to be run locally when developing, but the integration tests not, because they install packages, etc).

>       Pytest on test collection needs that test containing files has different names or be in different python modules. I change names. Other option, for save `cibuildwheel_test.py` name is to add `__init__.py` file in each test sub directory.

I don't understand this.

>     * add test of pytest calling to azure test case and add installation of `pytest-azurepipelines` which integrate information about test passing in test tab.

bin/run_tests.py
YannickJadoul: 1. I meant in general. But I now read your comment that you don't want to stop on the first error.

2. Oh, I didn't see that. But then I vote for changing them all to fmt-style, if there aren't too many :-)

Czaki:
>Why? Was this a problem? 

Tests of `cibuildwheel ` takes long time. It is better to know each tests which fail. 
Ex. your change broke test 4 and 6, but in different approach. With old code you got only information about problem with test 4. Fix it and then got information that test 6 also fail.  

> That didn't work before? Couldn't you call python bin/run_test.py tests/02_test? What's the advantage of your changes?

This allow call single test, not group of test. `pytest test/02_test/cibuildwheel_test_test.py::test_failing_test` 
Or allow to run on this groups of test that are interesting 
`pytest test/02_test test/04_build_skip` 

Also some of python IDE (like PyCharm) have support for using pytest module with better reporting of problems.

> Well, again, you could do that before with python bin/run_tests.py. And unit tests are arguably different, so I don't think we need to be able to call them all in 1 go (for example, the unit tests need to be run locally when developing, but the integration tests not, because they install packages, etc).

pytest has here two arguments. You can ex. call only `pytest unit_test` or `pytest test` to have executed only part of test 

>I don't understand this.

> >    * add test of pytest calling to azure test case and add installation of `pytest-azurepipelines` which integrate information about test passing in test tab.

See here. Integration of test result reporting. 
https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=439&view=ms.vss-test-web.build-test-results-tab

YannickJadoul:
> > Why? Was this a problem?
> 
> Tests of `cibuildwheel ` takes long time. It is better to know each tests which fail.
> Ex. your change broke test 4 and 6, but in different approach. With old code you got only information about problem with test 4. Fix it and then got information that test 6 also fail.

Right. But then most often, when I do break a test, I break all tests, so it's pretty handy that they're all stopped immediately. So I'm not sure how I feel about this. Maybe @joerick had a reasoning when he created this original script?

> > That didn't work before? Couldn't you call python bin/run_test.py tests/02_test? What's the advantage of your changes?
> 
> This allow call single test, not group of test. `pytest test/02_test/cibuildwheel_test_test.py::test_failing_test`
> Or allow to run on this groups of test that are interesting
> `pytest test/02_test test/04_build_skip`
>
> Also some of python IDE (like PyCharm) have support for using pytest module with better reporting of problems.

Pfff, those are some corner cases, I believe. Because again, we never run these tests locally, right; or do we? @mayeut, @joerick?
So IDEs should be able to run the unit tests easily, but I don't know about the integration tests.
 
> > Well, again, you could do that before with python bin/run_tests.py. And unit tests are arguably different, so I don't think we need to be able to call them all in 1 go (for example, the unit tests need to be run locally when developing, but the integration tests not, because they install packages, etc).
> 
> pytest has here two arguments. You can ex. call only `pytest unit_test` or `pytest test` to have executed only part of test

I know. But my point is that having these two types of tests separated was good. Because you don't want to accidentally run the integration tests on your local development machine.
 
> > I don't understand this.
> 
> > > * add test of pytest calling to azure test case and add installation of `pytest-azurepipelines` which integrate information about test passing in test tab.
> 
> See here. Integration of test result reporting.
> https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=439&view=ms.vss-test-web.build-test-results-tab

Thanks for that; that looks cool indeed! But I meant I didn't understand the thing above "Pytest on test collection needs that test containing files has different names or be in different python modules. I change names. Other option, for save `cibuildwheel_test.py` name is to add `__init__.py` file in each test sub directory."

Czaki:
> Pfff, those are some corner cases, I believe. Because again, we never run these tests locally, right; or do we? @mayeut, @joerick?

I work on linux machine and run test locally.

> Thanks for that; that looks cool indeed! But I meant I didn't understand the thing above "Pytest on test collection needs that test containing files has different names or be in different python modules. I change names. Other option, for save cibuildwheel_test.py name is to add __init__.py file in each test sub directory."

I responded on this question in #214

YannickJadoul:
> I work on linux machine and run test locally.

Ah, yes, right, because it's with docker. But if someone develops on macOS or Windows, they should not accidentally run this, right?

Also, how do you fool `cibuildwheel` into thinking it's on a CI?

> I responded on this question in #214

I saw. Thanks a lot! It makes more sense now :-)

Czaki:
> Ah, yes, right, because it's with docker. But if someone develops on macOS or Windows, they should not accidentally run this, right?
If you do not set `CIBW_PLATFORM` then it not run.

> Also, how do you fool cibuildwheel into thinking it's on a CI?
I set `CIBW_PLATFORM`

YannickJadoul:
Yesss, OK, good point.

Czaki:
I remove two biggest weakness of this PR. 
There is no modifying path in each test, and wrap listing wheels in output folder. 

circleci test fail because of network problem. It need to be manually restarted.

Issue #143, opened 2019-05-31
PyPy support


YannickJadoul:
Are we interested in building PyPy wheels with `cibuildwheel`, @joerick and @mayeut? I would be. Apparently, PyPy now also offers [manylinux images with PyPy executables](https://github.com/pypy/manylinux).

I'm happy to look into this, if agreed this would be a useful addition to `cibuildwheel`. It shouldn't even be a huge effort, I'd guess. But then should also maybe have a think about whether to make this an opt-in feature, and how?

YannickJadoul:
Hmmm, only now noticed this is for the `manylinux2010` standard, so this semi-depends on #135

joerick:
Oooh, that looks great! Yes that feels like a logical step for `cibuildwheel`. I actually didn't realise Cpython extensions would work with Pypy, but it looks like [they support that](http://doc.pypy.org/en/latest/faq.html#do-cpython-extension-modules-work-with-pypy). I'd be happy to just add it to CIBW_BUILD by default, assuming the builds are painless enough :D

altendky:
cpython c-api isn't the only thing to compile.  there's cffi too.

YannickJadoul:
Definitely! And even more advisable for PyPy, as far as I know.
But of course, since the C API is now +- complete in PyPy, the first step could already just easily allow developers to target PyPy with their existing code?

Either way, I think `cibuildwheel` is quite agnostic about the underlying mechanism, or am I forgetting something?

altendky:
Oh yes, cffi vs c-api shouldn't really matter to cibuildwheel I don't think.  I was just mentioning the other sort of project that needs binary wheels. Anyways, I would tend towards adding it to the defaults.  I think encouraging more PyPy awareness and support is a good thing.

altendky:
fyi, https://github.com/antocuni/pypy-wheels/tree/master/docker/image

YannickJadoul:
@altendky Thanks!
Was I just hallucinating when I thought I'd seen a ready and published `manylinux2010` image for PyPy, then, a few months ago?

PR #215, opened 2019-11-16
Add manylinux2014 x86_64/i686 support

mayeut:
Add manylinux2014 x86_64/i686 support.

Partial fix for #216


test/08_manylinuxXXXX_only/spam.c
```diff
@@ -1,7 +1,5 @@
 #include <Python.h>
-#if defined(__linux__)
```
YannickJadoul: Nice! Good point; we should only be building this on Linux :-)

test/08_manylinuxXXXX_only/cibuildwheel_test.py
```diff
@@ -12,9 +14,11 @@ def test():
     # rather than when dynamically loading the Python
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env={
         'CIBW_ENVIRONMENT': 'CFLAGS="$CFLAGS -Werror=implicit-function-declaration"',
+        'CIBW_SKIP': 'cp27*',  # not available on manylinux2014
```
YannickJadoul: Agreed. I'm fine dropping 2.7 for this specific test.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
```diff
@@ -12,9 +14,11 @@ def test():
...
+        'CIBW_MANYLINUX_X86_64_IMAGE': manylinux_image,
+        'CIBW_MANYLINUX_I686_IMAGE': manylinux_image,
     })
 
-    # also check that we got the right wheels
-    expected_wheels = [w for w in utils.expected_wheels('spam', '0.1.0')
-                       if not '-manylinux' in w or '-manylinux2010' in w]
+    expected_wheels = [w.replace('manylinux2010', manylinux_image) for w in utils.expected_wheels('spam', '0.1.0')
+                       if '-cp27' not in w and '-manylinux2010' in w]
```
YannickJadoul: Meh. Is there nothing better we can do? I already didn't like the previous lines.

`if '-cp27' not in w and '-{}_'.format(manylinux_image) in w` ?

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: Wait, actually... I agree we don't need to test 2.7 in this particular test, but ... do we force the user to put `CIBW_SKIP` to `cp27*` when using `manylinux2014`? Because running with only `CIBW_MANYLINUX_..._IMAGE=manylinux2014` will result in an error, now.

mayeut: I really don't want to clutter `linux.py` with specific code. I was thinking about wether documentation only or adding a check for python existence (not specific to manylinux2014). Since it's opt-in, I did not mind not changing `linux.py`. Documentation shall be updated though.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
mayeut: > if '-cp27' not in w and '-{}_'.format(manylinux_image) in w

Won't work with `manylinux2014`

If you find something easier, feel free to commit.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: > Documentation shall be updated though.

OK, good point. I guess we'll soon enough be dropping 2.7 anyway? We should maybe start thinking about how to phase out support?

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: Why so?

Actually, now realizing: you're not updating the list of `utils.expected_wheels`, because it's not our default yet.
Maybe we should add an extra argument to `utils.expected_wheels` to select the right `manylinux` version?

mayeut: When I saw `utils.expected_wheels` while refactoring tests, I indeed thought that it could be improved. Shall be in another PR though don't you think ?

YannickJadoul: Hmmm, for refactoring, yes. But the main thing I'm noticing for this PR is that we're breaking the current rule/habit that all the wheel names are to be found in `utils.expected_wheels`, and that we filter out the ones that are not expected in a test.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
mayeut: Documentation updated. For phasing-out support of 2.7, we should start thinking about how to do it but this PR is probably not the place.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
mayeut: I've hacked something together to get the pattern 'only filter out'

YannickJadoul: Nice! I've wanted to do something like this before, but I never was happy with the way to handle the 2-phase format-strings, but ... this could work. I should see how #143 gets rebased onto this but ... :-)

Just a suggestion; up to you to judge whether you like it or not: what about only keeping the loop over `manylinux_images`, and keep the listing of different Python versions and architectures explicit? This would keep consistency with the other platforms, and keep a nice mapping to the build identifiers. But again, just a suggestion. Up to your judgment.

docs/options.md
```diff
@@ -235,19 +235,26 @@ CIBW_REPAIR_WHEEL_COMMAND_LINUX: "auditwheel repair --lib-sdir . -w {dest_dir} {
...
 
 Note that `auditwheel` detects the version of the `manylinux` standard in the Docker image through the `AUDITWHEEL_PLAT` environment variable, as `cibuildwheel` has no way of detecting the correct `--plat` command line argument to pass to `auditwheel` for a custom image. If a Docker image does not correctly set this `AUDITWHEEL_PLAT` environment variable, the `CIBW_ENVIRONMENT` option can be used to do so (e.g., `CIBW_ENVIRONMENT="manylinux2010_$(uname -m)"`).
 
+Note that `manylinux2014` doesn't provide any python 2.7 implementation and shall be used together with `CIBW_SKIP` option
```
```suggestion
Note that `manylinux2014` doesn't support builds with Python 2.7 - when building with `manylinux2014`, skip Python 2.7 using `CIBW_SKIP` (see example below).
```

test/08_manylinuxXXXX_only/cibuildwheel_test.py
joerick: Could this change to only skip 2.7 when `manylinux_image == 'manylinux2014'`? While we're still supporting 2.7 we should check it still works on manylinux1. Want to be sure that our last 2.7 release is good!

test/08_manylinuxXXXX_only/spam.c
```diff
@@ -12,7 +10,9 @@ spam_system(PyObject *self, PyObject *args)
     if (!PyArg_ParseTuple(args, "s", &command))
         return NULL;
 
-#if defined(__linux__)
+#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 17)
+    sts = (int)secure_getenv("NON_EXISTING_ENV_VARIABLE");
+#elif defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 10)
     sts = malloc_info(0, stdout);
 #endif
```
joerick: Sorry, I might be a little ignorant here - what does this section do? Maybe some comments would help :)

test/shared/utils.py
```diff
@@ -66,37 +66,17 @@ def cibuildwheel_run(project_path, env=None, add_env=None, output_dir=None):
     return wheels
 
 
-def expected_wheels(package_name, package_version):
+def expected_wheels(package_name, package_version, manylinux_versions=['manylinux1', 'manylinux2010']):
```
joerick: Why is the default both 'manylinux1' and 'manylinux2010'? I thought the default was to build `manylinux2010` only?

test/08_manylinuxXXXX_only/spam.c
YannickJadoul: Part of the `manylinux` standard is a maximum version of glibc from which symbols can be used. So to force a wheel to be e.g. `manylinux2010`-compatible but not `manylinux1`-compatible, we need to use a symbol from `glibc` that's in `glibc` [version 2.12](https://www.python.org/dev/peps/pep-0571/#the-manylinux2010-policy) but not in [version 2.5](https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy).

test/shared/utils.py
YannickJadoul: Our simple, default `spam.c` is so simple that even when we build it on a `manylinux2010` image (where it can use more recent versions of `glibc` etc.), it produces a shared library that's also still `manylinux1` compatible.

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: Eh, I also thought so for a second, but then in principle, 2.7 is tested in the other tests already? So for the `manylinuxXXXX_only` tests, I'm not sure it's very urgent.

joerick: I don't think it's covered on the manylinux1 image, but I might be wrong?

test/shared/utils.py
joerick: Whoops. Ah yes. Realised this last night after I left this comment!

test/08_manylinuxXXXX_only/spam.c
joerick: Oh I see. A comment to that effect would be great.

YannickJadoul: @mayeut The one thing we're actually not testing this way is if during compilation we're building in the right `manylinux` image, since it now doesn't compile things that would fail if `glibc`'s version is smaller? I know we're testing it indirectly, when checking the wheels, but ...

What about having `CIBW_ENVIRONMENT='CFLAGS="-DMANYLINUX_IMAGE={}"'.format(manylinux_image)` and using that for the precprocessor statements? That would solve both problems, if it's not too messy?

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: > I don't think it's covered on the manylinux1 image, but I might be wrong?

Hmmm, true.

mayeut: restored tests with `cp27` on `manylinux1` and `manylinux2010`

test/08_manylinuxXXXX_only/cibuildwheel_test.py
mayeut: I finally kept consistency by rewriting the way identifiers are built on all architectures using PEP 425 as a baseline.

test/08_manylinuxXXXX_only/spam.c
mayeut: This one feels a bit messy. I'd rather rely on `auditwheel` and checking wheels doing their job.
The simple comments added in `spam.c` should be enough don't you think ?

test/08_manylinuxXXXX_only/cibuildwheel_test.py
YannickJadoul: Nice! :-)

test/08_manylinuxXXXX_only/spam.c
YannickJadoul: Works as well, for me.

mayeut: @joerick, good for you ?

test/08_manylinuxXXXX_only/spam.c
```diff
@@ -12,7 +18,9 @@ spam_system(PyObject *self, PyObject *args)
     if (!PyArg_ParseTuple(args, "s", &command))
         return NULL;
 
-#if defined(__linux__)
+#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 17)  /* manylinux2014 is glibc 2.17 */
+    sts = (int)secure_getenv("NON_EXISTING_ENV_VARIABLE");
```
joerick:
```suggestion
    // secure_getenv is only available in manylinux2014
    sts = (int)secure_getenv("NON_EXISTING_ENV_VARIABLE");
```

test/08_manylinuxXXXX_only/spam.c
```diff
@@ -12,7 +18,9 @@ spam_system(PyObject *self, PyObject *args)
...
+#elif defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 10)  /* manylinux2010 is glibc 2.12 */
     sts = malloc_info(0, stdout);
```
joerick:
```suggestion
    // malloc_info is only available on manylinux2010+
    sts = malloc_info(0, stdout);
```

test/08_manylinuxXXXX_only/spam.c
joerick: I added a little more explanation as comment, but otherwise i'm happy :)

test/08_manylinuxXXXX_only/spam.c
YannickJadoul: To be pedantic:

"ensuring a manylinux2014-only wheel is produced", or "ensuring a manylinux2014 wheel is produced that is not manylinux2010-compatible"

joerick: good point. I've edited the above

mayeut:
@joerick, @YannickJadoul,
It think this is now good to go, all comments addressed.

PR #217, opened 2019-11-17
Run each test using a temporary directory

mayeut:
Simpler way of getting each test to run in a unique temporary directory than #213 and #214 which also attempts to make all tests being run with a single `pytest` invocation.

### DIFF
--- a/test/shared/utils.py
+++ b/test/shared/utils.py
@@ -4,10 +4,24 @@ Utility functions used by the cibuildwheel tests.
 This file is added to the PYTHONPATH in the test runner at bin/run_test.py.
 '''
 
-import subprocess, sys, os
+import subprocess, sys, os, shutil
+from tempfile import mkdtemp
+from contextlib import contextmanager
+
 
 IS_WINDOWS_RUNNING_ON_AZURE = os.path.exists('C:\\hostedtoolcache')
 IS_WINDOWS_RUNNING_ON_TRAVIS = os.environ.get('TRAVIS_OS_NAME') == 'windows'
+
+
+# Python 2 does not have a tempfile.TemporaryDirectory context manager
+@contextmanager
+def TemporaryDirectoryIfNone(path):
+    _path = path or mkdtemp()
+    try:
+        yield _path
+    finally:
+        if path is None:
+            shutil.rmtree(_path)
 
 
 def cibuildwheel_get_build_identifiers(project_path, env=None):
@@ -24,12 +38,18 @@ def cibuildwheel_get_build_identifiers(project_path, env=None):
     return cmd_output.strip().split('\n')
 
 
-def cibuildwheel_run(project_path, env=None, add_env=None):
+def cibuildwheel_run(project_path, env=None, add_env=None, output_dir=None):
     '''
     Runs cibuildwheel as a subprocess, building the project at project_path.
 
     Uses the current Python interpreter.
-    Configure settings using env.
+
+    :param project_path: path of the project to be built.
+    :param env: full environment to be used, os.environ if None
+    :param add_env: environment used to update env
+    :param output_dir: directory where wheels are saved. If None, a temporary
+    directory will be used for the duration of the command.
+    :return: list of built wheels (file names).
     '''
     if env is None:
         env = os.environ.copy()
@@ -37,10 +57,13 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
     if add_env is not None:
         env.update(add_env)
 
-    subprocess.check_call(
-        [sys.executable, '-m', 'cibuildwheel', project_path],
-        env=env,
-    )
+    with TemporaryDirectoryIfNone(output_dir) as _output_dir:
+        subprocess.check_call(
+            [sys.executable, '-m', 'cibuildwheel', '--output-dir', str(_output_dir), project_path],
+            env=env,
+        )
+        wheels = os.listdir(_output_dir)
+    return wheels
 
 
 def expected_wheels(package_name, package_version):
--- a/bin/run_test.py
+++ b/bin/run_test.py
@@ -3,27 +3,13 @@
 from __future__ import print_function
 import os, sys, subprocess, shutil
 
-project_root = os.path.dirname(os.path.dirname(__file__))
-test_utils_dir = os.path.join(project_root, 'test', 'shared')
 
 def single_run(test_project):
-    # set up an environment that gives access to the test utils
-    env = os.environ.copy()
-
-    if 'PYTHONPATH' in env:
-        env['PYTHONPATH'] += os.pathsep + test_utils_dir
-    else:
-        env['PYTHONPATH'] = test_utils_dir
-
     # run the test
     subprocess.check_call(
         [sys.executable, '-m', 'pytest', '-vv', os.path.join(test_project, 'cibuildwheel_test.py')],
-        env=env,
     )
 
-    # clean up
-    if os.path.exists('wheelhouse'):
-        shutil.rmtree('wheelhouse')
 
 if __name__ == '__main__':
     import argparse
@@ -33,7 +19,7 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
     project_path = os.path.abspath(args.test_project_dir)
-    
+
     if not os.path.exists(project_path):
         print('No test project not found.', file=sys.stderr)
         exit(2)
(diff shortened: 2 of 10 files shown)


test/shared/utils.py
```diff
@@ -37,10 +50,17 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
...
-        env=env,
-    )
+    with TemporaryDirectory() as output_dir:
+        try:
+            subprocess.check_call(
+                [sys.executable, '-m', 'cibuildwheel', '--output-dir', output_dir, project_path],
+                env=env,
+            )
+        except:
+            assert len(os.listdir(output_dir)) == 0
```
YannickJadoul: Hmmm, interesting. Are we sure about this?

mayeut: I was hesitant about this and thought I'd get feedback.
While all tests are passing as of now, this check is not applicable in the general case.
Wheels are put in the output directory config per config on Windows/macOS or all at once on Linux.
The behavior is thus different depending on the platform. On Linux, you either get all or none of the wheels depending of a single error. On macOS/Windows, you get all wheels with tests successful (if applicable) up until the error.
Either we consider the linux behavior to be the right one and this check can be left as-is (requiring to add further tests later to ensure this behavior) or I can make this check optional for the one test that does use it (and we keep a different behavior depending on the platform).

YannickJadoul: > The behavior is thus different depending on the platform.

Good point. I hadn't even considered that. But that's a good point.
I was more thinking in terms that it should be the test asserting (because asserting no wheels are built is part of the test-), rather than here in `utils`.

In the end, the difference between OSs doesn't matter that much, I'd say? The error is thrown anyway?

YannickJadoul: There are two more solutions, though:

1. Catch all errors, and just return the list? Normally, the test should catch that the wrong wheels are built? If necessary, we can return a few things, like error status, output, and built wheels; or return built wheels and `None`/caught exception?

2. Throw a custom exception on catching this one, that "includes the list of built wheels", as a way of having an exception ánd return value?

Neither of these two solutions give me a very warm feeling, though.

test/shared/utils.py
```diff
@@ -37,10 +50,17 @@ def cibuildwheel_run(project_path, env=None, add_env=None):
...
         env.update(add_env)
 
-    subprocess.check_call(
-        [sys.executable, '-m', 'cibuildwheel', project_path],
-        env=env,
-    )
+    with TemporaryDirectory() as output_dir:
+        try:
+            subprocess.check_call(
+                [sys.executable, '-m', 'cibuildwheel', '--output-dir', output_dir, project_path],
```
YannickJadoul: Other annoying detail: in #179 (about which @joerick is not fully conviced yet, either, as far as I understood), @Czaki needs `subprocess.run` rather than `subprocess.check_call`. Then again, he already wasn't using `utils.cibuildwheel_run`, so maybe that's not relevant.

Czaki: But now this test (from #179) has symilar structure to other. After this changes, even after refactor to new style it will have "different structure"

test/shared/utils.py
mayeut: There's yet another solution which provides flexibility (but consistency-wise would be a bit lower).
Add an `output_dir=None` default arg to `cibuildwheel_run`, much like @Czaki proposed at one time in order to use `tmp_path` only on selected project before saying all tests shall use a temp folder.

My feeling is that all tests needs to run in a temp folder.
90% of them won't need to bother with an `output_dir` and just need the resulting wheel names and what's here just works. By adding `output_dir=None`, we allow selected tests that require access to `output_dir` to provide it.

Any thoughts ?

Czaki: > The behavior is thus different depending on the platform. On Linux, you either get all or none of the wheels depending of a single error.
No. If you build 64 an 32 bits, then it copy wheels in two packs.

YannickJadoul: > No. If you build 64 an 32 bits, then it copy wheels in two packs.

Huh? We don't mount a directory into the docker image anymore, do we? Because CircleCI doesn't allow that?

YannickJadoul: @mayeut 

> There's yet another solution which provides flexibility (but consistency-wise would be a bit lower).
> [...]
> Any thoughts ?

Hmmm, myes, interesting; I hadn't thought of that, indeed. I guess that's as good of a solution as the other two things I suggested (or even better), since it's less of a hack, and doesn't try to "return a value from a function that's throwing an exception". (Or maybe this is the sign that the abstraction of having `run_cibuildwheel` return a list of wheels is not completely right?)

Czaki: But you run separated docker containers for x86_64 and i686 wheels. 
and when container finish work then result is copied. So you will have two transactions.

YannickJadoul: @Czaki Ooooh, yes, right. I thought you meant you get pairs of 32- and 64-bit wheels. But you get two separate batches of all 32- and all 64-bit wheels.

Czaki:
@YannickJadoul @mayeut 
I would like to argue why I prefer #213 or #218 rather the #214 or this.
My points of view is point of view of person who short time ago join to developing package. 

In my opinion test should be self explaining. So person who do not have perfect knowledge about project should understand what scenario is executed. For me this PR and #214 do not satisfy this. 

Reading test under this PR I rather expect that `utils.cibuildwheel_run` clean wheelhouse rather than use `TemporaryDirectory`. And I do not expect that it will return list of wheelhouse.  

I think that all my remarks can be removed by rename functions. Ex. `cibuildwheel_run -> cibuildwheel_run_and_list_wheels`, but this can produce very long names. Maybe this approach needs rethink structure of tests.

test/shared/utils.py
mayeut: > No. If you build 64 an 32 bits, then it copy wheels in two packs.

You're right. Doesn't change the fact that behavior is different though. Anyway, this is not the matter here.

> Or maybe this is the sign that the abstraction of having run_cibuildwheel return a list of wheels is not completely right?

I think the option doesn't hurt too much the abstraction.

mayeut:
> My points of view is point of view of person who short time ago join to developing package.
In my opinion test should be self explaining. So person who do not have perfect knowledge about project should understand what scenario is executed. For me this PR and #214 do not satisfy this.

Thanks for that feedback, that's valuable. I do understand this point of view and I guess the current size of this part of the project does allow for what you're asking. @joerick ?

From a generic maintenance point of view, duplicate code is most often a nightmare. Things can grow out of proportions quite fast and, I really dislike having to change something at many places when it could have been factored. Yes, it does requires a better understanding of the code base but it's a trade-off between maintainability and ease of discovery. It does require to spend time thinking this refactor in order not to create something too convoluted either. (and if you think this PR is, maybe I'm not leaning towards a solution that's good enough)

May I ask if the current state of tests in master could have benefited from a bit more docs (in the specific README probably) ?
Explaining the structure of a test perhaps ? or linking to a reference test that's more commented than other (maybe test 02 which has 3 tests)

YannickJadoul:
> Reading test under this PR I rather expect that `utils.cibuildwheel_run` clean wheelhouse rather than use `TemporaryDirectory`. And I do not expect that it will return list of wheelhouse.

Hmmm, I don't fully understand why, because I feel the opposite is more intuitive: If you are not very familiar with the code or `cibuildwheel` in general, this hides away the details. `actual_wheels = utils.cibuildwheel_run(...)` and then `assert set(actual_wheels) == set(expected_wheels)` feels very clear to me, hiding away all the details you at first don't care about? (This kind of code that's almost readable as text, is why we love Python, no? :) )

My resistance against making `tmp_dir` explicit is 1) it increases maintenance and the possibility for inconsistencies ([Don't repeat yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself)!). But 2) it also adds complexity by specifying a directory we don't care about. By that I mean: we care about it to give correct results, but the actual test does not depend on the value of the directory and does not e.g. test that if you pass `-w wheelhouse` `cibuildwheel` will not place them in a different folder `wheelhouse_dir` or so. So, if you want to look at the implementation details, you'll have to look at `utils.py` anyway, but if you take a quick look at what a certain test is testing, why bother the reader with `tmp_dir`?

> In my opinion test should be self explaining. So person who do not have perfect knowledge about project should understand what scenario is executed. 

So I completely agree on this; I just seem to have a different solution for it :-)

PR #220, opened 2019-11-19
Only build x86_64 / Intel 10.9+ on macOS

mayeut:
Only build x86_64 / Intel 10.9+ on macOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
 What does it do?
 ----------------
 
-|   | macOS 10.6+ intel | macOS 10.9+ x86_64 | manylinux i686 | manylinux x86_64 |  Windows 32bit | Windows 64bit |
-|---|---|---|---|---|---|---|
-| Python 2.7 | ✅ |    | ✅ | ✅ | ✅¹ | ✅¹ |
-| Python 3.5 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
-| Python 3.6 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
-| Python 3.7 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
-| Python 3.8 |    | ✅ | ✅ | ✅ | ✅  | ✅  |
+|   | macOS 10.9+ x86_64 | manylinux i686 | manylinux x86_64 |  Windows 32bit | Windows 64bit |
+|---|---|---|---|---|---|
+| Python 2.7 | ✅ | ✅ | ✅ | ✅¹ | ✅¹ |
+| Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
 
 > ¹ Not supported on Travis
 
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -100,8 +100,14 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
         # setup target platform, only required for python 3.5
         if config.version == '3.5':
-            env['_PYTHON_HOST_PLATFORM'] = 'macosx-10.9-x86_64'  # cross-compilation platform override
-            env['ARCHFLAGS'] = '-arch x86_64'  # https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260
+            if '_PYTHON_HOST_PLATFORM' not in env:
+                # cross-compilation platform override
+                env['_PYTHON_HOST_PLATFORM'] = 'macosx-10.9-x86_64'
+            if 'ARCHFLAGS' not in env:
+                # https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260
+                env['ARCHFLAGS'] = '-arch x86_64'
+            if 'MACOSX_DEPLOYMENT_TARGET' not in env:
+                env['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
 
         # run the before_build command
         if before_build:
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -14,11 +14,11 @@ from .util import prepare_command, get_build_verbosity_extra_flags
 def get_python_configurations(build_selector):
     PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
     python_configurations = [
-        PythonConfiguration(version='2.7', identifier='cp27-macosx_intel', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.6.pkg'),
-        PythonConfiguration(version='3.5', identifier='cp35-macosx_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
-        PythonConfiguration(version='3.6', identifier='cp36-macosx_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_intel', url='https://www.python.org/ftp/python/3.7.5/python-3.7.5-macosx10.6.pkg'),
-        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.0/python-3.8.0-macosx10.9.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_x86_64', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.9.pkg'),
+        PythonConfiguration(version='3.5', identifier='cp35-macosx_x86_64', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.6', identifier='cp36-macosx_x86_64', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.9.pkg'),
+        PythonConfiguration(version='3.7', identifier='cp37-macosx_x86_64', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.9.pkg'),
+        PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.1/python-3.8.1-macosx10.9.pkg'),
     ]
 
     # skip builds as required
@@ -32,6 +32,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
     repaired_wheel_dir = os.path.join(temp_dir, 'repaired_wheel')
 
     python_configurations = get_python_configurations(build_selector)
+
     get_pip_url = 'https://bootstrap.pypa.io/get-pip.py'
     get_pip_script = '/tmp/get-pip.py'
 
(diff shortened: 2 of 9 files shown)


mayeut:
It probably should be the default. `intel` can even be dropped I think, that would simplify things greatly as well as going 10.9+. Then this PR would become "fix python 3.5 on macOS to produce a macos_10_9_x86_64 wheel" and changes will be dropped when python 3.5 goes EOL. We can also close #202 / #203 if we go that way.
```
SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(details.distro.version, r"^(\d+.\d+)") as system_version,
  details.cpu
WHERE
  details.system.name = 'Darwin' AND
  file.filename like '%macos%' AND
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY system_version, details.cpu
ORDER BY num_downloads DESC
```
image
image

PS: I'm not sure that i686/x86_64 actually reflect the flavor of python running (I download one of my wheel from a 10.14 system from python3.7-32 and I can't see it in the stats while I'm sure it wasn't cached). Given that python runs x86_64 as a default when possible, it shouldn't change much the outcome.

joerick:
Tbh I'm on board with 64bit only default on Mac too, if there are no major objections. I can't remember my last 32bit mac, but it was a loong time ago! The only thing I don't know (agreeing with your suspicion with those figures) is if people are still running python-32 on a 64bit machine (for binary compat reasons), but I'd say it was an edge case. 

I think `CIBW_PREFERRED_ARCHITECTURE_MACOS` might be a good name? Keeping the platform as a suffix for conformity.

Also, I'm thinking this PR could remove the arch from build identifier, so it'd be just `cp38-macosx_10_14`.

What do you think @mayeut?

joerick:
That build identifier would actually become even simpler `cp37-macosx` when #156 lands, because the specific version of macOS depends on an environment variable that's set when building. I should probably have made them that simple from the start! Hindsight 20/20 I suppose 🙄

mayeut:
> I think CIBW_PREFERRED_ARCHITECTURE_MACOS might be a good name? Keeping the platform as a suffix for conformity.

I think your proposal is much clearer for the end user (and even for me) but for consistency sake, I went with `MACHINE` which is used throughout `CPython` when building platform tags. I guess my explanation doesn't change your preference ?

> Tbh I'm on board with 64bit only default on Mac too, if there are no major objections. I can't remember my last 32bit mac, but it was a loong time ago! The only thing I don't know (agreeing with your suspicion with those figures) is if people are still running python-32 on a 64bit machine (for binary compat reasons), but I'd say it was an edge case.

Ok, so default to `x86_64` but still allow `intel` builds on demand ? You don't want to just drop 32-bit and go to 10.9 as a minimum ?

YannickJadoul:
> You don't want to just drop 32-bit and go to 10.9 as a minimum ?

In combination with #156, aren't we leaving the minimum version up to the user to decide and to be checked by `wheel`?

mayeut:
Let's wait for #156 to be merged-in, I'll rework this after.

joerick:
#156 is waiting on a pypa/wheel release, which might be a while yet... 

@mayeut:
> it probably should be the default. intel can even be dropped I think, that would simplify things greatly as well as going 10.9+. Then this PR would become "fix python 3.5 on macOS to produce a macos_10_9_x86_64 wheel" and changes will be dropped when python 3.5 goes EOL. We can also close #202 / #203 if we go that way.

For the sake of simplicity, I'm happy to just do the above - drop `intel` and go to `10_9` where we can. 32-bit isn't something our users have asked for, and it'll be less code to maintain. Would you be able to go ahead with that @mayeut ?

cibuildwheel/macos.py
```diff
@@ -14,10 +14,10 @@
...
-        PythonConfiguration(version='3.5', identifier='cp35-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
-        PythonConfiguration(version='3.6', identifier='cp36-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.6.pkg'),
-        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_6_intel', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.6.pkg'),
+        PythonConfiguration(version='2.7', identifier='cp27-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/2.7.17/python-2.7.17-macosx10.9.pkg'),
+        PythonConfiguration(version='3.5', identifier='cp35-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.5.4/python-3.5.4-macosx10.6.pkg'),
+        PythonConfiguration(version='3.6', identifier='cp36-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.9.pkg'),
+        PythonConfiguration(version='3.7', identifier='cp37-macosx_10_9_x86_64', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.9.pkg'),
```
joerick: Aha, so we have a little snag here - the build identifiers are changing - both the machine and the macos system. 

Previously I've mused about renaming these e.g. `cp37-macosx_10_9_x86_64` →  `cp37-macosx`. Thinking about this fresh...

The architecture isn't needed any longer, since we only build x86_64. Should we remove it? Will we ever need an architecture switch on macos again? Hm, well, [ARM Macs might exist](https://www.theinquirer.net/inquirer/news/3082867/apple-arm-macbook-2020-bloomberg) one day! And, all other platforms contain an architecture flag, so keeping it retains consistency.

Then there's the `10_9` bit - we expect to remove this part because #156 should be making this conditional on the MACOSX_DEPLOYMENT_TARGET var instead.

So here's what I'm thinking... rather than making users update their settings from `cp37-macosx_10_6_intel` to `cp37-macosx_10_9_x86_64` and then from `cp37-macosx_10_9_x86_64` to `cp37-macosx_x86_64` when #156 lands, let's pull in the [build selector migration](https://github.com/joerick/cibuildwheel/pull/156/files#diff-dfdf4eb7a2b7f84841e9d286a4063e03R204-R208) and go to the `cp37-macosx_x86_64` format in this PR.

Thoughts @YannickJadoul @Czaki ?

README.md
```diff
@@ -15,13 +15,13 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
...
-| Python 3.6 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
-| Python 3.7 | ✅ |    | ✅ | ✅ | ✅  | ✅  |
-| Python 3.8 |    | ✅ | ✅ | ✅ | ✅  | ✅  |
+|   | macOS 10.9+ x86_64 | manylinux i686 | manylinux x86_64 |  Windows 32bit | Windows 64bit |
+|---|---|---|---|---|---|
+| Python 2.7 | ✅ | ✅ | ✅ | ✅¹ | ✅¹ |
+| Python 3.5 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.6 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.7 | ✅ | ✅ | ✅ | ✅  | ✅  |
+| Python 3.8 | ✅ | ✅ | ✅ | ✅  | ✅  |
```
joerick: Ahhhh :)

YannickJadoul: Don't merge #185, then :-/

cibuildwheel/macos.py
YannickJadoul: Depends a bit on when we can merge #156, i.e., when a new wheel version will be released.

If that's before you plan a new release, I'd argue to keep this as it is now and not mix different PRs. If wheel won't release too soon, then yeah, maybe. But maybe that's still another clean PR, then?
(I've asked, btw: https://github.com/pypa/wheel/pull/314#issuecomment-569087959; let's see what comes out of that.)

On keeping the architecture: I don't think that's really bothering? It even keeps things clear, that wheels will only be built for 64-bit, I'd say. And with the wildcard matching on `*-macosx_*`, most users might not even need to change too much (I'd hope users don't depend on this 10_6/10_9 too much?). And indeed: keeping it would be good for the sake of consistency with `manylinux_x86_64`/`manylinux_i686`/ `win32`/`win_amd64`, where also now always have architecture, but not the platform's version :-)

YannickJadoul: Afterthought: is there any reason to use the `10.9` packages from Python.org? We should be able to build 64-bit-only wheels that are still 10.6-compatible, or not, @mayeut? In that case, we can still keep our old build identifiers for a little while, until #156.

joerick: Thanks for getting in touch with the wheel folks! I had been trying to unblock a few PRs but hopefully that's moving now.

> Afterthought: is there any reason to use the 10.9 packages from Python.org?

I _think_ this is more convenient because the 10.9 packages build a x86_64 wheel by default, but I'll let @mayeut correct me if I'm wrong...

joerick:
I think this PR can get moving again! @mayeut, are you able to rebase/merge master?

mayeut:
@joerick, now rebased.

cibuildwheel/macos.py
```diff
@@ -98,6 +98,11 @@ def call(args, env=None, cwd=None, shell=False):
         call(['pip', '--version'], env=env)
         call(['pip', 'install', '--upgrade', 'setuptools', 'wheel', 'delocate'], env=env)
 
+        # setup target platform, only required for python 3.5
+        if config.version == '3.5':
+            env['_PYTHON_HOST_PLATFORM'] = 'macosx-10.9-x86_64'  # cross-compilation platform override
```
Czaki:
```python
if "MACOSX_DEPLOYMENT_TARGET" not in env:
    env["MACOSX_DEPLOYMENT_TARGET"]="10.9"
```

And maybe some checek if variables are not aleredy defined?

cibuildwheel/macos.py
```diff
@@ -50,20 +51,19 @@ def call(args, env=None, cwd=None, shell=False):
...
-    call(['curl', '-L', '--retry', '3', '--retry-delay', '3', '-o', get_pip_script, get_pip_url])
+    call(['curl', '--retry', '3', '--retry-delay', '3', '-sSLo', get_pip_script, get_pip_url])
 
     for config in python_configurations:
         # if this version of python isn't installed, get it from python.org and install
         python_package_identifier = 'org.python.Python.PythonFramework-%s' % config.version
         if python_package_identifier not in installed_system_packages:
             # download the pkg
-            call(['curl', '-L', '-o', '/tmp/Python.pkg', config.url])
+            call(['curl', '--retry', '3', '--retry-delay', '3', '-sSLo', '/tmp/Python.pkg', config.url])
```
YannickJadoul: Should we not turn this into a function, then, similar to the thing on Windows?

mayeut:
@YannickJadoul, @joerick, @Czaki,
Thanks for the feedback. I unfortunately have very limited time to work on Open Source projects at the moment so feel free to push in this branch.

Czaki:
@YannickJadoul @joerick, I'm not a maintainer so I cannot push changes to branch. Which of You I should mention if I have some remark?

YannickJadoul:
OK, will do; then we can get this merged soon :-) Thanks for all the work you've already put into this before, @mayeut!

@Czaki Go ahead and mention me, if you want. I'm subscribed to this thread, anyway, so I'll implement the remarks made.

joerick:
I've made the changes you suggested @Czaki @YannickJadoul, and I think I'm happy with it. Any further comments?

YannickJadoul:
Looks good to me!

One thing I can still think of: now that `macosx_intel` do not work anymore, should we add a warning/error/something else to `__main__.py`, just like we did before when we replaced old build tags?

If not, go ahead and merge, as far as I'm concerned :-)

Czaki:
I try to build quite big python 3.5 depend project. I takes about an hour to finish build. In few hour i will give information if it pass.

Czaki:
It works.

YannickJadoul:
Hurray! Thanks for testing, @Czaki!

> One thing I can still think of: now that `macosx_intel` do not work anymore, should we add a warning/error/something else to `__main__.py`, just like we did before when we replaced old build tags?

Then it's just this. Any thoughts, @Czaki? Because maybe this should have already be done in #156, actually.
If not, let's just leave it to @joerick's judgement :-)

joerick:
> One thing I can still think of: now that macosx_intel do not work anymore, should we add a warning/error/something else to __main__.py, just like we did before when we replaced old build tags?

Ah! Yes, good catch, thank you @YannickJadoul !

YannickJadoul:
> Ah! Yes, good catch, thank you @YannickJadoul !

But so, worth taking into account that it also already changed in #156. So you might want to check for `macosx_10_6` or `macosx_10_9` as well!

joerick:
> So you might want to check for `macosx_10_6` or `macosx_10_9` as well!

Indeed. We haven't released that migration, so they can probably be merged together.

Czaki:
in #156 system version is removed only. I do not do any changes with `intel`/`x86_64` suffixes.

YannickJadoul:
@joerick @Czaki Just found that the docs still say

> To get C++11 and C++14 support, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.9"`.

In `cpp_standards.md`

Since we've now bumped everything to at least 10.9, we don't need that anymore?

Czaki:
Yes

YannickJadoul:
Fixed; quickly pushed ffec9ae to master.

McSinyx:
Hi, is it possible to have a release with this in a near future?  Our project needs 10.9 for C++11 support and the previous release only have it for Python 3.8.

Czaki:
Hi @McSinyx. Only thing you need is to define `MACOSX_DEPLOYMENT_TARGET` to `10.9`. See more:  https://cibuildwheel.readthedocs.io/en/latest/cpp_standards/

McSinyx:
Thank you @Czaki for the tip.  How do I specify the Python version then?  Is the following OK?

    CIBW_BUILD=cp37-macosx_10_6_intel

I see the linked documentation is built on GitHub master instead of the released one though.  I've just checked out `v1.1.0` and `MACOSX_DEPLOYMENT_TARGET` is not handled anywhere.

Czaki:
> I see the linked documentation is built on GitHub master instead of the released one though. I've just checked out v1.1.0 and MACOSX_DEPLOYMENT_TARGET is not handled anywhere.

Handling of MACOSX_DEPLOYMENT_TARGET is in `wheel` package form release 0.34.0 

> Thank you @Czaki for the tip. How do I specify the Python version then? Is the following OK?

because next release should be in short time I suggest usage 

```
CIBW_BUILD="cp37-macosx*"
```

McSinyx:
> Handling of `MACOSX_DEPLOYMENT_TARGET` is in wheel package form release 0.34.0

Thanks, I was wondering why it isn't prefixed with `CIBW`.  Your suggestion [works very well](https://travis-ci.com/McSinyx/palace/builds/151226333) too!

Czaki:
> Thanks, I was wondering why it isn't prefixed with CIBW. 

Because this is MacOs system variable used to control compatibility . 

> Your suggestion works very well too!

Why you use separate job for each build? Maybe try: `CIBW_BUILD="cp3[6-8]-macosx*"`

McSinyx:
> Why you use separate job for each build?

We want to make use of some parallelism, but mainly it's for better Python compatibility test (e.g. if 3.6 fails we would still want to know if 3.7 passes).

PR #221, opened 2019-11-20
fix UnboundLocalError instead of proper error message

Czaki:
fix for
```
  File "/home/czaki/.pyenv/versions/3.6.8/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/czaki/.pyenv/versions/3.6.8/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/czaki/Projekty/cibuildwheel/cibuildwheel/__main__.py", line 254, in <module>
    main()
  File "/home/czaki/Projekty/cibuildwheel/cibuildwheel/__main__.py", line 76, in main
    if platform is None:
UnboundLocalError: local variable 'platform' referenced before assignment
```

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -61,25 +61,29 @@ def main():
     args = parser.parse_args()
 
     detect_obsolete_options()
-
     if args.platform != 'auto':
         platform = args.platform
     else:
         ci = strtobool(os.environ.get('CI', 'false')) or 'BITRISE_BUILD_NUMBER' in os.environ or 'AZURE_HTTP_USER_AGENT' in os.environ
-        if ci:
-            if sys.platform.startswith('linux'):
-                platform = 'linux'
-            elif sys.platform == 'darwin':
-                platform = 'macos'
-            elif sys.platform == 'win32':
-                platform = 'windows'
-        if platform is None:
+        if not ci:
             print('cibuildwheel: Unable to detect platform. cibuildwheel should run on your CI server, '
                   'Travis CI, AppVeyor, Azure Pipelines and CircleCI are supported. You can run on your '
                   'development machine or other CI providers using the --platform argument. Check --help '
                   'output for more information.',
                   file=sys.stderr)
             exit(2)
+        if sys.platform.startswith('linux'):
+            platform = 'linux'
+        elif sys.platform == 'darwin':
+            platform = 'macos'
+        elif sys.platform == 'win32':
+            platform = 'windows'
+        else:
+            print('cibuildwheel: Unable to detect platform from "sys.platform" in a CI environment. You can run '
+                  'cibuildwheel using the --platform argument. Check --help output for more information.',
+                  file=sys.stderr)
+            exit(2)
+
 
     output_dir = args.output_dir
     test_command = get_option_from_environment('CIBW_TEST_COMMAND', platform=platform)
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -61,6 +61,7 @@ def main():
     args = parser.parse_args()
 
     detect_obsolete_options()
+    
     if args.platform != 'auto':
         platform = args.platform
     else:
@@ -178,7 +179,7 @@ def main():
     elif platform == 'macos':
         cibuildwheel.macos.build(**build_options)
     else:
-        raise Exception('Unsupported platform')
+        raise Exception('Unsupported platform: {}'.format(platform))
 
 
 def detect_obsolete_options():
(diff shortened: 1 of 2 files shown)


cibuildwheel/__main__.py
```diff
@@ -73,6 +72,8 @@ def main():
                 platform = 'macos'
             elif sys.platform == 'win32':
                 platform = 'windows'
+            else:
+                platform = None
```
mayeut: the assignment shall be moved to the same level as the `ci` assignment to fix all cases. This only fixes the case where we're running in CI but `sys.platform` returns an unsupported value.
If we wan't a different message for each case, could be modified as follow:
```
        ci = strtobool(os.environ.get('CI', 'false')) or 'BITRISE_BUILD_NUMBER' in os.environ or 'AZURE_HTTP_USER_AGENT' in os.environ
        if not ci:
            print('cibuildwheel: Unable to detect platform. cibuildwheel should run on your CI server, '
                  .........,
                  file=sys.stderr)
            exit(2)
        if sys.platform.startswith('linux'):
            platform = 'linux'
        elif sys.platform == 'darwin':
            platform = 'macos'
        elif sys.platform == 'win32':
            platform = 'windows'
        else:
            print('cibuildwheel: Unable to detect platform from "sys.platform"'
                  .........,
                  file=sys.stderr)
            exit(2)

```

Czaki: I change it, but in another way. Your proposition create code repetition.

mayeut: @Czaki, in fact you're using my exact first proposition. The second proposition doesn't duplicate code.
Full picture:
```

    detect_obsolete_options()

    if args.platform != 'auto':
        platform = args.platform
    else:
        ci = strtobool(os.environ.get('CI', 'false')) or 'BITRISE_BUILD_NUMBER' in os.environ or 'AZURE_HTTP_USER_AGENT' in os.environ
        if not ci:
            print('cibuildwheel: Unable to detect platform. cibuildwheel should run on your CI server, '
                  'Travis CI, AppVeyor, Azure Pipelines and CircleCI are supported. You can run on your '
                  'development machine or other CI providers using the --platform argument. Check --help '
                  'output for more information.',
                  file=sys.stderr)
            exit(2)
        if sys.platform.startswith('linux'):
            platform = 'linux'
        elif sys.platform == 'darwin':
            platform = 'macos'
        elif sys.platform == 'win32':
            platform = 'windows'
        else:
            print('cibuildwheel: Unable to detect platform from "sys.platform" in a CI environment. You can run '
                  'cibuildwheel using the --platform argument. Check --help output for more information.',
                  file=sys.stderr)
            exit(2)

    output_dir = args.output_dir
```

YannickJadoul:
Oh, whoops, regression introduced in #204?

mayeut:
> Oh, whoops, regression introduced in #204?

Yes, as mentioned in [this comment](https://github.com/joerick/cibuildwheel/pull/214/files#r347129283)

Czaki:
I'm writing tests for this now.

cibuildwheel/__main__.py
Czaki: Ok, I see the difference and agree that it is better.

mayeut:
@Czaki,

Those tests are almost exactly what I had in mind. Can you move them in a new file please (maybe `unit_test/platform_test.py` or `unit_test/parse_platform_test.py`) ? the `unit_test/environment_test.py` file is used for `cibuildwheel.environment` unit tests.

cibuildwheel/__main__.py
```diff
@@ -61,25 +61,29 @@ def main():
     args = parser.parse_args()
 
     detect_obsolete_options()
-
```
mayeut: Maybe this empty should be kept for readability ? allows the whole logic block for platform detection to be on its own.

Czaki: Ok.  maybe this function should be renamed to `detect_and_fix_obsolete_environment_variables` or `fix_deprecated_environment_variables"?

mayeut: I like the second one `fix_deprecated_environment_variables`,  it's clear enough. No need to add `detect_and_`.

Czaki: this PR or another? it is not connected with main problem solved by this PR.

mayeut: Let's say another. Since @joerick approved, I'll merge this one.

Czaki:
@mayeut  A moment ago one question show in my head. If the `build` functions from `linux` `macos` and `windows` should be monkeypatched? For security?

mayeut:
@Czaki, I'm not sure to understand what you mean. You can Open an Issue if you have a concern, we'll exchange on the subject in that issue.

PR #222, opened 2019-11-20
make tests pytest callable

Czaki:
Make whole test set executable by pytest. 

This is not removing original test run scripts.


joerick:
Having thought about it, I'm pretty neutral on this. @YannickJadoul and @mayeut do you have a preference?

YannickJadoul:
Hmmm, I don't really know either. It's a lot of changes, but they're all pretty simple (i.e., mostly renaming files).
But then I still don't see the need to run these tests with `pytest`, especially after having extended our unit tests in #232. After all, aren't the unit tests are rather meant for development, while these integration tests are rather meant to simulate what happens on the CI server? So for that purpose, I don't see why `pytest` would be better than `bin/run_tests.py`.

Then again, there's no real harm merging this, either. I'd just suggest that if we merge this, we still add `testpaths = unit_test` to `pytest.ini`, so that if you run `pytest` without any arguments, it will only run the unit tests by default.

*Linking to original discussion in #213.*

joerick:
Closing, because the advantages are not clear

Issue #224, opened 2019-11-21
How strong tests should protect developer against himself 


Czaki:
Just after PR #221 was merged I realized that there is scenario that some developer, which make stupid changes in `cibuildwheel.__main__.py` it is possible that unit test can call `build` for some platform and  make changes on developer machine. 

Such person can be protected by add some monkeypatch which replace all `build` function with empty call. 
But it will increase code to be maintained. 

What you think about this?

@mayeut

YannickJadoul:
> Just after PR #221 was merged I realized that there is scenario that some developer, which make stupid changes in `cibuildwheel.__main__.py` it is possible that unit test can call `build` for some platform and make changes on developer machine.

If I understand correctly, this is probably a rather rare case, or not? Only very rarely do we actually interact with platform detection, right?

> Such person can be protected by add some monkeypatch which replace all build function with empty call.
But it will increase code to be maintained.

Do you have any estimate on the amount of code this would be? Would it just be a few lines in `conftest.py` (that seems reasonable), or a whole amount extra in `__main__.py` (less enthusiast about that).

> What you think about this?

Alternatively: what about making sure the integration tests do not run by default, but only when you pass an extra argument to `pytest` or so? As a result, 1) we'd maybe write more unit tests rather than full integration tests, which is good, and 2) that makes it explicit to the developer when starting the tests that he should be careful (and should make him realize if he has been working on `__main__.py`), as you cannot run these tests "by accident".

This is all only relevant if/when we replace `run_tests.py` by a `pytest` invocation, right? (@joerick didn't seem very happy about that, before.)

Czaki:
> If I understand correctly, this is probably a rather rare case, or not? Only very rarely do we actually interact with platform detection, right?

This is rare (I have hope really rare) case and tests will detect this. But before test fail it will introduce changes in running system  

> Do you have any estimate on the amount of code this would be? Would it just be a few lines in conftest.py (that seems reasonable), or a whole amount extra in __main__.py (less enthusiast about that).

My way to solve this will only increase amounts of lines in tests. For safety purpose it will mock `macos.build`, `linux.build`, `windows.build`, `subprocess.check_call` and maybe few others functions which calls can create some system changes. 

Following suggestion of @mayeut tests introduced in #221 are unit test, not integration tests. And same problem may show with other unit tests. If they are tests `main()` function then there is possible that someone modify code and this test will fire building process. 

Whole problem is not connected with making tests pytest callable.

[...]

PR #226, opened 2019-11-24
Add better protection to unittest to be shure that it will create no changes on developer machine

Czaki:
This PR cover really rare case case mentioned in #224. 
This may happen if user do some changes in `main` flow (ex set default platform) then it is possible that build will be launched. 

This PR introduces protection o two levels. Set build directory to temporary, empty dir. and monkeypatch `build`, `subprocess.Popen` and `urlopen`.  

`monkeypatch` will fail if module has no such attribute.

### DIFF
--- a/unit_test/platform_test.py
+++ b/unit_test/platform_test.py
@@ -1,14 +1,29 @@
 import sys
 import os
+import subprocess
 
 import pytest
 
 from cibuildwheel.__main__ import main
+from cibuildwheel import windows, linux, macos
 
 
-def test_unknown_platform_non_ci(monkeypatch, capsys):
+def not_call_mock(*args, **kwargs):
+    raise RuntimeError("This should never be called")
+
+
+def apply_mock_protection(monkeypatch):
+    monkeypatch.setattr(subprocess, "Popen", not_call_mock)
+    monkeypatch.setattr(windows, "urlopen", not_call_mock)
+    monkeypatch.setattr(windows, "build", not_call_mock)
+    monkeypatch.setattr(linux, "build", not_call_mock)
+    monkeypatch.setattr(macos, "build", not_call_mock)
+
+
+def test_unknown_platform_non_ci(monkeypatch, capsys, tmp_path):
     monkeypatch.setattr(os, 'environ', {})
-    monkeypatch.setattr(sys, "argv", ["python", "."])
+    monkeypatch.setattr(sys, "argv", ["python", str(tmp_path)])
+    apply_mock_protection(monkeypatch)
     with pytest.raises(SystemExit) as exit:
         main()
     assert exit.value.code == 2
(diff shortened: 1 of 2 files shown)


YannickJadoul:
Nice! Just two minor remarks:

> Set build directory to temporary, empty dir. 

Why is this necessary, if we're already making sure `build` does not work?

> and monkeypatch `build`, `subprocess.Popen` and `urlopen`.

What's the idea behind mocking `subprocess.Popen` and `urlopen`; are these called in `main`? Again, I thought we were just going to mock `build`, and keep it simple.

YannickJadoul:
Oh, by the way, if we're doing this and monkeypatching `build`. How about we test that it also gets called when we expect it to be called?

Czaki:
> >    Set build directory to temporary, empty dir.
> Why is this necessary, if we're already making sure build does not work?
>>  and monkeypatch build, subprocess.Popen and urlopen.
> What's the idea behind mocking subprocess.Popen and urlopen; are these called in main? Again, I thought we were just going to mock build, and keep it simple.

Idea is to protect all possible way. E.g if someone would like to split build process on two functions.

In typical use cast this tests are obsolete, because there is no scenario of call `build` without code modification. 

> Oh, by the way, if we're doing this and monkeypatching build. How about we test that it also gets called when we expect it to be called?

In integration test?

YannickJadoul:
> > > Set build directory to temporary, empty dir.
> 
> > Why is this necessary, if we're already making sure build does not work?
> 
> > > and monkeypatch build, subprocess.Popen and urlopen.
> 
> > What's the idea behind mocking subprocess.Popen and urlopen; are these called in main? Again, I thought we were just going to mock build, and keep it simple.
> 
> Idea is to protect all possible way. E.g if someone would like to split build process on two functions.

I see, but ... pfff, that might be a bit too much? I don't know, but I've never known the `build` function to be changed/split/renamed/..., and íf anyone is going to do that, it'll probably be very clear when that's done.

> In typical use cast this tests are obsolete, because there is no scenario of call `build` without code modification.
> 
> > Oh, by the way, if we're doing this and monkeypatching build. How about we test that it also gets called when we expect it to be called?
> 
> In integration test?

Hmm, myeah, maybe. I was just thinking that if we have these unit test anyway, we could add something. But you're right; it's probably not worth it.

Czaki:
> I see, but ... pfff, that might be a bit too much? I don't know, but I've never known the build function to be changed/split/renamed/..., and íf anyone is going to do that, it'll probably be very clear when that's done.

This is core of my question from #224. How strong this protection should be. In this PR I propose maximum level which is not connected with moving from unit test to integration test. 

> Hmm, myeah, maybe. I was just thinking that if we have these unit test anyway, we could add something. But you're right; it's probably not worth it.

I can propose more unit test which will allow test more part of cibuildwheel without running integration test.

unit_test/platform_test.py
```diff
@@ -1,14 +1,29 @@
...
 
 
-def test_unknown_platform_non_ci(monkeypatch, capsys):
+def not_call_mock(*args, **kwargs):
+    raise RuntimeError("This should never be called")
+
+
+def apply_mock_protection(monkeypatch):
+    monkeypatch.setattr(subprocess, "Popen", not_call_mock)
+    monkeypatch.setattr(windows, "urlopen", not_call_mock)
```
joerick: What's this urlopen mock for - could it be removed?

Czaki: on windows urlopen is used to do download nuget on disc
https://github.com/joerick/cibuildwheel/blob/054be2eff2293dfa9f3c380a9d20b9c91786b2e1/cibuildwheel/windows.py#L70

https://github.com/joerick/cibuildwheel/blob/054be2eff2293dfa9f3c380a9d20b9c91786b2e1/cibuildwheel/windows.py#L95-L99

and I decide to mock any function that can introduce change on machine.
I can remove it if you wish.

joerick:
Thanks!

Issue #223, opened 2019-11-21
Update Travis-CI example in README


SethMMorton:
First, I want to say that part of the reason I am opening this issue is to say how awesome this project is. Thank you for doing this!

---

There are some things that I would recommend changing in the Travis-CI example

- `sudo: required` is no longer needed, and is requested to be omitted now (https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)
- The whole `PIP=` things can be omitted if calls to `pip` use `python -m pip`
- Installation of dependencies could be moved to a `install` step
- Instead of setting an environment variable with the user name, it could be added to the `twine` call (this is just a preference)
- Windows could be added

It could be updated to look like

```yaml
language: python

matrix:
  include:
    # perform a linux build
    - services: docker
    # and a mac build
    - os: osx
      language: generic
    # and a windows build
    - os: windows
      language: generic
      before_install:
        - choco install python --version 3.8.0
        - export PATH="/c/Python38:/c/Python/Scripts:$PATH"

install:
  - python -m pip -U pip
  - python -m install twine cibuildwheel==1.0.0

script:
  # build the wheels, put them into './wheelhouse'
  - python -m cibuildwheel --output-dir wheelhouse
  # if the release was tagged, upload them to PyPI
  - |
    if [[ $TRAVIS_TAG ]]; then
      python -m twine upload --user joerick wheelhouse/*.whl
      # Note: TWINE_PASSWORD is set in Travis settings
    fi
```

There are even ways to make `cibuildwheel` get called in a separate "deploy" stage from one's tests, which is quite nice. That could be added as an alternate example, but then I suppose you support multiple services and you don't want to seem like you favor Travis too much.

I'd be happy to make a PR if you are interested.

YannickJadoul:
Thanks for these suggestions, @SethMMorton. Except for adding Windows, none of these are really high on the list of things to urgently do, but ... yeah, I'm OK with these; it makes things a little bit cleaner indeed :-)

I expect others to quick like the cleanup as well. Do you have time to make a PR? If not, I can do it.
And if you feel like it, feel free to have a look at the other examples in the as well: https://cibuildwheel.readthedocs.io/en/latest/setup/

One thing I'm a bit concerned about: are these actually tested? Are these part of our test suite, somehow, @joerick, @mayeut? And should they?

mayeut:
Seems cleaner, thanks.

[...]

PR #228, opened 2019-11-26
Update travis examples

SethMMorton:
This is intended to close issue #223. 

The example in the README has been validated against https://github.com/joerick/cibuildwheel-autopypi-example/pull/5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,40 +48,40 @@ Usage
 Example setup
 -------------
 
-To build manylinux and macOS wheels on Travis CI, and upload them to PyPI whenever you tag a version, you could use this `.travis.yml`:
+To build manylinux, macOS, and Windows  wheels on Travis CI and upload them to PyPI whenever you tag a version, you could use this `.travis.yml`:
 
 ```yaml
 language: python
 
-matrix:
+jobs:
   include:
     # perform a linux build
-    - sudo: required
-      services:
-        - docker
-      env: PIP=pip
+    - services: docker
     # and a mac build
     - os: osx
-      language: generic
-      env: PIP=pip2
+      language: shell
+    # and a windows build
+    - os: windows
+      language: shell
+      before_install:
+        - choco install python --version 3.8.0
+        - export PATH="/c/Python38:/c/Python38/Scripts:$PATH"
 
 env:
   global:
     - TWINE_USERNAME=joerick
       # Note: TWINE_PASSWORD is set in Travis settings
 
+install:
+  - python -m pip install twine cibuildwheel==1.0.0
+
 script:
-  - $PIP install cibuildwheel==1.0.0
-
   # build the wheels, put them into './wheelhouse'
-  - cibuildwheel --output-dir wheelhouse
+  - python -m cibuildwheel --output-dir wheelhouse
 
+after_success:
   # if the release was tagged, upload them to PyPI
-  - |
-    if [[ $TRAVIS_TAG ]]; then
-      python -m pip install twine
-      python -m twine upload wheelhouse/*.whl
-    fi
+  - if [[ $TRAVIS_TAG ]]; then python -m twine upload wheelhouse/*.whl; fi
 ```
 
 For more information, including how to build on Appveyor, Azure, CircleCI, check out the [documentation](https://cibuildwheel.readthedocs.org).
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ after_success:
   - if [[ $TRAVIS_TAG ]]; then python -m twine upload wheelhouse/*.whl; fi
 ```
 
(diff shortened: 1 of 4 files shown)


SethMMorton:
In addition to updating the README, I also added an `examples/` folder and included two different Travis-CI configuration examples - one is identical to the README, but the other demonstrates a configuration that allows a user to do both testing and deployment through CI.

Before I explain my motivation in adding the `examples/` I want to say that I added this in its own commit, and would be happy to revert the commit if the maintainers are against the content or the idea.

Now, for motivation.

**Motivation A**

All of the examples in the `cibuildwheel` docs show using the CI to *only* build wheels (no tests). When I first encountered `cibuildwheel`, I was much less experienced than I am now at talking to TravisCI and I assumed that the `cibuildwheel` examples were canonical examples and that the proper use was a deployment-only configuration. I'm not saying mine was the correct assumption, but at the time I was a "I just want to release some wheels and not have to think about it too much" user so I just went with it and created a [wheel-deployment-only project](https://github.com/SethMMorton/fastnumbers-wheels). This methodology turned out to be neither convenient nor elegant.

Of course, now I know that the separate project is not needed, but I'm not sure that is 100% obvious to everyone, especially those who only have a cursory knowledge of CI and what is possible. For this reason I think it is important that `cibuildwheel` demonstrate that users can both perform their automated testing as well as deployment in one CI configuration file.

So, one of the examples I added does just that. It uses TravisCI stages to first run automated testing, and *then* deployment. The deployment only occurs if a) the tests pass, b) the branch is "master", c) the commit is a tagged commit and d) the current repo is the "main" repo. [Here](https://travis-ci.com/SethMMorton/cibuildwheel-autopypi-example/builds/138156911) is an example of what the build might look like when the deploy stage happens, and [here](https://travis-ci.com/SethMMorton/cibuildwheel-autopypi-example/builds/138157640) is what it might look like when some condition is met that prevents deployment from occurring. There are ways to construct this that look a bit cleaner that I can also add if desired.

What about `CIBW_TEST_COMMAND`? I considered using this, but I wanted to have my tests be a precondition for wheel building/deployment, not a post-condition. Perhaps I am thinking about it wrong - if so, an example of using `CIBW_TEST_COMMAND` would go a long way :smile:

**Motivation B**

I don't use other CI provides besides TravisCI (well, I've used AppVeyor in the past), so I've only included TravisCI configuration examples. But I see some value in maintainers or users showing what is possible with `cibuildwheel` through examples (it can get tedious to do so in documentation). I think that an `examples/` folder that grows over time would be an asset to this project.

YannickJadoul:
Great! I'd be up for including this :-)

2 more remarks; not necessarily to be fixed in this PR:

- We should not duplicate these example configurations. We have something that can include other files into the docs, right? So it would be good to be able to include some (not per se all) of the files from `examples/` into the docs.
- It would also be nice to not only focus on Travis CI, but also have other examples. If the previous bullet point can be made to work, that would be perfect.

SethMMorton:
> We should not duplicate these example configurations. We have something that can include other files into the docs, right? So it would be good to be able to include some (not per se all) of the files from examples/ into the docs.

This is actually how I wanted to make this commit, but according to [Stack Overflow](https://stackoverflow.com/q/4779582/1399279) it's not possible with straight Markdown. However, I did see that there is some black-magic JavaScript going on in `setup.md` so maybe it is possible. If someone wants to let me know how it is done I can update the PR.

> It would also be nice to not only focus on Travis CI, but also have other examples. If the previous bullet point can be made to work, that would be perfect.

Agreed.  That is my "Motivation B".  I originally wanted to add examples for all CI, but as I mentioned before I have no familiarity with them so did not feel confident in doing so. I was thinking that even if *I* do not include those as examples, someone can easily enough after this gets merged.

YannickJadoul:
(to both points) Great!
Thanks a lot for doing this :-)

joerick:
@YannickJadoul there is a way to include another file -  a mkdocs extension I wrote :) you can include a tag like this:

```
{% 
   includemarkdown "../README.md"
   start="<!--intro-start-->" 
   end="<!--intro-end-->"
%}
```

The start and end parts are optional - I use them to extract part of the readme to the root of the docs page. But I _think_ these should work inside a fenced code block too. 

I don't directly how it would be used in this PR - what were you thinking?

joerick:
@SethMMorton this is a great PR, thank you! love getting docs contributions.

SethMMorton:
@joerick I interpreted it as using the content of the examples in the `examples` directory directly in the `setup.md` file so that the two are always in sync.

joerick:
Cool. We can consider something like that separately. Cheers!

YannickJadoul:
> Cool. We can consider something like that separately. Cheers!

OK, I'll put it on my TODO list :-)
Maybe I'll also make an issue for it, to remind ourselves?

PR #231, opened 2019-11-29
Including CI configurations from examples into docs

YannickJadoul:
Follow up on #228: rather than repeating the samples configurations both as standalone file in `examples/` and in the "Setup" page of the docs, we can (should?) include the example files into the docs with the custom extension we already have.

This is a first request for comments. @joerick, @mayeut, @Czaki, @SethMMorton, any remarks or suggestions? Should I continue doing the same for the other configurations?

Two problems, I've encountered:

1. I needed to add two arguments to the plugin, `before` and `after`, since the `<!-- BEGIN INCLUDE %s %s %s -->` and `<!-- END INCLUDE -->` wouldn't disappear inside a triple-backtick code block.

2. I was thinking to also like to make `.travis.yml` (the line before the actual content) a link, but that's not too straightforward, since the `examples/` directory is outside the `docs/` directory. Maybe this link isn't crucial, though?


SethMMorton:
One thing to keep in mind (and may require a rewording of the `setup.md` page) is that the examples are completely self-contained, e.g. they show using `cibuildwheel` *as well as* showing how to actually perform the deploy, whereas `setup.md` currently only shows how to execute `cibuildwheel`, and then leaves it as an exercise to the user to figure out how to deploy in the manner that best fits them.

If the documentation were simply updated to pull directly from the examples and nothing else, the wording surrounding the examples would not make sense. I think some of the wording should be updated to reflect the change in the example (easy), or the extension should be modified to filter the deployment section (hard).

For the other providers, the documentation seems to imply that users should just download the wheels from the artifacts and not deploy from the CI. If examples are updated to do automated deployments, this should be changed as well.

joerick:
I think this would be more simply remedied by creating a new travis-ci example config in /examples that matched the version currently in docs - something like `examples/travis-ci-minimal.yml`, and importing that.

We could break out the existing docs examples for each CI service using the same pattern (`<provider>-minimal.yml`). 

As for the include, could we do something like this @YannickJadoul ?

~~~markdown
> .travis.yml
```yaml
{% include "../examples/travis-ci-minimal.yml" %}
```
~~~

`include` being a different directive that doesn't add the comments. 

It would require changes to the extension but might be a bit clearer. Happy to help with this if you like!

YannickJadoul:
> I think this would be more simply remedied by creating a new travis-ci example config in /examples that matched the version currently in docs - something like `examples/travis-ci-minimal.yml`, and importing that.

Great, I like that! We should refer to the directory for further examples, though :-)

> As for the include, could we do something like this @YannickJadoul ?

@joerick, you mean making our own second extension/directive within the extension? It somehow feels as if this and `includemarkdown` could just be different uses of the same functionality, or is there something that sets them apart?

Besides that, do you know about this extension: https://github.com/cmacmackin/markdown-include/? Should we continue to roll our own, or could we start depending on this external code?

joerick:
> @joerick, you mean making our own second extension/directive within the extension? It somehow feels as if this and includemarkdown could just be different uses of the same functionality, or is there something that sets them apart?

Yeah, a second directive within the extension. We needed a separate version that doesn't add the comments, which don't work because the code we're including isn't markdown - hence the idea to call it something else.

> Besides that, do you know about this extension: https://github.com/cmacmackin/markdown-include/? Should we continue to roll our own, or could we start depending on this external code?

I can't remember if I've seen this one before! It looks good, but my original use case was to include _parts_ of the README into the docs, which isn't supported by this. The mkdocs extension was pretty easy/fun to write anyway :)

YannickJadoul:
Right, something like this, then? :-)
Feel free to push corrections, if you had something else in mind.

Once that's done, I'll transfer the other examples from `setup.md` to their own file in `examples/`, and refer to these examples from the docs.

joerick:
Looks perfect!

YannickJadoul:
There we go. I'd still like to somehow test the examples, but I'm not sure how to force that :-/

joerick:
Nice work @YannickJadoul !

PR #232, opened 2019-11-30
Test main function with unit tests

Czaki:
Testing if main function properly set `build` function parameters with unit test could speed up finding errors.


Czaki:
@YannickJadoul Any remarks to this?

YannickJadoul:
@Czaki Hmmm, I don't know. I'm a bit on the edge. Isn't this maybe a bit too detailed? Especially the `argtest` part seems very ... strict, e.g. when we rename a parameter (then again; yes, there is this interface that all three platforms need to have, so ... myeah, maybe)? And all of these things are already tested by the integration tests?

Apart from that, it's quite hard to see what you added, because the file was renamed. Just a thought/idea: how possible is it to split up the file into two or three separate ones?

Czaki:
The problem with split is that I also rewrite original test. 
Eg.
```
monkeypatch.setattr(os.path, "exists", lambda x: True)
``` 
instead of creating temporary dir with dummy `setup.py` 

I have no idea how test if some arguments are calculated properly without `argtest`. when change name then it need to be done for all systems (except for manylinux) 

They are partially tested by integration test. I do not remember test checking if environment variable with platform suffix properly overwrite generic one (or not even test if they are used). There is no test for CIBW_REPAIR_WHEEL_COMMAND,  CIBW_BUILD_VERBOSITY. 

Also they take very long time in compare to unittest and cannot be run locally. 

I try to prepare other PR with some split.

YannickJadoul:
Hmmm, I see what you mean. But I'm still quite concerned about having to maintain all of this in the future, somehow. These tests aren't too easy to read and aren't very simple to add when implementing a new feature

@joerick, @mayeut, any thoughts?

joerick:
Always good to have more tests (especially enforcing the stability of the user API), but I agree that these are quite hard to read and reason about, and could be tricky to maintain.  I wonder if some of that could be better test/variable names. I'll see if I can suggest some improvements now

Czaki:
I can try rewrite test to be easier to read. 
But this will introduce more code repetition.

YannickJadoul:
I was also thinking about something like that:

E.g.,
```
@pytest.fixture
def mock_protection_main(monkeypatch):
    monkeypatch.setattr(subprocess, "Popen", not_call_mock)
    monkeypatch.setattr(windows, "urlopen", not_call_mock)
    monkeypatch.setattr(windows, "build", not_call_mock)
    monkeypatch.setattr(linux, "build", not_call_mock)
    monkeypatch.setattr(macos, "build", not_call_mock)
    monkeypatch.setattr(os.path, "exists", lambda x: True)
    monkeypatch.setattr(sys, "argv", ["python", "abcabc"])

with_mock_protection_main = pytest.mark.usefixtures("mock_protection_main")
```

already allows to decorate a function with `@with_mock_protection_main`; maybe there are a few other things that will improve readability and allow the tests themselves to focus on what they're actually testing.

Czaki:
I do this with `@pytest.fixture(autouse=True)` to be sure that protection  is enabled all time.

YannickJadoul:
> I do this with `@pytest.fixture(autouse=True)` to be sure that protection is enabled all time.

I thought about that, but I was wondering if we always want it to be applied?

Czaki:
> > I do this with @pytest.fixture(autouse=True) to be sure that protection is enabled all time.
> I thought about that, but I was wondering if we always want it to be applied?

In this case yes. This is protection of introducing changes on developer machine.

YannickJadoul:
> In this case yes. This is protection of introducing changes on developer machine.

Yes, but what about the other tests? Because `autouse` will enforce this on all tests. E.g., if we want to test `windows.build` later, separately, it will be mocked.

Czaki:
It force only in this module, not in all tests. 
I think that such test should be put in other file.

YannickJadoul:
> It force only in this module, not in all tests.
> I think that such test should be put in other file.

OK, I missed that part. I thought it would be applied to all tests!

YannickJadoul:
OK, I've had some ideas (mostly consisting of moving repeated/complicated code into fixtures) and tried to apply them. Could you have a look at the branch https://github.com/YannickJadoul/cibuildwheel/tree/main_function_tests? Currently it's only this commit: https://github.com/YannickJadoul/cibuildwheel/commit/94c579cbb571e259a745a49395eecea4f348c5d8

I already think it already provides a better overview, and allows the test functions to focus on the actual tests (rather than the `monkeypatch`ing overhead, etc), but I'm of course biased ;-) So I'm happy to get some feedback on what you think, @Czaki, @joerick. If you agree, I'll push the extra commit onto this PR; if not, maybe there's some things you want to salvage :-)

Czaki:
I see what you try to do. 
I put three comments with small problems. They should be easy to fix. 

Main problem is that your code do not work on clean environment. You remove some known variables. I think that much better is to set environment to be clean. This is why I use set`setattr` instead of `setenv` 

I also found that your and my version do not test if build do not use variable for other platform. But maybe it is to much?

YannickJadoul:
> I see what you try to do.
> I put three comments with small problems. They should be easy to fix.

Right, good point, thanks! I hope they're fixed in https://github.com/YannickJadoul/cibuildwheel/commit/dac5734b8badcfd355a87c962dca18d668c78d3f ?

> Main problem is that your code do not work on clean environment. You remove some known variables. I think that much better is to set environment to be clean. This is why I use set`setattr` instead of `setenv`

Yeah, I thought about that, but I tried out `monkeypatch.setenv` for three reasons: 1) I didn't know if just replacing `os.environ` was a good idea (since I know it's normally kind of a special object, not júst a dict), 2) I then saw that `monkeypatch` had these special things dedicated to environment variables (so there's probably a reason they created that?), and 3) it's a bit shorter and more straightfoward like this?

I thought about resetting the whole environment, but then ... I wasn't sure it made a difference. Do you have a scenario in mind where the current approach goes wrong? If really necessary, I think we should still be able to do this in the current approach.

> I also found that your and my version do not test if build do not use variable for other platform. But maybe it is to much?

Pfffff, myeah, I don't know if that's worth testing? Would such a thing really go wrong without messing up the current tests? I mean: if `CIBW_..._MACOS` is used for `CIBW_..._LINUX`, I'd guess one of the two tests, either for Linux or for macOS should fail? :-/

Czaki:
> I thought about resetting the whole environment, but then ... I wasn't sure it made a difference. Do you have a scenario in mind where the current approach goes wrong? If really necessary, I think we should still be able to do this in the current approach.

Someone have `CIBW_BEFORE_BUILD_LINUX` in his environment and then run tests. Tests will fail without proper information. 

Maybe use `delenv` on for given platform variable?

YannickJadoul:
> Someone have `CIBW_BEFORE_BUILD_LINUX` in his environment and then run tests. Tests will fail without proper information.
> 
> Maybe use `delenv` on for given platform variable?

Ah, yes, something like that. That specific case should be easily fixed, but the question is if there's more of these, ofc?

But then .. people shouldn't really set `CIBW_...` variables locally (and if they do, they should remember it?), normally, so I don't find this a huge problem.
I was rather thinking whether there are cases when tests succeed when they shouldn't. That would be a major problem.

Czaki:
> But then .. people shouldn't really set `CIBW_...` variables locally (and if they do, they should remember it?), normally, so I don't find this a huge problem.

On linux, with docker it is easier to debug build than pushing commits and waiting. 

> I was rather thinking whether there are cases when tests succeed when they shouldn't. That would be a major problem. 

In current code no. But if some next environment variables checks are added and more environment variables can change values of variable. 

Eg. changes in name convention of environment variable. But I'm not sure if we should worry this case.

YannickJadoul:
> Hi @Czaki , I've had a good look through and I do think these will be valuable, thank you. But I have some suggestions, mainly that the argtest system you have can be replaced by a [mock](https://docs.python.org/3/library/unittest.mock.html).
> 
> I suggest [pytest-mock](https://pypi.org/project/pytest-mock/) for this, which will inject a `mocker` object into tests and manage clean up.

@joerick Have you had a look at [my branch](https://github.com/YannickJadoul/cibuildwheel/tree/main_function_tests) which adds two commits to @Czaki's work (https://github.com/YannickJadoul/cibuildwheel/commit/94c579cbb571e259a745a49395eecea4f348c5d8 & https://github.com/YannickJadoul/cibuildwheel/commit/dac5734b8badcfd355a87c962dca18d668c78d3f) ? Except for the comment on `os.path.exists` and `mock` (I didn't know whether we wanted to introduce a new development dependency?), those already address all of your comments.

unit_test/main_function_test.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def apply_mock_protection(monkeypatch):
+    monkeypatch.setattr(subprocess, "Popen", not_call_mock)
+    monkeypatch.setattr(windows, "urlopen", not_call_mock)
+    monkeypatch.setattr(windows, "build", not_call_mock)
+    monkeypatch.setattr(linux, "build", not_call_mock)
+    monkeypatch.setattr(macos, "build", not_call_mock)
+    monkeypatch.setattr(os.path, "exists", lambda x: True)
+    monkeypatch.setattr(sys, "argv", ["python", "abcabc"])
```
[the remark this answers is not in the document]
Czaki: This also protect, that when other protection falls then it not point to directory with `setup.py` so next steps will fail (this is reason of using strange name). 

But of course it is more need to workaround protection from `main` function.

unit_test/main_function_test.py
```diff
@@ -0,0 +1,263 @@
...
+def test_platform_chose(system, choose_platform_method, argtest, monkeypatch):
+    apply_mock_protection(monkeypatch)
+    monkeypatch.setattr(globals()[system], "build", argtest)
+    if choose_platform_method == "partameter":
+        monkeypatch.setattr(sys, "argv", sys.argv + ["--platform", system])
+    else:
+        monkeypatch.setattr(os, 'environ', {"CIBW_PLATFORM": system})
+    main()
+    assert argtest.kwargs["project_dir"] == "abcabc"
+
```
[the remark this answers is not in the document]
Czaki: I use this  `monkeypatch.setattr(globals()[system], "build", argtest)` 

instead of this `build_func = mocker.patch('cibuildwheel.{}.build'.format(system))`

Because it allow to not increase project dependency. Bout if you ok with this it may be better.

joerick:
> @joerick Have you had a look at my branch

ah! I had not! sorry, I've trying to fit in reviews and I missed that update before I posted my review. 

Yes, I totally agree, your branch is a huge improvement, and addresses the majority of my comments :). The way you've written the mock makes it much more readable, bit I do still think using a proper mock might be a little better, since Mocks are designed to do exactly this.

YannickJadoul:
Yeah, sorry for the confusion. I didn't want to push, but creating a new PR was also a bit weird, so ... anyway.

@Czaki, what do you think? You still had some concerns, but should I push my branch to this PR, such that we can further discuss the remaining problems (and maybe introduce `mock`) ?

Czaki:
@YannickJadoul Push it.

YannickJadoul:
Thanks; done!

Azure Pipelines doesn't seem to like this push though...

Czaki:
It looks like problem with checkout. So I force to rerun.

Czaki:
@YannickJadoul Could you restart Travis. Against problem with download get_pip.py 

Maybe add some code fix which will retry download before fail?

YannickJadoul:
> Maybe add some code fix which will retry download before fail?

Done. I don't know; most of the time it goes well. Do we have any idea why and when it fails?

Czaki:
From my observation it is some network error. See #237

unit_test/main_options_test.py
```diff
@@ -0,0 +1,202 @@
...
+    ('i686', 'custom_image', None),
+])
+def test_manylinux_images(architecture, image, full_image, platform, intercepted_build_args, monkeypatch):
+    if image is not None:
+        monkeypatch.setenv('CIBW_MANYLINUX_' + architecture.upper() + '_IMAGE', image)
+
+    main()
+
+    if platform == 'linux':
+        assert intercepted_build_args.kwargs['manylinux_images'][architecture] == full_image or image
```
joerick: Just checking this logic condition... It looks to me like `(intercepted_build_args.kwargs['manylinux_images'][architecture] == full_image) or (image is not None)`, is that the intention? If so, it could be written as an `if` more clearly, I think

unit_test/main_options_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def get_default_repair_command(platform):
+    if platform == 'linux':
+        return 'auditwheel repair -w {dest_dir} {wheel}'
+    elif platform == 'macos':
+        return 'delocate-listdeps {wheel} && delocate-wheel -w {dest_dir} {wheel}'
+    elif platform == 'windows':
+        return ''
+
```
joerick:
```suggestion
    else:
        raise ValueError('unknown platform', platform)

```

unit_test/main_options_test.py
```diff
@@ -0,0 +1,202 @@
...
+        monkeypatch.setenv('CIBW_ENVIRONMENT_' + platform.upper(), env_string)
+        monkeypatch.setenv('CIBW_ENVIRONMENT', 'overwritten')
+    else:
+        monkeypatch.setenv('CIBW_ENVIRONMENT', env_string)
+
+    main()
+
+    intercepted_environment = intercepted_build_args.kwargs['environment']
+    assert isinstance(intercepted_environment, ParsedEnvironment)
+    assert intercepted_environment.as_dictionary({}) == environment
```
joerick:
```suggestion
    assert intercepted_environment.as_dictionary(prev_environment={}) == environment
```

unit_test/main_options_test.py
```diff
@@ -0,0 +1,202 @@
...
+    if test_command is not None:
+        if platform_specific:
+            monkeypatch.setenv('CIBW_TEST_COMMAND_' + platform.upper(), test_command)
+            monkeypatch.setenv('CIBW_TEST_COMMAND', 'overwritten')
+        else:
+            monkeypatch.setenv('CIBW_TEST_COMMAND', test_command)
+
+    main()
+
+    assert intercepted_build_args.kwargs['test_command'] == test_command or ''
```
joerick: This condition might be wonky, also, I think it's equivalent to 

    assert (intercepted_build_args.kwargs['test_command'] == test_command) or ''

which is equivalent to:

    assert intercepted_build_args.kwargs['test_command'] == test_command

the logical operators e.g. `and` `or` are evaluated after the comparison ones... e.g. 

```python
>>> if 'asd' == 1 or 2:
...    print('wat')
... 
wat
```

Maybe this should be 
```python
    assert intercepted_build_args.kwargs['test_command'] == (test_command or '')
```

unit_test/main_options_test.py
```diff
@@ -0,0 +1,202 @@
...
+    if before_build is not None:
+        if platform_specific:
+            monkeypatch.setenv('CIBW_BEFORE_BUILD_' + platform.upper(), before_build)
+            monkeypatch.setenv('CIBW_BEFORE_BUILD', 'overwritten')
+        else:
+            monkeypatch.setenv('CIBW_BEFORE_BUILD', before_build)
+
+    main()
+
+    assert intercepted_build_args.kwargs['before_build'] == before_build or ''
```
joerick: Same here - `(before_build or '')`

YannickJadoul:
> Few minor comments, but this is nearly there!

Thanks! I guess these are mostly sloppy things from my commits, not from @Czaki's anymore.
I'm currently writing docs for #185, but I'll give this a look after :-)

unit_test/main_options_test.py
YannickJadoul: OK, yes, something went wrong, here. I think the full check should be:

```
if image is not None:
    assert intercepted_build_args.kwargs['manylinux_images'][architecture] == full_image
else:
    assert intercepted_build_args.kwargs['manylinux_images'][architecture] == image
```

unit_test/main_options_test.py
YannickJadoul: OK, yes, indeed. Weird, I don't know how that from my head onto the screen! Even more so because `test_command` will be `None` if not set, rather than `''`. Thanks for catching that!

YannickJadoul:
Fixed your remarks, @joerick.

There's still a few unresolved discussions, though:
1. Do we still want to use `mock` instead of `ArgsInterceptor`? It's slightly better hidden away now, but still possible to replace it with `mock`.
2. What about the `monkeypatch` of `os.path.exists` that you didn't like, @joerick?
3. What about the fact that we don't fully clear the environment, but only set the variables we need with `monkeypatch.setenv`? See @Czaki's remarks above.

joerick:
Thanks for the summary @YannickJadoul , I'd forgotten about them!

> Do we still want to use mock instead of ArgsInterceptor? It's slightly better hidden away now, but still possible to replace it with mock.

I'm happy either way, we can leave this as-is now and move to using Mocks down the line if/when we need more features from them.

> What about the monkeypatch of os.path.exists that you didn't like, @joerick?

What if we move them out of mock_protection, like...

```python
@pytest.fixture(autouse=True)
def mock_protection(monkeypatch):
    '''
    ensure that a unit test will never actually run a cibuildwheel 'build'
     function, which shouldn't be run on a developer's machine
    '''
    def fail_on_call(*args, **kwargs):
        raise RuntimeError("This should never be called")

    monkeypatch.setattr(subprocess, 'Popen', fail_on_call)
    monkeypatch.setattr(windows, 'urlopen', fail_on_call)
    monkeypatch.setattr(windows, 'build', fail_on_call)
    monkeypatch.setattr(linux, 'build', fail_on_call)
    monkeypatch.setattr(macos, 'build', fail_on_call)

@pytest.fixture(autouse=True)
def fake_project_dir(monkeypatch):
    '''
    Monkey-patch enough for the main() function to run
    '''
    real_os_path_exists = os.path.exists
    def mock_os_path_exists(path):
        if path == os.path.join(MOCK_PROJECT_DIR, 'setup.py'):
            return True
        else:
            return real_os_path_exists(path)
        
    monkeypatch.setattr(os.path, 'exists', mock_os_path_exists)
    monkeypatch.setattr(sys, 'argv', ['cibuildwheel', MOCK_PROJECT_DIR])
```

> What about the fact that we don't fully clear the environment, but only set the variables we need with monkeypatch.setenv? See @Czaki's remarks above.

I see good arguments on both sides! If forced to choose, I slightly prefer this version with `monkeypatch.setenv`, because it reads more clearly in the code.

YannickJadoul:
> I'm happy either way, we can leave this as-is now and move to using Mocks down the line if/when we need more features from them.

Agreed!
 
> > What about the monkeypatch of os.path.exists that you didn't like, @joerick?
> 
> What if we move them out of mock_protection, like...
> 
> ```python
> @pytest.fixture(autouse=True)
> def mock_protection(monkeypatch):
>     '''
>     ensure that a unit test will never actually run a cibuildwheel 'build'
>      function, which shouldn't be run on a developer's machine
>     '''
>     def fail_on_call(*args, **kwargs):
>         raise RuntimeError("This should never be called")
> 
>     monkeypatch.setattr(subprocess, 'Popen', fail_on_call)
>     monkeypatch.setattr(windows, 'urlopen', fail_on_call)
>     monkeypatch.setattr(windows, 'build', fail_on_call)
>     monkeypatch.setattr(linux, 'build', fail_on_call)
>     monkeypatch.setattr(macos, 'build', fail_on_call)
> 
> @pytest.fixture(autouse=True)
> def fake_project_dir(monkeypatch):
>     '''
>     Monkey-patch enough for the main() function to run
>     '''
>     real_os_path_exists = os.path.exists
>     def mock_os_path_exists(path):
>         if path == os.path.join(MOCK_PROJECT_DIR, 'setup.py'):
>             return True
>         else:
>             return real_os_path_exists(path)
>         
>     monkeypatch.setattr(os.path, 'exists', mock_os_path_exists)
>     monkeypatch.setattr(sys, 'argv', ['cibuildwheel', MOCK_PROJECT_DIR])
> ```

I could be happy with this as well. What do you think, @Czaki?

> I see good arguments on both sides! If forced to choose, I slightly prefer this version with `monkeypatch.setenv`, because it reads more clearly in the code.

Any input here, @Czaki? You're the one that wanted to fully clear the environment? What if ... we make sure all `CIBW_...` variables and other variables we depend on get cleared in the some fixture, but don't clear all the others? After all, if tests break by depending on some environment variables set, that's then a problem that's maybe good to detect. We can always change this later, if it turns out we forgot about something?

Czaki:
I think that in this moment all main developers of `cibuildwheel` know possible problems without cleared environment, so if anyone report any problem with this test set, then it will be fast solved. 

So I think that `setenv` is good option. Without manual cleaning.

YannickJadoul:
> So I think that `setenv` is good option. Without manual cleaning.

OK, so adding something to the `mock_protection` fixture is not necessary, you think? I was thinking something like

```python
ENVIRONMENT_VARIABLES = [
    'CIBW_PLATFORM',
    'CIBW_ENVIRONMENT',
    'CIBW_ENVIRONMENT_LINUX',
    'CIBW_ENVIRONMENT_MACOS',
    'CIBW_ENVIRONMENT_WINDOWS',
    # ...
]


# ...

@pytest.fixture
def mock_protection():
    # ...

    for var in ENVIRONMENT_VARIABLES:
        monkeypatch.delenv(var, raising=False)
```

YannickJadoul:
> What if we move them out of mock_protection, like...

Done. Unless anyone likes the proposed series of `monkeypatch.delenv` above, this is ready to merge, as far as I'm concerned.

YannickJadoul:
Hurray! Thanks for all the work, @Czaki!

PR #233, opened 2019-12-01
Add example of usage more than one environment variable

Czaki:
In the documentation of `CIBW_ENVIRONMENT` there is no example with more than one environment variable.

There are some problem witch `chocolatey` service

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -148,6 +148,8 @@ You must set this variable to pass variables to Linux builds (since they execute
 
 You can use `$PATH` syntax to insert other variables, or the `$(pwd)` syntax to insert the output of other shell commands.
 
+To specify more than one environment variable separate it by space. 
+
 Platform-specific variants also available:<br/>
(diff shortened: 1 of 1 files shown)


joerick:
Thanks @czaki!

YannickJadoul:
@joerick, @Czaki This was already in there.

> "A space-separated list of environment variables to set during the build. Bash syntax should be used, even on Windows."

I like the extra example as demonstration, but maybe we should still remove the superfluous line of documentation?

Czaki:
@YannickJadoul @joerick My bad. But example is missing. What now?

YannickJadoul:
I don't know; we could just keep it like this, of course. It's not wrong. I just noticed.
Or if you prefer I can push a correction removing the line but keeping the example?

YannickJadoul:
OK, let's just keep it then, unless you prefer to remove the line? Rephrased in https://github.com/joerick/cibuildwheel/commit/795413e5c0aae49e8c727856e22a693495abce58.

joerick:
I think it's okay to duplicate, since it's easy to miss 'space-separated' at the top. Thanks for fixing the grammar @YannickJadoul !

PR #234, opened 2019-12-03
change directory to not conflict with setup.cfg of package

Czaki:
During development of my package I meet the problem, that get-pip.py script fails if it is run from directory with `setup.cfg` in which version if defined with `attr` and to low version of setuptools (not properly support of `attr`). 
```
version = attr: PartSeg.__version__
```
Because this script is called before `before_build` is executed, then I found that solution is to change directory. 
https://dev.azure.com/PartSeg/PartSeg/_build/results?buildId=474&view=results


YannickJadoul:
Huh, that seems weird? Why is `get-pip.py` accessing `setup.cfg`?

Czaki:
For me it looks like `setuptools` look for `setup.cfg` for locations in path and `get-pip.py` and
unpack location on begin of path, not replace it. 

line: 195
```
sys.path.insert(0, pip_zip)
```

But this patch is result of experiments, not supported by knowledge.

joerick:
Thanks Czaki

PR #236, opened 2019-12-10
Test main function with unit tests with files separation

Czaki:
@YannickJadoul  I separate and left old test in `platform_test.py` file.

### DIFF
--- a/unit_test/main_function_test.py
+++ b/unit_test/main_function_test.py
@@ -1,0 +1,212 @@
+import sys
+import os
+import subprocess
+
+import pytest
+
+from main_function_test_utils import argtest, apply_mock_protection
+from cibuildwheel.__main__ import main
+from cibuildwheel import windows, linux, macos
+from cibuildwheel.util import BuildSelector
+from cibuildwheel.environment import ParsedEnvironment
+
+
+@pytest.mark.parametrize("system", ["macos", "linux", "windows"])
+@pytest.mark.parametrize("choose_platform_method", ["partameter", "environment"])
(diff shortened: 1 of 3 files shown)


YannickJadoul:
@Czaki Thanks; good to compare with #232!

There's still more noise than I expected, though, so maybe #232 is equally good? I find it hard to judge, actually; sorry :-/

unit_test/main_function_test.py
```diff
@@ -0,0 +1,212 @@
...
+def test_build_selector(system, build, build_set, skip, skip_set, monkeypatch, argtest):
+    apply_mock_protection(monkeypatch)
+    monkeypatch.setattr(globals()[system], "build", argtest)
+    env = {"CIBW_PLATFORM": system, "CIBW_BUILD": build, "CIBW_SKIP": skip}
+    monkeypatch.setattr(os, 'environ', env)
+    main()
+    assert isinstance(argtest.kwargs["build_selector"], BuildSelector)
+    selector = argtest.kwargs["build_selector"]
+    selected = set([x for x in ["aaaa", "abcd", "abab", "bcde"] if selector(x)])
+    assert selected == build_set.difference(skip_set)
```
YannickJadoul: We already have `build_selector_test.py` for this, for example, and integration tests to see if `CIBW_BUILD`/`CIBW_SKIP` work. Do we need this here again?

Czaki: I'm not sure. This check it faster, but of course there is redundancy.

YannickJadoul:
Closing in favour of #232

PR #237, opened 2019-12-17
Repeat get-pip download on fail

Czaki:
Repeat try of download `get-pip.py` if some error occurrences.  From my observation sometimes there is network error.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -1,5 +1,4 @@
 from __future__ import print_function
-from time import sleep
 import tempfile
 import os, subprocess, shlex, sys, shutil
 from collections import namedtuple
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -52,7 +52,7 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
 
     # get latest pip once and for all
     
-    call(['curl', '-L', '-o', '--retry', '3', '--retry-delay', '3', get_pip_script, get_pip_url])
+    call(['curl', '-L', '--retry', '3', '--retry-delay', '3', '-o', get_pip_script, get_pip_url])
 
     for config in python_configurations:
         # if this version of python isn't installed, get it from python.org and install
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -51,14 +51,8 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
         return subprocess.check_call(args, env=env, cwd=cwd, shell=shell)
 
     # get latest pip once and for all
-    for _ in range(3):
-        try:
-            call(['curl', '-L', '-o', get_pip_script, get_pip_url])
-        except subprocess.CalledProcessError:
-            sleep(3)
-            continue
-        break
-    assert os.path.exists(get_pip_script)
+    
+    call(['curl', '-L', '-o', '--retry', '3', '--retry-delay', '3', get_pip_script, get_pip_url])
 
     for config in python_configurations:
         # if this version of python isn't installed, get it from python.org and install
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from time import sleep
 import tempfile
 import os, subprocess, shlex, sys, shutil
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


YannickJadoul:
2 remarks:
- 10 tries seems like a lot. What about 3? Maaaybe 5, but ... if it doesn't work after 3, something else is wrong.
- I'd make this loop part of `download`. If it can't hurt for `get-pip.py`, then it shouldn't hurt for other things? And it will keep the main code more readable.

Czaki:
> 10 tries seems like a lot. What about 3? Maaaybe 5, but ... if it doesn't work after 3, something else is wrong.
the network connection problem take long time. If decrease download repetition then maybe increase sleep value? What do you think? 

>  I'd make this loop part of download. If it can't hurt for get-pip.py, then it shouldn't hurt for other things? And it will keep the main code more readable.

Done. This change also give better protection for other downloads, but I do not see any problems with other downloads.

YannickJadoul:
> Done. This change also give better protection for other downloads, but I do not see any problems with other downloads.

Yeah, it's not about protecting them that's most important; it's keeping the main code in `build` readable and clear. Thanks :-)

> I do not understand problems with windows on travis.
> https://travis-ci.org/joerick/cibuildwheel/jobs/625959545?utm_medium=notification&utm_source=github_status

Me neither. I restarted, but if this doesn't work, maybe you should push a temporary commit that shows the output of the log file that's mentioned when the build fails.

YannickJadoul:
> the network connection problem take long time. If decrease download repetition then maybe increase sleep value? What do you think?

Travis is currently timing out because it doesn't get output in time. Not sure if it's related to too many/too long tries, but let's not exaggerate here. 3 seconds is ages for a computer, and if it doesn't work after 3 tries, something else is probably wrong?

I am actually still not convinced retrying will work. When the builds fail currently, this might just be because something else goes wrong in the Travis container (a misconfiguration, for example). Trying again without restarting the full build isn't going to help, in such a case.

Czaki:
> Travis is currently timing out because it doesn't get output in time. Not sure if it's related to too many/too long tries, but let's not exaggerate here. 3 seconds is ages for a computer, and if it doesn't work after 3 tries, something else is probably wrong?

Ok. If this problem still occur then it will be easy to increase repetition number. 

> I am actually still not convinced retrying will work. When the builds fail currently, this might just be because something else goes wrong in the Travis container (a misconfiguration, for example). Trying again without restarting the full build isn't going to help, in such a case.

This Wat I observe it is that sometimes (rare, but happen) server with `get-pip.py` do not resposne properly. 

Because services like Azure Pipelines do not allow to restart selected works its worth to me to minimize probability of fail from reason independent from my code.

YannickJadoul:
Right. Tests are succeeding again, so let's see how things go on, if this PR will indeed solve most of the failures.

> Because services like Azure Pipelines do not allow to restart selected works its worth to me to minimize probability of fail from reason independent from my code.

Azure does allow to do so, though. CircleCI and AppVeyor as well, but I don't have the rights to restart jobs there.

Czaki:
I have few projects on Azure Pipelines and in some I can restart failed jobs. In some i need to restart whole pipeline..

cibuildwheel/macos.py
```diff
@@ -50,7 +51,14 @@ def call(args, env=None, cwd=None, shell=False):
         return subprocess.check_call(args, env=env, cwd=cwd, shell=shell)
 
     # get latest pip once and for all
-    call(['curl', '-L', '-o', get_pip_script, get_pip_url])
```
mayeut: `curl` has a retry option. This would be simpler to use it I guess:

```
Matt$ curl --help | grep retry
     --retry NUM   Retry request NUM times if transient problems occur
     --retry-connrefused  Retry on connection refused (use with --retry)
     --retry-delay SECONDS  Wait SECONDS between retries
     --retry-max-time SECONDS  Retry only within this period
```

cibuildwheel/macos.py
```diff
@@ -50,7 +51,8 @@ def call(args, env=None, cwd=None, shell=False):
...
+    
+    call(['curl', '-L', '-o', '--retry', '3', '--retry-delay', '3', get_pip_script, get_pip_url])
```
mayeut:
```suggestion
    call(['curl', '-L', '--retry', '3', '--retry-delay', '3', '-o', get_pip_script, get_pip_url])
```

cibuildwheel/windows.py
```diff
@@ -62,17 +63,29 @@ def simple_shell(args, env=None, cwd=None):
...
-        response = urlopen(url)
+        for _ in range(3):
+            try:
+                response = urlopen(url)
+            except:
+                sleep(3)
+                continue
+            break
+        else:
+            print("Download from url " + url + "failed", file=sys.stderr)
```
joerick: Could we log the final exception here, too? Could be useful for debugging if the problem isn't a transient one

Czaki: I change behaviour to re raise last exception.

cibuildwheel/macos.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from time import sleep
```
mayeut: You can remove this line now that all is handled by `curl`
```suggestion
```

PR #242, opened 2020-01-07
Add CIBW_BEFORE_TEST to customize test environment

Czaki:
I need this change to overwrite some environment variable (use `clang` compiler instead of `gcc` on MacOS) 
It allow also use much more elegant solution than
```yaml
CIBW_TEST_REQUIRES: "scikit_build; pip install -r requirements_test.txt"
```
to perform multi step requirements installation (in this case it is needed because there is no wheel for python 3.8).

I think that it is better solution than #240 which is my first try to solve this problem.


YannickJadoul:
Restarted the test (cfr. https://github.com/joerick/cibuildwheel/issues/243#issuecomment-572994895).

I still need to think about this and review this, but I don't have time, right now. I'll get a look somewhere next week or so.

Czaki:
One additional remark. From PR #164 test are executed in separated environment. So  `CIBW_BEFORE_BUILD` does not allow to modify test environment.

joerick:
As noted in #243 , I'm +1 on this, in principle. Is this ready for review now @Czaki ?

Czaki:
@joerick  Yes

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
 CIBW_TEST_EXTRAS: test,qt
 ```
 
+### `CIBW_BEFORE_TEST` {: #before-test}
+> Execute a shell command preparing test environment
```
```suggestion
> Execute a shell command before testing each wheel
```

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+
+A shell command to run after setup test environment. This option allows you to run a command in **each** Python environment before the `pip wheel` command. This is useful if you need to install non pip package, change values of environment variables
```
```suggestion
A shell command to run in **each** test virtual environment, before your wheel is installed and tested. This is useful if you need to install non pip package, change values of environment variables
```

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+or perform multi step pip installation (like install `scikit-build` or `cython` before install test package)
```
joerick: These reasons are a little curious, actually.

- Non-pip packages don't care about the virtualenv, so can be installed in BEFORE_BUILD. 
- Changing environment variables won't work, since the command is run in a subshell
- Maybe I'm missing something, but I can't see the need to install anything _before_ installing the wheel? Wheel installation is just an unzip and move, no code within the package is run. So why is there a need to install dependencies before that? It seems that TEST_REQUIRES would be sufficient for these cases.

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+
+If dependencies are required to build your wheel (for example if you include a header from a Python module), set this to `pip install .`, and the dependencies will be installed automatically by pip. However, this means your package will be built twice - if your package takes a long time to build, you might wish to manually list the dependencies here instead.
```
```suggestion
```

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+#### Examples
+```yaml
+# install packages needed to build test dependencies
+CIBW_BEFORE_TEST: pip install cmake scikit-build
+
+# install test dependencies with overwritten environment variables. 
+CIBW_BEFORE_TEST: CC=gcc CXX=g++ pip install -r requirements.txt
+
+# chain commands using &&
+CIBW_BEFORE_TEST: yum install -y libffi-dev && pip install .
```
joerick: `pip install .` would be an anti-pattern, since it would build and install from code, rather than from the wheel.

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+
+On Linux and macOS, the command is run in a shell, so you can write things like `cmd1 && cmd2`.
+
+Platform-specific variants also available:<br/>
+ `CIBW_BEFORE_TEST_MACOS` | `CIBW_BEFORE_TEST_WINDOWS` | `CIBW_BEFORE_TEST_LINUX`
+
+#### Examples
+```yaml
+# install packages needed to build test dependencies
+CIBW_BEFORE_TEST: pip install cmake scikit-build
```
joerick: See above, I'm not sure why these aren't TEST_REQUIRES

docs/options.md
```diff
@@ -327,6 +327,32 @@ Platform-specific variants also available:<br/>
...
+
+# install test dependencies with overwritten environment variables. 
+CIBW_BEFORE_TEST: CC=gcc CXX=g++ pip install -r requirements.txt
```
joerick: Curious - why would this be necessary?

test/10_before_test/cibuildwheel_test.py
```diff
@@ -0,0 +1,22 @@
...
+import utils
+
+def test():
+    project_dir = os.path.dirname(__file__)
+
+    # build the wheels
+    actual_wheels = utils.cibuildwheel_run(project_dir, add_env={
+        # write python version information to a temporary file, this is
+        # checked in setup.py
+        'CIBW_BEFORE_TEST': '''python -c "import sys; open('/tmp/pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('/tmp/pythonexecutable.txt', 'w').write(sys.executable)"''',
```
joerick: The `pythonversion.txt` stuff is overkill for this test, I think. Good to check we're in the right virtualenv, though. Maybe we could do it just with `pythonexecutable.txt`?

docs/options.md
Czaki: Example situation:

 - PyQt5 from 5.14.0 has wheel with `manylinux2014` platform tag. Earlier are published with `manylinux1`. `manylinux2014` needs pip  in version 19.3. Example solution with `BEFORE_BUILD` is to manually install `PyQt5==5.13.2`
- [cmake](https://pypi.org/project/cmake/) package does not provide wheels for x86 python 3.8. So only option is to build it from source. But cmake package needs `scikit-build` for build and this dependency needs to be importable (installed in previous transatction). And package [blosc](https://pypi.org/project/blosc/) does not provide wheel and need cmake package to be installed. And tests need `blosc` package. 
- i do not have given example but it is possible to have similar story with `Cython` 

Also non pip package can care about virtualenv (location of python executable). Here example: https://github.com/LLNL/zfp

And change environment variable I think about `ENV1=1 ENV2=7 command`

docs/options.md
Czaki: Agree. I copy it from `CIBW_BEFORE_BUILD` documentation. Whole example is bad. Installig things system side during testis also bad idea. Maybe `pip install scikit-build && pip install cmake`?

docs/options.md
Czaki: As I mentioned before. Some dependencies need to be importable in `setup.py`. And still some authors of packages does not provide wheels or it takes over month since release of python 3.8 to wheel for this python. And such things allow maintainer to have full test case without waiting on other persons. 

```
File "/tmp/pip-install-f9rmc9hk/cmake/setup.py", line 7, in <module>
          from skbuild import setup
      ModuleNotFoundError: No module named 'skbuild'
```

docs/options.md
Czaki: My case is that i want to build package with OMP support for MacOS. MacOS clang does not support this. So I need to use brew g++ for this. But one of test dependencies need to be compiled with system `clang`. Otherwise I got such message:

```
python(53258,0x7fff9481d380) malloc: *** error for object 0x10f526c00: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

Other case which I can imagine is that my package is build against newest version of some library. And test dependency is build against older (system one). And I need to change `LD_LIBRARY_PATH` etc to avoid collision.

test/10_before_test/cibuildwheel_test.py
Czaki: Why it is overkill here and not overkill in `CIBW_BEFORE_BUILD` test?

docs/options.md
YannickJadoul: This confusion makes me wonder; wouldn't a different example be better for the docs, then?

docs/options.md
YannickJadoul: @joerick I know `setup_requires` doesn't always work, for some reason. So that would be good.

But indeed, this might not be the perfect example to put in the docs, either?

cibuildwheel/windows.py
```diff
@@ -58,7 +58,7 @@ def get_python_configurations(build_selector):
     return python_configurations
 
 
-def build(project_dir, output_dir, test_command, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment):
+def build(project_dir, output_dir, test_command, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment, before_test):
```
YannickJadoul: Ugh, these arguments are becoming a mess.

At any rate, I don't think `before_test` should be alone at the end of the argument list. Can you put it maybe after `test_command ` or so?

README.md
```diff
@@ -100,7 +100,8 @@ Options
 | **Testing** | [`CIBW_TEST_COMMAND`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-command)  | Execute a shell command to test each built wheel |
 |   | [`CIBW_TEST_REQUIRES`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-requires)  | Install Python dependencies before running the tests |
 |   | [`CIBW_TEST_EXTRAS`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-extras)  | Install your wheel for testing using extras_require |
-| **Other** | [`CIBW_BUILD_VERBOSITY`](https://cibuildwheel.readthedocs.io/en/stable/options/#build-verbosity)  | Increase/decrease the output of pip wheel |
+|   | [`CIBW_BEFORE_TEST`](https://cibuildwheel.readthedocs.io/en/stable/options/#before-test)  | Execute shell command to prepare test environment |
```
YannickJadoul: Let's reorder this. This line should probably be the line after `CIBW_TEST_COMMAND`?

test/10_before_test/cibuildwheel_test.py
```diff
@@ -0,0 +1,22 @@
+import os
+import utils
+
+def test():
```
YannickJadoul: What about not adding yet another extra test folder, but just add it to `02_test` or `03_before_build`? Now that we're using `pytest` and are able to have multiple test functions, we can group the together much better?

@joerick ?

docs/options.md
Czaki: What do you think about 
`CFLAGS="-I ... -L ..." pip install package name` 
for custom library location?

test/10_before_test/cibuildwheel_test.py
Czaki: Maybe yes. This is global question about test granulation. @joerick What you think about this?

cibuildwheel/windows.py
Czaki: Agree. But there is question if use `kwargs` or reorder arguments? @joerick ?

docs/options.md
YannickJadoul: Huh, maybe. But .. that's again something where you'd ask the question where a user would ask why this is not `CIBW_TEST_ENVIRONMENT`...

cibuildwheel/windows.py
YannickJadoul: I'd say just reorder now such that it make +- logical sense. Refactoring this (maybe `**kwargs`, but I don't know what the best solution is) is for another time/PR.

docs/options.md
Czaki: Totally different example. Install non pip package like: https://github.com/LLNL/zfp

For this...  (original example)
MacOS system clang does not support OpenMP. So if package needs OpenMP support then there is need to compile it with gcc/g++ (ex from brew). 

For second. I meet problem when test for one package fail on Azure Pipeline but not fail on my computer. And this fail was connected with package used in test and build from source. And for me possibility to control compilation of single package is big profit.

test/10_before_test/cibuildwheel_test.py
joerick: It would be nice to unify a little, but `10_before_test` is different than the others because a test within the project checks the existence of a file, rather than `setup.py`.

test/10_before_test/cibuildwheel_test.py
joerick: I think the `pythonversion.txt` check in BEFORE_BUILD is there through an abundance of caution (paranoia?) about which version of Python is running. Maybe it's not necessary there either. But we certainly don't need it here.

test/10_before_test/setup.py
```diff
@@ -0,0 +1,19 @@
+from setuptools import setup, Extension
+import sys, os
+
+# assert that the Python version as written to pythonversion.txt in the CIBW_BEFORE_BUILD step
+# is the same one as is currently running.
```
joerick: this comment and the one below are wrong. remove please

test/10_before_test/setup.py
```diff
@@ -0,0 +1,19 @@
...
+version_file = 'c:\\pythonversion.txt' if sys.platform == 'win32' else '/tmp/pythonversion.txt'
+if os.path.exists(version_file):
+    os.remove(version_file)
+
+# check that the executable also was written
+executable_file = 'c:\\pythonexecutable.txt' if sys.platform == 'win32' else '/tmp/pythonexecutable.txt'
+if os.path.exists(executable_file):
+    os.remove(executable_file)
```
joerick: Just thinking- I'm not sure we need these delete steps here in `setup.py`..?

unit_test/main_options_test.py
```diff
@@ -204,3 +204,18 @@ def test_build_verbosity(build_verbosity, platform_specific, platform, intercept
...
+    if before_test is not None:
+        if platform_specific:
+            monkeypatch.setenv('CIBW_BEFORE_TEST_' + platform.upper(), before_test)
+            monkeypatch.setenv('CIBW_BEFORE_TEST', 'overwritten')
+        else:
+            monkeypatch.setenv('CIBW_BEFORE_TEST', before_test)
+
+    main()
+
+    assert intercepted_build_args.kwargs['before_test'] == before_test
```
YannickJadoul: While we're at the level of details. Maybe an extra newline, here?
And ... I don't know; do the tests follow the order of arguments in `build`, or not? If so, I'd say move this up a bit; otherwise not necessary.

test/10_before_test/setup.py
YannickJadoul: I agree. The reason these were in `CIBW_BEFORE_BUILD` is probably because only `setup.py` gets run here. Now there's a test suite that's run, so things should be checked there.

test/10_before_test/cibuildwheel_test.py
YannickJadoul: Right, I see now :-)

test/10_before_test/cibuildwheel_test.py
```diff
@@ -0,0 +1,22 @@
...
+    project_dir = os.path.dirname(__file__)
+
+    # build the wheels
+    actual_wheels = utils.cibuildwheel_run(project_dir, add_env={
+        # write python version information to a temporary file, this is
+        # checked in setup.py
+        'CIBW_BEFORE_TEST': '''python -c "import sys; open('/tmp/pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('/tmp/pythonexecutable.txt', 'w').write(sys.executable)"''',
+        'CIBW_BEFORE_TEST_WINDOWS': '''python -c "import sys; open('c:\\pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('c:\\pythonexecutable.txt', 'w').write(sys.executable)"''',
+        'CIBW_TEST_REQUIRES': 'nose',
```
YannickJadoul: Just curious: why depend on `nose`? Why not just `pytest` (since we're already using that) of Python's own unit tests?

test/10_before_test/test/spam_test.py
```diff
@@ -0,0 +1,24 @@
+import sys, os
+from unittest import TestCase
+
+
+class TestBeforeTest(TestCase):
+    def test_version(self):
+        # assert that the Python version as written to pythonversion.txt in the CIBW_BEFORE_BUILD step
```
YannickJadoul: `CIBW_BEFORE_TEST`

test/10_before_test/cibuildwheel_test.py
Czaki: @joerick ? I copy it from `CIBW_BEFORE_BUILD`

docs/options.md
Czaki: yes

test/10_before_test/setup.py
Czaki: In macos build cibuildwheel use symlink for choosing python versions. Then all python version has same executable path. So we need to delete this files to be sure that `CIBW_BEFORE_BUILD` is executed properly.

unit_test/main_options_test.py
Czaki: done

test/10_before_test/cibuildwheel_test.py
joerick: At this point, historical reasons, it could equally be pytest. But there's something quite nice about using something that requires installation, that wouldn't be available in PATH otherwise.

test/10_before_test/setup.py
joerick: This is an *excellent* point about `pythonversion.txt`. I previously said that `pythonversion.txt` should be removed in this test - I was wrong, for the symlink reason above, we should keep that. Apologies @Czaki!

However, I still think the `setup.py` 'remove files' step is unnecessary. Each test will overwrite the previous version.

test/10_before_test/cibuildwheel_test.py
YannickJadoul: Ah, yes, I overlooked that in this case, it's pretty good to have an external dependency, as we can also test `CIBW_TEST_REQUIRES` that way :-)

It's just that apparently `nose` is not getting maintenance anymore.

test/10_before_test/setup.py
Czaki: I revert this change.

test/10_before_test/cibuildwheel_test.py
joerick: good point! i didn't realise. Might be worth changing to `nose2` at some point!

Czaki: So what you suggest? nose2? https://docs.nose2.io/en/latest/

Czaki:
I'm a little lost with comments to documentation. I try to apply all, but I'm not sure if I properly understand all suggestions. Could You check?

test/10_before_test/cibuildwheel_test.py
YannickJadoul: Maybe it's not urgent, though? Maybe we shouldn't bother with it in this PR?

Czaki: ok

YannickJadoul: (Unless you or @joerick think it's a small thing and want to, ofc; I don't know. It just sounded as a future thing. And for our purposes, `nose` is still good enough.)

joerick:
Agreed

joerick:
Would be nice to get this into the next release - @Czaki could you merge/rebase please? @YannickJadoul did you have any further comments before it goes in?

README.md
```diff
@@ -98,9 +98,10 @@ Options
...
 | **Testing** | [`CIBW_TEST_COMMAND`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-command)  | Execute a shell command to test each built wheel |
+|   | [`CIBW_BEFORE_TEST`](https://cibuildwheel.readthedocs.io/en/stable/options/#before-test)  | Execute shell command to prepare test environment |
 |   | [`CIBW_TEST_REQUIRES`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-requires)  | Install Python dependencies before running the tests |
 |   | [`CIBW_TEST_EXTRAS`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-extras)  | Install your wheel for testing using extras_require |
-| **Other** | [`CIBW_BUILD_VERBOSITY`](https://cibuildwheel.readthedocs.io/en/stable/options/#build-verbosity)  | Increase/decrease the output of pip wheel |
+| **Other** | [`CIBW_BUILD_VERBOSITY`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-extras)  | Increase/decrease the output of pip wheel |
```
YannickJadoul: Something went wrong here!

Czaki:
@YannickJadoul again pypy problem with virtualenv separation. 
From travis windows: 

```
which python

/tmp/tmpeoj7ehk3/Scripts/python

+ pip install C:\Users\travis\AppData\Local\Temp\cibuildwheel4ut18dye\repaired_wheel\spam-0.1.0-pp36-pypy36_pp73-win32.whl

Requirement already satisfied: spam==0.1.0 from file:///C:/Users/travis/AppData/Local/Temp/cibuildwheel4ut18dye/repaired_wheel/spam-0.1.0-pp36-pypy36_pp73-win32.whl in c:\cibw\pypy3.6-v7.3.0-win32\site-packages (0.1.0)
```

YannickJadoul:
@Czaki Sigh :-( But this time on Windows? I'll have a look tomorrow, if I can. Could you meanwhile test if `virtualenv<20` works?

Czaki:
I think that `02_test` should be somehow rewritten to check if is running in virtualenv...

Lest wait on result.

Czaki:
@YannickJadoul Yes. Again problem with virtualenv.

YannickJadoul:
> @YannickJadoul Yes. Again problem with virtualenv.

Args. Sorry, today I don't have time to look at it. Maybe tomorrow, I don't know.
Do you think you can try make a minimal example that shows the error? virtualenv 20 is still young, so if you can show what's happening, maybe it could be a virtualenv issue.

Is it also an issue on master, btw, or is it linked to these changes?

> Travis is somehow buggy.
> 
> ```
> write /var/lib/docker/tmp/GetImageBlob188542173: no space left on device
> ```

I restarted everything in the latest build that failed.

Czaki:
> I restarted everything in the latest build that failed.
Again same behaviour. I think that this need to wait till Monday when person from Travis goes back to work. 

> Is it also an issue on master, btw, or is it linked to these changes?

There is no test in master that cover this case... 

I will think about this and maybe produce some test which maybe can be used as example.

Czaki:
@YannickJadoul Done

Czaki:
@joerick Ready for merge. appveyor queue is rely long.

test/11_before_test/test/spam_test.py
```diff
@@ -0,0 +1,28 @@
...
+        # check that the executable also was written
+        executable_file = 'c:\\pythonexecutable.txt' if sys.platform == 'win32' else '/tmp/pythonexecutable.txt'
+        with open(executable_file) as f:
+            stored_executable = f.read()
+        print('stored_executable', stored_executable)
+        print('sys.executable', sys.executable)
+        #  Works around path-comparison bugs caused by short-paths on Windows e.g.
+        #  vssadm~1 instead of vssadministrator
+
+        assert os.stat(stored_executable) == os.stat(stored_executable)
```
joerick: one of these should be sys.executable?

Czaki: yes

Czaki:
Not ready. Need to somehow fix this. What you think about compare folders instead of executable (pypy3.exe vs python.exe)

joerick:
what's the difference between the paths? is python.exe a symbolic link to pypy3.exe?

joerick:
ah, yes it is, cibuildwheel creates it when it installs pypy.

[this](https://stackoverflow.com/a/48416561/382749) might help

joerick:
if that doesn't work, you could consider a [hard link](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/mklink) from `python.exe` to `pypy3.exe` in install_pypy, instead of a symbolic one - maybe os.stat would return the same in that case?

Czaki:
> if that doesn't work, you could consider a hard link from python.exe to pypy3.exe in install_pypy, instead of a symbolic one - maybe os.stat would return the same in that case?

It is not so easy. this executables are cleated by `virtualenv`, not by `instal_pypy` And both files are not links. 
Possible solutions which I see:

* check if `bin` dir is same (`os.path.dirname`)
* use `sys.prefix` [link](https://docs.python.org/3/library/sys.html#sys.prefix) instead of `sys.executable` 
* use information form `sys.implementation` and calculate name of executable?

joerick:
Ah, I hadn't realised that it was virtualenv creating the executables!

hmm. hmm! Of the 3 choices above, my preference would be to compare `sys.prefix`, instead.

Czaki:
So ready.

Czaki:
Hm. Using circleci and appveyor really slow testing PR :(

PR #249, opened 2020-01-18
Switch examples to use PyPI API tokens

### LINKED ISSUE #154: Update usage examples to use PyPI tokens
https://pypi.org/help/#apitoken

PyPI tokens are a much nicer way to auth with PyPI instead of username/password. It's still beta right now, but once it exits we should update all sample code to use this instead.

joerick:
API tokens came out of beta yesterday, so this PR fixes the examples to use those instead of PyPI passwords

https://pypi.org/help/#apitoken

Fix #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,8 @@ jobs:
 
 env:
   global:
-    - TWINE_USERNAME=joerick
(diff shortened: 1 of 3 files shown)


YannickJadoul:
Seems simple enough; and better, indeed! :-)

Issue #239, opened 2019-12-26
Deterministic builds


indygreg:
I think `cibuildwheel` should strive to be deterministic in its behavior. i.e. if you run `cibuildwheel` tomorrow, it will have identical behavior to today. Put another way, the user expectation of `cibuildwheel` is that it only changes in significant ways when its version is updated. Having deterministic behavior makes CI and release pipelines more predictable and reproducible. This reduces frustration and is better from a security perspective.

Fully deterministic output is hard to achieve. Especially when you don't control the base VM image being executed on. But this doesn't mean `cibuildwheel` shouldn't strive to be deterministic wherever possible.

One of the areas where `cibuildwheel` isn't deterministic today is downloading 3rd party dependencies.

For example, installing pip on macOS always retrieves the latest stable version of pip (https://github.com/joerick/cibuildwheel/blob/651f6a9172020aa9a2b0c9eb50dfca06d865ace4/cibuildwheel/macos.py#L35). A better solution here is to fetch an explicit version of `get-pip.py` from e.g. https://github.com/pypa/get-pip/raw/309a56c5fd94bd1134053a541cb4657a4e47e09d/get-pip.py (corresponds to pip 19.2.3).

Another example of non-deterministic behavior is with `pip install`. I think `cibuildwheel` should be pinning versions universally (ideally with hashes for additional security protections). Otherwise, the exact installed package version could vary over time. An example where versions aren't being pinned is https://github.com/joerick/cibuildwheel/blob/651f6a9172020aa9a2b0c9eb50dfca06d865ace4/cibuildwheel/linux.py#L89 and https://github.com/joerick/cibuildwheel/blob/651f6a9172020aa9a2b0c9eb50dfca06d865ace4/cibuildwheel/windows.py#L144.

Is the `cibuildwheel` project receptive to making behavior more deterministic (and secure) by making downloads (and possibly other behavior) more deterministic?

joerick:
Hi @indygreg ! Thanks for your proposal. I definitely agree with the spirit of the suggestion - deterministic software is a good thing™. This was something that I always thought "Oh, we'll get to v1.0 and then think about it", but perhaps because we haven't had many reports of dependencies breaking things it's fallen by the wayside.

This is where we are now, as I see it

```
                      ┌─────────────────┐
 Now                  │                 │
 cibuildwheel v1.1    │      PyPI       │
                      │                 │
                      └─────────────────┘
                               ▲
                               │
                               ▼
[...]

PR #252, opened 2020-01-24
Add option to specify pip versions on windows

### LINKED ISSUE #250: Add option to specify pip version on windows
As part of the initial setup for an environment `python -m pip install --upgrade pip` is run unconditionally on windows:

 https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/windows.py#L144

However, sometimes it's necessary to install a specific or capped version of pip to workaround issues. This has come up recently because of the pip 20.0.0 and pip 20.0.1 releases which are causing issues trying to install some packages (and in my local case running `pip install pip<20.0.0` as a `CIBW_BEFORE_BUILD` command). It would be nice to have an option to be able to specify a version so that we don't have to install pip multiple times.

mtreinish:
As part of the initial setup for an environment on windows:
'python -m pip install --upgrade pip' is run unconditionally. However,
there are situtations where a specific version of pip is needed
(normally to avoid issues introduced in new pip releases). This commit
adds a new config env variable flag, CIBW_PIP_VERSION_WINDOWS, which
takes a pip version in that will be used for this command. For example,
if CIBW_PIP_VERSION_WINDOWS=19.3.1 then the windows environment setup
will run 'python -m pip install --upgrade pip=19.3.1'.

Fixes #250

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -164,7 +164,9 @@ def main():
     elif platform == 'macos':
         pass
     elif platform == 'windows':
-        pass
+        pip_version = os.environ.get('CIBW_PIP_VERSION_WINDOWS', None)
+        if pip_version:
+            build_options['pip_version'] = pip_version
 
     # Python is buffering by default when running on the CI platforms, giving problems interleaving subprocess call output with unflushed calls to 'print'
     sys.stdout = Unbuffered(sys.stdout)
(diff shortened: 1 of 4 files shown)


YannickJadoul:
On an initial look, this is quite specific. Why ónly Windows, and ónly `pip`?
Personally, I think I would prefer some kind of more general mechanism, like `CIBW_PIP_CONSTRAINT` discussed in #250 and #239 (or something similar, that doesn't just limit itself to the initial installation of `pip` on Window).

joerick:
Thanks for this @mtreinish ! I'm working on something that achieves this more generally in #256, so closing this in favour of that.

PR #257, opened 2020-01-30
Clean up temporary workaround in test/06_docker_images/cibuildwheel_test.py

YannickJadoul:
There is now a `manylinux2010-x86` dockcross image, so we can remove a few hacks :-)


joerick:
👍 thanks

PR #264, opened 2020-02-03
Checking python in PATH is the one `cibuildwheel` installed/expects

### LINKED ISSUE #261: Changing `PATH` in `CIBW_ENVIRONMENT` can cause wrong Python to be used
See https://github.com/samuelcolvin/rtoml/pull/8#issuecomment-580880863: if someone changes `PATH` in `CIBW_ENVIRONMENT`, the newly installed Python (from nuget on Windows, but I'm guessing the same is possible on macOS or Linux; still need to check) is not used, but some other Python version of the system is used.

So I think we should add some kind of check after applying `CIBW_ENVIRONMENT` that makes sure we're using the `python` executable we think we're using.

(I'm happy to actually do it myself, actually, if no one else is eager to do this; but I currently don't have time to do it immediately, so I thought I'd make sure to report it.)

YannickJadoul:
Fixes #261

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -70,9 +70,14 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
                 {environment_exports}
 
                 # check the active python and pip are in PYBIN
-                # if `test` returns false, the script will exit due to errexit
-                test "$(which pip)" = "$PYBIN/pip"
-                test "$(which python)" = "$PYBIN/python"
+                if [ "$(which pip)" != "$PYBIN/pip" ]; then
+                    echo "cibuildwheel: python available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python above it."
+                    exit 1
+                fi
+                if [ "$(which python)" != "$PYBIN/python" ]; then
+                    echo "cibuildwheel: pip available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert pip above it."
+                    exit 1
+                fi
 
                 if [ ! -z {before_build} ]; then
                     sh -c {before_build}
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -63,17 +63,25 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             mkdir /output
             cd /project
 
-            {environment_exports}
+            for PYBIN in {pybin_paths}; do (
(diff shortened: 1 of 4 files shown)


YannickJadoul:
Once tests pass, two more things to consider:

- Should we resolve the symlinks with `os.path.realpath`?
- Should we throw an exception and/or print an error message? The main problem is that throwing a Python exception from the bash script in manylinux is not that easy...

joerick:
This looks good to me. Probably worth adding an error message to save people raising issues. How about 

    cibuildwheel: python / pip available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python/pip above it.

YannickJadoul:
> This looks good to me. Probably worth adding an error message to save people raising issues. How about
> 
> ```
> cibuildwheel: python / pip available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python/pip above it.
> ```

Sure, sounds good. But just print it as part, have it as the assert message, throw an exception, ...?

joerick:
Whatever's convenient - assert message or just a simple print would be fine. As you mentioned, getting an exception out of the Linux script is hard, so `echo error message; exit 1` might be the easiest. Or, I guess you could use a special return code from the bash script, if you wanted to get fancy! Up to you

YannickJadoul:
(currently waiting to do this until the build gets fixed; see #270)

YannickJadoul:
@joerick OK, ready for further review/merge, I think. Though it's probably best if this gets rebased onto #185, rather than the other way around.

YannickJadoul:
@Czaki If you still want to have a quick look, I'll wait merging this. (Not sure there is a lot to see, though. It's a pretty small PR. But still, if you spot something, do tell.)

Czaki:
Tomorrow.

YannickJadoul:
> Tomorrow.

Great; no rush to get this merged!

cibuildwheel/linux.py
```diff
@@ -63,17 +63,30 @@ def build(project_dir, output_dir, test_command, test_requires, test_extras, bef
             mkdir /output
             cd /project
 
-            {environment_exports}
+            for PYBIN in {pybin_paths}; do (
+                # Temporary hack/workaround, putting loop body in subshell; fixed in PR #256
+
+                export PATH="$PYBIN:$PATH"
+                {environment_exports}
```
Czaki: Wrong order. Environment can still overwrite path. 

Maybe:
```
export PATH="$PYBIN:$PATH"
{environment_exports}
export PATH="$PYBIN:$PATH"
```

YannickJadoul:
Thanks, @Czaki! :-)

Issue #245, opened 2020-01-10
drop support for python 2.7


samuelcolvin:
It looks like `cibuildwheel` has a significant amount of logic dedicated to python 2.7.

Isn't it time to drop support for 2.7 and simplify the library?

Czaki:
> Being the last of the 2.x series, 2.7 will receive bugfix support until 2020. Support officially stops January 1 2020, but the final release will occur after that date.
> 
> Planned future release dates:
> 
>     2.7.18 code freeze January, 2020
>     2.7.18 release candidate early April, 2020
>     2.7.18 mid-April, 2020

from https://www.python.org/dev/peps/pep-0373/

samuelcolvin:
humm, so when would you suggest `cibuildwheel` can drop 2.7 support?

Many other libraries have already dropped python 2 and rely on prior releases for those still (inexplicably) using python 2.7.

YannickJadoul:
Personally, I think it would be nice to have one final Python 2.7 compatible release with the current merged PRs and some of the pending ones.

Note also that PyPy (which we'll support soon: #185) won't drop 2.7 support (since large parts of their toolchain are built upon the 2.7 version of the language), so I'm not sure we can remove all the code that relates to 2.7.

Czaki:
I agree. Also there is Tauthon which may be supported in future.

YannickJadoul:
Urgh, "Tauthon" what a silly pun :-D

Czaki:
Adding Thauton is an option only if they provide bundled version for all systems. Not when it need to be compiled from source.

joerick:
I'd be okay supporting Python 2 wheels for a little while longer, too, but curious to opinions.

I had thought that we'd get a good, stable release of cibuildwheel out, and then would refer anyone who needs Python 2 support to that version. But... now that I think about it, it would be good to get deterministic builds #239 in before this! At some point, that pinned version of cibuildwheel will be pulling in dependencies that don't support or are buggy on Python 2, for example, the manylinux images, `setuptools`, or `virtualenv`.

However we could definitely remove Python 2 support for running cibuildwheel itself now, I think it's totally fine to require it on the build machine.

jtrmal:
For what it is worth coming from someone just passing by, Py 2.7 won't disappear in two weeks. There will still be (IMO) maintainers, especially in the LTS distributions. So my opinion is to support Py2.7 as long as (reasonably) feasible.

YannickJadoul:
[...]

PR #265, opened 2020-02-03
Dropping Python 2.7 host support

YannickJadoul:
See #245

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,9 +3,9 @@ image:
   - Visual Studio 2015
 
 build_script:
-  - cmd: "C:\\Python27\\python.exe -m pip install -r requirements-dev.txt"
+  - cmd: "C:\\Python37\\python.exe -m pip install -r requirements-dev.txt"
   - sh: "${HOME}/.localpython3.7.4/bin/python3 -m pip install -r requirements-dev.txt"
   # the '-u' flag is required so the output is in the correct order.
   # See https://github.com/joerick/cibuildwheel/pull/24 for more info.
-  - cmd: "C:\\Python27\\python.exe -u ./bin/run_tests.py"
+  - cmd: "C:\\Python37\\python.exe -u ./bin/run_tests.py"
   - sh: "${HOME}/.localpython3.7.4/bin/python3 ./bin/run_tests.py"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,11 @@
 version: 2
 
 jobs:
-  osx-python2:
+  osx-python3.6:
     macos:
-      xcode: "10.0.0"
+      xcode: "9.4.1"
     environment:
-      PYTHON: python2
+      PYTHON: python3
     steps:
       - checkout
 
@@ -16,7 +16,7 @@ jobs:
           name: Test.
           command: venv/bin/python ./bin/run_tests.py
 
-  osx-python3:
+  osx-python3.7:
     macos:
       xcode: "10.0.0"
     environment:
@@ -31,23 +31,7 @@ jobs:
           name: Test.
           command: venv/bin/python ./bin/run_tests.py
 
-  linux-python2:
-    docker:
-      - image: circleci/python:2.7
-    environment:
-      PYTHON: python2
-    steps:
-      - checkout
-      - setup_remote_docker
-
-      - run:
-          name: Prepare the environment.
-          command: bash .circleci/prepare.sh
-      - run:
-          name: Test.
-          command: venv/bin/python ./bin/run_tests.py
-
-  linux-python3:
+  linux-python3.6:
     docker:
       - image: circleci/python:3.6
     environment:
@@ -67,7 +51,6 @@ workflows:
   version: 2
   all-tests:
     jobs:
-      - osx-python2
-      - osx-python3
-      - linux-python2
-      - linux-python3
+      - osx-python3.6
+      - osx-python3.7
+      - linux-python3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,6 @@ language: generic
 
 matrix:
   include:
-    # Linux Python 2
-    - sudo: required
-      language: python
-      python: 2.7
-      services: docker
-      env: PYTHON=python
-
     # Linux Python 3
     - sudo: required
       language: python
--- a/CI.md
+++ b/CI.md
@@ -1,0 +1,9 @@
+This is a summary of the Python versions and platforms covered by the different CI platforms:
+
+|          | 3.5              | 3.6              | 3.7                   | 3.8              |
+|----------|------------------|------------------|-----------------------|------------------|
+| Linux    | Travis CI        | CircleCI         | AppVeyor              | Azure Pipelines  |
+| macOS    | Azure Pipelines  | CircleCI         | Travis CI¹ / CircleCI | Azure Pipelines  |
+| Windows  | TravisCI         | Azure Pipelines  | AppVeyor              | Azure Pipelines  |
+
+> ¹ Python version not really pinned, but dependent on the (default) version of image used.
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,5 @@
 jobs:
-- job: linux
+- job: linux_38
   pool: {vmImage: 'Ubuntu-18.04'}
   steps:
     - task: UsePythonVersion@0
@@ -9,7 +9,17 @@ jobs:
         python -m pip install -r requirements-dev.txt
         python ./bin/run_tests.py
 
-- job: macos
+- job: macos_35
+  pool: {vmImage: 'macOS-10.13'}
+  steps:
+    - task: UsePythonVersion@0
+      inputs:
+        versionSpec: '3.5'
+    - bash: |
+        python -m pip install -r requirements-dev.txt
+        python ./bin/run_tests.py
+
+- job: macos_38
   pool: {vmImage: 'macOS-10.13'}
   steps:
     - task: UsePythonVersion@0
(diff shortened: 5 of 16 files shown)


YannickJadoul:
Can't seem to (easily & quickly) install a Python 3.5 on a macOS CI, but following @mayeut's example see `CI.md`:
```

This is a summary of the Python versions and platforms covered by the different CI platforms:

|          | 3.5        | 3.6                | 3.7                   | 3.8              |
|----------|------------|--------------------|-----------------------|------------------|
| Linux    | Travis CI  | CircleCI           | AppVeyor              | Azure Pipelines  |
| macOS    |            | CircleCI           | Travis CI¹ / CircleCI | Azure Pipelines  |
| Windows  | TravisCI   | AppVeyor (32-bit)  | AppVeyor (64-bit)     | Azure Pipelines  |

> ¹ Python version not really pinned, but dependent on the (default) version of image used.
```

YannickJadoul:
@Czaki I remember there was a situation where `subprocess.run` would have been nice to use in one of your PRs, but it wasn't possible because of Python 2.7. Do you still remember where that was?

Czaki:
I need to check. If i good remember it is connected with checking if proper messages are produced in #156. But now this check are in unit_test part (or can be done here). You can check it or wart few hours.

YannickJadoul:
> If i good remember it is connected with checking if proper messages are produced in #156.

I think you are right. But I cannot immediately find it.

There's not hurry. And don't worry if you do not find it; it's not very important. Just wondering if you remembered.

appveyor.yml
```diff
@@ -1,11 +1,15 @@
-image:
-  - Ubuntu
-  - Visual Studio 2015
+environment:
+  matrix:
+    - image: Ubuntu
+    - image: Visual Studio 2015
+      PYTHON_VERSION: 36
```
joerick: Very minor (sorry!), `PYTHON_DIR: C:\\Python36` might be clearer than this?

YannickJadoul:
Thanks, @joerick! All done, committed, and pushed, except for:

> https://github.com/joerick/cibuildwheel/blob/master/setup.py#L1 (change to `python3`?)

I just removed the shebang line, there; it feels a bit weird for `setup.py` to have that, and StackOverflow seems to agree with me: https://stackoverflow.com/questions/35701131/why-does-setup-py-usually-not-have-a-shebang-line

test/07_ssl/setup.py
```diff
@@ -1,17 +1,9 @@
 import ssl
-import sys
+import sys, urllib.request
 from setuptools import setup, Extension
 
-if sys.version_info[0] == 2:
```
Czaki: This file need to be python 2.7 compatibile. We still support build python 2.7 wheels.

YannickJadoul: Huh.
Something went wrong there. I had already undone this change...

YannickJadoul: Too many rebases, I guess.

YannickJadoul:
If you would have said something here, I could have quickly done it here, instead of with the PR on a PR, @Czaki...

Czaki:
It was easier to test then suggest change. But when i have code then it is easier to make PR than write. Unfortunatelly github does not allow to add comments to lines which are not changed...

I promise to suggest such simple changes as comment.

YannickJadoul:
Ah, I see :-) Sorry, then. But I would've found those `(object)` things also easily, if you had told me. That's what I meant.
Anyway, good catch!

Czaki:
When I starting I think that it will need more changes. But I agree that this change is prety simple.

Czaki:
I go thorough history of PR and cannot found this changes.  

>@Czaki I remember there was a situation where subprocess.run would have been nice to use in one of your PRs, but it wasn't possible because of Python 2.7. Do you still remember where that was?

YannickJadoul:
> I go thorough history of PR and cannot found this changes.

Then it was probably removed, some point. Thank you for the effort! :-)

test/07_ssl/setup.py
YannickJadoul: OK, I know what went wrong. I rebased and merged an old version of the commit, after I had fixed it, by merging your PR :-D

My bad. Too much rebasing!

YannickJadoul:
Found another obvious one: `from __future__ import print_function`
I guess we're really living in the `__future__` now!

Apart from that, I don't know; we can probably keep on finding stuff that would look slightly better in Python 3, but I would suggest to merge soon enough, if other PRs start depending on this (cfr. #256) ? If there's something else important, we can easily make another PR; if we come across something during another PR, we can change it on the go.

YannickJadoul:
Python 3.5 doesn't have the "file system path protocol" ([PEP 519](https://www.python.org/dev/peps/pep-0519/)) yet, meaning that you indeed need a bunch of `str(...)` calls when passing it as argument to e.g. `subprocess` or so. So I'll try to continue a bit to see how bad it is, but I expect it might not be worth it until we can drop 3.5.

joerick:
> Python 3.5 doesn't have the "file system path protocol" ([PEP 519](https://www.python.org/dev/peps/pep-0519/)) yet, meaning that you indeed need a bunch of `str(...)` calls when passing it as argument to e.g. `subprocess` or so. So I'll try to continue a bit to see how bad it is, but I expect it might not be worth it until we can drop 3.5.

I agree with this - the pathlib stuff will be nice, but only when we can use them as conveniently as string paths.

Just had another look over it, the rest looks good - thanks for the suggestions @Czaki. Merge when you're ready!

joerick:
The Appveyor fail is... weird... it looks like the Ubuntu image is trying to run the Windows `cmd` lines, in contradiction with [their docs](https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/#running-windows-and-linux-builds-side-by-side). However, it's not happening on `master` so there must be something in this config that's triggering it...

appveyor.yml
```diff
@@ -1,11 +1,15 @@
...
+  matrix:
+    - image: Ubuntu
+    - image: Visual Studio 2015
+      PYTHON_DIR: C:\\Python36
+    - image: Visual Studio 2015
+      PYTHON_DIR: C:\\Python37-x64
 
 build_script:
-  - cmd: "C:\\Python27\\python.exe -m pip install -r requirements-dev.txt"
+  - cmd: "%PYTHON_DIR%\\python.exe -m pip install -r requirements-dev.txt"
```
Czaki: here is difference. On Ubuntu the variable PYTHON_DIR is not defined so it not eval to string starting from C:\\ and this may make fail.

Czaki:
Why adding step to appveyor? This one is not pararell so it will increase time with waiting on test result.

YannickJadoul:
> Why adding step to appveyor? This one is not pararell so it will increase time with waiting on test result.

Because Windows wasn't covered that well, yet. And because I also thought it would be good to try a 64-bit and 32-bit test. Does it take so long? AppVeyor is always the slowest anyway, because it doesn't cancel old builds (although that can be configured).

YannickJadoul:
> The Appveyor fail is... weird... it looks like the Ubuntu image is trying to run the Windows `cmd` lines, in contradiction with [their docs](https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/#running-windows-and-linux-builds-side-by-side). However, it's not happening on `master` so there must be something in this config that's triggering it...

I'm trying again. It used to work, because some build succeeded, before, I thought. Or maybe it just didn't fail yet.

YannickJadoul:
> I'm trying again. It used to work, because some build succeeded, before, I thought. Or maybe it just didn't fail yet.

OK, maybe it never worked :-(

Czaki:
Maybe better is to add this test to azure pilenes? It has python 32-bits and parallelism work (10 workers)

I think that problem is with using matrix.

YannickJadoul:
@Czaki ^ See what I just did?

YannickJadoul:
> @Czaki ^ See what I just did?

Didn't fancy debugging the AppVeyor config, again.

The proposed CI table now looks like this:


|          | 3.5              | 3.6                | 3.7                   | 3.8              |
|----------|------------------|--------------------|-----------------------|------------------|
| Linux    | Travis CI        | CircleCI           | AppVeyor              | Azure Pipelines  |
| macOS    | Azure Pipelines  | CircleCI           | Travis CI¹ / CircleCI | Azure Pipelines  |
| Windows  | TravisCI         | Azure Pipelines    | AppVeyor              | Azure Pipelines  |

@joerick Do you agree with fully trying to cover those different versions, or are those extra two Azure builds too much?

Czaki:
You write about Windows 32 bits test. I do not see them anywhere. It can be done on azure pipelines with :
```yaml
- task: UsePythonVersion@0
  inputs:
    versionSpec: '3.6' 
    #addToPath: true 
    architecture: 'x86' # Options: x86, x64 (this argument applies only on Windows agents)
```
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version?view=azure-devops#yaml-snippet

YannickJadoul:
> You write about Windows 32 bits test. I do not see them anywhere.

AppVeyor is 32-bit by default (https://www.appveyor.com/docs/windows-images-software/#python)
And actually, after your comment, I thought about it again, and I don't think the host Python version makes a big difference in 64- vs 32-bit?

joerick:
I'm not concerned about host 32/64 bit tbh, that's only a concern on the build side. 

As far as the matrix goes, i think it's good to cover all platforms on 3.5 and the latest, I'm not too worried about gaps in the middle. But where possible, it's good to spread them out and avoid duplicates.

YannickJadoul:
OK, then let's keep it like this. Azure is often fast and parallel enough, and if builds start being too slow, we can easily drop it again.

Let's also try to keep `CI.md` up to date. I think it's quite nice to have that overview somewhere? :-)

PR #277, opened 2020-02-26
[PoC] Test refactoring to use a project template approach

joerick:
For discussion - I've been playing around with this alternative test structure, which has a common 'test template project', which is customised for each test. 

The result is much more code-sharing, and it's easier to see what's different between each test. But it's maybe harder to understand? Curious what people think.

It's very incomplete, Sorry! You'll have to excuse some incomplete refactorings etc.


test/test_02_testing.py
```diff
@@ -0,0 +1,91 @@
+import os, subprocess
+import pytest, textwrap
+from . import utils
+
+
+def test(tmpdir):
+    project_dir = str(tmpdir)
+
+    # generate sample project with a test
+    utils.generate_project(
```
joerick: FYI, there is an incomplete refactor here, I had in mind to migrate these all over to the TemplateProjectC version as seen in 01_basic

test/test_02_testing.py
```diff
@@ -0,0 +1,91 @@
...
+    expected_wheels = utils.expected_wheels("spam", "0.1.0")
+    assert set(actual_wheels) == set(expected_wheels)
+
+
+def test_extras_require(tmpdir):
+    project_dir = str(tmpdir)
+
+    utils.generate_project(
+        path=project_dir,
+        setup_py_setup_args_add='extras_require={"test": ["nose"]},',
```
joerick: This sort of thing might be easier by adding lines to a `setup.cfg`, instead of the setup() function?

test/utils.py
```diff
@@ -120,15 +116,66 @@ def get_platform_tags(python_abi_tag):
     return templates
 
 
+def generate_project(path,
```
joerick: An alternate version of `project_generator.py` - my thinking is that `project_generator.py` would replace this function

Czaki:
I would like to rather use `{% extends ... %}` than long string with code.

joerick:
> I would like to rather use `{% extends ... %}` than long string with code.

Could you explain what you mean here? I don't see how `extends` would fit into the current approach

Czaki:
>>    I would like to rather use {% extends ... %} than long string with code.

> Could you explain what you mean here? I don't see how extends would fit into the current approach

I do not like long part of code in string. Especially with indentation change.  

for example for file `test_03_before build I wold prefer: 
```python 
setup_string = """
import sys
...
"""

def test(tmpdir):
    project_dir = str(tmpdir)

    utils.generate_project(
        path=project_dir,
        setup_py_add=setup_string
)
```

Or use files setup.py with:
```
{% extends `setup.py` %} 
{% block setup_header %}
import sys, os
....
assert (
                os.path.realpath(stored_executable).lower()
                == os.path.realpath(sys.executable).lower()
            )
{% endblock %}
```

Otherwise it will create similar problem like in #256 in linux.py that beacuse of reformatting is hard to see if something in bash script is changed.

YannickJadoul:
Finally took some time to give this a look.

Interesting solution with the templates, indeed. But quite a code-heavy one, it seems (cfr. the `utils.generate_project`)? :-(

On the one hand, I quite like having having everything together in a single pytest test file. On the other hand, I can also see @Czaki's remark about the confusion of having all files together. Though maybe that could already be solved by putting multiine strings like `spam_c_top_level_add` at the top of the files, separate from the test functions?

Another thing: this nicely reduces code duplication, but will it still allow easily for tests that do not follow the same framework? I don't know what's coming up, but maybe we'll need something else than the `spam.c` base, one day.
From that perspective, keeping folders with files (that can still include jinja magic to get rid of code duplication, cfr. @Czaki's suggestion) might be better future-proof?

Also, `spam.c` and `spam.cpp` seem so similar. Is there a reason they cannot start from the same template (and rename the `c` extension to `cpp` somehow)? Or again, however much I like the single-file tests now, having a file `spam.cpp` with `{% extends 'spam.c' %}` might solve this as well?

joerick:
Not convinced on an `{% extends %}` approach. An extends approach would require an extra 2-3 files for each test - even if that file was just `{% extends "base/setup.py" %}`, you'd have to check it while reading the test. I'd rather make the tests more self-contained, so the relevant information needed to understand each one is kept together.

Currently working on a new approach, incorporating the rest of the feedback above. Quite a big job! but I'll ping you when I've got another draft.

joerick:
Work continues in #334

PR #279, opened 2020-02-27
Change twine references to be immediately preceded by pip install in examples

### LINKED ISSUE #266: Travis CI with suggested .travis.yml cannot find "twine"
Thanks for writing cibuildwheel! It was the key to starting back up my old habit of releasing C-accelerated Python libraries, by letting me support platforms I don't have locally.

I was especially happy to see the recent support for 3-platform builds using only Travis CI. The feature worked great the first time I used it, but earlier this week ran into a snag —

https://travis-ci.org/brandon-rhodes/python-sgp4/jobs/646174937

```
/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: No module named twine
```

My `.travis.yml` is essentially the example one shown in your docs, except I trigger upload under slightly different conditions:

https://github.com/brandon-rhodes/python-sgp4/blob/632b3e60d7a4f6ba5662830811c4bf614cfba741/.travis.yml#L1

Do you have any insight as to why `twine` cannot be found by `python -m` under osx? If not then I can begin iterating with `which` and `find` and `echo` and so forth, but I thought that I would start by asking if there were anything obviously wrong with how I pasted and used the example YAML. Thanks!

joerick:
This ensures that the python that installs is the same python that
should execute the module (this sometimes isn't true on macOS because
the python.org installs that cibuildwheel makes take precedence over
the system versions)

Fix #266

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ env:
     # Note: TWINE_PASSWORD is set to a PyPI API token in Travis settings
 
 install:
-  - python3 -m pip install twine cibuildwheel==1.1.0
+  - python3 -m pip install cibuildwheel==1.1.0
 
 script:
   # build the wheels, put them into './wheelhouse'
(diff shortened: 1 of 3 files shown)


YannickJadoul:
Ugh, that's quite counterintuitive, that running cibuildwheel changes `python3`.
Should we not make a note about this in the docs somewhere? (I know there's a warning about not running this on your local mac, but I had never realized #266 could happen.)

joerick:
Good point - it could use an entry in the 'Tips and Tricks' part of the docs. I just wish there was a better way around it!! I suppose we could recommend `PYTHON=$(which python3)`to keep the reference the same?

YannickJadoul:
> I suppose we could recommend `PYTHON=$(which python3)`to keep the reference the same?

Possibly, yes. The most important part is to let users know, I think. Once you know, there's plenty of solutions. In the end, if it's just on a CI image, it doesn't matter too much which Python you're using, I guess.

joerick:
See bae011950c1c7d50d98515afb56b3e123ef01255 for the extra docs update

PR #282, opened 2020-03-01
change 02_test to check if virtualenv is properly used

Czaki:
During rebase of #242 I found that creating virtualenv for pypy on windows is buggy. This PR contains test which check if proper virtualenv is used. On travis error is good visible:

```
AIL: test_virtualenv (spam_test.TestSpam)
----------------------------------------------------------------------
  File "C:\Users\travis\build\Czaki\cibuildwheel\test\02_test\test\spam_test.py", line 18, in test_virtualenv
    self.assertTrue(virtualenv_path in sys.executable.lower())
AssertionError: False is not true

-------------------- >> begin captured stdout << ---------------------
=[executable] c:\cibw\pypy3.6-v7.3.0-win32\python.exe
=[spam location] c:\cibw\pypy3.6-v7.3.0-win32\site-packages\spam.pypy36-pp73-win32.pyd
=[virtualenv path] c:\users\travis\appdata\local\temp\tmpinjk7b6y
--------------------- >> end captured stdout << ----------------------
```

@YannickJadoul Here is minimal example of problem with virtualenv. I think that this test should be merged.

In contrast to previous one this problem occurs only in pypy so I call also @mattip

### DIFF
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -18,9 +18,11 @@ class TestSpam(TestCase):
 
     def test_virtualenv(self):
         virtualenv_path = normalize_path(os.environ.get("__CIBW_VIRTUALENV_PATH__"))
+        if not virtualenv_path:
+            self.fail("No virtualenv path defined in environment variable __CIBW_VIRTUALENV_PATH__")
+
         print("=[executable]", sys.executable)
         print("=[spam location]", spam.__file__)
         print("=[virtualenv path]", virtualenv_path)
         self.assertTrue(virtualenv_path in normalize_path(sys.executable))
         self.assertTrue(virtualenv_path in normalize_path(spam.__file__))
-
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -1,9 +1,26 @@
+from __future__ import print_function
+import os
+import sys
 from unittest import TestCase
 
 import spam
+
+
+def normalize_path(path_str):
+    """because of windows short path"""
+    return os.path.normcase(path_str).replace("vssadm~1", "vssadministrator")
 
 
 class TestSpam(TestCase):
     def test_system(self):
         self.assertEqual(0, spam.system('python -c "exit(0)"'))
         self.assertNotEqual(0, spam.system('python -c "exit(1)"'))
+
+    def test_virtualenv(self):
+        virtualenv_path = normalize_path(os.environ.get("__CIBW_VIRTUALENV_PATH__"))
+        print("=[executable]", sys.executable)
+        print("=[spam location]", spam.__file__)
+        print("=[virtualenv path]", virtualenv_path)
+        self.assertTrue(virtualenv_path in normalize_path(sys.executable))
+        self.assertTrue(virtualenv_path in normalize_path(spam.__file__))
+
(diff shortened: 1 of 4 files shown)


YannickJadoul:
Hmmm, this still only catches if the wrong `env` dictionary is passed. But in principle, if something went wrong setting up the virtualenv, this will still not fail?

test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,28 @@
...
+    def test_virtualenv(self):
+        virtualenv_path = normalize_path(os.environ.get("__CIBW_VIRTUALENV_PATH__"))
+        if not virtualenv_path:
+            self.fail("No virtualenv path defined in environment variable __CIBW_VIRTUALENV_PATH__")
+
+        print("=[executable]", sys.executable)
+        print("=[spam location]", spam.__file__)
+        print("=[virtualenv path]", virtualenv_path)
+        self.assertTrue(virtualenv_path in normalize_path(sys.executable))
+        self.assertTrue(virtualenv_path in normalize_path(spam.__file__))
```
Czaki: @YannickJadoul This two lines check if used python executable is from virtualenv. Also it check if package is installed in virtualenv.

Czaki:
> Hmmm, this still only catches if the wrong env dictionary is passed. But in principle, if something went wrong setting up the virtualenv, this will still not fail?

I do not understand. In the test there is check if python and package is installed in directory created with mktemp.

YannickJadoul:
> I do not understand. In the test there is check if python and package is installed in directory created with mktemp.

Sorry, I looked too fast.
Thanks for making this test/minimal example! :-)

Czaki:
@YannickJadoul @mattip
I finally determine source of problem. when pypy creates virtualenv it create executables in `bin` not `Scripts`. 
For pypy 2.7 it happens for virtualenvs 19 and 20. For pypy 3.6 and virtualenv 20 it puts part of files in Scripts (like python.exe) but part still in bin (like pip.exe) 

This PR forces virtualenv version bellow 20. I create also PR #283 where I put both paths in PATH environment variable.

test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,33 @@
...
+            self.fail("No virtualenv path defined in environment variable __CIBW_VIRTUALENV_PATH__")
+
+        print("=[executable]", sys.executable)
+        print("=[spam location]", spam.__file__)
+        print("=[virtualenv path]", virtualenv_path)
+        print("=[listdir]", os.listdir(virtualenv_path))
+        if os.path.exists(os.path.join(virtualenv_path, 'Scripts')):
+            print("=[listdir]2", os.listdir(os.path.join(virtualenv_path, 'Scripts')))
+        if os.path.exists(os.path.join(virtualenv_path, 'bin')):
+            print("=[listdir]2", os.listdir(os.path.join(virtualenv_path, 'bin')))
```
Czaki: This print are for debug when pypy virtualenv creating was fixed.

mattip:
This should be fixed in the next release of pypy, see [issue 3155](https://foss.heptapod.net/pypy/pypy/issues/3155)

Czaki:
Poited issue does not contain information about pypy 2.7. It will be also fixed?

mattip:
yes, the same fix was applied to all versions

joerick:
Closed in favour of #283

PR #283, opened 2020-03-02
change 02_test to check if virtualenv is properly used another version

Czaki:
copy of #282 with virtualenv not forced to be bellow 20.

### DIFF
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -18,9 +18,11 @@ class TestSpam(TestCase):
 
     def test_virtualenv(self):
         virtualenv_path = normalize_path(os.environ.get("__CIBW_VIRTUALENV_PATH__"))
+        if not virtualenv_path:
+            self.fail("No virtualenv path defined in environment variable __CIBW_VIRTUALENV_PATH__")
+
         print("=[executable]", sys.executable)
         print("=[spam location]", spam.__file__)
         print("=[virtualenv path]", virtualenv_path)
         self.assertTrue(virtualenv_path in normalize_path(sys.executable))
         self.assertTrue(virtualenv_path in normalize_path(spam.__file__))
-
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -1,9 +1,26 @@
+from __future__ import print_function
+import os
+import sys
 from unittest import TestCase
 
 import spam
+
+
+def normalize_path(path_str):
+    """because of windows short path"""
+    return os.path.normcase(path_str).replace("vssadm~1", "vssadministrator")
 
 
 class TestSpam(TestCase):
     def test_system(self):
         self.assertEqual(0, spam.system('python -c "exit(0)"'))
         self.assertNotEqual(0, spam.system('python -c "exit(1)"'))
+
+    def test_virtualenv(self):
+        virtualenv_path = normalize_path(os.environ.get("__CIBW_VIRTUALENV_PATH__"))
+        print("=[executable]", sys.executable)
+        print("=[spam location]", spam.__file__)
+        print("=[virtualenv path]", virtualenv_path)
+        self.assertTrue(virtualenv_path in normalize_path(sys.executable))
+        self.assertTrue(virtualenv_path in normalize_path(spam.__file__))
+
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -6,9 +6,15 @@ from unittest import TestCase
 import spam
 
 
-def normalize_path(path_str):
-    """because of windows short path"""
-    return os.path.normcase(path_str).replace("vssadm~1", "vssadministrator")
+def path_contains(parent, child): 
+    parent = os.path.abspath(parent) 
+    child = os.path.abspath(child) 
+
+    while child != os.path.dirname(child): 
+        child = os.path.dirname(child)  
+        if os.stat(parent) ==  os.stat(child): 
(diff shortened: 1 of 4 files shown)


test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,33 @@
...
+import sys
 from unittest import TestCase
 
 import spam
 
 
+def normalize_path(path_str):
+    """because of windows short path"""
+    return os.path.normcase(path_str).replace("vssadm~1", "vssadministrator")
+
```
joerick: Eesh, is there any way around this? It seems like this could break quite easily.

Maybe instead of the 'path in path' below we could use a function `path_contains(parent, child)`, like:

```python
def path_contains(parent, child): 
    parent = os.path.abspath(parent) 
    child = os.path.abspath(child) 

    while child != os.path.dirname(child): 
        child = os.path.dirname(child)  
        if os.path.samefile(parent, child): 
            return True 
    return False
```

Czaki: I will check this but i use operator `in` because on macos path looks like (`/private` prefix)
```
('=', '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp0gr8ip8m/bin/python')
('=', '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp0gr8ip8m/lib/python2.7/site-packages/spam.so')
('=', '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp0gr8ip8m')
```

test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,39 @@
...
 import spam
 
 
+def path_contains(parent, child): 
+    parent = os.path.abspath(parent) 
+    child = os.path.abspath(child) 
+
+    while child != os.path.dirname(child): 
+        child = os.path.dirname(child)  
+        if os.stat(parent) ==  os.stat(child):
```
Czaki: `os.path.samefile` is from python 3.2

test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,39 @@
...
 
 
+def path_contains(parent, child): 
+    parent = os.path.abspath(parent) 
+    child = os.path.abspath(child) 
+
+    while child != os.path.dirname(child): 
+        child = os.path.dirname(child)  
+        if os.stat(parent) ==  os.stat(child): 
+            return True
```
joerick:
```suggestion
            # parent and child refer to the same directory on the filesystem
            return True 
```

test/02_test/test/spam_test.py
```diff
@@ -1,9 +1,39 @@
+from __future__ import print_function
+import os
+import sys
 from unittest import TestCase
 
 import spam
 
 
+def path_contains(parent, child):
```
joerick:
```suggestion
def path_contains(parent, child): 
    ''' returns True if `child` is inside `parent`.

    Works around path-comparison bugs caused by short-paths on Windows e.g.
    vssadm~1 instead of vssadministrator
    '''
```

test/02_test/test/spam_test.py
joerick: 👍

joerick:
I'm guessing if you had a project at CircleCI when the PR was raised, it associated it with your project instead of mine. I've somehow got them running now.

Btw, you can run the flake8 tests locally by just doing `$ flake8`

Czaki:
> Btw, you can run the flake8 tests locally by just running flake8 - it will pick up the same config as the CI runner.

But this need additional actions so one could create PR that pass all test showed in web UI but finally makes project failing. 

But it looks like it finally working.

joerick:
thank you @Czaki !

YannickJadoul:
@Czaki @joerick Sorry, only now catching up with this PR.

Two remarks/thoughts:
- Do we really want to pollute the environment (sorry, pun really not intended) with `__CIBW_VIRTUALENV_PATH__` in the main code, just for a test? Or is it a feature, and should we document it?
- Is there a way to isolate lines https://github.com/joerick/cibuildwheel/pull/283/files#diff-bf885b72fecfdf7485cab147f7957d5cR198-R200 to somewhere in the code where PyPy is installed? Currently, all our patches are there. If not, can we put a check for PyPy 2.7, 7.3.0 there? I think I read it's going to be fixed in the next release, no? So it might be nice to have it extra specific, to 1) know when to remove it, and 2) catch other unexpected problems.

Czaki:
> Is there a way to isolate lines https://github.com/joerick/cibuildwheel/pull/283/files#diff-bf885b72fecfdf7485cab147f7957d5cR198-R200 to somewhere in the code where PyPy is installed? Currently, all our patches are there. If not, can we put a check for PyPy 2.7, 7.3.0 there? I think I read it's going to be fixed in the next release, no? So it might be nice to have it extra specific, to 1) know when to remove it, and 2) catch other unexpected problems.

No so easy. This files are created during by virtualenv not during installation. 
Also I think that this bugfix has smallest possible impact on code. And after next release of pypy it could be removed. 

> Do we really want to pollute the environment (sorry, pun really not intended) with __CIBW_VIRTUALENV_PATH__ in the main code, just for a test? Or is it a feature, and should we document it?

Have you any other idea? Only what I can imagine is to create other environment variable that is checked before setting this environment variable, 

I also find that in virtualenv there should be defined `sys.base_prefix` but I found it only on stackoverflow. Do not found documentation for this.

PR #290, opened 2020-03-06
Try removing appveyor_run_with_env.cmd

joerick:
I think it was only really needed for Python 3 < 3.4. This would simplify the windows script, removing the shell/simple_shell stuff.

Let's see if the tests pass...


joerick:
The Travis failure is in fact the s390 build, which I've noticed is failing quite often with a 'no space on device' error. I'll contact Travis support about it, but I'm thinking we should label that platform <sup>BETA</sup> when we release the next version.

mayeut:
@joerick, maybe the s390x shall be marked as allowed to fail for now.
travis-ci knows about the issue but nothing new on this end: https://travis-ci.community/t/no-space-left-on-device-for-system-z/5954/11

joerick:
Looks like this is common complaint for Travis CI.

https://travis-ci.community/t/no-space-left-on-device-for-system-z/5954/11

I'm going to keep an eye on that thread, but in the meantime, try setting that platform as 'allow_failures' in a separate PR.

joerick:
haha, thanks @mayeut ! We posted the same thing at the same time :)

PR #291, opened 2020-03-07
Update CPython 3.8 from 3.8.1 to 3.8.2 on macOS/Windows

mayeut:
Changelog: https://docs.python.org/release/3.8.2/whatsnew/changelog.html#python-3-8-2-final


joerick:
Thanks @mayeut !

Issue #289, opened 2020-03-05
multiple commands in CIBW_BEFORE_BUILD on windows 


louisabraham:
I cannot get CIBW_BEFORE_BUILD to execute multiple commands on windows. I tried separating them with `;` but get an error (https://travis-ci.org/louisabraham/pydivsufsort/jobs/658762195).

joerick:
Anyone know enough about Windows shell to do this? We run the command using:

```python
subprocess.run(['cmd', '/E:ON', '/V:ON', '/C', '<command>'])
```

joerick:
According to https://ss64.com/nt/cmd.html, it will support joining commands with '&&' if surrounded with double-quotes. If #290 works that should make this pretty simple to add :)

louisabraham:
nice! It could be useful to add to the documentation.

joerick:
Oh, I mean we'd surround with double-quotes internally to cibuildwheel, not that users should do that

joerick:
Whoops, fumbled the close button

louisabraham:
The meaning of the doc is not very clear: I think it's the commands that should be surrounded by quotes, not the ampersand.

joerick:
This is fixed in #293 - no quotes required!

PR #293, opened 2020-03-07
Allow chaining of commands on Windows

joerick:
Experiment - by removing the explicit `cmd` call for a Windows shell (we don't need this anymore since run_with_env was removed, can we chain commands on windows using `&&` or `||`, like we do on Mac/Linux.


joerick:
Would fix #289

PR #296, opened 2020-03-09
Referencing gmic-py as Working example (for the macosx Github Actions…

myselfhimself:
… builds)


joerick:
Thanks @myselfhimself!

PR #297, opened 2020-03-09
Use yaml anchors and references to not repeat full job in allow_failures

YannickJadoul:
Minor detail, but I'd still like to see CI pass, rather than pushing to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ jobs:
       env:
         - PYTHON=C:\\Python35\\python
 
-    - name: Linux | s390x | Python 3.5
+    - &linux_s390x_35
+      name: Linux | s390x | Python 3.5
       language: python
       python: 3.5
       services: docker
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Random remark: @joerick, mind if we clean up some of your merged branches?
I keep on typing `git checkout ma<tab>`, and it doesn't work, because there's still an ancient `mayeut-patch-1` branch on `upstream` (i.e. `joerick/cibuildwheel`)

YannickJadoul:
> Random remark: @joerick, mind if we clean up some of your merged branches?
> I keep on typing `git checkout ma<tab>`, and it doesn't work, because there's still an ancient `mayeut-patch-1` branch on my `upstream` remote.

(Actually, just noticed. That branch might not be stale, but some other are definitely merged.-

joerick:
Nice! I didn't know YAML could do this

joerick:
I've had a purge of branches - thanks for the reminder

YannickJadoul:
Great, thanks! :-)

YannickJadoul:
@joerick Travis now lists 2 "info messages" on the config, as well:

> - root: missing os, using the default linux
> - language: value minimal is an alias for shell, using shell

Might be worth changing, while we're at it?

PR #298, opened 2020-03-13
Separate python install process to avoid timeout problem

Czaki:
Installing multiple python version takes long time. Pytest hide stdout produced during test run. 
So maybe it is a good idea to separate python installation tasks to get information to stdout. Hence I create `test_single` in first groups of test which run cibuildwheel on each python version separately. 

Python is installed once per test session

### DIFF
--- a/test/01_basic/cibuildwheel_test.py
+++ b/test/01_basic/cibuildwheel_test.py
@@ -1,9 +1,23 @@
 import os
 import platform
 
+import pytest
+
 import utils
 
 project_dir = os.path.dirname(__file__)
+
+
+@pytest.mark.parametrize("build_prefix", ["cp27", "cp35", "cp36", "cp37", "cp38", "pp27", "pp36"])
+def test_single(build_prefix):
+    """To force installation all python for test"""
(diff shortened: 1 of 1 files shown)


joerick:
Ah, cool, that could be a good workaround. Is that what you think the problem is on CircleCI over on #256? Because I increased the timeout to 30m and still had an error - I figured the problem must be something related to CircleCI's docker setup.

Have you seen the timeout problem outside of #256?

joerick:
Ah, I see, this was a PR to my PR! Well it looks like the problem was the use of stdin and `docker exec` without the `-i` flag. 

It seems CircleCI still isn't building your PRs - might be worth checking that or contacting their support about it 🤔

Czaki:
Yes. This is PR to PR. Yes. in this case workaround with extends time of waiting on stdout is working. But adding support of next Python version will increase this time may produce this problem again.

This PR is suggestion how to remove this problem also in future.

joerick:
In fact I didn't need to increase the timeout, the problem was that the script had hung. But we can keep this in mind if we get a timeout happen in regular operation.

Issue #301, opened 2020-03-31
Docs updates


henryiii:
I have started testing this after the addition of GHA and am quite excited to see it seems to cover most of the "special" things I had to do for boost-histogram, and almost everything else I've had to do for other packages. I just have a few documentation bug reports/requests:

* The [environment docs](https://cibuildwheel.readthedocs.io/en/stable/options/#setting-options) do not show examples for GHA (I can help add those)
* The [docs claim](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions)  that MACOS_DEPLOYMENT_TARGET is set to 10.9 by default, but it does not seem to be set at all. It is quite important to set at least 10.9, but in order to make it easy to change, maybe it's best not set by default, but the docs should be corrected.
* The docs mention the [PyBind Win + 2.7 workaround](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#windows-and-python-27), but don't show how to do that in cibuildwheel.

joerick:
> * The [docs claim](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions)  that MACOS_DEPLOYMENT_TARGET is set to 10.9 by default, but it does not seem to be set at all. It is quite important to set at least 10.9, but in order to make it easy to change, maybe it's best not set by default, but the docs should be corrected.

Yeah... we currently don't set `MACOS_DEPLOYMENT_TARGET`, we just use a version of Python that guarantees the resulting wheel will be 10.9+. I had thought that setuptools was setting this, but I guess I'm wrong. I'd propose setting `MACOS_DEPLOYMENT_TARGET=10.9`, if the user hasn't set it. What do you think, @Czaki?

> * The docs mention the [PyBind Win + 2.7 workaround](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#windows-and-python-27), but don't show how to do that in cibuildwheel.

To be completely honest, I don't know. @YannickJadoul might be more help here.

henryiii:
> we just use a version of Python that guarantees the resulting wheel will be 10.9+

10.7, actually, which is why I noticed the docs were not correct. To support macOS 10.14/10.15, I think you have to set 10.9 or higher.

YannickJadoul:
Hey, @henryiii! Thanks for the sharp eye and remarks!

[...]

PR #302, opened 2020-03-31
Add GHA example

henryiii:
First point of #301.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -42,6 +42,13 @@ variables:
   CIBW_TEST_COMMAND: "nosetests {project}/tests"
 ```
 
+> .github/workflows/*.yml ([docs](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)) (can be global, in job, or in step)
+```yaml
+env:
+  CIBW_TEST_REQUIRES: nose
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -42,6 +42,13 @@ variables:
   CIBW_TEST_COMMAND: "nosetests {project}/tests"
 ```
 
+> .github/workflows/*.yml ([docs](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)) (can be global, in job, or in step)
+```yaml
+env:
+  CIBW_TEST_REQUIRES: nose
+  CIBW_TEST_COMMAND: "nosetests {project}/tests"
```
```suggestion
  CIBW_TEST_COMMAND: nosetests {project}/tests
```

Quotes are technically not needed here.

henryiii: Also, shouldn't we change the test command example to pytest?

joerick: Quotes aren't needed, but I think they're useful to people unfamiliar to YAML. But yes, pytest would probably be more familiar to most readers. If you change, could you change all the CI platform examples?

henryiii:
Travis failure seems to be unrelated, this only changes docs.

YannickJadoul:
Restarted the test, and it's indeed green now! :-)

PR #304, opened 2020-04-03
PyPy 7.3.1

YannickJadoul:
This is still a release candidate, but let's already check if everything works and if we can remove those few workarounds! :-)

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -36,8 +36,8 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.7', identifier='cp37-macosx_x86_64', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.9.pkg'),
         PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.2/python-3.8.2-macosx10.9.pkg'),
         # PyPy
-        PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='https://bitbucket.org/pypy/pypy/downloads/pypy2.7-v7.3.1-osx64.tar.bz2'),
-        PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='https://bitbucket.org/pypy/pypy/downloads/pypy3.6-v7.3.1-osx64.tar.bz2'),
+        PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='https://github.com/joerick/cibuildwheel-resources/releases/download/1/pypy2.7-v7.3.1-osx64.tar.bz2'),
+        PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='https://github.com/joerick/cibuildwheel-resources/releases/download/1/pypy3.6-v7.3.1-osx64.tar.bz2'),
     ]
 
     # skip builds as required
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -36,8 +36,8 @@ def get_python_configurations(build_selector):
         PythonConfiguration(version='3.7', identifier='cp37-macosx_x86_64', url='https://www.python.org/ftp/python/3.7.6/python-3.7.6-macosx10.9.pkg'),
         PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.2/python-3.8.2-macosx10.9.pkg'),
         # PyPy
-        PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='https://bitbucket.org/pypy/pypy/downloads/pypy2.7-v7.3.1rc1-osx64.tar.bz2'),
-        PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='https://bitbucket.org/pypy/pypy/downloads/pypy3.6-v7.3.1rc2-osx64.tar.bz2'),
+        PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='http://buildbot.pypy.org/nightly/trunk/pypy-c-jit-99266-aee0dd4e0712-osx64.tar.bz2'),
+        PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='http://buildbot.pypy.org/nightly/py3.6/pypy-c-jit-99273-42fb7e063523-osx64.tar.bz2'),
     ]
 
     # skip builds as required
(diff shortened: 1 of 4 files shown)


YannickJadoul:
The macOS failure here is reported to the PyPy developers, and they're working on it. Something seems to be wrong with the compilation on macOS 10.14, while our failing builds are on 10.13.

@joerick, up to you to decide whether you want to wait for this to make a release. My expectation is that the release candidate from PyPy will soon (within a few days/week) be fixed and then get released. The final thing then is to have an official PyPy manylinux wheel, but I already have a PR ready for that (cfr. the `yannickjadoul/manylinux2010-pypy_x86_64:7.3.1` docker image I'm using to test the release candidates).

YannickJadoul:
PyPy 7.3.1 has been released! Rebased onto master as well.

Now just waiting for pypy/manylinux#5 to be merged so it can be used in this PR.

@joerick The first three commits are just different release candidates, and temporary use of https://hub.docker.com/r/yannickjadoul/manylinux2010-pypy_x86_64 instead of PyPy's official images. Do you want to keep this history, or should I just remove it (and get rid of the potentially confusing `git blame` of `cibuildwheel/resources/pinned_docker_images.cfg`)?

YannickJadoul:
Oh, annoying detail: in 1.5 months, the PyPy download links from bitbucket.org will die. So I guess we might need to make a small patch release, once there's a new download location for PyPy :-(

Czaki:
> Oh, annoying detail: in 1.5 months, the PyPy download links from bitbucket.org will die. So I guess we might need to make a small patch release, once there's a new download location for PyPy :-(

Maybe add proper message to this PR? Printed on download fail. To allow user understand why it fail.

YannickJadoul:
> Maybe add proper message to this PR? Printed on download fail. To allow user understand why it fail.

That's a good idea. But any idea where I would do so? Just in `install_pypy`?

Czaki:
> That's a good idea. But any idea where I would do so? Just in install_pypy?

I think it is best place.

joerick:
> Oh, annoying detail: in 1.5 months, the PyPy download links from bitbucket.org will die

Is this because they are moving away from Bitbucket?

joerick:
> Do you want to keep this history, or should I just remove it (and get rid of the potentially confusing `git blame` of `cibuildwheel/resources/pinned_docker_images.cfg`)?

In this case it's of questionable value, but I'm starting to think that keeping history is best, on the whole. It shows a thought process, Github provides useful features if we don't force-push (like 'changes since you last viewed'). Up to you, ultimately though!

YannickJadoul:
> Is this because they are moving away from Bitbucket?

Yeah. And because those people from Bitbucket don't even want to provide some kind of archived version of stuff... Grmbl.

I asked on IRC if I could switch to a new one, but:

> <mattip> use the bitbucket ones for another couple of weeks. We are working on getting a CDN in front of buildbot, and making that an https link

> > 
> 
> In this case it's of questionable value, but I'm starting to think that keeping history is best, on the whole. It shows a thought process, Github provides useful features if we don't force-push (like 'changes since you last viewed').

Myeah, true, I like history, but only the clean version. With too many commits, history isn't worth anything either. This change back and forth was really only to test and prepare the single commit in the end.

> Up to you, ultimately though!

I don't agree with that, though, since in the end, you'll still be the main maintainer and might need to look at history. Personally here, I'd just squash things nicely into one commit, but if you want, I'll leave it like this. At any rate, I'll try to be a bit more conservative towards history, in the future ;-)

joerick:
> > Up to you, ultimately though!
> 
> I don't agree with that, though, since in the end, you'll still be the main maintainer and might need to look at history. Personally here, I'd just squash things nicely into one commit, but if you want, I'll leave it like this. At any rate, I'll try to be a bit more conservative towards history, in the future ;-)

I appreciate that! I'll try to remember to squash when I merge this one.

Let's get this in, and then #310, and I think I'll do a release :)

YannickJadoul:
Yeah, sure. just waiting for pypy/manylinux#5. But that's close, I think!

YannickJadoul:
Oh, and I was waiting for that timing to know whether to put a warning as @Czaki suggested, or whether I can already use the new PyPy download links. Do you want this release urgent? I can go poke around a bit?

joerick:
> Do you want this release urgent? I can go poke around a bit?

No rush, I think I'd rather wait for the new Pypy links than do a release that might break within weeks!

YannickJadoul:
pypy/manylinux#5 got merged. Reverted `pinned_docker_images.cfg` back to `pypywheels/manylinux2010-pypy_x86_64`, rebased onto master, and fixed the (initially forgotten) unit test.

So it's just waiting for the new links, now (or putting the warning into place).

joerick:
> So it's just waiting for the new links, now (or putting the warning into place).

Nice! By the way, how did you find out about the bitbucket links going away? I can see the announcements about the move to hepapod but I can't find anything about the bitbucket links dying.

YannickJadoul:
> Nice! By the way, how did you find out about the bitbucket links going away?

Yes, I popped onto the #pypy IRC, and they were going to poke the PSF about it again.

> I can see the announcements about the move to hepapod but I can't find anything about the bitbucket links dying.

BitBucket is dropping support for Mercurial, so that's why they moved in the first place (though heptapod seems have a UI that's a lot saner and faster than BitBucket as well). New Mercurial repositories can already not be created anymore, and by June all existing Mercurial repos will be deleted (yes, you read that correctly, deleted; not even archived!). See https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket

On the moving download links, see also https://mail.python.org/pipermail/pypy-dev/2020-April/015924.html, https://mail.python.org/pipermail/pypy-dev/2020-April/015926.html, and https://mail.python.org/pipermail/pypy-dev/attachments/20200413/6a77159f/attachment.html

If you want to release sooner than that, I'd like to implement @Czaki's suggestion, rather than not having PyPy 7.3.1, btw. There are a few fixes I'd like to use myself :-)

joerick:
Figured I might as well speed this along! You okay with this approach, @YannickJadoul ? We can switch back to pypy official links when they appear.

YannickJadoul:
> Figured I might as well speed this along! You okay with this approach, @YannickJadoul ? We can switch back to pypy official links when they appear.

Sure, yes, I think it's a nice out-of-the-box workaround! :-)
Is this also OK for @mattip? Or is the https://downloads.python.org/pypy really almost there?

mattip:
The https://github.com/joerick/cibuildwheel-resources/releases/download/1/pypy2.7-v7.3.1-osx64.tar.bz2 seems like a good solution. The link is almost there but now shows a 404 message.

YannickJadoul:
Alright, thanks, @mattip!

As you see fit, then, @joerick! It's your release schedule, but this seems fine to me :-)

mattip:
News flash: https://downloads.python.org/pypy/ (with the trailing slash) is live.

YannickJadoul:
Amazing! Thanks for the notification, @mattip!

@joerick, I pushed a new commit. Let's see if this works, but if so, we're definitely ready to merge :-)

joerick:
Hahah what timing :)

YannickJadoul:
Wow, I'm also still not used to seeing AppVeyor finish as first CI :smile:

PR #306, opened 2020-04-05
Set MACOSX_DEPLOYMENT_TARGET to 10.9, if not set by the user

joerick:
See discussion in https://github.com/joerick/cibuildwheel/issues/301#issuecomment-606821847 - this would standardise builds across the Pythons ~and remove warnings for pypy~.


YannickJadoul:
Personally, I'm still not entirely convinced this is the way to go. Why should we force to build for 10.9 while PyPy 2.7 can support 10.7? And meanwhile, it seems like the builds are working just fine?

The file name of the wheels is correct (10.7), so I'd somehow rather find out where the 10.3 comes originates? If this is a PyPy issue, it might be better to fix it on their side.

YannickJadoul:
Also, what about PyPy 3.6? I seem to remember that the actual minimal macOS version there is 10.13? So setting `MACOSX_DEPLOYMENT_TARGET` target to 10.9 seems weird, for that?

joerick:
> Personally, I'm still not entirely convinced this is the way to go. Why should we force to build for 10.9 while PyPy 2.7 can support 10.7? And meanwhile, it seems like the builds are working just fine?

Allow me to try to convince you! If we're willing to accept a minimum of 10.9 on CPython, we don't need to support a wider range on a more niche implementation - the chances of anyone needing that are even slimmer. And a minimum of 10.9 across the board reduces inconsistency, simplifies our code, the docs, and also lets users use more modern C++ features by default - see the removed config from those tests. 

> Also, what about PyPy 3.6? I seem to remember that the actual minimal macOS version there is 10.13? So setting MACOSX_DEPLOYMENT_TARGET target to 10.9 seems weird, for that?

The pypy website agrees with you-
image

Which makes their 10.7 wheels even more surprising. I don't suppose setting that to 10.9 wouldn't hurt, but there's no point going that far back if they won't ever run on those systems. Worth defaulting MACOSX_DEPLOYMENT_TARGET=10.13 for that build? Mayybe?

Czaki:
I try to identify this problem with something like this https://github.com/Czaki/cibuildwheel/commit/19368188175709bc0e091385b33e1630c83b40ec, but it pass test. But maybe symilar check somewhere else may simplify find.

> Which makes their 10.7 wheels even more surprising. I don't suppose setting that to 10.9 wouldn't hurt, but there's no point going that far back if they won't ever run on those systems. Worth defaulting MACOSX_DEPLOYMENT_TARGET=10.13 for that build? Mayybe?

I go  through this. `libpypy3-c.dylib` is compiled against macos 10.7. but other runtime libraries (from `lib_pypy` dir are compiled against 10.13. Wheels are compiled against  `libpypy3-c.dylib` so they provide pypy implementation that work on macos 10.13 and newer, but user may compile pypy on against older macos version to use it.

YannickJadoul:
Sorry for the delay in answering!

> Allow me to try to convince you! If we're willing to accept a minimum of 10.9 on CPython, we don't need to support a wider range on a more niche implementation - the chances of anyone needing that are even slimmer. And a minimum of 10.9 across the board reduces inconsistency, simplifies our code, the docs, and also lets users use more modern C++ features by default - see the removed config from those tests.

Yes, I might agree, íf you would still want to make this change without the intention of fixing the weird 10.3 warning from #301 :-)
It just feels like this is not the right solution for what might be something important to report to PyPy (I'm planning to look into this and see if I can reproduce it, if I can find the time), if that makes sense? Even more so since @Czaki seems to have problems to reproduce the same warning.

> > Also, what about PyPy 3.6? I seem to remember that the actual minimal macOS version there is 10.13? So setting MACOSX_DEPLOYMENT_TARGET target to 10.9 seems weird, for that?
> 
> The pypy website agrees with you-
> image
> 
> Which makes their 10.7 wheels even more surprising. I don't suppose setting that to 10.9 wouldn't hurt, but there's no point going that far back if they won't ever run on those systems. Worth defaulting MACOSX_DEPLOYMENT_TARGET=10.13 for that build? Mayybe?

> I go through this. `libpypy3-c.dylib` is compiled against macos 10.7. but other runtime libraries (from `lib_pypy` dir are compiled against 10.13. Wheels are compiled against `libpypy3-c.dylib` so they provide pypy implementation that work on macos 10.13 and newer, but user may compile pypy on against older macos version to use it.

Thanks again, @Czaki! :-) Indeed, I think you already told us before, but I forgot.
So in that case, I think there should be no exception for PyPy 3.6. Sorry about the confusion!

YannickJadoul:
@joerick I dug a bit deeper: https://github.com/joerick/cibuildwheel/issues/301#issuecomment-610087191.

So, `cibuildwheel` is not to blame, but I'm starting to change my mind. Maybe it is indeed `cibuildwheel`'s goal to avoid all this build target/environment rubbish and give the user a clean build environment that will "just work". And of course, it can be easily overridden by a power user with the necessary knowledge.

joerick:
> Maybe it is indeed cibuildwheel's goal to avoid all this build target/environment rubbish and give the user a clean build environment that will "just work"

I think so - and moreover, try to make things more consistent between the platforms. I think this helps.

Issue #158, opened 2019-08-30
Support for gitlab-CI


pmav99:
https://docs.gitlab.com/ce/ci/

Prier:
Would really like to have support for Gitlab CI.

Czaki:
@joerick there is gitlab ci for external repositories https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/index.html

And gitlab provide runners for Linux and Windows https://docs.gitlab.com/ee/user/gitlab_com/index.html#shared-runners

So there is an option to add this support. What dou You think about this? Could you try to setup initial integration to check if it works?

YannickJadoul:
Ow, that does seem interesting, if we could test it from here!

joerick:
Implemented (linux-only) in #419. See #427 for an issue regarding Windows support.

Issue #281, opened 2020-02-28
Code refactor suggestions


Czaki:
During contribution to cibuildwheel I found that some functions are to long and have many lines inside which are not connected with main task of this function.

My suggestions are:

- [ ] Pack `build` arguments in some namedtuple (class based with typing) or namespace. `manylinux_images` can be keep as separated arguement

- [ ] extract everything connected with setup python to install_python. It should:
  * install python
  * install pip if missed  (I do not know why `get_pip.py`is always called)
  * install base packages (setuptools, wheel) 

  something like:
```
def install_python(python_configuration, constrains):
    ...
```      

- [ ] extract some variables from main function to module level. For example `repair_command_default` can be changed to 
```python
repair_command_default_dict = {
    "linux": 'auditwheel repair -w {dest_dir} {wheel}',
    "macos": 'delocate-listdeps {wheel} && delocate-wheel --require-archs x86_64 -w {dest_dir} {wheel}'
}
``` 
and usage:
```python
repair_command = get_option_from_environment('CIBW_REPAIR_WHEEL_COMMAND', platform=platform, default=repair_command_default_dict.get(platform, ""))
```

- [ ] share common commands between scripts (optional). To be sure that build commands (like `pip wheel ... ` are same on every system maybe it is a good idea to  put it in separat module and import?

joerick:
Thanks @Czaki, these are good suggestions. 

I think I agree with all the points above, except the last one because I like the readability of having the commands right there in the script.

Czaki:
> I think I agree with all the points above, except the last one because I like the readability of having the commands right there in the script.

From one side I agree with you (and this is why I put optional). Form other side I still think how to reduce chance of inconsistency between. systems. Maybe add something in unit test  which will mock `call/shell`cals.

When #256 will be finished I can prepare PR with my proposition of this changes.

joerick:
> Maybe add something in unit test which will mock `call/shell`cals.

Sounds good. Yes I'd rather have a unit test that enforces consistency.

YannickJadoul:
I agree the build options need to be refactored, yes. I've actually been trying out the type annotations and mypy, as suggested in #258, and this dictionary is one of the things that just doesn't seem to work with typing (I'll make a PR soon, just to be able to judge how things look, and to be able to further decide on it).

[...]

Issue #208, opened 2019-11-12
manylinux builds failing with cibuildwheel 1.0.0


msanders:
Seeing the following error after upgrading `cibuildwheel` from 0.10.2 to 1.0.0:

```
ERROR: running bdist_wheel
running build
running build_py
creating build
creating build/lib
creating build/lib/autopy
copying autopy/__init__.py -> build/lib/autopy
running build_ext
running build_rust
error: Can not find Rust compiler
```

Full stack trace: https://travis-ci.org/autopilot-rs/autopy/jobs/610426289
Commit (previous builds were passing): https://github.com/autopilot-rs/autopy/commit/244e532cfd99217aed83c952907c9d1f824d9771

Only thing I found was this thread, so I tried removing `sudo: required` in travis.yml (which is currently just copy-pasted from the `cibuildwheel` README) but it had no effect. https://github.com/PyO3/setuptools-rust/issues/54

mayeut:
I can see this in the log before the mentioned failure:
```
++ curl --proto =https --tlsv1.2 -sSf https://sh.rustup.rs -o /root/rust-installer/rustup.sh
317curl: option --proto: is unknown
318curl: try 'curl --help' or 'curl --manual' for more information
319++ sh /root/rust-installer/rustup.sh --default-toolchain=nightly-2019-10-05 -y
320/usr/bin/yum
321sh: /root/rust-installer/rustup.sh: No such file or directory
```

Manylinux images are now using `manylinux2010` instead of `manylinux1` images which might explain the difference in behavior. You can still use `manylinux1` images using [`CIBW_MANYLINUX_X86_64_IMAGE`, `CIBW_MANYLINUX_I686_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image) options.

joerick:
Agreed. Alternatively, at a guess, you could remove the `--proto '=https' --tlsv1.2` from this command, since curl on the newer manylinux2010 images would be more up-to-date.

As a side note, I'd recommend adding `set -o errexit` to your build scripts - that'll help catch this sort of thing sooner.

msanders:
It looks like this is due to this issue: https://github.com/rust-lang/rustup/issues/1794. Removing the `--proto` flag from the curl command indeed fixed it.

Oddly, the version of curl on manylinux2010 was downgraded from 7.64.1 on manylinux1 to 7.19.7. Will be opening a separate issue upstream for that.

[...]

PR #307, opened 2020-04-05
Setup python extracted from build function

Czaki:
For windows and macos extract python setup to separate function and rename `windows.shell` to `windows.call` to unify `build` function between systems.

Main advantage of this PR is separate build wheel logic from python setup logic. Build function is shorter and easier to analyse. 

This also allow to simplify user porting on CI which are not possible to test from github like gitlab (#158)

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -108,6 +108,69 @@ def install_pypy(version, url):
     return installation_bin_path
 
 
+def setup_python(python_configuration, dependency_constraints, environment):
+    if python_configuration.identifier.startswith('cp'):
+        installation_bin_path = install_cpython(python_configuration.version, python_configuration.url)
+    elif python_configuration.identifier.startswith('pp'):
+        installation_bin_path = install_pypy(python_configuration.version, python_configuration.url)
+    else:
+        raise ValueError("Unknown Python implementation")
+    
+    env = os.environ.copy()
+    env['PATH'] = os.pathsep.join([
+        SYMLINKS_DIR,
+        installation_bin_path,
+        env['PATH'],
+    ])
+
+    # Fix issue with site.py setting the wrong `sys.prefix`, `sys.exec_prefix`,
+    # `sys.path`, ... for PyPy: https://foss.heptapod.net/pypy/pypy/issues/3175
+    # Also fix an issue with the shebang of installed scripts inside the
+    # testing virtualenv- see https://github.com/theacodes/nox/issues/44 and
+    # https://github.com/pypa/virtualenv/issues/620
+    # Also see https://github.com/python/cpython/pull/9516
+    env.pop('__PYVENV_LAUNCHER__', None)
+    env = environment.as_dictionary(prev_environment=env)
+
+    # check what version we're on
+    call(['which', 'python'], env=env)
+    call(['python', '--version'], env=env)
+    which_python = subprocess.check_output(['which', 'python'], env=env, universal_newlines=True).strip()
+    if which_python != '/tmp/cibw_bin/python':
+        print("cibuildwheel: python available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python above it.", file=sys.stderr)
+        exit(1)
+
+    dependency_constraint_flags = []
+    if dependency_constraints:
+        dependency_constraint_flags = [
+            '-c', dependency_constraints.get_for_python_version(python_configuration.version)
+        ]
+
+    # install pip & wheel
+    call(['python', get_pip_script] + dependency_constraint_flags, env=env, cwd="/tmp")
+    assert os.path.exists(os.path.join(installation_bin_path, 'pip'))
+    call(['which', 'pip'], env=env)
+    call(['pip', '--version'], env=env)
+    which_pip = subprocess.check_output(['which', 'pip'], env=env, universal_newlines=True).strip()
+    if which_pip != '/tmp/cibw_bin/pip':
+        print("cibuildwheel: pip available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert pip above it.", file=sys.stderr)
+        exit(1)
+    call(['pip', 'install', '--upgrade', 'setuptools', 'wheel', 'delocate'] + dependency_constraint_flags, env=env)
+
+    # setup target platform, only required for python 3.5
+    if python_configuration.version == '3.5':
+        if '_PYTHON_HOST_PLATFORM' not in env:
+            # cross-compilation platform override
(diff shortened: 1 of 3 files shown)


cibuildwheel/macos.py
```diff
@@ -117,64 +180,8 @@ def build(project_dir, output_dir, test_command, before_test, test_requires, tes
...
-        if config.version == '3.5':
-            if '_PYTHON_HOST_PLATFORM' not in env:
-                # cross-compilation platform override
-                env['_PYTHON_HOST_PLATFORM'] = 'macosx-10.9-x86_64'
-            if 'ARCHFLAGS' not in env:
-                # https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260
-                env['ARCHFLAGS'] = '-arch x86_64'
-            if 'MACOSX_DEPLOYMENT_TARGET' not in env:
-                env['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
+        env, dependency_constraint_flags = setup_python(config, dependency_constraints, environment)
```
joerick, line 183: I don't like the multiple return here, the `env` makes a lot of sense, but the dependency_constraint_flags seems too trivial. Could that go elsewhere? I think a method `DependencyConstraints.pip_flags(python_version)` would neaten this up.

cibuildwheel/windows.py
```diff
@@ -19,7 +19,7 @@
 IS_RUNNING_ON_TRAVIS = os.environ.get('TRAVIS_OS_NAME') == 'windows'
 
 
-def shell(args, env=None, cwd=None):
+def call(args, env=None, cwd=None):
```
joerick, line 22: `call()` on macOS doesn't use a shell by default, here Windows always does. I think that's why the names are different. If we're changing it, I'd recommend using a shell only where necessary - when running user commands.

YannickJadoul:
To be fair, personally, I don't really see the advantage of moving that one piece of code to its own function. We're not calling it multiple times, anyway, so it seems to only clutter up stuff more, as demonstrated by the multiple return values, no?

cibuildwheel/windows.py
YannickJadoul: It might actually be nice to keep the distinction, then?

YannickJadoul:
> This also allow to simplify user porting on CI which are not possible to test from github like gitlab (#158)

I don't really understand how this is going to be easier, either?

cibuildwheel/macos.py
Czaki: I move `dependecy_constrains_flags` outside `python_setup` function.

Czaki:
This changes are connected with my propositions from #281.

> To be fair, personally, I don't really see the advantage of moving that one piece of code to its own function. We're not calling it multiple times, anyway, so it seems to only clutter up stuff more, as demonstrated by the multiple return values, no?

For me first reason is that function build is to long. So I suggest to extract from it thinks not connected with build. 

> > This also allow to simplify user porting on CI which are not possible to test from github like gitlab (#158)
> 
> I don't really understand how this is going to be easier, either?

```python
def custom_main():
    from cibuildwheel import windows
    windows.setup_python = custom_setup

    from cibuildwheel.__main__ import main
    main()
```

For example someone have installed python version and would like to setup python create virtualenv instead of install new python versions on machine.

YannickJadoul:
> This changes are connected with my propositions from #208 

I know, but I wasn't really convinced before, either. Anyway, just voicing my opinion; it's up to @joerick to decide

> For me first reason is that function build is to long. So I suggest to extract from it thinks not connected with build.

I don't see how it's too long, because for me needing to go up and down might be more annoying than just seeing the series of commands. But that's just taste, I guess; I'm not saying this is horrible, either :-)

> ```python
> def custom_main():
>     from cibuildwheel import windows
>     windows.setup_python = custom_setup
> 
>     from cibuildwheel.__main__ import main
>     main()
> ```
> 
> For example someone have installed python version and would like to setup python create virtualenv instead of install new python versions on machine.

Pffff, we don't have a public API, so let's not count on this! If we want to allow these kinds of things, lots of other things need to happen on the level of architecture and design. For now, if anyone wants a custom cibuildwheel, it'll have to be a fork anyway, I suspect.

cibuildwheel/windows.py
Czaki: it is not possible because of this bug https://bugs.python.org/issue8557 (with `shell=False` environment variable PATH passed with `env=env` is ignored). 

Change name allow to use same build function for both systems.

joerick: Ah, that is a problem. Probably that means we should stick with `shell`, for now.

joerick:
For me, the neatness of having a `setup_python()` that returns an `env` with the correct `python` available is pretty neat. (Also, seeing it like that, you could imagine chucking an `lru_cache` on there and potentially only having to run that function a few times per build, which could speed up our tests a bit. We'd need to eliminate any global state, though.)

cibuildwheel/windows.py
Czaki: Ok. I revert this change.

Czaki:
> For me, the neatness of having a `setup_python()` that returns an `env` with the correct `python` available is pretty neat. (Also, seeing it like that, you could imagine chucking an `lru_cache` on there and potentially only having to run that function a few times per build, which could speed up our tests a bit. We'd need to eliminate any global state, though.)

`lru_cache` will not work if we ftill call next tests with `subprocess.check_call` in `cibuildwheel_run` from `test/shared/uitils`.

But I see other option for optimalization. Wrote code which will check environment and avoid obsolete `pip install` call?

YannickJadoul:
> But I see other option for optimalization. Wrote code which will check environment and avoid obsolete `pip install` call?

`pip install` takes only a few seconds, but the whole setup is already fragile, sometimes. I don't think that's going to be worth it?

Czaki:
> > But I see other option for optimalization. Wrote code which will check environment and avoid obsolete `pip install` call?
> 
> `pip install` takes only a few seconds, but the whole setup is already fragile, sometimes. I don't think that's going to be worth it?

Extract to separate function should simplify of profiling. We could test how many times it consume in next steps.

joerick:
> lru_cache will not work if we ftill call next tests with subprocess.check_call in cibuildwheel_run from test/shared/uitils.

Good point - it's a subprocess, so the cache would be thrown away each time.

Plus, there is tons of global state - the python installs themselves, and the whole filesystem, which could be modified by CIBW_ENVIRONMENT. Erm, so yeah, not my finest idea 😳!

> `pip install` takes only a few seconds, but the whole setup is already fragile, sometimes. I don't think that's going to be worth it?

It could be `pip install`, it could be `pip wheel`, I honestly don't know! It's a bit of a mystery to me why our tests are so slow. Line profiling would be helpful.

Issue #258, opened 2020-01-30
Concidering the use of type annotations in the code base


MomIsBestFriend:
Type annotations (see [pep 484](https://www.python.org/dev/peps/pep-0484/)) can be really powerful, and now when the code base is relatively small, IMO it's worth considering using it.

This is of course deserves a discussion (just my opinion)

Czaki:
Just 2 days ago in #245 there is decision to drop support execution of cibuildwheel under python 2.7. I think that adding type annotation should be part of migration.

YannickJadoul:
@MomIsBestFriend @Czaki Could you explain the advantage? I don't see the point of adding and maintaining type annotations without any good reason.

Czaki:
Better support for statistical analysis of code. Some of IDE inform that you use wrong type. Better information fore someone who would like to participate.

YannickJadoul:
I like typing, but I'm not convinced about the advantages it would bring to `cibuildwheel`, in the limited amount of code we have? But I was also kind of avoiding answering here, and just hoping @joerick would say "yes"/"no" :D

MomIsBestFriend:
@YannickJadoul I mean, why not? there are a lot of reasons why to use static types, I will name just a few.

- Better integration with IDE environments (as mentioned earlier), when I have static types for example in ```Pycharm``` it will give me much better suggestions for method, that are actually related to the type of my variable.

- Has (small) speed boost

- Better code readability, lemme give you an example.

Consider this piece of code:

```python

def func(lst):
    for thing in lst:
        validate(thing)
```

Clearly you cannot say what this is doing, but if you consider this piece of code:

```python
def func(lst: List[Account]):
    for thing in lst:
        validate(thing)
````

You can somewhat tell what this is doing (even if this example is stupid and meaningless).

Another reason why you should have type annotations is to find bugs in your programs before you even run it, for example.

```python
holiday_bonus = "50.0"

def add_bonus(cur_bonus: float):
    return cur_bonus + 50

def should_get_bonus(emp):
    if emp in good_employees:
        add_bonus(emp)
````

As you can clearly see where is the bug, when having type annotations ```mypy``` will complain even before running the program, and having to find the specific line that's causing the problem when dealing with larger code base, is just way more difficult rather than running ```mypy```

- Having type annotations also help when refactoring the code.


Basically, I can't think of a a reason why not too annotate the code, especially when speaking of a code base that's relatively small.

joerick:
[...]

PR #311, opened 2020-04-06
Pack build arguments

Czaki:
This is PR with packing all argument of build function into single structure. This is extreme version of shortening it signature. I'm not sure if it is best solution, but also partially show how it look if there is try of introduce inheritance. 

I mark it as draft to ask you about opinion. Maybe it is better to create few structures and pack argument in few groups (like setup, test, ...).

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -76,7 +76,7 @@ def get_python_configurations(build_selector):
     return [c for c in python_configurations if matches_platform(c.identifier) and build_selector(c.identifier)]
 
 
-def build(opt: BuildOptions):
+def build(options: BuildOptions):
     try:
         subprocess.check_call(['docker', '--version'])
     except Exception:
@@ -86,14 +86,14 @@ def build(opt: BuildOptions):
               file=sys.stderr)
         exit(2)
 
-    python_configurations = get_python_configurations(opt.build_selector)
+    python_configurations = get_python_configurations(options.build_selector)
     platforms = [
-        ('cp', 'manylinux_x86_64', opt.manylinux_images['x86_64']),
-        ('cp', 'manylinux_i686', opt.manylinux_images['i686']),
-        ('cp', 'manylinux_aarch64', opt.manylinux_images['aarch64']),
-        ('cp', 'manylinux_ppc64le', opt.manylinux_images['ppc64le']),
-        ('cp', 'manylinux_s390x', opt.manylinux_images['s390x']),
-        ('pp', 'manylinux_x86_64', opt.manylinux_images['pypy_x86_64']),
+        ('cp', 'manylinux_x86_64', options.manylinux_images['x86_64']),
+        ('cp', 'manylinux_i686', options.manylinux_images['i686']),
+        ('cp', 'manylinux_aarch64', options.manylinux_images['aarch64']),
+        ('cp', 'manylinux_ppc64le', options.manylinux_images['ppc64le']),
+        ('cp', 'manylinux_s390x', options.manylinux_images['s390x']),
+        ('pp', 'manylinux_x86_64', options.manylinux_images['pypy_x86_64']),
     ]
 
     for implementation, platform_tag, docker_image in platforms:
@@ -112,14 +112,14 @@ def build(opt: BuildOptions):
                   '/bin/bash'])
 
             call(['docker', 'cp',
-                  os.path.abspath(opt.project_dir) + '/.',
+                  os.path.abspath(options.project_dir) + '/.',
                   container_name + ':/project'])
 
             call(['docker', 'start', container_name])
 
             for config in platform_configs:
-                if opt.dependency_constraints:
-                    constraints_file = opt.dependency_constraints.get_for_python_version(config.version)
+                if options.dependency_constraints:
+                    constraints_file = options.dependency_constraints.get_for_python_version(config.version)
 
                     # `docker cp` causes 'no space left on device' error when
                     # a container is running and the host filesystem is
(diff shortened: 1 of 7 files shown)


cibuildwheel/linux.py
```diff
@@ -75,7 +76,7 @@ def get_python_configurations(build_selector):
     return [c for c in python_configurations if matches_platform(c.identifier) and build_selector(c.identifier)]
 
 
-def build(project_dir, output_dir, test_command, before_test, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment, manylinux_images, dependency_constraints):
+def build(opt: BuildOptions):
```
joerick, line 79: I'm not a fan of abbreviations like this, I'd prefer `options` I think

cibuildwheel/macos.py
```diff
@@ -108,13 +109,13 @@ def install_pypy(version, url):
     return installation_bin_path
 
 
-def build(project_dir, output_dir, test_command, before_test, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment, dependency_constraints):
-    abs_project_dir = os.path.abspath(project_dir)
+def build(opt: BuildOptions):
```
joerick, line 112: same here- ~opt~options

cibuildwheel/util.py
```diff
@@ -103,5 +107,39 @@ def get_for_python_version(self, version):
             return self.base_file_path
 
 
+BuildOptions = NamedTuple("BuildOptions", [
+    ("project_dir", str),
+    ("output_dir", str),
+    ("test_command", str),
```
joerick, line 113: Should be optional?

cibuildwheel/util.py
```diff
@@ -103,5 +107,39 @@ def get_for_python_version(self, version):
             return self.base_file_path
 
 
+BuildOptions = NamedTuple("BuildOptions", [
```
joerick, line 110: Ew, I didn't know 3.5 only had this style of NamedTuple... 

You know, I've been considering dropping Python 3.5 support on the host side - 3.6 has so many great features (proper pathlib.Path support, f-strings, better typing)...

cibuildwheel/util.py
```diff
@@ -103,5 +107,39 @@ def get_for_python_version(self, version):
...
+    ("project_dir", str),
+    ("output_dir", str),
+    ("test_command", str),
+    ("test_requires", List[str]),
+    ("test_extras", str),
+    ("before_build", str),
```
joerick, line 116: Also should be optional?

cibuildwheel/util.py
Czaki: Strongly agree. It may also reduce test time.

YannickJadoul: Are the types important?

Maybe we should keep the types for a separate PR? I have some changes stashed away that are probably completely outdated after all the refactoring, but I could bring try to rebase them. And types are pretty useless without tooling to check them, anyway?

At any rate, these were the types I found
```
def build(project_dir: str, output_dir: str, build_selector: Callable[[str], bool],
          manylinux_images: Dict[str, str], environment: ParsedEnvironment,
          before_build: Optional[str], build_verbosity: int, repair_command: str,
          test_command: Optional[str], test_requires: List[str], test_extras: str) -> None:
```

cibuildwheel/util.py
Czaki: yes. I think it is empty string when undefined. Or maybe it should be setted to default as empty sting (in main)?

cibuildwheel/linux.py
YannickJadoul: If that's too long to type every time, then we could say `project_dir=options.project_dir`.
Then again, what's the point of passing 1 argument, if you immediately expand it again :-/

cibuildwheel/util.py
Czaki: > Are the types important?
>
> Maybe we should keep the types for a separate PR? 

I do not see big difference between current NamedTuple and untyped one.  
Removing python 3.5 should be in separate PR. 

> And types are pretty useless without tooling to check them, anyway (as demonstrated by the tests passing without the correct Optional[str]s

Even vscode have basic type support and provide hinting base on them. Also marking type mismatch is avaliable.

YannickJadoul: > Even vscode have basic type support and provide hinting base on them. Also marking type mismatch is avaliable.

We didn't have that before, either, did we? That's why I'm wondering if that shouldn't be discussed in the context of #258 and a related PR.

joerick: I'm increasingly open to typing, having had better experiences with it through the VS Code Python extension (you need to *disable* Jedi in their settings, but it's fantastic).

Typing offers three benefits as I see it
  - inline documentation of types (for others reading the source code)
  - metadata for editors to provide auto-complete
  - type checking, to ensure the above is accurate, and to find bugs before the code is run

So for the sake of the first two, I'm happy for this to be a typing.NamedTuple. It's feeling increasingly likely that types will eventually come to the project, so hopefully we'll be able to check them soon, too.

@YannickJadoul you've mentioned in the past that you had played around with project-level types - I'd be really curious to see how that looks if you're able to share the branch?

cibuildwheel/util.py
```diff
@@ -103,5 +107,39 @@ def get_for_python_version(self, version):
...
+    test_requires: List[str]
+    test_extras: str
+    before_build: Optional[str]
+    build_verbosity: int
+    build_selector: BuildSelector
+    repair_command: str
+    environment: ParsedEnvironment
+    before_test: str
+    dependency_constraints: Optional[DependencyConstraints]
+    manylinux_images: Optional[Dict[str, str]]
```
joerick, line 141: Remove this comment to prevent the code and the comment getting out of sync
```suggestion
Replace this definition with a class-style NamedTuple in the
PEP526 style when Python 3.5 host support is dropped
```

cibuildwheel/util.py
YannickJadoul: OK, fair enough. I just feels weird to introduce typing in this PR.

@joerick I'll try to push what I had later today, even though it'll be outdated.

Czaki: But this PR introduces types only in new part. Does not type existing one.

YannickJadoul:
How comes tests here passed (I assume), but `flake8` fails when merged into master?

YannickJadoul:
Also, it's really a bad one:

`assert intercepted_build_args.args[0],test_extras == ('[' + test_extras + ']' if test_extras else '')`

So why did the CI let this pass? Really weird.

joerick:
Ah! It's that problem where CircleCI doesn't run on Czaki's PRs. Would you mind investigating/contacting them about it @Czaki? It's only your PRs that don't run, so I suspect it's something to do with your account

YannickJadoul:
Argh. I thought I'd fixed it in 25f2e8cb461c6bb3ee7569cebb8498bd87c9ade7, but then I'd only pushed it to my repo, and not to `upstream`.
Anyway, flake8 should now be fixed, after merging with your fix.

Czaki:
> Ah! It's that problem where CircleCI doesn't run on Czaki's PRs. Would you mind investigating/contacting them about it @Czaki? It's only your PRs that don't run, so I suspect it's something to do with your account

When I prepare next PR I will try to find the reason. It may be connected with setup of circleCI on my account. I disable it for cibuildwheel So maybe in next PR it will work.

PR #312, opened 2020-04-06
Refactor expected_wheels test function

joerick:
This function had grown a little unwieldy, so I took some time to refactor it. It's still a little messy, but I think it's a little better than before.

### DIFF
--- a/test/shared/utils.py
+++ b/test/shared/utils.py
@@ -89,7 +89,7 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
 
     python_abi_tags = ['cp35-cp35m', 'cp36-cp36m', 'cp37-cp37m', 'cp38-cp38']
 
-    if pm.machine() == 'x86_64':
+    if pm.machine() in ['x86_64', 'AMD64', 'x86']:
         python_abi_tags += ['cp27-cp27m', 'pp27-pypy_73', 'pp36-pypy36_pp73']
 
         if platform == 'linux':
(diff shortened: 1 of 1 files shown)


test/shared/utils.py
```diff
@@ -81,60 +81,61 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
...
-    if platform == 'linux':
-        if pm.machine() not in ['x86_64', 'i686']:
-            if manylinux_versions is None:
-                manylinux_versions = ['manylinux2014']
-            architectures = {'cp': [pm.machine()]}
-        else:
-            if manylinux_versions is None:
-                manylinux_versions = ['manylinux1', 'manylinux2010']
-            python_abi_tags += extra_x86_python_abi_tags
+    if pm.machine() in ['x86_64', 'AMD64', 'x86']:
```
YannickJadoul, line 92: Why not just the `['x86_64', 'i686']` from before?

YannickJadoul: Ah, wait, I think I see. Because Windows and macOS give different values? :-|
So `'i686'` was never used, then?

joerick: Correct, yes. `i686` was never used (you can't use Docker on i686). Windows refers to `x86_64` as `AMD64`. But Windows reports architecture weirdly - even on a 64-bit machine, if Python is running in 32-bit mode, it reports `x86`. I'm not sure if we have to support 32-bit Python, but I think 32-bit might be more common on Windows so I added that too.

YannickJadoul: Nice, thanks!
Then let's just get it merged! :-)

PR #313, opened 2020-04-06
Only run unit tests and 01_basic on AppVeyor

YannickJadoul:


YannickJadoul:
Note (mainly to self): there should probably still be a condition checking the current branch to see if these tests should be skipped.

Czaki:
Hmm. I remember that installing python is really time expensive. Maybe This PR should also limit number of  produced wheels in 01_basic?

YannickJadoul:
Let's check, first. But I've already made two typos, and everyone's pushing commits, so it's slow.

YannickJadoul:
If you look at other builds, it only takes 10 minutes to complete the first test. So that seems already more acceptable than 25?

joerick:
Sigh. 4 hours later, and Appveyor is still pending? I appreciate the effort, but I think we go with the `appveyor-*` style.

joerick:
Okay, we can try this if you like. I've enabled rolling builds, we'll see if it makes a difference.

YannickJadoul:
AppVeyor builds are now 3-4 minutes per job, like this! :-) 

Now just to figure why setting the `PYTEST_ADDOPTS` doesn't work from the powershell shell?

YannickJadoul:
Look at that, AppVeyor is done before CircleCI ;-)

joerick:
Cool! so, erm, does that powershell command also modify the environment on Linux and macOS? 

Side note- sometimes it would be nice to see what you've tried and the CI runs for those attempts, but those commits are lost to force-push. what do you think?

YannickJadoul:
> Cool! so, erm, does that powershell command also modify the environment on Linux and macOS?

Yes, [it seems so](https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/#bash-and-powershell). Microsoft has been pushing their powershell to different platforms, as far as I know, so luckily we have a single line that works for all.

> Side note- sometimes it would be nice to see what you've tried and the CI runs for those attempts, but those commits are lost to force-push. what do you think?

Yeah, sorry. Most of the time it's just a stupid mistake that I'd like to see erased from history, though :-D Like 2 tries before noticing `PYEST_ADDOPTS` instead of `PYTEST_ADDOPTS`, and then 3 more to finally notice I copy-pasted too quickly from the docs, and it's of course `$env:PYTEST_ADDOPTS` and not `$PYTEST_ADDOPTS`.

The most important things (but I had noticed them locally, while trying out stuff) are:
-  `_basic`, because the expression after `-k` needs to be a valid Python expression :-(
- `--suppress-no-test-exit-code` (and the associated pytest plugin), because if all tests get skipped in a single `pytest` invocation (i.e., for all other tests than `01_basic`), by default the return code is non-zero.

joerick:
Wonderful! Feel free to hit the big green button when you're ready :)

YannickJadoul:
:-)
Still want to try to build in full when on `master` or when the branch name contains `appveyor-`. Given that we can use powershell, that's hopefully not too hard. Or is that not necessary?

joerick:
Good plan. Yes please do.

joerick:
Another suggestion - could we mark the macOS build as 'allow failure'? It's constantly failing at the moment with random errors when downloading dependencies. 

(I'd link the builds but they have disappeared - force push strikes again!)

@Czaki tried adding `--retries 3` to that dependency download line but it didn't help.

Czaki:
@YannickJadoul You have some projects build on appveyor? You do not meet this problems with download? 

It will be nice to distinguish fail from cibuildwheel and network problem. But I have no idea how to do this. So maybe it is best to mark as "allow faliture"

joerick:
Here's an example Mac failure:
https://ci.appveyor.com/project/joerick/cibuildwheel/builds/31984928

I don't think it's a network failure, because we'd see pip logs about that. It seems like the VM just... dies.

Czaki:
@joerick 
did you tray to ask them? Maybe they have more verbose log?

YannickJadoul:
> Another suggestion - could we mark the macOS build as 'allow failure'? It's constantly failing at the moment with random errors when downloading dependencies.

Yes, I like that idea! For now, we have enough mac tests, anyway, I think. I mostly want to keep the Windows build for AppVeyor, given that it's Windows.

> @Czaki tried adding `--retries 3` to that dependency download line but it didn't help.

OK, then I'll remove this again?

YannickJadoul:
> @YannickJadoul You have some projects build on appveyor? You do not meet this problems with download?

I only use AppVeyor for Windows, from back when there was more or less only Travis CI and AppVeyor. Lots of problems, mostly with MSVC, but not in this way ;-)

joerick:
I've raised an issue with Appveyor on the macOS failures: https://github.com/appveyor/ci/issues/3376

> > @Czaki tried adding `--retries 3` to that dependency download line but it didn't help.
> 
> OK, then I'll remove this again?

Yes please. I just looked it up and the default for that param is 5 anyway, so it won't be helping.

YannickJadoul:
Argh. Missed the "not" in that condition :-/ Let's try again.

YannickJadoul:
Done! And it wasn't AppVeyor we were waiting for, this time :-)
Let's merge and see if it now indeed fully tests on `master`!

YannickJadoul:
Seems to be testing fully now, indeed: https://ci.appveyor.com/project/joerick/cibuildwheel/builds/32041763

PR #315, opened 2020-04-07
Add type annotations and drop Python 3.5 host support

### LINKED ISSUE #258: Concidering the use of type annotations in the code base

YannickJadoul:
Finally pushing this, after this came up again in #307. As noted, this is based on an old version of master, before a bunch of refactoring PRs were merged, so it still needs to be rebased (which might be as hard as just doing it from scratch again?).

To check, install mypy and run `mypy --strict cibuildwheel/`.

Mainly meant to open the discussion of which aspects of type annoations are useful in the context of cibuildwheel and which are probably not. Python's typing is not all or nothing, so we can surely find middle ground.

@joerick, @Czaki, I think you were both interested to see this? :-)

EDIT:
Closes #258

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,22 +6,22 @@ branches:
 
 jobs:
   include:
-    - name: Linux | x86_64 + i686 | Python 3.5
+    - name: Linux | x86_64 + i686 | Python 3.6
       language: python
-      python: 3.5
+      python: 3.6
       services: docker
       env: PYTHON=python
 
-    - name: Linux | arm64 | Python 3.5
+    - name: Linux | arm64 | Python 3.6
       language: python
-      python: 3.5
+      python: 3.6
       services: docker
       arch: arm64
       env: PYTHON=python
 
-    - name: Linux | ppc64le | Python 3.5
+    - name: Linux | ppc64le | Python 3.6
       language: python
-      python: 3.5
+      python: 3.6
       services: docker
       arch: ppc64le
       env: PYTHON=python
@@ -30,25 +30,25 @@ jobs:
       os: osx
       env: PYTHON=python3
 
-    - name: Windows | x86_64 | Python 3.5
+    - name: Windows | x86_64 | Python 3.6
       os: windows
       language: shell
       before_install:
-       - choco install python3 --version 3.5.4 --no-progress -y
+       - choco install python3 --version 3.6.8 --no-progress -y
       env:
-        - PYTHON=C:\\Python35\\python
+        - PYTHON=C:\\Python36\\python
 
-    - &linux_s390x_35
-      name: Linux | s390x | Python 3.5
+    - &linux_s390x_36
+      name: Linux | s390x | Python 365
       language: python
-      python: 3.5
+      python: 3.6
       services: docker
       arch: s390x
       env: PYTHON=python
 
   allow_failures:
     # must repeat the s390x job above exactly to match
-    - *linux_s390x_35
+    - *linux_s390x_36
 
 install: $PYTHON -m pip install -r requirements-dev.txt
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
         - PYTHON=C:\\Python36\\python
 
     - &linux_s390x_36
-      name: Linux | s390x | Python 365
+      name: Linux | s390x | Python 3.6
       language: python
       python: 3.6
       services: docker
--- a/CI.md
+++ b/CI.md
@@ -1,12 +1,12 @@
 This is a summary of the Python versions and platforms covered by the different CI platforms:
 
-|          | 3.5              | 3.6              | 3.7                                                | 3.8              |
-|----------|------------------|------------------|----------------------------------------------------|------------------|
-| Linux    | Travis CI        | CircleCI         | AppVeyor² / GitHub Actions                         | Azure Pipelines  |
-| macOS    | Azure Pipelines  | CircleCI         | AppVeyor² / Travis CI¹ / CircleCI / GitHub Actions | Azure Pipelines  |
-| Windows  | TravisCI         | Azure Pipelines  | AppVeyor² / GitHub Actions                         | Azure Pipelines  |
+|          | 3.6                          | 3.7                                                 | 3.8              |
+|----------|------------------------------|-----------------------------------------------------|------------------|
+| Linux    | Travis CI / CircleCI         | AppVeyor² / GitHub Actions                          | Azure Pipelines  |
+| macOS    | CircleCI                     | AppVeyor² / Travis CI¹ / CircleCI / GitHub Actions  | Azure Pipelines  |
+| Windows  | Travis CI / Azure Pipelines  | AppVeyor² / GitHub Actions                          | Azure Pipelines  |
 
 > ¹ Python version not really pinned, but dependent on the (default) version of image used.
 > ² AppVeyor only runs the "basic" test to reduce load.
 
-Non-x86 architectures are covered on Travis CI using Python 3.5.
+Non-x86 architectures are covered on Travis CI using Python 3.6.
--- a/CI.md
+++ b/CI.md
@@ -1,10 +1,10 @@
 This is a summary of the Python versions and platforms covered by the different CI platforms:
 
-|          | 3.6                          | 3.7                                                 | 3.8              |
-|----------|------------------------------|-----------------------------------------------------|------------------|
-| Linux    | Travis CI / CircleCI         | AppVeyor² / GitHub Actions                          | Azure Pipelines  |
-| macOS    | CircleCI                     | AppVeyor² / Travis CI¹ / CircleCI / GitHub Actions  | Azure Pipelines  |
-| Windows  | Travis CI / Azure Pipelines  | AppVeyor² / GitHub Actions                          | Azure Pipelines  |
+|          | 3.6                          | 3.7                                      | 3.8              |
+|----------|------------------------------|------------------------------------------|------------------|
+| Linux    | Travis CI / CircleCI         | AppVeyor² / GitHub Actions               | Azure Pipelines  |
+| macOS    | CircleCI                     | AppVeyor² / Travis CI¹ / GitHub Actions  | Azure Pipelines  |
+| Windows  | Travis CI / Azure Pipelines  | AppVeyor² / GitHub Actions               | Azure Pipelines  |
 
 > ¹ Python version not really pinned, but dependent on the (default) version of image used.
 > ² AppVeyor only runs the "basic" test to reduce load.
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -134,7 +134,7 @@ def main() -> None:
 
     dependency_versions = get_option_from_environment('CIBW_DEPENDENCY_VERSIONS', platform=platform, default='pinned')
     if dependency_versions == 'pinned':
-        dependency_constraints = DependencyConstraints.with_defaults()
+        dependency_constraints = DependencyConstraints.with_defaults()  # type: Optional[DependencyConstraints]
     elif dependency_versions == 'latest':
         dependency_constraints = None
     else:
@@ -169,6 +169,7 @@ def main() -> None:
         print_build_identifiers(platform, build_selector)
         exit(0)
 
+    manylinux_images = None  # type: Optional[Dict[str, str]]
     if platform == 'linux':
         pinned_docker_images_file = os.path.join(
             os.path.dirname(__file__), 'resources', 'pinned_docker_images.cfg'
@@ -181,8 +182,7 @@ def main() -> None:
         #   'pypy_x86_64': {'manylinux2010': '...' }
         #   ... }
 
-        manylinux_images = {}  # type: Optional[Dict[str, str]]
-        assert manylinux_images is not None  # Weird problem with mypy
+        manylinux_images = {}
 
         for build_platform in ['x86_64', 'i686', 'pypy_x86_64', 'aarch64', 'ppc64le', 's390x']:
             pinned_images = all_pinned_docker_images[build_platform]
@@ -199,9 +199,6 @@ def main() -> None:
                 image = config_value
 
             manylinux_images[build_platform] = image
-
-    else:
-        manylinux_images = None
 
     build_options = BuildOptions(
         package_dir=package_dir,
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -5,7 +5,7 @@ import textwrap
 import traceback
 from configparser import ConfigParser
 
-from typing import Any, Dict, List, Optional, Union, overload
+from typing import Any, Dict, List, Optional, overload
 
 import cibuildwheel
 import cibuildwheel.linux
@@ -24,12 +24,10 @@ from cibuildwheel.util import (
 
 
 @overload
-def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str:
-    ...
+def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str: ...  # noqa: E704
 @overload
-def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: None = None) -> Optional[str]:
-    ...
-def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]:
+def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: None = None) -> Optional[str]: ...  # noqa: E704 E302
+def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]:  # noqa: E302
     '''
     Returns an option from the environment, optionally scoped by the platform.
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -5,6 +5,8 @@ import textwrap
 import traceback
 from configparser import ConfigParser
 
+from typing import Any, Dict, List, Optional, Union, overload
+
 import cibuildwheel
 import cibuildwheel.linux
 import cibuildwheel.macos
@@ -14,14 +16,20 @@ from cibuildwheel.environment import (
     parse_environment,
 )
 from cibuildwheel.util import (
+    BuildOptions,
     BuildSelector,
     DependencyConstraints,
     Unbuffered,
-    BuildOptions
 )
 
 
-def get_option_from_environment(option_name, platform=None, default=None):
+@overload
+def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str:
+    ...
+@overload
+def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: None = None) -> Optional[str]:
+    ...
+def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]:
     '''
     Returns an option from the environment, optionally scoped by the platform.
 
@@ -39,13 +47,13 @@ def get_option_from_environment(option_name, platform=None, default=None):
     return os.environ.get(option_name, default)
 
 
-def strtobool(val):
+def strtobool(val: str) -> bool:
     if val.lower() in ('y', 'yes', 't', 'true', 'on', '1'):
         return True
     return False
 
 
-def main():
+def main() -> None:
     parser = argparse.ArgumentParser(
         description='Build wheels for all the platforms.',
         epilog='''
@@ -114,7 +122,7 @@ def main():
     test_extras = get_option_from_environment('CIBW_TEST_EXTRAS', platform=platform, default='')
     package_dir = args.package_dir
     before_build = get_option_from_environment('CIBW_BEFORE_BUILD', platform=platform)
-    build_verbosity = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
+    build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
     build_config, skip_config = os.environ.get('CIBW_BUILD', '*'), os.environ.get('CIBW_SKIP', '')
     if platform == 'linux':
         repair_command_default = 'auditwheel repair -w {dest_dir} {wheel}'
@@ -138,7 +146,7 @@ def main():
         test_extras = '[{0}]'.format(test_extras)
 
     try:
-        build_verbosity = min(3, max(-3, int(build_verbosity)))
+        build_verbosity = min(3, max(-3, int(build_verbosity_str)))
     except ValueError:
         build_verbosity = 0
 
@@ -175,7 +183,8 @@ def main():
         #   'pypy_x86_64': {'manylinux2010': '...' }
         #   ... }
 
-        manylinux_images = {}
+        manylinux_images = {}  # type: Optional[Dict[str, str]]
+        assert manylinux_images is not None  # Weird problem with mypy
 
         for build_platform in ['x86_64', 'i686', 'pypy_x86_64', 'aarch64', 'ppc64le', 's390x']:
             pinned_images = all_pinned_docker_images[build_platform]
@@ -213,7 +222,7 @@ def main():
     )
 
     # Python is buffering by default when running on the CI platforms, giving problems interleaving subprocess call output with unflushed calls to 'print'
-    sys.stdout = Unbuffered(sys.stdout)
+    sys.stdout = Unbuffered(sys.stdout)  # type: ignore
 
     print_preamble(platform, build_options)
 
@@ -231,7 +240,7 @@ def main():
         exit(2)
 
 
-def detect_obsolete_options():
+def detect_obsolete_options() -> None:
     # Check the old 'MANYLINUX1_*_IMAGE' options
     for (deprecated, alternative) in [('CIBW_MANYLINUX1_X86_64_IMAGE', 'CIBW_MANYLINUX_X86_64_IMAGE'),
                                       ('CIBW_MANYLINUX1_I686_IMAGE', 'CIBW_MANYLINUX_I686_IMAGE')]:
@@ -258,7 +267,7 @@ def detect_obsolete_options():
                 os.environ[option] = os.environ[option].replace(deprecated, alternative)
 
 
-def print_preamble(platform, build_options):
+def print_preamble(platform: str, build_options: BuildOptions) -> None:
     print(textwrap.dedent('''
              _ _       _ _   _       _           _
          ___|_| |_ _ _|_| |_| |_ _ _| |_ ___ ___| |
@@ -282,9 +291,9 @@ def print_preamble(platform, build_options):
     print('\nHere we go!\n')
 
 
-def print_build_identifiers(platform, build_selector):
+def print_build_identifiers(platform: str, build_selector: BuildSelector) -> None:
     if platform == 'linux':
-        python_configurations = cibuildwheel.linux.get_python_configurations(build_selector)
+        python_configurations = cibuildwheel.linux.get_python_configurations(build_selector)  # type: List[Any]
     elif platform == 'windows':
         python_configurations = cibuildwheel.windows.get_python_configurations(build_selector)
     elif platform == 'macos':
@@ -296,7 +305,7 @@ def print_build_identifiers(platform, build_selector):
         print(config.identifier)
 
 
-def detect_warnings(platform, build_options):
+def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
     warnings = []
 
     # warn about deprecated {python} and {pip}
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -134,7 +134,7 @@ def main() -> None:
 
     dependency_versions = get_option_from_environment('CIBW_DEPENDENCY_VERSIONS', platform=platform, default='pinned')
     if dependency_versions == 'pinned':
-        dependency_constraints = DependencyConstraints.with_defaults()  # type: Optional[DependencyConstraints]
+        dependency_constraints: Optional[DependencyConstraints] = DependencyConstraints.with_defaults()
     elif dependency_versions == 'latest':
         dependency_constraints = None
     else:
@@ -169,7 +169,7 @@ def main() -> None:
         print_build_identifiers(platform, build_selector)
         exit(0)
 
-    manylinux_images = None  # type: Optional[Dict[str, str]]
+    manylinux_images: Optional[Dict[str, str]] = None
     if platform == 'linux':
         pinned_docker_images_file = os.path.join(
             os.path.dirname(__file__), 'resources', 'pinned_docker_images.cfg'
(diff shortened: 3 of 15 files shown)


cibuildwheel/bashlex_eval.py
```diff
@@ -37,11 +41,11 @@ def evaluate_node(node, context):
         raise ValueError('Unsupported bash construct: "%s"' % node.word)
 
 
-def evaluate_word_node(node, context):
+def evaluate_word_node(node: bashlex.ast.node, context: NodeExecutionContext) -> str:
     word_start = node.pos[0]
     word_end = node.pos[1]
     word_string = context.input[word_start:word_end]
-    letters = list(word_string)
+    letters = list(word_string)  # type: List[Optional[str]]
```
Czaki, line 48: why optional?

cibuildwheel/environment.py
```diff
@@ -74,8 +70,14 @@ def as_dictionary(self, prev_environment):
 
         return environment
 
-    def as_shell_commands(self):
+    def as_shell_commands(self) -> List[str]:
         return [a.as_shell_assignment() for a in self.assignments]
 
-    def __repr__(self):
+    def __repr__(self) -> str:
```
Czaki, line 76: I do not think that typing `__repr__` this is needed

cibuildwheel/linux.py
```diff
@@ -6,14 +6,21 @@
 import uuid
 from collections import namedtuple
 
+from typing import Callable, Dict, List, Optional
+
+from .environment import (
+    ParsedEnvironment,
+)
```
Czaki, line 13: single line?

```python
from .environment import ParsedEnvironment,
```

cibuildwheel/linux.py
```diff
@@ -6,14 +6,21 @@
...
 from .util import (
     get_build_verbosity_extra_flags,
     prepare_command,
 )
 
 
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['identifier', 'path'])
+PythonConfiguration = namedtuple('PythonConfiguration', ['identifier', 'path'])
```
Czaki, line 20:
```python
 PythonConfiguration = NamedTuple('PythonConfiguration', [('identifier', str), ('path', str)])
```

cibuildwheel/macos.py
```diff
@@ -7,14 +7,19 @@
 from collections import namedtuple
 from glob import glob
 
+from typing import Callable, Dict, List, Optional, Union
+
+from .environment import (
+    ParsedEnvironment,
+)
```
Czaki, line 14:
```python
from .environment import ParsedEnvironment,
```

cibuildwheel/windows.py
```diff
@@ -7,6 +7,11 @@
 from glob import glob
 from zipfile import ZipFile
 
+from typing import Callable, Dict, List, Optional
+
+from .environment import (
+    ParsedEnvironment,
+)
```
Czaki, line 14:
```python
from .environment import ParsedEnvironment,
```

cibuildwheel/macos.py
```diff
@@ -24,8 +29,10 @@ def call(args, env=None, cwd=None, shell=False):
     return subprocess.check_call(args, env=env, cwd=cwd, shell=shell)
 
 
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
+PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
```
Czaki, line 32:
```python
PythonConfiguration = NamedTuple('PythonConfiguration', [('version', str), ('identifier', str), ('url, str)'])
```

cibuildwheel/util.py
```diff
@@ -24,37 +26,37 @@ def get_build_verbosity_extra_flags(level):
...
 
-    def __call__(self, build_id):
-        def match_any(patterns):
+    def __call__(self, build_id: str) -> bool:
+        def match_any(patterns: List[str]) -> bool:
             return any(fnmatch(build_id, pattern) for pattern in patterns)
         return match_any(self.build_patterns) and not match_any(self.skip_patterns)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
```
Czaki, line 38: unnecessary

cibuildwheel/windows.py
```diff
@@ -31,22 +36,24 @@ def simple_shell(args, env=None, cwd=None):
...
+def get_nuget_args(version: str, arch: str) -> List[str]:
     python_name = 'python' if version[0] == '3' else 'python2'
     if arch == '32':
         python_name = python_name + 'x86'
     return [python_name, '-Version', version, '-OutputDirectory', 'C:\\cibw\\python']
 
 
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'url'])
+PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'url'])
```
Czaki, line 53:
```python
PythonConfiguration = NamedTuple('PythonConfiguration', [('version', str), ('arch', str), ('identifier', str), ('url', Optional[str])])
```

cibuildwheel/bashlex_eval.py
YannickJadoul: Because I got an error that `None` is assigned further down. If you remove the annotation, it should show up.

cibuildwheel/linux.py
YannickJadoul: Maybe, yes. I guess I copied this, and removed an unused import afterwards.

cibuildwheel/environment.py
YannickJadoul: It probably gave me warnings if I didn't do it. But I'll try.

YannickJadoul: `cibuildwheel/environment.py:56: error: Function is missing a return type annotation`

cibuildwheel/util.py
YannickJadoul: Seems necessary to shut up `mypy`? See above.

cibuildwheel/linux.py
YannickJadoul: Isn't it nicer in style to have both imports the same?

```python
from .environment import (
    ParsedEnvironment,
)
from .util import (
    download,
    get_build_verbosity_extra_flags,
    prepare_command,
)
```

Czaki: and line above?
```python
 from typing import Callable, Dict, List, Optional
```

Czaki: Yes. this dicusssion is not strongly connected with PR clue.

YannickJadoul:
> Great job.

Thanks! Thanks for the quick review as well.

There's many things I'm not convinced of that I like, though:



> Looking on this code I think that `installl_cpython` and `install_pypy` should get as argument `PythonConfiguration`, not only selected fields. This will give better flexibility.

I don't know if we need that flexibility currently, but I'll have a look after rebasing. Too many things have changed anyway.

> And when drop execution on python 3.5 they maybe it is a good idea to return `Path` object from them.

There are lots of things to do when dropping 3.5 support. `pathlib` is not really related to types, I'd say (though types will help refactoring). For types, most importantly, there's `var: Type = abc` rather than `var = abc  # type: Type` and the NamedTuple. I'll make an issue with a list!

YannickJadoul:
There are still quite a few things I don't like, though:

- I don't mind importing `typing` for type annotations, but it's weird/ugly that it suddenly takes over the role of `collections.namedtuple`. `typing` is for types, not for functionality!?
- The `@overload`s of `__main__.get_option_from_environment` are exaggerated (I couldn't find another way to specify the return type correctly than with `@overload`). There's more code to define/type the function than in it's body. And flake8 acts annoying about style, as it wants me to put `...` on the next line ánd have 2 lines of whitespace in between (so I added `# noqa`). Kind of defeats the purpose of a quick utility function?
- Importing `ParsedEnvironment` is quite annoying as well? Where's the nice duck-typing gone?

Czaki:
> There are still quite a few things I don't like, though:
> 
>    * I don't mind importing `typing` for type annotations, but it's weird/ugly that it suddenly takes over the role of `collections.namedtuple`. `typing` is for types, not for functionality!?

For me best is dataclass, but it is introduced in python 3.7. Current option is only one which gives type support and do not force us to write own constructor. 

>    * The `@overload`s of `__main__.get_option_from_environment` are exaggerated (I couldn't find another way to specify the return type correctly than with `@overload`). There's more code to define/type the function than in it's body. And flake8 acts annoying about style, as it wants me to put `...` on the next line ánd have 2 lines of whitespace in between (so I added `# noqa`). Kind of defeats the purpose of a quick utility function?

maybe use `pyi` for some files (it need to move all typing to this file) to hide typing from main file. 

>    * Importing `ParsedEnvironment` just for typing is quite annoying as well? Where's the nice duck-typing gone?

Something for something. You got type cheeking, but need to write more code. 

You do not think about introduce type check task (as flake8) to tests?

YannickJadoul:
> You do not think about introduce type check task (as flake8) to tests?

Yes, I will. But figuring that out was not my main concern. Something to think about after rebase, and once we can make decisions.
That's why I just wrote the `mypy --strict cibuildwheel/` command, for now.

cibuildwheel/__main__.py
```diff
@@ -14,11 +16,15 @@
...
-    Unbuffered
+    Unbuffered,
 )
 
 
-def get_option_from_environment(option_name, platform=None, default=None):
+@overload
+def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str: ...  # noqa: E704
+@overload
+def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: None = None) -> Optional[str]: ...  # noqa: E704 E302
```
joerick, line 26: Is this overload redundant? It feels like the one below captures this one

YannickJadoul: The problem then is that `mypy` gets confused that even if you pass a default (and the return value can thus never be `None`), it will complain about assigning an `Optional[str]` to a `str` (e.g, calling `parse_environment(environment_config)` which expects a `str`).

cibuildwheel/environment.py
joerick: I noticed above that some functions were marked as `-> None`, which seemed... too obvious and cluttery. As does this. Maybe we can get a bit more granular with the mypy rules - `--strict` just enables them all

cibuildwheel/linux.py
YannickJadoul: Nono, it's a valid point; I appreciate the comment! :-)
But I don't really know what's the best way to solve it.

cibuildwheel/windows.py
```diff
@@ -101,7 +106,10 @@ def install_pypy(version, arch, url):
     return installation_path
 
 
-def build(project_dir, output_dir, test_command, test_requires, test_extras, before_build, build_verbosity, build_selector, repair_command, environment):
+def build(project_dir: str, output_dir: str, build_selector: Callable[[str], bool],
```
joerick, line 109: build_selector should be a BuildSelector object, I think.

cibuildwheel/__main__.py
joerick: yes, I meant that specific line - I'll try it

cibuildwheel/windows.py
YannickJadoul: If you prefer, yes. I made it this way, such the code here (and in `macos.py` and `linux.py`) is agnostic about the actual type (and doesn't need to import it), but just knows it need to be able to be called.

cibuildwheel/environment.py
YannickJadoul: `mypy.ini` or `setup.cfg` should allow for that, indeed.

cibuildwheel/__main__.py
joerick: (you probably already know this, but) The way that `@overload` works is really confusing... the signatures marked with @overload are the 'public' signatures of the function, but the last one specifies the types of the arguments inside. So this is equivalent:

```python
@overload
def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str: ...  # noqa: E704
@overload
def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]: ...  # noqa: E704 E302
def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]:  # noqa: E302
    '''
    Returns an option from the environment, optionally scoped by the platform.
```

According to the docs, this is the intended way to use `@overload`:

```python
@overload
def get_option_from_environment(option_name: str, platform: Optional[str], default: str) -> str: ...  # noqa: E704
@overload
def get_option_from_environment(option_name: str, platform: Optional[str] = None, default: Optional[str] = None) -> Optional[str]: ...  # noqa: E704 E302
def get_option_from_environment(option_name, platform=None, default=None):
    '''
    Returns an option from the environment, optionally scoped by the platform.
```

But in this form, `--strict` complains `Function is missing a type annotation`. Without `--strict`, mypy has those params as `Any`, so isn't type-checking the interior. Frustrating.

YannickJadoul:
Is there any PR I should wait for before rebasing? If not, I'll try to do it somewhere over the next few days. I've noticed it doesn't make a lot of sense to discuss code too much, if it might still change substantially.

joerick:
So. In general, I'm happy with the way this is heading. Yes, a detailed review would have to wait until we have an up-to-date PR. Let's try to get a few of the open PRs in before doing that - I'm thinking about #295 and #311 mainly, but some of the smaller ones might as well land, too.

joerick:
> * I don't mind importing `typing` for type annotations, but it's weird/ugly that it suddenly takes over the role of `collections.namedtuple`. `typing` is for types, not for functionality!?

Strong agree. Always felt weird to me. Why didn't they just type-enable the version in `collections`? But it's not a big deal to me.

> * The `@overload`s of `__main__.get_option_from_environment` are exaggerated (I couldn't find another way to specify the return type correctly than with `@overload`). There's more code to define/type the function than in it's body. And flake8 acts annoying about style, as it wants me to put `...` on the next line ánd have 2 lines of whitespace in between (so I added `# noqa`). Kind of defeats the purpose of a quick utility function?

Yeah. I don't think there's anything we can do to mypy to improve this, but maybe we can tweak the flake8 rules to help here. Let's explore in the new/rebased PR?

> * Importing `ParsedEnvironment` just for typing is quite annoying as well? Where's the nice duck-typing gone?

Will presumably be helped by the BuildOptions struct in #311.

joerick:
Another note - I vote we drop Python 3.5 support in the same PR that we do the typing work. It will make things much easier. But let's leave the `pathlib` refactor for now - that will be much easier once we have types.

cibuildwheel/__main__.py
YannickJadoul: Yes, I tried lots of stuff, when writing that. Also with `Union[]` and generics, but I didn't get anything to work (and pass with `--strict`), so best I could do was what's there.
Very frustrating for such a simple use case, indeed :-(

YannickJadoul:
Side note, after all the complaining: on the bright side, running `mypy` and annotating and fixing these issues is a lot of fun and satisfaction as well ;-) (I just wish there'd be less clutter remaining in the code afterwards.)

YannickJadoul:
First rebase. Got curious, and it actually went a lot quicker than expected. Juts a few extra things to annotate, now.

cibuildwheel/bashlex_eval.py
joerick: We could set them to `''` empty string instead of None, and make it non-optional.

cibuildwheel/environment.py
joerick: It looks like the function types really are required, if we want type inference-

> Type inference is not used in dynamically typed functions (those without a function type annotation) — every local variable type defaults to Any in such functions. Any is discussed later in more detail.

https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#type-inference

cibuildwheel/util.py
```diff
@@ -2,12 +2,13 @@
...
 from time import sleep
-from typing import NamedTuple, List, Optional, Dict
+
+from typing import Dict, List, NamedTuple, Optional, Type, TypeVar
 
 from .environment import ParsedEnvironment
 
 
-def prepare_command(command, **kwargs):
+def prepare_command(command: str, **kwargs: str) -> str:
```
Czaki, line 11: str? `Dict`

cibuildwheel/util.py
```diff
@@ -81,18 +82,21 @@ def download(url, dest):
...
+
 class DependencyConstraints:
-    def __init__(self, base_file_path):
+    def __init__(self, base_file_path: str):
         assert os.path.exists(base_file_path)
         self.base_file_path = os.path.abspath(base_file_path)
 
     @classmethod
-    def with_defaults(cls):
+    def with_defaults(cls: Type[DependencyConstraints_T]) -> DependencyConstraints_T:
```
joerick, line 94: We don't have to worry about subclassing, maybe lets just make this a staticmethod? Or do-

```suggestion
    def with_defaults(cls: Type[DependencyConstraints]) -> DependencyConstraints:
```

cibuildwheel/windows.py
```diff
@@ -62,20 +70,21 @@ def get_python_configurations(build_selector):
...
     nuget_args = get_nuget_args(version, arch)
     installation_path = os.path.join(nuget_args[-1], nuget_args[0] + '.' + version, 'tools')
     shell([nuget, 'install'] + nuget_args)
     return installation_path
 
 
-def install_pypy(version, arch, url):
+def install_pypy(version: str, arch: str, url: Optional[str]) -> str:
     assert arch == '32'
+    assert url is not None
```
Czaki, line 87: Optional and assert? Maybe add to if before call function?

CI.md
```diff
@@ -1,11 +1,11 @@
...
-|          | 3.5              | 3.6              | 3.7                                               | 3.8              |
-|----------|------------------|------------------|---------------------------------------------------|------------------|
-| Linux    | Travis CI        | CircleCI         | AppVeyor / GitHub Actions                         | Azure Pipelines  |
-| macOS    | Azure Pipelines  | CircleCI         | AppVeyor / Travis CI¹ / CircleCI / GitHub Actions | Azure Pipelines  |
-| Windows  | TravisCI         | Azure Pipelines  | AppVeyor / GitHub Actions                         | Azure Pipelines  |
+|          | 3.6                          | 3.7                                               | 3.8              |
+|----------|------------------------------|---------------------------------------------------|------------------|
+| Linux    | Travis CI / CircleCI         | AppVeyor / GitHub Actions                         | Azure Pipelines  |
+| macOS    | CircleCI                     | AppVeyor / Travis CI¹ / CircleCI / GitHub Actions  | Azure Pipelines  |
+| Windows  | Travis CI / Azure Pipelines  | AppVeyor / GitHub Actions                         | Azure Pipelines  |
```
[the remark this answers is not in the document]
YannickJadoul: I added it in #265 (see https://github.com/joerick/cibuildwheel/pull/265#issuecomment-581493314 and https://github.com/joerick/cibuildwheel/pull/265#issuecomment-582149255) and and I think it was updated after GitHub Actions was added.

It's not part of the docs and not meant for users, but just something I added to keep track of which CI providers our own tests run on and which Python versions are run (after @mayeut made such a nice table once or twice at the start of a PR; sorry for the useless mention, @mayeut, but it was a great idea :-) ).

cibuildwheel/util.py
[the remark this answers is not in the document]
YannickJadoul: Because `cls` is a parameter and could be different if there's a subclass. If this isn't necessary, this should probably not be `@classmethod` but `@staticmethod`?

cibuildwheel/util.py
YannickJadoul: No: https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

cibuildwheel/windows.py
YannickJadoul: I didn't want to break the line `installation_bin_path = install_pypy(python_configuration.version, python_configuration.url)`. But yeah, maybe.

cibuildwheel/util.py
YannickJadoul: OK, let's go for `@staticmethod`, then?
I know I overdid it a little bit, here. But just wanted do get `--strict` to pass, and worry about what was too much later :-)

.travis.yml
```diff
@@ -30,25 +30,25 @@ jobs:
...
-       - choco install python3 --version 3.5.4 --no-progress -y
+       - choco install python3 --version 3.6.8 --no-progress -y
       env:
-        - PYTHON=C:\\Python35\\python
+        - PYTHON=C:\\Python36\\python
 
-    - &linux_s390x_35
-      name: Linux | s390x | Python 3.5
+    - &linux_s390x_36
+      name: Linux | s390x | Python 365
```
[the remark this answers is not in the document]
YannickJadoul: Whoops. Thanks :-)

cibuildwheel/util.py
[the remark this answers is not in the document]
Czaki: I still does not understand. If it is subclass it still return instance of `DependecyConstrains` and cls still be `DependecyConstrains` class (with extension)

cibuildwheel/windows.py
Czaki: I think about ` elif python_configuration.identifier.startswith('pp') and python_configuration.url is not None:

YannickJadoul: Yeah, just did that. Just give me a second to commit :-)

cibuildwheel/util.py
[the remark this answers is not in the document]
YannickJadoul: No, because you expect `SubclassOfDependencyConstraints.with_defaults` to return a `SubclassOfDependencyConstraints` object, and not a `DependencyConstraints` object. So the return statement would be OK, but the code using the classmethod wouldn't be OK.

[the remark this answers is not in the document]
YannickJadoul: Anyway, it's solved. See https://github.com/joerick/cibuildwheel/pull/315#discussion_r406854654.

cibuildwheel/environment.py
YannickJadoul: Yuk :-(
Somehow it isn't necessary for '__init__`, I think?

cibuildwheel/util.py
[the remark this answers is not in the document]
Czaki: So still typing system in python have weakness. Declaration that it return something in type of current class should be easy.

cibuildwheel/bashlex_eval.py
YannickJadoul: If you prefer? It does clean up `''.join(l for l in letters if l is not None)` to just `''.join(letters)` as well.

cibuildwheel/windows.py
YannickJadoul: Oh, wait no. I did it slightly differently.

But if `python_configuration.identifier.startswith('pp')` and `python_configuration.url is None`, that's a mistake. So we should assert on that.

Czaki: so tests will fail on unsuported python interpreter
Or I have maybe introduce CPythonConfiguration and PyPythonConfiguration? To force at the level of code url?

YannickJadoul: What? No. `python_configuration.url` is something we list ourselves. How can this fail?

YannickJadoul: This is literally what asserts are made for, right? To check programmatic assumptions that are not user input.

YannickJadoul: (Or I'm really missing something, and then I'm curious what obvious thing I overlooked. Sorry, then, for being too unbelieving :-) )

cibuildwheel/util.py
joerick: Yep staticmethod is good

cibuildwheel/windows.py
```diff
@@ -107,6 +106,7 @@ def setup_python(python_configuration: PythonConfiguration, dependency_constrain
     if python_configuration.identifier.startswith('cp'):
         installation_path = install_cpython(python_configuration.version, python_configuration.arch, nuget)
     elif python_configuration.identifier.startswith('pp'):
+        assert python_configuration.url is not None
         installation_path = install_pypy(python_configuration.version, python_configuration.arch, python_configuration.url)
```
joerick, line 110: Really cool that this works!

cibuildwheel/util.py
YannickJadoul: Should be done, already (but GitHub shows an old version if you click the `cibuildwheel/util.py` here).

cibuildwheel/windows.py
YannickJadoul: Yes, indeed. Makes perfect sense, but cool feature of mypy type inference :-)

Czaki:
Progress?

YannickJadoul:
> Progress?

On what? I think there aren't a lot of open discussions, for the moment? But this is not for the upcoming release, yet, if I understood @joerick correctly.

Seems recent PRs resulted in conflicts, though. I'll rebase, in a bit!

Czaki:
@YannickJadoul I do not want to create next PR until types are finished/decided.

YannickJadoul:
Rebased

Czaki:
There is backport dataclasses module to python 3.6 https://pypi.org/project/dataclasses/

YannickJadoul:
> There is backport dataclasses module to python 3.6 https://pypi.org/project/dataclasses/

Is that worth it? Not a big advantage over `NamedTuple`, I thought?

Czaki:
dataclass allow to create modifiable container. 
It comes outside typing module: 

> I don't mind importing `typing` for type annotations, but it's weird/ugly that it suddenly takes over the role of `collections.namedtuple`. `typing` is for types, not for functionality!?

It also have few nice features like `dataclasses.field`

YannickJadoul:
> dataclass allow to create modifiable container.

But if we don't need mutable containers, it might be better to not do this?

> It comes outside typing module:
> 
> > I don't mind importing `typing` for type annotations, but it's weird/ugly that it suddenly takes over the role of `collections.namedtuple`. `typing` is for types, not for functionality!?
> 
> It also have few nice features like `dataclasses.field`

OK, yes, agreed there's some nice things, definitely. But not sure it's worth adding the extra dependency to the core of `cibuildwheel`? :-/

Czaki:
This is dependency only for python 3.6. From Python 3.7 it is built-in module. 

The question is If nice features of dataclass will be useful. If Yes then I not think it is good to wait with this till python 3.6 is dropped if it can be introduced earlier.

joerick:
I'm not so interested in the dataclass - config should be an immutable object anyway.

I think we should try to get this in soon, before #334, if that works for you @YannickJadoul! I'll take another look over now, but I think it's pretty close, it probably just needs a merge of upstream master.

cibuildwheel/linux.py
```diff
@@ -42,8 +43,13 @@ def matches_platform(identifier):
...
 
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'path'])
+class PythonConfiguration(NamedTuple):
+    version: str
+    identifier: str
+    path: str
+
+
+def get_python_configurations(build_selector: Callable[[str], bool]) -> List[PythonConfiguration]:
```
joerick, line 52: I think we'd be better to use BuildSelector as a type here - it'll help readability for people new to the codebase.

cibuildwheel/macos.py
```diff
@@ -26,8 +28,13 @@ def call(args, env=None, cwd=None, shell=False):
...
 
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'identifier', 'url'])
+class PythonConfiguration(NamedTuple):
+    version: str
+    identifier: str
+    url: str
+
+
+def get_python_configurations(build_selector: Callable[[str], bool]) -> List[PythonConfiguration]:
```
joerick, line 37: BuildSelector here?

YannickJadoul:
Alright. I'll see if I can fit in some time to rebase today or tomorrow. Still want to/need to find some time to decently review #334 as well.

cibuildwheel/windows.py
```diff
@@ -3,37 +3,45 @@
...
-def get_python_configurations(build_selector):
-    PythonConfiguration = namedtuple('PythonConfiguration', ['version', 'arch', 'identifier', 'url'])
+class PythonConfiguration(NamedTuple):
+    version: str
+    arch: str
+    identifier: str
+    url: Optional[str]
+
+
+def get_python_configurations(build_selector: Callable[[str], bool]) -> List[PythonConfiguration]:
```
joerick, line 44: BuildSelector?

cibuildwheel/linux.py
YannickJadoul: OK, yeah, that might be a good argument. Doing so.

CI.md
```diff
@@ -1,11 +1,11 @@
...
 
-|          | 3.5              | 3.6              | 3.7                                               | 3.8              |
-|----------|------------------|------------------|---------------------------------------------------|------------------|
-| Linux    | Travis CI        | CircleCI         | AppVeyor / GitHub Actions                         | Azure Pipelines  |
-| macOS    | Azure Pipelines  | CircleCI         | AppVeyor / Travis CI¹ / CircleCI / GitHub Actions | Azure Pipelines  |
-| Windows  | TravisCI         | Azure Pipelines  | AppVeyor / GitHub Actions                         | Azure Pipelines  |
+|          | 3.6                          | 3.7                                                | 3.8              |
+|----------|------------------------------|----------------------------------------------------|------------------|
+| Linux    | Travis CI / CircleCI         | AppVeyor / GitHub Actions                          | Azure Pipelines  |
+| macOS    | CircleCI                     | AppVeyor / Travis CI¹ / CircleCI / GitHub Actions  | Azure Pipelines  |
```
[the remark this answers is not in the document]
YannickJadoul: Not sure why that was originally, there, but yes. If it works for 3.6 on CircleCI, and works on other CIs for 3.7, then we should be safe!

YannickJadoul:
> Just a few notes. Then we'd need:

Thanks! Rebased, solved the conflicts, and fixed the final few things

> * some mypy config somewhere (equivalent to `--strict`), either in setup.cfg or in mypy.ini

Done. Added to `setup.cfg` since that already contains `flake8` config anyway. Or any reason to prefer `mypy.ini`?

There's [a bunch of other flags/config values to be set](https://mypy.readthedocs.io/en/stable/config_file.html), with various more errors/warnings. Nothing immediately catches my eye, but might be interesting to go over? Depends a bit on how strict you want to be; e.g. disallowing `Any` could be interesting in some places, to ensure annotations will be added in new code?

> * Type checking in CI. I guess either this is an explicit job, or maybe this would be convenient to roll it into flake8 https://pypi.org/project/flake8-mypy/

I was planning to just install `mypy` and run it in the same job (but different call) as `flake8`. Do you prefer `flake8-mypy`? To be fair, it's neat that it can be integrated, but I don't really see the advantage to add it as extra dependency (but it should be a minor/stable dependency, ofc).

YannickJadoul:
Hèhè, thanks! :-)

PR #318, opened 2020-04-09
Try Appveyor config workaround

joerick:
APPVEYOR_CONSOLE_DISABLE_PTY, as suggested in https://github.com/appveyor/ci/issues/3376#issuecomment-611314299


joerick:
Seems to be working. I'll disable the allowed_failure.

PR #319, opened 2020-04-10
Provide `package_dir` for setting subdirectory of project [alternative]

joerick:
I hope you don't mind @filips123 - I've taken this further. It's turned out to be quite a fundamental change! So maybe easier for me to do it.

Still to go are tests, but the comments in the previous PR have been addressed.


cibuildwheel/__main__.py
```diff
@@ -48,26 +48,34 @@ def strtobool(val):
...
-                              'run in Windows, and it will build and test for all versions of '
-                              'Python. Default: auto.'))
+                        help='''
+                            Platform to build for. For "linux" you need docker running, on Mac
+                            or Linux. For "macos", you need a Mac machine, and note that this
+                            script is going to automatically install MacPython on your system,
+                            so don't run on your development machine. For "windows", you need to
+                            run in Windows, and it will build and test for all versions of
+                            Python. Default: auto.
+                        ''')
```
joerick, line 66: These changes are mostly unrelated - I just realised that the way I had been writing long strings was a little crazy. (argparse reformats all text, it seems)

joerick:
Tests are now in! Ready for review.

YannickJadoul:
Will review, but I'll take my time for a thorough comparison, this weekend.

test/13_subdir_package/bin/before_build.py
```diff
@@ -0,0 +1 @@
+print('before_build.py executed!')
```
Czaki, line 1: I do not see any check for this.

joerick:
Thanks for the review!

> But maybe move this test to be one of first and then use this ability to reduce number of spam.c files? (in other PR)

I'm not sure what you mean... are you referring to #277 ?

Czaki:
yes. but I thin that for many test it can be used one sample package so maybe instead of of templates and dynamic write of files use this PR result.

test/13_subdir_package/bin/before_build.py
joerick: I don't check that it ran, that's covered by a different test. The test just verifies that it's looking up the correct path - if it didn't find the file at `{project}/bin/before_build.py`, we'd have a failed test.

As an aside, how would I test for this? It would be nice to be able to assert inside tests based on the stdout of cibuildwheel. I was thinking we could return a named tuple from run_cibuildwheel, but would pytest `capfd` work perhaps?

Czaki: I prefer to use fixture `capsys` as in `main_patform_test.py`

joerick: would that capture the output of a subprocess?

Czaki: From my experience yes.

cibuildwheel/linux.py
```diff
@@ -96,12 +96,18 @@ def build(options: BuildOptions):
         ('pp', 'manylinux_x86_64', options.manylinux_images['pypy_x86_64']),
     ]
 
+    if not os.path.realpath(options.package_dir).startswith(os.path.realpath('.')):
```
YannickJadoul, line 99: I guess this is strictly speaking only necessary on Linux because of the docker image?
But it does feel like the abstraction gets lost a bit by only enforcing it here, and not in `__main__.py`.

Then again, the old `project_dir` is now only relevant on Linux, I guess? Or does it still have a function on Windows/macOS?

cibuildwheel/macos.py
```diff
@@ -229,7 +228,7 @@ def build(options: BuildOptions):
             call(['which', 'python'], env=virtualenv_env)
 
             if options.before_test:
-                before_test_prepared = prepare_command(options.before_test, project=abs_project_dir)
+                before_test_prepared = prepare_command(options.before_test, project='.', package=options.package_dir)
```
YannickJadoul, line 231: `{package}` does seem a bit like it would be the name of the package rather than a directory? So why wouldn't this be `package_dir` (as we already call it in the arguments to cibuildwheel) ?
But we already had the same with `{project}` ofc, so maybe it's good to keep it consistent like this.

docs/options.md
```diff
@@ -348,7 +351,10 @@ CIBW_DEPENDENCY_VERSIONS: ./constraints.txt
 ### `CIBW_TEST_COMMAND` {: #test-command}
 > Execute a shell command to test each built wheel
 
-Shell command to run tests after the build. The wheel will be installed automatically and available for import from the tests. `{project}` can be used as a placeholder for the absolute path to the project's root and will be replaced by `cibuildwheel`.
+Shell command to run tests after the build. The wheel will be installed automatically and available for import from the tests. To ensure the wheel is imported by your tests (instead of your source copy), tests are run from a different directory. Use the placeholders `{project}` and `{package}` when specifying paths in your project.
+
+- `{project}` is an absolute path to the project root - the working directory where cibuildwheel was called.
```
YannickJadoul, line 356: Do we still want to advertise this?

test/13_subdir_package/bin/before_build.py
YannickJadoul: If it's easy enough, it can't hurt to check it again?

YannickJadoul:
> I'm not sure what you mean... are you referring to #277 ?

I would keep that separate, btw. We should try to solve this, but the difference between 12 test folders before this PR and 13 folders after isn't going to make a difference here. I do agree with the thing that somehow tests are less and less in a "logical" order (but that's also another discussion).

cibuildwheel/linux.py
joerick: 'project dir' does still have a function on Windows/macOS, but only as the working dir - it's the cwd when commands like CIBW_BEFORE_BUILD run and CIBW_ENVIRONMENT are executed...

I was wondering about checking this in `__main__.py`, but it just felt a little punitive. It's still documented as 'must be a subpath of cwd', as if it were a global rule. But I don't think it needs to be enforced that strictly.

cibuildwheel/macos.py
joerick: good point... yes, the option is called `package_dir`... how would you feel about renaming the command line option to `package`?

YannickJadoul: Hmmm, not too sure about that. I kind of like that it still refers to a directory, actually.

docs/options.md
joerick: Before putting this PR together I thought we could have completely removed it. But it is still necessary in CIBW_TEST_COMMAND, because that command is run from a different cwd, so there still needs to be a way to reference something that's inside the project but outside the package.

cibuildwheel/linux.py
YannickJadoul: > 'project dir' does still have a function on Windows/macOS, but only as the working dir

But we don't really change the working dir? So somehow, it makes sense that that's just where commands are executed. (But true, it's a good point that these are not executed from `package_dir`, indeed!)

Anyway, I'm happy with this as well. Just a minor remark, when seeing this.

joerick:
> To be fair, I don't see a lot of differences with #295. Maybe some extra cleaning up and the added test, or am I missing something fundamental?

No, that's all it is - I just took @filips123's PR and fixed a couple things.

test/13_subdir_package/bin/before_build.py
joerick: I've added these asserts using `capfd` - `capsys` didn't capture anything FYI.

docs/options.md
YannickJadoul: Ah, yes. I forgot the exact reasons, but fair enough, then we need to keep it :-/

test/13_subdir_package/bin/before_build.py
YannickJadoul: Somehow the output of the subcommand would need to be forwarded to `sys.stdout`/`sys.stderr` to have `capsys` work, I guess?
But that's apparently non-trivial to do: https://stackoverflow.com/questions/18421757/live-output-from-subprocess-command. Or you'd need to have the `subprocess.check_call` replaced by `subprocess.run` and return the output, which would also not be so clean anymore, since now the list of wheels is returned.
So hurray for `capfd`! :-)

cibuildwheel/macos.py
joerick: Probably best to keep this consistent with `{project}` then, as you say.

PR #320, opened 2020-04-10
Add back missing 'troubleshoot' function call

joerick:
Noticed this function wasn't called anywhere - I think it got lost in a bad merge on #256. This PR adds it back.


Czaki:
Any test for this?

joerick:
No, not currently.

PR #322, opened 2020-04-16
Travis-CI Example: Fix aarch64

ax3l:
Thank you for the amazing `cibuildwheel`! :rocket: :sparkles: 

The `arch:` `aarch64` is called `arm64` on Travis-CI.
https://docs.travis-ci.com/user/multi-cpu-architectures/

Otherwise this falls back to a generic Linux amd64 host.

### DIFF
--- a/examples/travis-ci-minimal.yml
+++ b/examples/travis-ci-minimal.yml
@@ -6,7 +6,7 @@ jobs:
     - services: docker
     # perform a linux ARMv8 build
     - services: docker
-      arch: aarch64
+      arch: arm64
     # perform a linux PPC64LE build
     - services: docker
       arch: ppc64le


joerick:
Thank you for the kind words ☺️

Good catch. Thanks for sending us a PR!

ax3l:
Anytime, thanks for making my deployment pipelines productive! :star_struck: 
[So many beautiful wheels! :100: ](https://pypi.org/project/openPMD-api/#files)

YannickJadoul:
@ax3l Is there a reason you have different macOS versions? For 3.5, I see a wheel with minimum 10.9, but for 3.6, 3.7, 3.8, you have two wheels, 10.13 and 10.15. This seems odd: if you have a 10.13 wheel, that should work for 10.15 as well.

Czaki:
If I good understand he compile c++ code separately without set MACOSX_DEPLOYMENT_TARGET. And produce both to use most modern compilers with best optimisation. 

So I suggest to read more about MACOSX_DEPLOYMENT_TARGET which can simplify build process.

ax3l:
> This seems odd: if you have a 10.13 wheel, that should work for 10.15 as well.

Jepp, I had the github action in place first which can only build for 10.15. Then I realized some of my colleagues are still at 10.14 and one even at 10.13, so I added another travis CI job to have them covered. will probably comment out the 10.15 build for the next release.

Czaki:
But if your code does not need c++17 you can build against 10.9. Set MACOSX_DEPLOYMENT_TARGET="10.9" environment variable before build libraries.

ax3l:
Thank you @Czaki, totally [overlooked this](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions). :hugs:

Czaki:
> Thank you @Czaki, totally [overlooked this](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#macos-and-deployment-target-versions). hugs
> 
> > To get C++11 and C++14 support, MACOSX_DEPLOYMENT_TARGET needs to be set to (at least) "10.9". By default, cibuildwheel already does this, building 64-bit-only wheels for macOS 10.9 and later.
> 
> This seams off, or I don't fully understand why it defaults to 10.15 on GitHub actions... Maybe because the python interpreter is build for 10.15?
> 
> Earlier in the docs:
> 
> > If not set, Python will set this variable to the version the Python distribution itself was compiled on (10.6 or 10.9, for the python.org packages), when creating the wheel.

yes - but target macos revision is set out of box for code called from `setup.py`. Version 10.15 was forced by `CIBW_BEFORE_BUILD: bash -x .github/library_builders.sh `  where binary files are build against current macos version. 

@YannickJadoul @joerick What you think? If the docs should be updated?

joerick:
I do not understand why the builds on Github were defaulting to 10.15, but (complete conincidence) I recently merged https://github.com/joerick/cibuildwheel/pull/306, which sets MACOSX_DEPLOYMENT_TARGET to 10.9 if not set. So the above confusion should be resolved in the next release of cibuildwheel.

Czaki:
> I do not understand why the builds on Github were defaulting to 10.15, but (complete coincidence) 

Because part of libraries are build outside `setup.py` so python cannot set minimum deployment target. And macos build by default against current system version.

joerick:
ah! thanks. So #306 would fix that as well, as long as that building is done inside CIBW_BEFORE_BUILD.

YannickJadoul:
Ah, yes, I had the same, indeed, having my compilation based on CMake rather than fully through `setup.py`. I happily blamed the custom build setup, though, rather than `cibuildwheel` ;-) But I'm quite happy `wheel` got patched and `cibuildwheel`'s got these docs to make me (and others) aware of this!

Czaki:
> ah! thanks. So #306 would fix that as well, as long as that building is done inside CIBW_BEFORE_BUILD.

Yes. But if Someone would like to build it one in steep before cibuildwheel. 
Unfortunately delocate  does not verify if included libs are in proper version. I start two PR https://github.com/matthew-brett/delocate/pull/61 and https://github.com/matthew-brett/delocate/pull/68 which comes from discussion from https://github.com/matthew-brett/delocate/issues/56 (there are two implementation for same problem). 

But my experience with macos and access to macos machines (only one laptop) I still do not have idea how to prepare proper tests.

Czaki:
> Ah, yes, I had the same, indeed, having my compilation based on CMake rather than fully through `setup.py`. I happily blamed the custom build setup, though, rather than `cibuildwheel` ;-) But I'm quite happy `wheel` got patched and `cibuildwheel`'s got these docs to make me (and others) aware of this!

But `wheel` patch still dos not deal with `delocate` gathered libraries.

YannickJadoul:
> But `wheel` patch still dos not deal with `delocate` gathered libraries.

Ah, yes, that's true.

Anyway, this isn't related to this PR anymore. So if there's still things to discuss, we should create a different issue.

ax3l:
> yes - but target macos revision is set out of box for code called from setup.py. Version 10.15 was forced by CIBW_BEFORE_BUILD: bash -x .github/library_builders.sh where binary files are build against current macos version.

Oh, that's good info. So if I am now setting `MACOSX_DEPLOYMENT_TARGET` globally that means I can be sure that the binary dependencies build in `CIBW_BEFORE_BUILD` will also know this? Just want to be 100% sure because `MACOSX_DEPLOYMENT_TARGET` is set on the "host" environment (and macOS is also build in the same context, without further starting or isolating the wheel build as we know it from the manylinux docker images, for instance).

Czaki:
> Oh, that's good info. So if I am now setting `MACOSX_DEPLOYMENT_TARGET` globally that means I can be sure that the binary dependencies build in `CIBW_BEFORE_BUILD` will also know this? Just want to be 100% sure because `MACOSX_DEPLOYMENT_TARGET` is set on the "host" environment (and macOS is also build in the same context, without further starting or isolating the wheel build as we know it from the manylinux docker images, for instance).

Yes. If you set it globally it would gather all builds, even this done before cibuildwheel (I build python independent libraries in such way).

ax3l:
Thanks a lot, same use case here. I'll stop side-tracking this issue now.

Czaki:
@ax3l On remark. I'm not sure if this works with `brew` installed things. 
If you would like to manually verify your wheel then you need to unpack it and check every `.dylib` and `.so` file with `otool -l` then check version showed in under `MACOSX_DEPLOYMENT_TARGET` or `LC_VERSION_MIN_MACOSX`key. (`delocate` collect libraries to hidden  dir)

ax3l:
Thanks, yes I compile from source for (statically linked) dependencies I ship.

PR #327, opened 2020-04-25
Update CPython 3.7 from 3.7.6 to 3.7.7

mayeut:
https://docs.python.org/3.7/whatsnew/changelog.html#python-3-7-7-final


mayeut:
This is simple housekeeping, I went ahead and merged.

PR #328, opened 2020-04-25
Update openssl patch from 1.0.2t to 1.0.2u on macOS

mayeut:
Update to latest 1.0.2 series (now EOL)
Unfortunately, CPython 3.5.4 doesn't seem to support 1.1.1g or I missed something: https://travis-ci.org/github/mayeut/patch-macos-python-openssl

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -74,7 +74,7 @@ def install_cpython(version, url):
         call(['sudo', 'installer', '-pkg', '/tmp/Python.pkg', '-target', '/'])
         # patch open ssl
         if version == '3.5':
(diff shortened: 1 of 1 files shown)


Czaki:
Here some patch for openssl 1.1 https://github.com/patch-exchange/openssl-1.1-transition/blob/master/python3/python-3.5.2-openssl110.patch

mayeut:
@Czaki,
Thanks but only the native module _ssl is rebuilt in my repo so this patch won't work (furthermore it seems it's for 1.1.0 series which is EOL as well).
I could find the correct patch in cpython GitHub (since the latest 3.5 source only support OpenSSL 1.1.1) but have no time to dig into this further.
Given that Python 3.5 EOL is planned on 2020-09-13, the effort is probably not worth it

Czaki:
I agree. Maybe best is to patch only python 3.6+ with newer ssl.

mayeut:
This is simple housekeeping, I went ahead and merged.

PR #330, opened 2020-04-29
Updating the description of C++17 support on macOS

### LINKED ISSUE #325: macOS target version and C++17
Here we go again. Another subtlety/weirdness on macOS target versions and modern C++. Mind helping me to clear this one out, @Czaki?

So, full story:
- I recently [updated the pybind11 dependency of my project to a new version](https://github.com/YannickJadoul/Parselmouth/commit/70c83bb0ff2f577010a804136c66ae0ff02ed09e), which allows to opt out of C++17's ["aligned new"](https://en.cppreference.com/w/cpp/memory/new/operator_new) and ["sized deallocation"](https://en.cppreference.com/w/cpp/memory/new/operator_delete).
- I started playing around with `MACOSX_DEPLOYMENT_TARGET`, seeing if I could lower it from 10.14 (which was necessary before because of the size deallocation) to 10.13, and [found I could now set it to 10.9](https://github.com/YannickJadoul/Parselmouth/commit/476e7483a1db59ccbb254f129d41711f9d114eba), even though [the docs say C++17 requires >=10.13](https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/) (while before ... I don't know, I needed 10.13 or 10.14.
- Fine, OK, I reckoned that maybe I'm still using a new compiler and maybe I don't really use any C++17-standard-library features, but just some STL and new compiler features. Just to be sure, I decided to test out on an old, pre-10.13 macOS version. Travis CI offers 10.10 as oldest version (first released in 2014, end of life in 2017; so this should be old enough to be pre-C++17 compilers? It's the closest I can easily get to 10.9 anyway, it seems, and should be way older than the 10.13/10.14 from the docs).
- I [made a repository today](https://github.com/YannickJadoul/Parselmouth-OSX-10.9), trying out the installation of an old and a new version, and, [it seems to work indeed](https://travis-ci.org/github/YannickJadoul/Parselmouth-OSX-10.9/jobs/679119747#L259) (hurray, good news!). The old version on my `stable` branch (from before the pybind11 update, that `wheel` labels `praat_parselmouth-0.3.3-cp37-cp37m-macosx_10_13_intel.whl`) doesn't install (of course, because the filename doesn't match the platform), does install when your rename it to `10_9_intel.whl` (also: of course, because you're lying to Python by changing the filename), but also [works](https://travis-ci.org/github/YannickJadoul/Parselmouth-OSX-10.9/jobs/679119747#L246)!?

So:
1. We should update our docs, right? Apparently it depends on the kinds of C++17 features you use, and if it's only compiler/STL features, you might be OK to target lower macOS versions? Or is my analysis wrong?
2. What the hell's going on with the supposedly incompatible wheel? Why does it work, even if `wheel` detects it's built against a newer macOS (version and labels it 10.13+)??

EDIT:
The `stable` branch is still using `cibuildwheel` 1.1.0 to build wheels, I believe. But that shouldn't matter, should it? The latest `wheel` normally still gets installed, and that's where @Czaki added the target version detection, if I remember correctly?

YannickJadoul:
What a mess.

See #325: it is sometimes possible to build C++17-only code with `MACOSX_DEOPLOYMENT_TARGET` set to a value lower than `"10.13"`.

It's horrible that there seems to be no better documentation around, and that we have to figure this out. But thinking about it further, it does make sense to me: you need a new compiler (e.g. to compile [structure bindings](https://en.cppreference.com/w/cpp/language/structured_binding)) and a new version of the standard template library (e.g., for [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional)), but if neither of these require support of the C++ runtime library, after compilation an older version of OS X/macOS can still be supported.

Is this formulation clear, or will it only attract more questions/issues/...? Any improvements or things that need to be rephrased? @Czaki @joerick 

What a mess.

closes #325

### DIFF
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -22,9 +22,11 @@ However, to enable modern C++ standards, the deploment target needs to be set hi
 
 To get C++11 and C++14 support, `MACOSX_DEPLOYMENT_TARGET` needs to be set to (at least) `"10.9"`. By default, `cibuildwheel` already does this, building 64-bit-only wheels for macOS 10.9 and later.
 
-To get C++17 support, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
+To get C++17 support, a minimal version of Xcode 9.3 is needed, requiring at least macOS 10.13. To be able to use new C++17 library features and link against the C++ runtime library, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
 
-For more details see https://en.cppreference.com/w/cpp/compiler_support and https://xcodereleases.com/: Xcode 10 needs macOS 10.13 and Xcode 11 needs macOS 10.14.
+However, if only C++17 compiler and standard template library features are used, that don't require a C++17 runtime library, it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`. To find out if this is the case, try compiling and running with a lower `MACOSX_DEPLOYMENT_TARGET`: if C++17 features are used that need a more recent macOS SDK, building the wheel should fail.
+
+For more details see https://en.cppreference.com/w/cpp/compiler_support, https://en.wikipedia.org/wiki/Xcode, and https://xcodereleases.com/: Xcode 10 needs macOS 10.13 and Xcode 11 needs macOS 10.14.
 
 ## Windows and Python 2.7
 
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -22,9 +22,9 @@ However, to enable modern C++ standards, the deploment target needs to be set hi
 
 To get C++11 and C++14 support, `MACOSX_DEPLOYMENT_TARGET` needs to be set to (at least) `"10.9"`. By default, `cibuildwheel` already does this, building 64-bit-only wheels for macOS 10.9 and later.
 
-To get C++17 support, a minimal version of Xcode 9.3 is needed, requiring at least macOS 10.13. To be able to use new C++17 library features and link against the C++ runtime library, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
+To get C++17 support, Xcode 9.3+ is needed, requiring at least macOS 10.13 on the build machine. To use C++17 library features and link against the C++ runtime library, set `MACOSX_DEPLOYMENT_TARGET` to `"10.13"` or `"10.14"` (or higher) - macOS 10.13 offers partial C++17 support (e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`); macOS 10.14 has full C++17 support.
 
-However, if only C++17 compiler and standard template library features are used, that don't require a C++17 runtime library, it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`. To find out if this is the case, try compiling and running with a lower `MACOSX_DEPLOYMENT_TARGET`: if C++17 features are used that need a more recent macOS SDK, building the wheel should fail.
+However, if only C++17 compiler and standard template library (STL) features are used (not needing a C++17 runtime) it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`. To find out if this is the case, try compiling and running with a lower `MACOSX_DEPLOYMENT_TARGET`: if C++17 features are used that require a more recent deployment target, building the wheel should fail.
 
 For more details see https://en.cppreference.com/w/cpp/compiler_support, https://en.wikipedia.org/wiki/Xcode, and https://xcodereleases.com/: Xcode 10 needs macOS 10.13 and Xcode 11 needs macOS 10.14.
 


Czaki:
For me this formulation is not clear. If I good understad your previosus message it is possible to use c++17 code on older machines if it does not use new features from standard library binaries?

YannickJadoul:
> For me this formulation is not clear. If I good understad your previosus message it is possible to use c++17 code on older machines if it does not use new features from standard library binaries?

Yes. That is what I think, at least. I nowhere found this documented, but I have compiled my library with C++17 code for macOS 10.9.

Czaki:
So from changed documentation I silt understand that I need 10.13 as deployment target.

YannickJadoul:
The first paragraph indeed says what was there before (because 10.13/10.14 will always work), but then I say:

> However, if only C++17 compiler and standard template library features are used, that don't require a C++17 runtime library, it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`.

Is that not clear?

Czaki:
My bad. I focus on line 25.

YannickJadoul:
Maybe it's confusing to first say 10.13/10.14, and only then 10.9?

But I want to try avoiding people opening too many issues saying that it doesn't work with 10.9, but first saying 10.13/10.14. Is that OK?

docs/cpp_standards.md
```diff
@@ -22,9 +22,11 @@ However, to enable modern C++ standards, the deploment target needs to be set hi
 
 To get C++11 and C++14 support, `MACOSX_DEPLOYMENT_TARGET` needs to be set to (at least) `"10.9"`. By default, `cibuildwheel` already does this, building 64-bit-only wheels for macOS 10.9 and later.
 
-To get C++17 support, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
+To get C++17 support, a minimal version of Xcode 9.3 is needed, requiring at least macOS 10.13. To be able to use new C++17 library features and link against the C++ runtime library, set `MACOSX_DEPLOYMENT_TARGET` to (at least) `"10.13"` or `"10.14"` (macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`; macOS 10.14 has full C++17 support).
```
joerick, line 25:
```suggestion
To get C++17 support, Xcode 9.3+ is needed, requiring at least macOS 10.13 on the build machine. To use C++17 library features and link against the C++ runtime library, set `MACOSX_DEPLOYMENT_TARGET` to `"10.14"` or higher. You can also try `"10.13"` - macOS 10.13 offers partial C++17 support; e.g., the filesystem header is in experimental, offering `#include <experimental/filesystem>` instead of `#include <filesystem>`).
```

docs/cpp_standards.md
```diff
@@ -22,9 +22,11 @@ However, to enable modern C++ standards, the deploment target needs to be set hi
...
 
-For more details see https://en.cppreference.com/w/cpp/compiler_support and https://xcodereleases.com/: Xcode 10 needs macOS 10.13 and Xcode 11 needs macOS 10.14.
+However, if only C++17 compiler and standard template library features are used, that don't require a C++17 runtime library, it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`. To find out if this is the case, try compiling and running with a lower `MACOSX_DEPLOYMENT_TARGET`: if C++17 features are used that need a more recent macOS SDK, building the wheel should fail.
```
joerick, line 27:
```suggestion
However, if only C++17 compiler and standard template library (STL) features are required (not needing a C++17 runtime) it might be possible to set `MACOSX_DEPLOYMENT_TARGET` to a lower value, such as `"10.9"`. To find out if this is the case, try compiling and running with a lower `MACOSX_DEPLOYMENT_TARGET`: if C++17 features are used that require a more recent deployment target, building the wheel should fail.
```

YannickJadoul:
Alright, I made some changes to those changes, keeping the "10.13 or 10.14" (to avoid confusion with the 10.13 before) and making sure "macOS SDK" is still mentioned (useful to look up stuff; often this is how the "deployment target" is called).

If you're happy with this, go ahead and merge!

joerick:
Just on the SDK thing- to my understanding of the terms, changing MACOSX_DEPLOYMENT_TARGET doesn't change the SDK - the SDK is the compiler and the set of headers etc. that are used to compile the software. MACOSX_DEPLOYMENT_TARGET changes how the compiler reacts to some availability macros of the APIs the program uses. Does that agree with your understanding? That's my reasoning for removing the SDK reference - on Mac, the only way to change SDK is to use a different version of Xcode.

YannickJadoul:
> Just on the SDK thing- to my understanding of the terms, changing MACOSX_DEPLOYMENT_TARGET doesn't change the SDK - the SDK is the compiler and the set of headers etc. that are used to compile the software. MACOSX_DEPLOYMENT_TARGET changes how the compiler reacts to some availability macros of the APIs the program uses. Does that agree with your understanding? That's my reasoning for removing the SDK reference - on Mac, the only way to change SDK is to use a different version of Xcode.

Ah, hmmm, no. I had understood that `MACOSX_DEPLOYMENT_TARGET` basically chooses the SDK. I had to download older SDKs for `MACOSX_DEPLOYMENT_TARGET` to have an effect. But maybe that was an effect of CMake or so. I'll see if I can dig up something. If not, then it might be better to remove it, indeed!

YannickJadoul:
Alright, seems you're right. Somehow CMake uses  `MACOSX_DEPLOYMENT_TARGET` to set flags `-mmacosx-version-min=...` and (if the SDK is available) `-isysroot ...` (I'm not entirely sure what the latter then does). But builds also fail for versions for which the SDK is not found (tested with `<filesystem>`).

@Czaki, the errors say that `<filesystem>`, and more specifically `std::filesystem::current_path()`, is only available from 10.15. Where did you find the info about 10.14?

YannickJadoul:
> @Czaki, the errors say that `<filesystem>`, and more specifically `std::filesystem::current_path()`, is only available from 10.15. Where did you find the info about 10.14?

OK, I find the same though, if I check the table on https://en.cppreference.com/w/cpp/compiler_support and https://en.wikipedia.org/wiki/Xcode#Toolchain_versions. As far as I'm concerned, I don't really care anymore, ftm. Merge if you want to, @joerick.

Czaki:
@YannickJadoul  I only test compilation on macos 10.14. I do not test running...

joerick:
I did a bit of spelunking in the clang libcxx, here are the macros that define availability (and presumably dictate the error messages you've been seeing):

https://github.com/llvm/llvm-project/blob/llvmorg-10.0.0/libcxx/include/__config#L1373

The header files e.g. `<filesystem>` reference these macros. 

https://github.com/llvm/llvm-project/blob/llvmorg-10.0.0/libcxx/include/filesystem#L261

It looks like the entire filesystem header has the same rule - introduced in macOS 10.15.

Might be useful in future?

joerick:
Either way, this LGTM. Thank you for contributing your hard-earned knowledge back to the docs, @YannickJadoul !

YannickJadoul:
> @YannickJadoul I only test compilation on macos 10.14. I do not test running...

Yeah, but already compilation fails, for me. I tested compiling on macOS 10.15 with `MACOSX_DEPLOYMENT_TARGET=10.14`, and that failed, because of what @joerick points to.

> I did a bit of spelunking in the clang libcxx, here are the macros that define availability (and presumably dictate the error messages you've been seeing):

These are indeed what I found as well, in my local include files.

> It looks like the entire filesystem header has the same rule - introduced in macOS 10.15.

But so, this was only introduced in 10.15? Then I don't get how @Czaki managed to compile on 10.15? Or what that changed in 10.15? The confusion of different version numbers is quite horrible, with macOS, Xcode, LLVM/Clang, Apple's version of LLVM/Clang, ...

Anyway, we can update later if we figure out. For now, it seems that if compilation succeeds, things are OK, and `wheel` now picks up the correct things, so I think we're pretty safe :-)

> Either way, this LGTM. Thank you for contributing your hard-earned knowledge back to the docs, @YannickJadoul !

Thanks; glad to have (mostly) worked things out :-) I could only make sense of it all thanks to all the hard work from before, though. So thanks again to @Czaki for looking along now, as well, here and in #325!

Czaki:
> But so, this was only introduced in 10.15? Then I don't get how @Czaki managed to compile on 10.15? Or what that changed in 10.15? The confusion of different version numbers is quite horrible, with macOS, Xcode, LLVM/Clang, Apple's version of LLVM/Clang, ...

I compile it on my Laptop with MacOS 10.14 system.

YannickJadoul:
> I compile it on my Laptop with MacOS 10.14 system.

Myeah, weird :-/ If you want to figure out more, go ahead; I'm happy to look along and test things. But I think it's OK now, and I personally do not feel like spending another day to make sense of all version numbers. We can always update later if people have a problem or if anyone knows better.

PR #337, opened 2020-05-04
clean after bump to pypy 7.3.1

Czaki:
I realize that some cleaning are not done in #304


YannickJadoul:
Ah, good catch. I thought I had looked, but forgot about those!

Czaki:
Ready.

YannickJadoul:
Rerunning failed `ppc64le` (though there's no chance it's actually broken).

Czaki:
pasaed

Issue #336, opened 2020-05-04
manylinux architecture / 1.4.0 regression ?


jlaine:
Hello!

First of all, thanks again for all the hard work that goes into `cibuildwheel`, I'm using it successfully to build wheels for a growing number of packages (`aiortc`, `aioquic`, `PyAV`, `pylibsrtp`, `pylsqpack`) and it's a lifesaver.

I noticed my Linux builds have been failing for the past couple of days which seems to coincide with the release of `cibuildwheel` 1.4.0. To be more specific, the `i686` builds fail when linking against third-party libraries, which seems to be due to an architecture mismatch. The following output puzzles me, it looks like the `i686` build is in fact using an `x86_64` Python?

Look at the name of the docker image ` quay.io/pypa/manylinux2010_i686` vs the build directory `build/lib.linux-x86_64-3.6`:

```
2020-05-03T22:11:49.7883626Z Status: Downloaded newer image for quay.io/pypa/manylinux2010_i686:2020-04-29-0e1afc5
2020-05-03T22:11:58.3519923Z 5143d57f1cac28cce074d70d375b2601829ae6ae4600a11deeef4ffd6f938512
2020-05-03T22:11:58.3529495Z + docker cp . cibuildwheel-9a166a0e-1e64-4228-af7f-ebfe503b747f:/project
2020-05-03T22:11:58.6905700Z + docker start cibuildwheel-9a166a0e-1e64-4228-af7f-ebfe503b747f
2020-05-03T22:11:58.9714592Z cibuildwheel-9a166a0e-1e64-4228-af7f-ebfe503b747f
2020-05-03T22:11:58.9727256Z + docker exec -i cibuildwheel-9a166a0e-1e64-4228-af7f-ebfe503b747f sh -c 'cat > /constraints.txt'
2020-05-03T22:11:59.0792362Z + docker exec -i cibuildwheel-9a166a0e-1e64-4228-af7f-ebfe503b747f /bin/bash
2020-05-03T22:11:59.1929182Z     + mkdir -p /output
2020-05-03T22:11:59.1946204Z     + cd /project
2020-05-03T22:11:59.1953500Z     + PYBIN=/opt/python/cp36-cp36m/bin
2020-05-03T22:11:59.1954499Z     + export PATH=/opt/python/cp36-cp36m/bin:/opt/rh/devtoolset-7/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[...]

PR #338, opened 2020-05-04
add linux32 to i686 builds

Czaki:
Docker exex does not use entrypoint when using exec

https://stackoverflow.com/questions/43723777/how-to-execute-the-entrypoint-of-a-docker-images-at-each-exec-command

found in #336

### DIFF
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -53,4 +53,4 @@ class TestSpam(TestCase):
         # See #336 for more info.
         bits = struct.calcsize("P") * 8
         if bits == 32:
-            self.assertEqual(os.uname()[4], "i686")
+            self.assertEqual(platform.machine(), "i686")
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -1,6 +1,8 @@
 from __future__ import print_function
 import os
+import platform
 import sys
+import struct
 from unittest import TestCase
 
 import spam
@@ -43,3 +45,10 @@ class TestSpam(TestCase):
             print("=[listdir]2", os.listdir(os.path.join(virtualenv_path, 'bin')))
         self.assertTrue(path_contains(virtualenv_path, sys.executable))
         self.assertTrue(path_contains(virtualenv_path, spam.__file__))
+
+    def test_uname(self):
+        if platform.system() == "Windows":
+            return
+        bits = struct.calcsize("P") * 8
+        if bits == 32:
+            self.assertEqual(os.uname()[4], "i686")
--- a/test/02_test/test/spam_test.py
+++ b/test/02_test/test/spam_test.py
@@ -49,6 +49,8 @@ class TestSpam(TestCase):
     def test_uname(self):
         if platform.system() == "Windows":
             return
+        # if we're running in 32-bit Python, check that the machine is i686.
+        # See #336 for more info.
         bits = struct.calcsize("P") * 8
         if bits == 32:
             self.assertEqual(os.uname()[4], "i686")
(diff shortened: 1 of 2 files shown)


YannickJadoul:
That is a bit too fast, @Czaki. We have only just found the problem, and were just discussing the best solution. I think (see https://github.com/joerick/cibuildwheel/issues/336#issuecomment-623396974) that we should not put `linux32` ourselves, but use run instead?

Czaki:
in run you create new container. It is bad for me. because IT will return control flow to bash script (which is put as string in code) or add more steps of initialisation which create problem with performance (for me)

YannickJadoul:
Why can we not call `run` multiple times, instead of exec now? Setting up a docker container does not take so long, does it? Ah, no, you can't reuse the same `copy` then?

But it worked before, and we still created a container before, so we can use `start`?

YannickJadoul:
(See https://github.com/joerick/cibuildwheel/issues/336#issuecomment-623405271)

test/02_test/test/spam_test.py
```diff
@@ -43,3 +45,10 @@ def test_virtualenv(self):
             print("=[listdir]2", os.listdir(os.path.join(virtualenv_path, 'bin')))
         self.assertTrue(path_contains(virtualenv_path, sys.executable))
         self.assertTrue(path_contains(virtualenv_path, spam.__file__))
+
+    def test_uname(self):
+        if platform.system() == "Windows":
+            return
+        bits = struct.calcsize("P") * 8
+        if bits == 32:
+            os.uname().machine == "i686"
```
jlaine, line 54: Should this be an `assertEqual`?

test/02_test/test/spam_test.py
```diff
@@ -43,3 +45,10 @@ def test_virtualenv(self):
             print("=[listdir]2", os.listdir(os.path.join(virtualenv_path, 'bin')))
         self.assertTrue(path_contains(virtualenv_path, sys.executable))
         self.assertTrue(path_contains(virtualenv_path, spam.__file__))
+
+    def test_uname(self):
+        if platform.system() == "Windows":
+            return
+        bits = struct.calcsize("P") * 8
+        if bits == 32:
+            self.assertEqual(os.uname()[4], "i686")
```
joerick, line 54: platform.machine() will do the trick - it uses the same data as os.uname()

Czaki: It is somehow documented?

jlaine: https://github.com/python/cpython/blob/b88cd585d36d6285a5aeb0b6fdb70c134062181e/Lib/platform.py#L935

It literally calls `uname().machine`

YannickJadoul: I guess just writing `uname().machine` is also fine, then, here. But I'd replace the `[4]`, at least.

Czaki: > https://github.com/python/cpython/blob/b88cd585d36d6285a5aeb0b6fdb70c134062181e/Lib/platform.py#L935
> 
> It literally calls `uname().machine`

This is implementation, not documentation. So it can change in future. 

> I guess just writing `uname().machine` is also fine, then, here. But I'd replace the `[4]`, at least.

Not work on python 2.7 https://docs.python.org/2.7//library/os.html#os.uname

joerick: This is as good as I can do - https://github.com/python/cpython/blob/2.7/Lib/platform.py#L1328

Czaki: Ok. Changed.

YannickJadoul: > Not work on python 2.7 https://docs.python.org/2.7//library/os.html#os.uname

_Grrrrrrm, **sigh**_

PR #339, opened 2020-05-04
Add PyAV, aiortc and aioquic to example projects

jlaine:


jlaine:
FYI I also use `cibuildwheel` for a less common usecase, to generate binary builds of FFmpeg and its dependencies:

https://github.com/PyAV-Org/pyav-ffmpeg

The motivation here is to have binary builds which can then be used to build PyAV's wheels. The FFmpeg build used to be triggered as part of the wheel build process, but it is very time consuming to rebuild it for each PR against PyAV, so I split it out.

YannickJadoul:
Oh, cool!

> https://github.com/PyAV-Org/pyav-ffmpeg
> 
> The motivation here is to have binary builds which can then be used to build PyAV's wheels. The FFmpeg build used to be triggered as part of the wheel build process, but it is very time consuming to rebuild it for each PR against PyAV, so I split it out.

This sounds quite interesting, but I don't fully understand why you need `cibuildwheel` here, since it doesn't sound like you're building against every separate version of Python? So just curious: what benefits does `cibuildwheel` then offer if you don't need Python to be set up?

joerick:
Cheers @jlaine!

jlaine:
There are a couple of reasons I use cibuildwheel for the FFmpeg builds:

- for Linux I need to use the manylinux docker images to ensure I use the right libc / toolchains. Using cibuildwheel is considerably simpler for me than writing all the plumbing to operate docker myself. The recent issue I reported for instance is something I most likely would never have debugged myself, my docker-fu is very weak!

- cibuildwheel gives me a consistent framework to manage each supported architecture for each platform

- as the end goal is to produce wheels, cibuildwheel gives me a trivial way to include a smoke test in my builds: I build a wheel of a "dummy" package which links against the built libraries. This shakes out problems such as missing headers or linker errors

YannickJadoul:
> There are a couple of reasons I use cibuildwheel for the FFmpeg builds:

Ah, right, I see. Thanks for the nice and lengthy explanation!
I'd never thought of such a use case, so always good to keep in mind during future development.

Issue #316, opened 2020-04-08
Things to do once Python 3.5 host support is dropped


YannickJadoul:
List of things we could potentially do once we require a minimal Python version of 3.6 to execute `cibuildwheel`.

- f-strings!
- `pathlib` (`pathlib` already exists in 3.5, but the [path protocol](https://www.python.org/dev/peps/pep-0519/#provide-os-fspathb) doesn't, so lots of conversions to `str` are required).

If we have typing:
- Variable annotation syntax (local variables and `NamedTuple`)

Czaki:
Remove `object` inheritance eg in `DependencyConstraints` (it should be done earlier)

What do you think about work with async (or threading) for speedup work (start installing next python before previous build wheel is finished)?

YannickJadoul:
> Remove `object` inheritance eg in `DependencyConstraints` (it should be done earlier)

That's Python 2 vs. 3? I coul quickly commit that to `master` without PR, actually, if you want?

> What do you think about work with async (or threading) for speedup work (start installing next python before previous build wheel is finished)?

Not sure. I still feel that keeping things as slmple as possible is what we should aim for with `cibuildwheel`. I personally would rather have this tool work twice as slow but consistently, than have random failures because the installation accidentally added something the wheel did not expect?

Czaki:
There are special options to be sure that some tasks not start before the one its depend on it finish. 
But I'm not sure if speed-up if worth of higher complexity.  

> That's Python 2 vs. 3? I coul quickly commit that to `master` without PR, actually, if you want?

This is simple code cleaning from unnecessary parts. It should be done in some moment.

YannickJadoul:
> Remove `object` inheritance eg in `DependencyConstraints` (it should be done earlier)

@Czaki: not exactly the largest commit ever, but... b8dbedcdbed30d5f11dcbc9dcbc68a73040900ab

PR #341, opened 2020-05-09
Use f-strings

mayeut:
Now that minimum python supported is 3.6, we can use f-strings !

for tests, `setup.py` are not modified because running older python versions.

I left out 2 `.format(...)` I didn't see any interest in changing for now. 

Remaining `.format(...)` are in `linux.py`:
- in prepare_command function
- in the build function for the input string

xref #316

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -38,7 +38,7 @@ def get_option_from_environment(option_name: str, platform: Optional[str] = None
       CIBW_COLOR.
     '''
     if platform:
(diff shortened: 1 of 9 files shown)


YannickJadoul:
Ha, great!

Were we also going to use `pathlib`(#316) or is that a different PR/discussion?

mayeut:
> Were we also going to use pathlib (#316) or is that a different PR/discussion?

I guess it makes sense to get `pathlib` in different PR.
Keep it simple ! easier to review.

joerick:
Thanks mayeut!

PR #342, opened 2020-05-11
Add BEFORE_ALL options

### LINKED ISSUE #54: A BEFORE_BUILD command that's run only once per build environment
What is the right way to build (make && make install) dependencies only once at the beginning? If I run a script in CIBW_BEFORE_BUILD, everything gets built for every py version etc.

Czaki:
This PR add `BEFORE_ALL` option which is executed once per host. 

It is for install non python dependencies id docker images  

Closes #54 

I'm not sure if name of option is good. 

I'm not sure if test should be introduced as 14 group (as in PR) or added to other place (speed vs granularity)


Czaki:
- @YannickJadoul @joerick  ?

joerick:
Hey @Czaki, this looks pretty good so far. I think the name is good. The only thing slightly confusing me is that the CIBW_BUILD_ALL command is executed outside of the environment specified in CIBW_ENVIRONMENT. 

This feels like quite a gotcha, especially if people are setting things like CFLAGS in their CIBW_ENVIRONMENT option. Could we run BEFORE_ALL inside the environment?

It's probably also worth documenting what Python version is available in BEFORE_ALL, since it's not a cibuildwheel python - I guess it's the CI image Python on macOS/Windows, and whatever is default for manylinux.

---

(another option would be to choose a cibuildwheel CPython to execute BEFORE_ALL with, but that's probably overcomplicating the issue)

Czaki:
Changed

docs/options.md
```diff
@@ -191,6 +191,27 @@ CIBW_ENVIRONMENT: "BUILD_TIME=$(date) SAMPLE_TEXT=\"sample text\""
...
 
+### `CIBW_BEFORE_ALL` {: #before-all}
+> Execute a shell command preparing common part for each wheel. 
+
+Shell command to prepare common part of project (ex. build libraries which does not depend on python).
+This option is added mainly for linux build, because linux wheels are build in isolated from host docker containers.
+
+The placeholder `{package}` can be used here; it will be replaced by the path to the package being built by `cibuildwheel`.
+
+For windows and macos `CIBW_BEFORE_ALL` python interpreter is same as `cibuildwheel` is run. For linux it is default interpreter for docker image.
```
joerick, line 202:
```suggestion
The version of Python available inside `CIBW_BEFORE_ALL` is not controlled by cibuildwheel. On Windows and macOS, it's whatever is available on the host machine. For Linux, this comes from the Docker image, which on manylinux, is something rather old. Consider using a Python inside `/opt/python`, e.g. `/opt/python/cp38-cp38/bin/python`.
```

docs/options.md
```diff
@@ -191,6 +191,27 @@ CIBW_ENVIRONMENT: "BUILD_TIME=$(date) SAMPLE_TEXT=\"sample text\""
...
+Platform-specific variants also available:<br/>
+ `CIBW_BEFORE_ALL_MACOS` | `CIBW_BEFORE_ALL_WINDOWS` | `CIBW_BEFORE_ALL_LINUX`
+
+#### Examples
+```yaml
+# build third party library
+CIBW_BEFORE_ALL: make -C third_party_lib
+
+# install system library
+CIBW_BEFORE_ALL: yum install -y libffi-dev
```
joerick, line 213:
```suggestion
CIBW_BEFORE_ALL_LINUX: yum install -y libffi-dev
```

YannickJadoul:
I was quite happy using the `[ -d dependency ] || build_dependency` workaround (and given how old that issue was, I guess most users were?), but I guess it could be useful sometimes. It feels quite a heavy PR, though?

Not entirely convinced about the name, though. There was a whole discussion in #54 about the subtleties of such an option, so just `CIBW_BEFORE_BUILD` feels too simple and ambiguous, taking the comments #54 into account.
Related to that: we already have a whole series of `BEFORE_...` or `..._COMMAND` options, so we really should be careful that these don't become too confusing.
Just a suggestion, but what about `CIBW_SETUP_ARCHITECTURE`, or so? I guess this is run for every architecture.

> This feels like quite a gotcha, especially if people are setting things like CFLAGS in their CIBW_ENVIRONMENT option. Could we run BEFORE_ALL inside the environment?

It might also make sense to nót run this in the environment, to me. `CIBW_ENVIRONMENT` can add stuff to `PATH` etc, which is then built before. It should still be easy enough to set some custom environment if needed in this command?

> It's probably also worth documenting what Python version is available in BEFORE_ALL, since it's not a cibuildwheel python - I guess it's the CI image Python on macOS/Windows, and whatever is default for manylinux.
> 
> (another option would be to choose a cibuildwheel CPython to execute BEFORE_ALL with, but that's probably overcomplicating the issue)

That's a good catch. Not sure what the disadvantages of setting a Python are, but
1) This increases repeatability/determinism
2) Running plain `python` in a `manylinux2010` images gives you Python 2.6.6 (!!!), and `python3` does not exist. This seems useless to me:

```$ docker run -it --rm quay.io/pypa/manylinux2010_x86_64
[root@1a743514dd4e /]# python
Python 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-23)] on linux2
>>> 
[root@1a743514dd4e /]# python3
bash: python3: command not found
```

YannickJadoul:
Not sure this is a good idea, but just to think out of the box: what if instead of adding yet another option, we would start providing some more information to the user, through custom environment variables, just like CIs do?

E.g., I now set `CIBW_ENVIRONMENT='ARCH=$(uname -i)'`, but we could provide `CIBW_ARCH` (maybe with a different convention than `CIBW_` to avoid confusion with options).

Similarly, we could provide `CIBW_PYTHON_TAG` and `CIBW_BUILD_NR` that users can check themselves in `CIBW_BEFORE_BUILD`? (e.g., `if [ "$CIBW_BUILD_NR" == "0" ]; then ...; fi` in the before build script).

I'm not entirely convinced about this myself, so that's why I'm bringing it up for discussion, but this would stop us from everytime reconsidering different cases where we need to add an option for every single use case. Somehow, `CIBW_BEFORE_BUILD` has a lot of flexibility already, including the ability for conditionals that allow to check if something needs to be run. Shouldn't there be "one-- and preferably only one --obvious way to do it."? :-/

cibuildwheel/linux.py
```diff
@@ -130,6 +130,26 @@ def build(options: BuildOptions) -> None:
...
+                call(
+                    ['docker', 'exec', '-i', container_name] + shell_cmd,
+                    universal_newlines=True,
+                    input='''
+                        PS4='    + '
+
+                        set -o errexit
+                        set -o xtrace
+
+                        {environment_exports}
```
YannickJadoul, line 143: Do we want to run `CIBW_ENVIRONMENT` multiple times?

So far, we were only doing this once, so doing it multiple times seems unexpected. So if getting the result from some command is expensive, this will be done twice?
We might have users depending on this, or not?

For example, I have `CCACHE_BASEDIR=$(python -c "import tempfile; import os; print(os.path.realpath(tempfile.gettempdir())))` in `CIBW_ENVIRONMENT`. This is not so expensive in my case, but I don't know what's out there for other users.

Czaki: > Do we want to run `CIBW_ENVIRONMENT` multiple times?
> 
> So far, we were only doing this once, so doing it multiple times seems unexpected. So if getting the result from some command is expensive, this will be done twice?
> We might have users depending on this, or not?

Are you shure? For me enviroment is executed at least one per python version

Czaki:
> I'm not entirely convinced about this myself, so that's why I'm bringing it up for discussion, but this would stop us from everytime reconsidering different cases where we need to add an option for every single use case. Somehow, `CIBW_BEFORE_BUILD` has a lot of flexibility already, including the ability for conditionals that allow to check if something needs to be run. Shouldn't there be "one-- and preferably only one --obvious way to do it."? :-/

In my case. I do not vendor libraries but download it using scripts. Currently I first build docker container and then run cibuildwheel. Writing whole logic to avoid repetition code execution is problematic. 



> That's a good catch. Not sure what the disadvantages of setting a Python are, but
> 
>    1. This increases repeatability/determinism
> 
>    2. Running plain `python` in a `manylinux2010` images gives you Python 2.6.6 (!!!), and `python3` does not exist. This seems useless to me:

For me this option is rather to run python independent code. (like make, wget etc). If you see other things, then which version of python you suggest?

cibuildwheel/linux.py
YannickJadoul: Ah, yes, my mistake. That's already since the previous refactoring, then.
Maybe we should warn people in the docs, though?

YannickJadoul:
> In my case. I do not vendor libraries but download it using scripts. Currently I first build docker container and then run cibuildwheel. Writing whole logic to avoid repetition code execution is problematic.

I understand, yes. But surely there's a way of checking whether a docker container is already built/library is already downloaded and built?
But again, I can see it can be useful. I'd just like to be convinced that it's worth another option to maintain, and that all the `CIBW_BEFORE_...` and `CIBW_..._COMMAND` options won't confuse users :-/

Czaki:
change to new test system

joerick:
All good discussion here. I'm still generally in favour of the new option, because it's always been a little clumsy to `yum install` inside CIBW_BEFORE_BUILD. Like, it felt _wrong_ to keep calling it on every iteration, and writing around it using a bash if-statement felt like fighting the framework.

My argument for this is 'make the simple things simple'.

Regarding the name, we did discuss some more precise options in the past. I seem to recall `CIBW_BEFORE_BUILD_FOR_EACH_ARCH`, but that would mandate separate builds for win32/win_amd64, which I don't think is appropriate. We could go for something `CIBW_BEFORE_EACH_MACHINE` but it just seems clumsy - `CIBW_BEFORE_ALL` seems to capture the spirit of what the user might want to happen - then the docs can clarify the complicated bits.

> That's a good catch. Not sure what the disadvantages of setting a Python are, but
> 
> 1. This increases repeatability/determinism
> 2. Running plain `python` in a `manylinux2010` images gives you Python 2.6.6 (!!!), and `python3` does not exist. This seems useless to me:
> 
> ```
> [root@1a743514dd4e /]# python
> Python 2.6.6 (r266:84292, Jun 20 2019, 14:14:55) 
> [GCC 4.4.7 20120313 (Red Hat 4.4.7-23)] on linux2
> Type "help", "copyright", "credits" or "license" for more information.
> >>> 
> [root@1a743514dd4e /]# python3
> bash: python3: command not found
> ```

Ahhh... So options here are:

- a) document it and let users work around it
- b) run this command in the same Python as the first build (likely 2.7)
- c) run using the latest version in `CIBW_BUILD` - we could reverse the build order to make this more convenient? (does it feel slightly counter-intuitive to build 3.8, then 3.7, then 3.6...?)

I kinda don't love any of those. So I'm inclined to be lazy and do (a). Since I think the main use-case would be `make` or `yum` anyway.

YannickJadoul:
Alright, then. I was mostly surprised to see the long discussion from #54 on a name cut short to something as simple/short as `CIBW_BEFORE_ALL`, but if it makes sense to everyone, why not.

> I kinda don't love any of those. So I'm inclined to be lazy and do (a). Since I think the main use-case would be make or yum anyway.

Definitely not b), as far as I'm concerned. Running the latest one (but without reversing the order; should be possible, no?) would make sense to me, but maybe it's not a necessity. It also might give users bad ideas, running stuff that should potentially be Python-specific here. So I'm fine with a).

In the end, both of these things will boil down to clear documentation (and maybe add some diagnostics if we get common complaints/issues?).

Czaki:
@YannickJadoul As argument to introduce this change you can see this project:
https://github.com/czaki/imagecodecs_build

here I build and upload to dockerhub new images on every release, because dependencies build 7-10 minutes and there are multiple of them. Of course I can introduce special logic, or build docker images locally (as it is done in linux_build_docker), but managing multiple Dockerfiles  is annoying and problematic.  

Introduce this step will simplify pipeline and reduce maintenance. 
And this is example of project with multiple dependencies which are python independent.  

https://dev.azure.com/bokota/imagecodecs/_build/results?buildId=520&view=results

cibuildwheel/linux.py
joerick, line 143: Just had an idea re. available python, maybe i was overthinking this. it's always there on the docker images anyway - why not just do this?

```suggestion
                        export PATH=/opt/python/cp38-cp38:$PATH

                        {environment_exports}
```

Am I missing an obvious downside?

Czaki: Hm. I do not see any. @YannickJadoul ?

YannickJadoul:
> @YannickJadoul As argument to introduce this change you can see this project:
> https://github.com/czaki/imagecodecs_build
> 
> here I build and upload to dockerhub new images on every release, because dependencies build 7-10 minutes and there are multiple of them. Of course I can introduce special logic, or build docker images locally (as it is done in linux_build_docker), but managing multiple Dockerfiles is annoying and problematic.

@Czaki I think I more or less understand, yes. But my original point was that you could just write `CIBW_BEFORE_BUILD="[ -f dependencies_installed ] || ./install_dependencies.sh && touch dependencies_installed". Or you could even move the check to `./install_dependencies.sh`. Which is arguably not much more complex than adding a new option?

But I'm not saying we should do it. I just don't want to rush this, given that it seemed from #54 this was more complex than you might think at first.

And my main concern is that we need to get this right so that we don't have a future point where we need something in between of `CIBW_BEFORE_ALL` and `CIBW_BEFORE_BUILD`. For example: why are we running this only once for Windows? Someone might come along and say they need to execute a command to compile once per architecture, and that we need a `CIBW_BEFORE_ARCHITECTURE`, etc. Which is also why I'm still not completely convinced about `CIBW_BEFORE_ALL` as a name, because it doesn't tell you a lot. So I wanted to think this through before rushing to get this in.

cibuildwheel/linux.py
YannickJadoul: Yeah, this could work.

Advantage: a decent Python version available.
Disadvantages that I now see: we need to update it with every Python release, and we can't just document that you get the plain docker image without any changes.

YannickJadoul:
Actually. Thinking about this a bit further: the only reason we need this is because of the docker images on Linux, right? Because on Windows and macOS, you'd just be able to run some commands before `cibuildwheel`. Or am I missing something?

So, alternative idea (haven't thought this through): could we make this a Linux-only option, just like `CIBW_MANYLINUX_..._IMAGE` is Linux-only (e.g., `CIBW_BEFORE_MANYLINUX`, or `CIBW_MANYLINUX_SETUP`)?

Disadvantage is that we lose a bit of uniformity/transparency between platforms which `cibuildwheel` does tries to provide (but to some degree these docker images already have broken this). Advantage would be it's clear we don't promise anything to the users (i.e., we're not setting up `MACOSX_DEPLOYMENT_TARGET` on macOS, etc) and that this is run before each docker image (and there's none of the above discussion on whether to run it for every architecture on Windows, etc). Not entirely convinced myself, though.

Czaki:
> Actually. Thinking about this a bit further: the only reason we need this is because of the docker images on Linux, right? Because on Windows and macOS, you'd just be able to run some commands before `cibuildwheel`. Or am I missing something?

Yes, this is reason. Make it available on all platform may simplify script, but base need is support for such action on manylinux.

YannickJadoul:
> Make it available on all platform may simplify script, but base need is support for such action on manylinux.

Yes, true. But most likely, if it involves `yum` or `make` or so and no Python, it won't be cross-platform compatible anyway? So from that perspective, it does make sense to only provide `CIBW_BEFORE_ALL_LINUX` (and maybe give a different name, then, like the `CIBW_BEFORE_MANYLINUX`). If the others are just plainly run, it would provide more clarity for users to just run it in CI directly than as part of cibuildwheel.

Any thoughts, @joerick?

joerick:
> Yes, true. But most likely, if it involves `yum` or `make` or so and no Python, it won't be cross-platform compatible anyway? So from that perspective, it does make sense to only provide `CIBW_BEFORE_ALL_LINUX` (and maybe give a different name, then, like the `CIBW_BEFORE_MANYLINUX`). If the others are just plainly run, it would provide more clarity for users to just run it in CI directly than as part of cibuildwheel.

I definitely see the argument here, but I don't necessarily agree that it would create a better feature for users. Having CIBW_BEFORE_ALL be cross-platform is a very simple thing to do, and (for people familiar with cibuildwheel options), less surprising than the platform specific version.

> And my main concern is that we need to get this right so that we don't have a future point where we need something in between of `CIBW_BEFORE_ALL` and `CIBW_BEFORE_BUILD`. For example: why are we running this only once for Windows? Someone might come along and say they need to execute a command to compile once per architecture, and that we need a `CIBW_BEFORE_ARCHITECTURE`, etc

Well, we'll never _need_ such a command (everything in BEFORE_ALL is awkward but possible in BEFORE_BUILD). In my opinion, your specific example is unlikely to be a problem, because if setup for different architectures are needed, that can be done upfront in BEFORE_ALL. (The 32/64bit builds are interleaved so there's no crossover point between 32 and 64 bit.).

More broadly, I see the slippery slope argument, but I think we as maintainers can resolve it by keeping in mind that we can just say no to such feature requests (if they are niche enough), because the workarounds you suggest e.g. `CIBW_BEFORE_BUILD="[ -f dependencies_installed ] || ./install_dependencies.sh && touch dependencies_installed"`. But this one seems broad enough to warrant library support. I recognise that's a subjective judgement call!

YannickJadoul:
Alright, that's about the amount of discussion I wanted to have before continuing and merging this in! :-)

I think we're still missing a line in `README.md` table, though?

YannickJadoul:
Ah, also, @Czaki, if you merge/rebase on master, could you make sure the order of the options is +- contistent, after #348 got merged?

Czaki:
Rebased with checked order

joerick:
I think this is nearly done - @Czaki can you take a look at the unit test failures here?

Czaki:
Ok. After fix typing I forgot that `None` is changed to empty string.

Czaki:
I do not know what happen with circleci

YannickJadoul:
> I do not know what happen with circleci

All CircleCI Linux builds are broken. I have no idea why, but it's not linked to this PR :-(

Czaki:
```
Python 3.6.10
pip 20.1.1 from /usr/local/lib/python3.6/site-packages/pip (python 3.6)
ERROR: poetry 1.0.9 has requirement importlib-metadata<1.2.0,>=1.1.3; python_version < "3.8", but you'll have importlib-metadata 1.6.1 which is incompatible.
ModuleNotFoundError: No module named 'virtualenv.seed.via_app_data'
.circleci/prepare.sh: line 5: venv/bin/python: No such file or directory
.circleci/prepare.sh: line 6: venv/bin/python: No such file or directory
.circleci/prepare.sh: line 7: venv/bin/python: No such file or directory
 ```

so it looks like there is poetry installed which block virtualenv installation. 

But currently I do not have time to debug this 
but maybe something like `$PYTHON -m pip uninstall poetry` to will fix it https://github.com/joerick/cibuildwheel/blob/master/.circleci/prepare.sh

joerick:
Thank you @Czaki !

PR #346, opened 2020-05-16
Update the versions of our dependencies.

mayeut:
Mostly to update manylinux images & check everything still works smoothely.

### DIFF
--- a/cibuildwheel/resources/constraints-python27.txt
+++ b/cibuildwheel/resources/constraints-python27.txt
@@ -4,7 +4,7 @@
 #
 #    bin/update_constraints.py
 #
-appdirs==1.4.3            # via virtualenv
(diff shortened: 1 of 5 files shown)


Czaki:
Hm. @joerick said that he will do this just before release.

mayeut:
Doing this just before release exposes us to discover breaking things just before a release.
I don't mean for the PR to be merged right away.
It can be rebased/updated every now and then until the next release where @joerick can merge updates only once.

I just wanted to check if things are still working smoothely with manylinux images updates and the easiest way to do this is to open a PR.

Feel free to close once the CI status is known.

joerick:
Cheers @mayeut, I've no problem merging these if there's something specific you want tested

PR #347, opened 2020-05-17
Pre commit

Czaki:
Add pre commit configuration https://pre-commit.com/ to reduce number of coomits like "fix flake8", "fix mypy"

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
 
       - run:
           name: Install flake8 & mypy
-          command: sudo python -m pip install flake8==3.8.1 mypy
+          command: sudo python -m pip install flake8 mypy
       - run:
           name: flake8
           command: flake8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
 
       - run:
           name: Install flake8 & mypy
-          command: sudo python -m pip install flake8 mypy
+          command: sudo python -m pip install flake8==3.8.1 mypy
       - run:
           name: flake8
           command: flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,6 @@ repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0
     hooks:
-    - id: check-json
     - id: check-yaml
     - id: check-merge-conflict
     - id: end-of-file-fixer
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
     - id: check-yaml
     - id: check-merge-conflict
     - id: end-of-file-fixer
+      exclude: cibuildwheel/resources/pinned_docker_images.cfg
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: ''  # Use the sha / tag you want to point at
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,6 @@ repos:
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: ''  # Use the sha / tag you want to point at
     hooks:
-    -   id: mypy
-        args: [--strict]
-        files: cibuildwheel/*
-        exclude: test|cibuildwheel/resources
+    - id: mypy
+      files: cibuildwheel|test
+      exclude: cibuildwheel/resources
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,22 @@
+repos:
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.1
+    hooks:
+    - id: flake8
+      pass_filenames: true
+      # this seems to need to be here in addition to setup.cfg
+      exclude: cibuildwheel/resources/
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    - id: check-json
+    - id: check-yaml
+    - id: check-merge-conflict
+    - id: end-of-file-fixer
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: ''  # Use the sha / tag you want to point at
+    hooks:
+    -   id: mypy
+        args: [--strict]
+        files: cibuildwheel/*
+        exclude: test|cibuildwheel/resources
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,4 +42,3 @@ jobs:
     - bash: |
         python -m pip install -r requirements-dev.txt pytest-azurepipelines
         python ./bin/run_tests.py
-
--- a/cibuildwheel/resources/pinned_docker_images.cfg
+++ b/cibuildwheel/resources/pinned_docker_images.cfg
@@ -19,4 +19,3 @@ manylinux2014 = quay.io/pypa/manylinux2014_ppc64le:2020-05-17-2f8ac3b
 
 [s390x]
 manylinux2014 = quay.io/pypa/manylinux2014_s390x:2020-05-17-2f8ac3b
-
--- a/cibuildwheel/resources/pinned_docker_images.cfg
+++ b/cibuildwheel/resources/pinned_docker_images.cfg
@@ -19,3 +19,4 @@ manylinux2014 = quay.io/pypa/manylinux2014_ppc64le:2020-05-17-2f8ac3b
 
 [s390x]
 manylinux2014 = quay.io/pypa/manylinux2014_s390x:2020-05-17-2f8ac3b
+
--- a/docs/extra.js
+++ b/docs/extra.js
@@ -2,4 +2,3 @@
 $('.rst-content pre')
     .prev('blockquote')
     .addClass('code-block-filename');
-


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
       - run:
           name: Install flake8 & mypy
-          command: sudo python -m pip install flake8 mypy
+          command: sudo python -m pip install flake8==3.8.1 mypy
```
YannickJadoul, line 12: Why is this pinned? I thought we had decided to not pin dev requirements (as having them updated automaticall is too noisy)?
I also can't imagine the version of `flake8` to matter so much.

Czaki: I other project I manage last flake8 upgrade introduce many about 10 errors. Erors not connected with PR, but from upgrade of flake8.

also pre commit has pinned flake8 
But I do not thing this change is mandatory.

YannickJadoul: I think it's quite nice to sometimes get updates. I'm sure we otherwise just forget. If those errors are easy, why not fix them in the PR. If they are hard to fix, the PR can still be merged, even if flake fails, and fixed (or ignored) in another PR/commit.

I thought we had discussed pinning dev-dependencies and decided against. So I don't like adding this now, without thinking further about how to handle this later. And then if we do it, we should probably do it for all. But again, really not a fan. If static analysis gets better, we should benefit from it.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+repos:
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.1
+    hooks:
+    - id: flake8
+      pass_filenames: true
+      # this seems to need to be here in addition to setup.cfg
+      exclude: cibuildwheel/resources/
```
YannickJadoul, line 8: Why is this? This is set in `setup.cfg`, so why doesn't it work; does it have to do with this `pass_filenames`? Can't we run this the same way as on the CI?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    - id: check-json
+    - id: check-yaml
+    - id: check-merge-conflict
+    - id: end-of-file-fixer
```
YannickJadoul, line 15: What do these do? We're not checking them on CI either.

azure-pipelines.yml
```diff
@@ -42,4 +42,3 @@ jobs:
     - bash: |
         python -m pip install -r requirements-dev.txt
         python ./bin/run_tests.py
-
```
YannickJadoul, line 45: Why? It's good to have a newline at the end of the code, no? We have that in all files, I thought?

.pre-commit-config.yaml
Czaki: there is problem in pre commit and it does not handle configuration from setup.cfg

.pre-commit-config.yaml
Czaki: yaml is checked. If there is syntax bug then CI will not start.  Other I think that are nice.

azure-pipelines.yml
Czaki: This is result of `end-of-file-fixer` I think that this is nice. Few tools expect last empty line (ex cat two files)

.pre-commit-config.yaml
YannickJadoul: Yes, I read that comment. But is it a known bug? Or does it get solved it you don't set `pass_filenames`? I don't want to repeat all configuration multiple times, if possible.

.pre-commit-config.yaml
YannickJadoul: But they aren't check on the CI. So if I don't use this pre-commit hook (or someone else), it will be triggered for you (or someone else) using the pre-commit hook?
Or is it only for new changes, and not for all files? That would be ok for me :-)

azure-pipelines.yml
YannickJadoul: > Few tools expect last empty line (ex cat two files)

Yes, exactly. That's why I was wondering why you are removing the last line?

YannickJadoul: Oh, nevermind. There still is a final new line! I looked at the wrong file (`appveyor.yml` instead of `azure-pipelines.yml`). Sorry!

cibuildwheel/resources/pinned_docker_images.cfg
```diff
@@ -19,4 +19,3 @@ manylinux2014 = quay.io/pypa/manylinux2014_ppc64le:2020-05-15-5acda28
 
 [s390x]
 manylinux2014 = quay.io/pypa/manylinux2014_s390x:2020-05-15-5acda28
-
```
YannickJadoul, line 22: This is generated. Then we need to fix the script.

.pre-commit-config.yaml
Czaki: It is not solved

.pre-commit-config.yaml
Czaki: It check only changed files. 
I do not see other yaml files than CI configuration.

.pre-commit-config.yaml
YannickJadoul: Ah, too bad :-( Do you have a reference/link to add, so that it can later be checked whether it is fixed?

And are the other lines of `setup.cfg` honored, like `ignore = E501,W503`?

.pre-commit-config.yaml
YannickJadoul: > It check only changed files.

OK, good. Then I don't really mind too much!

> I do not see other yaml files than CI configuration.

No, but there's also `check-json`. Not sure we have that much json.

.pre-commit-config.yaml
Czaki: https://github.com/pre-commit/pre-commit-hooks/issues/112

pre commit check files and pass it to proper task. So it ignore all settings connected with include/exclude path and use others.

.pre-commit-config.yaml
Czaki: I remove json check

YannickJadoul: No, it can stay as well, if useful to you. I'm just thinking how to keep these checks as much equal to the ones running in CI.

.pre-commit-config.yaml
YannickJadoul: Thanks! That makes sense. I guess that's OK, then, if it's just about file selection and if other changes are propagated.

.pre-commit-config.yaml
Czaki: In this project there is no configuration files stored in json. So their wrong formatting does not make CI fail. I copy this list from other project where this is important.

joerick:
Hm. Personally, I'm not sure this has huge value for me. But I'm happy to add it if it's useful for people, but I'd like maintenance to be super-low, so I don't like that we have split config between setup.cfg and this new file.  For example, if you merge master, mypy now has a more complicated config now since we are typechecking tests in non-strict mode. For flake8, is it just the ignored directory that has to be specified here? Will it read the other options from setup.cfg?

Czaki:
This reduce number of commits like "fix flake8", "fix mypy" etc. This will also reduce probabylity of bottleneck because to many jobs on CI was created. 

Only include/exclude should be repeated in pre-commit config. 
Also if you write `flake8 path/to/excluded/file` then flake8 will check it. And this is reason why it need to be repeated. 

I will check if passing `--strict` to mypy is needed. If I good remember this option was hard-coded in calling command

YannickJadoul:
> I will check if passing `--strict` to mypy is needed. If I good remember this option was hard-coded in calling command

No, but that's not the most important part. The important part is that it picks up the options from `setup.cfg`; I do agree with @joerick on that (similar to the comments I made about `flake8`, but I somehow missed this).

Czaki:
> > I will check if passing `--strict` to mypy is needed. If I good remember this option was hard-coded in calling command
> 
> No, but that's not the most important part. The important part is that it picks up the options from `setup.cfg`; I do agree with @joerick on that (similar to the comments I made about `flake8`, but I somehow missed this).

For flake8 it picks all options except `files` and `exclude` (this I check previously). For mypy it should works same but have no idea how to verify it. @YannickJadoul may you suggest some error which should be cached by mypy with options from `setup.cfg` but will be ignored on default config?

YannickJadoul:
@Czaki, try a few of these? Or set another one to a non-default value?

For example, what about adding `strict_equality=False` and then adding a line `x: int = None`?

YannickJadoul:
Somehow, you need to add `strict_optional = False` to the `[mypy]` section, and not to `[mypy-cibuildwheel]`, I just found, playing around. Not sure why, but ...

Czaki:
I check it and changing of parameters in `setup.cfg` are applied both to pure mypy call and pre-commit call. 
So I can confirm that only files/exclude need to be separately updated in `.pre-commit-config.yaml`

.circleci/config.yml
joerick: If we're gonna pin, let's do it in dev-dependencies. But I'm okay leaving it unpinned, for the reason yannickjadoul mentions.

cibuildwheel/resources/pinned_docker_images.cfg
joerick: Agreed. This file is generated in `bin/update_dependencies.py`. I don't see an option to disable this last newline in `configparser`. Perhaps this hook could be removed also?

Czaki:
I exclude `cibuildwheel/resources/pinned_docker_images.cfg` from hook.

PR #348, opened 2020-05-17
Making order of options more consistent in docs and BuildOptions fields

YannickJadoul:
Reducing inconsistencies about the order in which options are applied.

Arguably, `CIBW_TEST_COMMAND` could be the first one withing the "Testing" section (as it's the central goal, and the others, `CIBW_TEST_REQUIRES`, `CIBW_TEST_EXTRA`, and `CIBW_BEFORE_TEST` wouldn't be there otherwise), though chronologically, it's the last thing happening.

Any thoughts? Is this too nitpicky?


joerick:
Personally I'd rather the CIBW_TEST_COMMAND is at the top of the Testing section in the docs - other than that these changes make sense :)

YannickJadoul:
Alright, agreed :-)

Really a minor PR, now that I look at it again, but sometimes it feels nice to clean up ;-)

Issue #351, opened 2020-05-25
Wheels built for linux cp27mu can't be imported


mayeut:
Issue originally reported to https://github.com/pypa/auditwheel/issues/243:

> Since a few days, binary wheels generated for Python2.7 on Travis CI do no longer work.
When importing, the error message undefined symbol: PyUnicodeUCS2_FromStringAndSize appears

The issue has been tracked down to `pip` (please refer to the original issue).

Since `cibuildwheel` now pins its dependencies, I thought I'd also report it as a bug here.

Test case in PR #352.

mayeut:
Fixed by #352

PR #334, opened 2020-05-03
Refactor tests to generate projects from template

joerick:
Intended to:

- reduce code duplication
- make differences between test projects easier to review
- simplify our test setup

This follows on from #277. Unfortunately it involves moving a lot of code around! But I think it's an improvement overall. I've used pathlib and type annotations quite heavily, so I've dropped 3.5 support here, too.

PR #352, opened 2020-05-25
Add test case for proper build isolation

mayeut:
If there's not a proper build isolation then, the wheel for `cp27mu` will pick up the `spam.so` built for `cp27m` resulting in an import error.

### DIFF
--- a/cibuildwheel/resources/constraints-python27.txt
+++ b/cibuildwheel/resources/constraints-python27.txt
@@ -15,12 +15,12 @@ importlib-resources==1.5.0  # via virtualenv
 pathlib2==2.3.5           # via importlib-metadata, importlib-resources, virtualenv
 scandir==1.10.0           # via pathlib2
 singledispatch==3.4.0.3   # via importlib-resources
-six==1.14.0               # via pathlib2, virtualenv
+six==1.15.0               # via pathlib2, virtualenv
 typing==3.7.4.1           # via importlib-resources
-virtualenv==20.0.20       # via -r cibuildwheel/resources/constraints.in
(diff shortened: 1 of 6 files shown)


mayeut:
The dependency update shall fix #351 by updating manylinux images.
Please note that `manylinux2014` does not have the [`pip` update yet](https://github.com/pypa/manylinux/pull/612) (`s390x` fails either because of travis-ci or upstream mirrors....) but is not impacted because it can't build python cp27 wheels anyway.

joerick:
Released as 1.4.2 :)

YannickJadoul:
Cool to see the new tests in action, like this! :-)

One thing we should perhaps be aware of in the long run is the time that each new integration test takes, though?

joerick:
> One thing we should perhaps be aware of in the long run is the time that each new integration test takes, though?

Yeah, good point. @mayeut is this test mostly useful on the `mu`/`m` linux builds? Might be worth adding `CIBW_BUILD=cp27-*` to that to save a bit of CI run time.

mayeut:
@joerick, `CIBW_BUILD=cp27-*` might be a good idea for this test yes (I would image the time to build the other wheels fairly small but I've not checked that).

YannickJadoul:
I seem to remember the testing times were more than expected, particularly in the cases with a test-phase? Ah yes, here: https://github.com/joerick/cibuildwheel/pull/334#issuecomment-623137531

But we might want to think how to improve this in general. It wasn't per se a comment on this PR. It has just become very (too?) easy to add integration tests, now ;-)

PR #354, opened 2020-05-27
Better help for GitHub Actions + Py27 workaround

henryiii:
Mentioned in #333 discussion.

### DIFF
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -40,5 +40,30 @@ Forcing `distutils` or `setuptools` to use a more recent version of MSVC that su
 2. Set up the build Visual Studio build environment you want to use, making sure that e.g. `PATH` contains `cl`, `link`, etc.
     - Usually, this can be done through `vcvarsall.bat x86` or `vcvarsall.bat x64`. The exact location of this file depends on the installation, but the default path for VS 2019 Community (e.g. used in AppVeyor's `Visual Studio 2019` image) is `C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat`. **Note**: `vcvarsall.bat` changes the environment variables, so this cannot be run in a subprocess/subshell and consequently running `vsvarsall.bat` in `CIBW_BEFORE_BUILD` does not have any effect.
     - In Azure Pipelines, [a `VSBuild` task is available](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/visual-studio-build)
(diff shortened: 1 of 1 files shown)


henryiii:
Also see https://github.com/microsoft/setup-msbuild/issues/13 - setup-msbuild is very specifically only a way to setup msbuild and not the rest of the env variables needed by distutils.

YannickJadoul:
Oops, sorry! I should have looked better, when adding that :-(

Does this work better, perhaps? https://github.com/ilammy/msvc-dev-cmd
I can't believe MS doesn't have a better way to set up the environment :-|

YannickJadoul:
> Does this work better, perhaps? https://github.com/ilammy/msvc-dev-cmd
> I can't believe MS doesn't have a better way to set up the environment :-|

YannickJadoul:
Sorry, wanted to comment but misclicked...

Was going to say I'd try myself, but I've not set up a GitHub Actions CI before, so it might be faster if you can try.

henryiii:
I can test that, I'm writing up docs anyway. They do mention the idea of having a setup action in the issue above, but they don't have one yet.

YannickJadoul:
Great, thanks! I'm really baffled there seems to be no official way to "activate" Visual Studio in GitHub Actions?

I also propose to maybe at the example code as a separate file to `examples/`, and just link to it? If we add this amount of example code for everything (i.e., not just for GitHub Actions, but also for Travis CI, etc), things might become messy.

henryiii:
Here's the example snippet:

```yaml
  - uses: ilammy/msvc-dev-cmd@v1

  - name: Build 64-bit wheel
    env:
      CIBW_BUILD: cp27-win_amd64
      DISTUTILS_USE_SDK: 1
      MSSdk: 1

  - uses: ilammy/msvc-dev-cmd@v1
    with:
      arch: x86

  - name: Build 32-bit wheel
    env:
      CIBW_BUILD: cp27-win32
      DISTUTILS_USE_SDK: 1
      MSSdk: 1
```

Only saw a minimal example, so didn't change that.

YannickJadoul:
@joerick Good to merge, if it's up to me. But maybe you want to have a say on whether and how to include the example?

joerick:
This looks good - would it also be appropriate to include that snippet after the `illamy` bullet point?

henryiii:
You mean the example snippet? I'd be happy to include it but wan't sure if it fit, since it's GHA only.

Is there a way to access `<details><summary></summary></details>` in the docs? That would be ideal :)

henryiii:
Added the snippet back in.

henryiii:
One travis build stalled, feel free to restart (or ignore, since this is documentation only).

joerick:
Thank you @henryiii!

Issue #350, opened 2020-05-24
Not reading `pyproject.toml` on Windows?


ajfriend:
That's for the awesome work on this project! I've been using it to build wheels for [our project here](https://github.com/uber/h3-py).

I'm able to build wheels across Windows/Mac/Linux. However, on Windows I need to specify `CIBW_BEFORE_BUILD: pip install scikit-build cython cmake` to get things to work. That doesn't seem to be necessary on Mac/Linux, as those build requirements are read from the `pyproject.toml` file. 

I'm guessing something is happening in the Windows flow that misses reading `pyproject.toml`?

For reference, my `cibuildwheel` commands for Windows on Github Actions look like:

```yaml
- name: Build wheels
  env:
    CIBW_SKIP: pp* cp27* *-win32
    CIBW_BEFORE_BUILD: pip install scikit-build cython cmake
    CIBW_TEST_REQUIRES: pytest numpy
    CIBW_TEST_COMMAND: pytest {project}/tests
  run: |
    pip install --upgrade cibuildwheel
    python -m cibuildwheel --output-dir wheelhouse
```

and for Mac, it's just:

```yaml
- name: Build wheels
  env:
    CIBW_SKIP: pp*
    CIBW_TEST_REQUIRES: pytest numpy
    CIBW_TEST_COMMAND: pytest {project}/tests
  run: |
    pip install --upgrade cibuildwheel
    python -m cibuildwheel --output-dir wheelhouse
```

henryiii:
Is this all versions of Windows, or just 3.5 (which will be the first one it tries)? I've noticed issues getting setuptools_scm to work with 3.5 on Windows, but I ended up just dropping 3.5 on Windows (and macOS) since those are probably quite rare. This might have indicated a pip version or setuptools version issue, though.

Czaki:
In my project `pyproject.toml` works. But I do not have python 3.5 build. So I also think that this is problem with setuptools/pip.

henryiii:
@ajfriend Could you try changing this line:

```diff
-    CIBW_SKIP: pp* cp27* *-win32
+    CIBW_SKIP: pp* cp27* *-win32 cp35-win_amd64
-    CIBW_BEFORE_BUILD: pip install scikit-build cython cmake
```

And dropping the `CIBW_BEFORE_BUILD` in a PR? Does that pass?

ajfriend:
That worked: https://github.com/uber/h3-py/actions/runs/114979866

[...]

PR #358, opened 2020-05-29
Adding PEP 518 test, fix 3.5 Python on Windows

henryiii:
As discussed in #350 

I'll fix the flake8 problem after the build. Wish this used pre-commit, I checked `pre-commit run` just to see.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -154,7 +154,7 @@ def setup_python(python_configuration: PythonConfiguration, dependency_constrain
     return env
 
 
-def pep_518_cp35_workaround(package_dir: str, env: Dict[str, str]) -> None:
+def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
     """
     Python 3.5 PEP 518 hack (see https://github.com/pypa/pip/issues/8392#issuecomment-639563494)
     Basically, nuget's Python is an embedded Python distribution, which is not supported by pip.
@@ -166,9 +166,9 @@ def pep_518_cp35_workaround(package_dir: str, env: Dict[str, str]) -> None:
     mostly "isolated".
     """
 
-    pyproject_path = os.path.join(package_dir, 'pyproject.toml')
+    pyproject_path = package_dir / 'pyproject.toml'
 
-    if os.path.exists(pyproject_path):
+    if pyproject_path.exists():
         data = toml.load(pyproject_path)
         requirements = (
             data['build-system'].get('requires', [])
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -149,25 +149,32 @@ def setup_python(python_configuration: PythonConfiguration, dependency_constrain
     shell(['pip', '--version'], env=env)
     shell(['pip', 'install', '--upgrade', 'setuptools', 'wheel'] + dependency_constraint_flags, env=env)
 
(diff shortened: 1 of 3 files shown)


YannickJadoul:
On the topic of pre-commit, there's a PR pending #347 ;-)

henryiii:
By the way, as predicted, this breaks on Python 3.5.4 + Windows, and works elsewhere!

henryiii:
It is possible this was a fix that went into 3.5.5, but it's really hard to find anything past 3.5.4 on Windows, since Python stopped providing binaries at 3.5.4, while the other platforms may have newer versions. Or it could be completely unrelated to the patches in Python.

YannickJadoul:
Nice! Thanks for clearing this up! :-)

Don't break your head over the Python 3.5 on Windows, though, I'd say. If it seems it's not a cibuildwheel thing, there's not a lot we can do (except report upstream if your debugging is successful, ofc). In a few months, 3.5 is end-of-life anyway, so ...

henryiii:
It's pretty easy to workaround, as well. It would be nice to simply say "3.5+" instead of adding caveats, and for it to work correctly out of the box. 3.5 on Linux is still around in a few places (Ubuntu 16.04, I last saw it on an old Raspbian), but I really doubt anyone is using it on Windows (or macOS).

Pip does check 3.5.4 on Windows in their testing matrix, though, which is quite odd that this is failing. I haven't checked their tests, but I know they at least test PEP 517 (this is technically 518, finally have the difference straight). It looks like on 3.5 + Windows it is not building in the environment pip set up for the build.

henryiii:
This is on a back burner for now for me; I bet it will need to be a report to pip, though, but not sure yet.

henryiii:
TODO: rename the test file 517 -> 518, as that's really what is being tested; the PEP 517 printout by Pip is only because that activates too.

YannickJadoul:
Hmmm, interesting, but quite annoying. If we could somehow pinpoint what goes wrong, we could apply a patch. But apart from that NuGet works quite well across different platforms, so it's probably not worth ditching it to solve this issue.

Is there a point in diffing the stdlib folders or so, between the two distributions?

henryiii:
I've patched for PEP 518 builds on Python 3.5. The problem is that the nuget distribution is based on the Embedded Python build. In 3.5, there wasn't a way to get it to behave like normal Python, and respect `PYTHON*` variables. This breaks pip's PEP 518 mechanism, which manipulates these variables. The embedded Python is not supported by pip (at least in embedded mode, which you can't disable on 3.5), so it's unlikely to be fixed by pip.

A workaround based on a suggestion by core Python and pip developer @pfmoore has been implemented, where the pyproject.toml requirements are injected into the prepared build environment when running Python 3.5 Windows builds. If this was eventually fixed, pip would start using build isolation like it does elsewhere, and the workaround does not affect the isolated environment.

@joerick, there is some flexibility in where this could go - I've added it to the setup function.

joerick:
Thank you for this @henryiii! This seems like a good workaround.

In terms of where it should go, `setup_python` probably isn't quite right, as it's doing stuff with the package. Actually, it should be using `options.package_dir` to find pyproject.toml (right now it's looking for a pyproject.toml in the working dir, but this should be `os.path.join(options.package_dir, 'pyproject.toml')`). So maybe we can create a function `pep_518_cp35_workaround(package_dir, env)` that is called after `setup_python` in the main build function?

joerick:
Thanks again @henryiii !

ajfriend:
This is great. Thanks all!

PR #362, opened 2020-06-01
example: for GH Actions use action to publish

### LINKED ISSUE #361: For GH Actions, use gh-action-pypi-publish instead of twine
There are some "problems"/tricks described when using `twine` with this action. 
Use dedicated GH action instead? 

```
- name: Publish to PyPI
  uses: pypa/gh-action-pypi-publish@master
  with:
    user: __token__
    password: ${{ secrets.pypi_password }}
    repository_url: https://test.pypi.org/legacy/ #TODO rm for real pypi
    packages_dir: ./wheelhouse/
```

breznak:
to PyPI, instead of calling twine directly.

Fixes #361

### DIFF
--- a/examples/github-deployment.yml
+++ b/examples/github-deployment.yml
@@ -35,3 +35,13 @@ jobs:
       with:
         name: wheels
         path: ./wheelhouse
+
+    - name: Publish to PyPI
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master #see https://github.com/pypa/gh-action-pypi-publish for setup
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        repository_url: https://test.pypi.org/legacy/ #TODO rm this line to use the "real" pypi
+        packages_dir: ./wheelhouse/
+
--- a/examples/github-deployment.yml
+++ b/examples/github-deployment.yml
@@ -1,4 +1,4 @@
-name: Build
+name: GH Deploy
 
 on: [push, pull_request]
 
@@ -36,7 +36,7 @@ jobs:
         name: wheels
         path: ./wheelhouse
 
-    - name: Publish to PyPI
+    - name: Publish to PyPI # (optional) automatically publish the whl to PyPI on a new tag/release
       if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@master #see https://github.com/pypa/gh-action-pypi-publish for setup
       with:
--- a/examples/github-minimal.yml
+++ b/examples/github-minimal.yml
@@ -37,6 +37,7 @@ jobs:
         path: ./wheelhouse
 
     - name: Publish to PyPI
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@master #see https://github.com/pypa/gh-action-pypi-publish for setup
       with:
         user: __token__
--- a/examples/github-minimal.yml
+++ b/examples/github-minimal.yml
@@ -35,3 +35,12 @@ jobs:
       with:
         name: wheels
         path: ./wheelhouse
+
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@master #see https://github.com/pypa/gh-action-pypi-publish for setup
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        repository_url: https://test.pypi.org/legacy/ #TODO rm this line to use the "real" pypi
+        packages_dir: ./wheelhouse/
+


YannickJadoul:
This is not a minimal example anymore. Could we just add this to a new example file `examples/github-deploy-only.yml` (cfr. `examples/travis-ci-deploy-only.yml`)?
Also, the Travis CI example only does this on tagged releases. Is it possible and easy to add to this example as well? Uploading every CI build to PyPI is not a good plan, as far as I know.

And just out of curiosity: what are the "problems/tricks" you mentioned in #361 that are avoided by this method?

breznak:
> Also, the Travis CI example only does this on tagged releases. Is it possible and easy to add to this example as well? Uploading every CI build to PyPI is not a good plan, as far as I know. 

sure, it's easily done and documented right here: 
https://github.com/pypa/gh-action-pypi-publish 
I didn't want to overcomplicate the example, but I can add the line to do that. 

> And just out of curiosity: what are the "problems/tricks" you mentioned in #361 that are avoided by this method?

https://cibuildwheel.readthedocs.io/en/stable/faq/#no-module-named-xyz-errors-after-running-cibuildwheel-on-macos

breznak:
> This is not a minimal example anymore. Could we just add this to a new example file  

sure, I can do that easily. 

But imho this is a part of a minimal & useful example. People typically would want to publish the new wheel to PyPI. And honestly, this is just a single step added.  
Imho, creating a second config for GH Actions would be more confusing (?)

joerick:
> But imho this is a part of a minimal & useful example. People typically would want to publish the new wheel to PyPI. And honestly, this is just a single step added.
> Imho, creating a second config for GH Actions would be more confusing (?)

The minimal example is [pulled into the docs](https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions) as a 'starting point' for configs, which, to be consistent with each other, show the minimum required to run cibuildwheel only. But we can link from that page to your example config showing how to upload to PyPI.

YannickJadoul:
> sure, it's easily done and documented right here:
> https://github.com/pypa/gh-action-pypi-publish
> I didn't want to overcomplicate the example, but I can add the line to do that.

Thanks! Given that this is a very sensible default (or stronger), I think it's good to include. This also nicely fits with the other examples we have.

> https://cibuildwheel.readthedocs.io/en/stable/faq/#no-module-named-xyz-errors-after-running-cibuildwheel-on-macos

Ah, yes, OK. I knew about this, but never made the link with `twine`, indeed

> But imho this is a part of a minimal & useful example.

Hmmm, not entirely sure. I still upload manually, just to be able to the build if there's an unexpected problem. Also, users might want to experiment first and only upload wheels afterwards. Or maybe they want to make wheels for private use. At any rate, it's not a core feature of cibuildwheel, so it's great to document but maybe not to include in the minimal example?

breznak:
> At any rate, it's not a core feature of cibuildwheel, so it's great to document but maybe not to include in the minimal example? 

ok, I'll revert and copy to an extended example.

breznak:
Ok, sorry for the hustle! :( Turns out the action is still only suppported on linux. 
https://github.com/htm-community/htm.core/runs/727475761?check_suite_focus=true#step:15:9 

There's a way I've been using (to upload & download artifacts and from linux to call the action), but for this example it's probably not worth it. 

Feel free to close this and the issue. Sorry for wasting your time.

henryiii:
That's actually the "correct" way to do it, ideally you want to make sure all jobs are successful before you upload, and you should upload once (there is now way to ignore existing in the action because of that).

YannickJadoul:
> Ok, sorry for the hustle! :( Turns out the action is still only suppported on linux.

Egh, that's annoying. We can keep this PR in mind for when it gets supported, though, so that your work wasn't for nothing. If you notice something, do feel free to let us know! :-)

(It's going to be quite interesting to see which one of all these newly create GitHub Actions is going to survive, in a few year, btw. I feel there's quite a few ad-hoc actions out there. But then again, this was from PyPA, so you'd think they'd support more than just Linux.)

breznak:
> . But then again, this was from PyPA, so you'd think they'd support more than just Linux.) 

yea, so was my thinking. The irony is it was me who brought the issue up a year ago. 
Ping to the issue for reference: https://github.com/pypa/gh-action-pypi-publish/issues/15

henryiii:
Couldn't this just be changed to the correct way to do it, with jobs and artifacts? It's not much longer, and doesn't risk uploading some wheels when other wheels break.

henryiii:
Also, if we had alternate arch support, like on Travis, then twine doesn't even work there.

henryiii:
This is reinvented by everyone using gha, see also https://github.com/htm-community/htm.core/blob/master/.github/workflows/release.yml#L17 from #364 - I think it should be in the examples and docs...

YannickJadoul:
> Couldn't this just be changed to the correct way to do it, with jobs and artifacts? It's not much longer, and doesn't risk uploading some wheels when other wheels break.

If something easily works, yes, we can make `examples/travis-ci-deploy-only.yml` (or `examples/travis-ci-upload.yml` or ...) and add it to the examples and link to it. I had understood it wasn't possible, but given your comment in #361, it would be nice to have that.

YannickJadoul:
@henryiii I've implemented your suggestion from #361; do you mind having a look?

Also, what did you mean by

> This is reinvented by everyone using gha, see also https://github.com/htm-community/htm.core/blob/master/.github/workflows/release.yml#L17 from #364 - I think it should be in the examples and docs...

Should something else still be added?

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - run: twine check wheelhouse/*
+
+    - uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        packages_dir: ./wheelhouse/
+        # To test: repository_url: https://test.pypi.org/legacy/
+      if: github.event_name == 'release' && github.event.action == 'published'
```
YannickJadoul, line 82: @henryiii I assume this is checking for git tags to only upload on a release, but what exactly do the two parts check/mean?

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Install Visual C++ for Python 2.7
+      if: startsWith(matrix.os, 'windows')
+      run: |
+        choco install vcpython27 -f -y
+
+    - name: Build wheels
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+
+    - uses: actions/upload-artifact@v1
```
breznak, line 34: nit: for a couple of these actions, a `@v2` is available and recommended.

examples/github-deploy-only.yml
breznak: I don't think this will work as you have `on: push, pull`
https://github.com/joerick/cibuildwheel/pull/362/files#diff-fe7f06db19278183fd8cb674f1009afcR3 

See https://github.com/htm-community/htm.core/actions/runs/122069673/workflow#L12-L16

examples/github-deploy-only.yml
YannickJadoul: At some point, there was a discussion: https://github.com/joerick/cibuildwheel/pull/194#discussion_r390019467

But feel free to update the ones that can be updated. It's your branch after all :-D But then we should also do this in the other example, and in cibuildwheels own CI configuration.

examples/github-deploy-only.yml
henryiii: You need to have `on: release` enabled too. You probably normally want to run this on tags & releases, and not on push and pull_requests. I don't have this fully tested in Scikit-HEP yet; several packages have adopted the pure-python version, but only one is on the cibuildwheel for now and hasn't had a release yet, but the idea was that only when a GitHub Release is made that you push to PyPI.

YannickJadoul: Oh, wait, so I change `on: push, pull` to `on: push, pull, release`?

Apart from that, we can merge it, and if you figure out more, could you let us know?

examples/github-deploy-only.yml
henryiii: Python v2 is easy, it just allows patch version pinning, which we don't use. You should be using upload/download v2, as they are simpler, will automatically merge, and don't even require a name (`artifact` is the default). Checkout v2 is the hardest, as it didn't support tag history at all until a week ago, and only supports tag history if you do a `fetch-depth: 0` checkout. If you use v2, `setuptools_scm` becomes trickier to use because of that. So not sure if that should change as the example yet. There's an ongoing discussion about how this could be improved here https://github.com/actions/checkout/issues/249.

examples/github-deploy-only.yml
YannickJadoul: Updated. Like this?

henryiii: Sure. I'll probably be trying to make a release before SciPy, and some of the other packages that have binary components may pick it up.

examples/github-deploy-only.yml
henryiii:
```suggestion
    - uses: actions/upload-artifact@v2
```
You should also drop the `name: wheels`, since this does not match the download below.

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+      with:
+        python-version: "3.8"
+
+    - name: Install dependencies
+      run: python -m pip install twine
+
+    - uses: actions/download-artifact@v2
+      with:
+        name: artifact
+        path: wheelhouse
```
henryiii, line 72:
```suggestion
        path: dist
```

Why not just put the wheels and sdist where twine expects them?

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - run: twine check wheelhouse/*
+
+    - uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        packages_dir: ./wheelhouse/
```
henryiii, line 80:
```suggestion
```
Then you can drop this line.

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - name: Install dependencies
+      run: python -m pip install twine
+
+    - uses: actions/download-artifact@v2
+      with:
+        name: artifact
+        path: wheelhouse
+
+    - run: twine check wheelhouse/*
```
henryiii, line 74:
```suggestion
    - run: twine check dist/*
```

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+    - uses: actions/checkout@v1
+
+    - uses: actions/setup-python@v1
+      name: Install Python
+      with:
+        python-version: '3.7'
+
+    - name: Build sdist
+      run: python setup.py sdist -d wheelhouse/
```
henryiii, line 51:
```suggestion
      run: python setup.py sdist
```

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - uses: actions/upload-artifact@v2
+      with:
+        path: wheelhouse/*.tar.gz
```
henryiii, line 55:
```suggestion
        path: dist/*.tar.gz
```

henryiii: Might as well leave this default too; the upload/download will put all these together for you anyway.

examples/github-deploy-only.yml
YannickJadoul: Thanks!

henryiii: Yes, that looks right.

examples/github-deploy-only.yml
YannickJadoul: Done! Thanks both :-)

examples/github-deploy-only.yml
YannickJadoul: I don't really know. This is where they were put in the minimal example as well, and in the examples from the start. Up to @joerick to explain if/why he prefers `wheelhouse` to `dist`?

examples/github-deploy-only.yml
YannickJadoul: Does this work if the wheels are still built in `wheelhouse`?

henryiii:
I should have made that an actual review, rather than several comments. If it helps, here is my 
wheel production https://github.com/scikit-hep/boost-histogram/blob/develop/.github/workflows/wheels.yml . It's tested except for the very final push to PyPY. You can download the artifacts and test them, for example, and it passes the twine check.

> Also, what did you mean by
> 
> ...
> 
> Should something else still be added?

No, just that most libraries that build multiple wheels (or eggs, not sure why setuptools_scm makes those) need this structure, to collect all the parts built and ensure that they all pass before uploading.

examples/github-deploy-only.yml
henryiii: I'm not recommending to move the wheels from wheelhouse, just to make the sdist in dist, and then to download all the wheels and the dist into dist before publish.

examples/github-deploy-only.yml
henryiii: Sure. This only uploads the dist.

examples/github-deploy-only.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+        choco install vcpython27 -f -y
+
+    - name: Build wheels
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+
+    - uses: actions/upload-artifact@v2
+      with:
+        path: ./wheelhouse
```
henryiii, line 36:
```suggestion
        path: ./wheelhouse/*.whl
```

You *might* need this to make sure it doesn't put the wheelhouse folder in the artifact, and to make sure these properly get merged. (I don't think so, but I've only tested it with the `*`)

examples/github-deploy-only.yml
henryiii: I think because they are wheelhouses? A wheelhouse contains *all* built wheels. If it had to build a wheel to make the current wheel, they get added here too. So ideally, you may want to do `wheelhouse/<package_name>*.whl` when you upload, do just get the part you want to distribute.

henryiii: A good way to check: require Pandas on Python 2.7. It's missing one of the types of manylinux wheels for 2.7 (`mu`?), so it will probably get built and added to the wheelhouse.

henryiii: And, no, Pandas has no interest in adding missing wheels or touching 2.7 at all. I've asked. :)

YannickJadoul: OK, I misunderstood, then. Thanks! Changed!

> And, no, Pandas has no interest in adding missing wheels or touching 2.7 at all. I've asked. :)

I can't blame them ;-)

examples/github-deploy-only.yml
YannickJadoul: Done as well. I guess it can't hurt, anyway. Let's stay as careful as possible, as these examples are not tested...

examples/github-deploy-only.yml
henryiii: > I can't blame them ;-)

I've used it to push (somewhat successfully) for some Pandas-dependent projects to be Python 3 only, so it's not only negative ;)

examples/github-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+        path: dist
+
+    - run: twine check dist/*
+
+    - uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        # To test: repository_url: https://test.pypi.org/legacy/
+      if: github.event_name == 'release' && github.event.action == 'published'
```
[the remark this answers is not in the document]
breznak: this workflow triggers on every push, pull, release : 
https://github.com/joerick/cibuildwheel/pull/362/files#diff-b4971e357cad581b8716b72ee38f58a5R3 

...which probably does not make sense, only tagged push & a release. 

Then this action filters only to "Release published"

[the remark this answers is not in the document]
joerick: I think people generally want CI to check that things build and to test resulting wheels. That's not what I'm asking - i'm wondering if pushing a tag creates a release in Github - such that this `if` statement evaluates to true.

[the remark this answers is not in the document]
breznak: > i'm wondering if pushing a tag creates a release in Github 

I think no, you have to have another action for that. Tag will show just as Tags.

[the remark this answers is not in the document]
henryiii: No, this should only release if you create a GitHub release. It will build on tags, since that's included by default in push, I think. 

```yaml
on:
  release:
  push:
    tags:
    - 'v*'
...
```

Will certainly build on both releases and tags. That way, you can tag, download and test the wheels, then release when you are ready. If you want to release on tags, this will push to PyPY even if you don't make a GitHub release.

```yaml
if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
```

Unlike Azure, there is no user interface for manually building so you can check wheels, so having the release mechanism stands in for that, and reduces your chances of making a bad release to PyPI. Depends on what you want to do.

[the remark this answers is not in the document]
henryiii: Like azure, though, if you upload artifacts, there's an option in the web interface to download them and play with them locally.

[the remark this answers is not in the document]
henryiii: I've also seen

```yaml
on: create
if: github.event_name == 'create' && github.event.ref_type == 'tag'
```

Suggested for tags, never tried it though.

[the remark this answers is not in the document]
joerick: Ah, cool. Gotcha. I'm more familiar with the tag-based approach, and I _think_ that is more common. Perhaps we could provide the user with a choice here? 

```suggestion
      # upload to PyPI on every tag starting with 'v'
      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
      # alternatively, to publish when a GitHub Release is created, use the following rule:
      # if: github.event_name == 'release' && github.event.action == 'published'
```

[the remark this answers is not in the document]
henryiii: That's fine - though I'd mention that `on: release` is needed too in the comment.

It's harder to do release based releases in non-GH tools, since they have to use the GitHub API to detect releases, while GHA has easy access to it.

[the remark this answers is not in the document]
YannickJadoul: Sounds all good to me

YannickJadoul:
OK, then I'm ready with this.

Merge when happy, @joerick. If @henryiii or @breznak learn more about the current example, we can still further update the example later :-)

henryiii:
Given https://github.com/pypa/gh-action-pypi-publish/pull/33 , we can simplify the example upload job by dropping the setup Python, install twine, and twine check, as it's now integrated into the gh action (only in master, but that's what is being used here anyway).

examples/github-deploy.yml
```diff
@@ -0,0 +1,83 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: "3.8"
+
+    - name: Install dependencies
+      run: python -m pip install twine
+
```
henryiii, line 67: These lines can be dropped now.

examples/github-deploy.yml
```diff
@@ -0,0 +1,83 @@
...
+    - uses: actions/download-artifact@v2
+      with:
+        name: artifact
+        path: dist
+
+    - run: twine check dist/*
```
henryiii, line 73: This line too.

henryiii:
I've made a release with a pure python package using a similar setup, and I think this should work, but here are some suggestions:

* Build only on published, full releases, rather than all release changes:
```yaml
release:
    types:
    - published
```

* Use the `jobs.<jobid>.if` key to disable/enable the whole job, rather than just the final publish (as it's only a download then publish anyway).

Pages [here](https://scikit-hep.org/developer/gha_basic) and [here](https://scikit-hep.org/developer/gha_wheels) have been updated.

examples/github-deploy.yml
```diff
@@ -0,0 +1,73 @@
...
+      with:
+        name: artifact
+        path: dist
+
+    - uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+        # To test: repository_url: https://test.pypi.org/legacy/
+      # upload to PyPI on every tag starting with 'v'
```
henryiii, line 70: This block should be moved to up to line 59.

YannickJadoul:
>    * Build only on published, full releases, rather than all release changes:
> 
> 
> ```yaml
> release:
>     types:
>     - published
> ```

Where should I add this?

breznak:
> Where should I add this? 

https://github.com/joerick/cibuildwheel/pull/362/files#diff-b4971e357cad581b8716b72ee38f58a5R3

YannickJadoul:
So ...

```
on:
  - push
  - pull_request
  - release:
      types:
        - published
```

... ?

YannickJadoul:
OK, done, both!

@henryiii Is there anything left to do? Or could we merge this soon?

examples/github-deploy.yml
```diff
@@ -1,6 +1,11 @@
 name: Build and upload to PyPI
 
-on: [push, pull_request, release]
+on:
+  - push
+  - pull_request
+  - release:
```
henryiii, line 6: I'd comment the release lines out then add above it:

```python
# alternatively, to publish when a GitHub Release is created, use the following rule:
```

The `push` should already trigger on all tags, so that's fine for the current push-on-tag example; this is for the push-on-release example that's commented out below.

I'm a fan of push on release because it a) Forces a GitHub release, which GitHub allows users to "watch". You don't get a notification when just a tag gets pushed. It also b) Lets you build on tag, verify it works, and then release on the existing tag - if it fails, you don't get a bad PyPI release, and you can move a GH tag if you need to.

henryiii:
Other than my one comment, I think it's ready. :)

examples/github-deploy.yml
joerick: Thank you @henryiii! Definitely hear you on this, I think deploy-on-release will be a great option for many projects. I just prefer build-on-tag as a more consistent experience across build providers.

Quick note on syntax- according to [this](https://github.community/t/array-of-configurable-trigger-events/16325) the contents of `on` is an object, not a list, so I -think- the syntax is

```yaml
on:
  push:
  pull_request:
  release:
    more_params: "something"
```

YannickJadoul: Huh, but before, we had `on: [push, pull_request, release]`? That's why I turned it into a list :-/

But yes, I'll change it!

henryiii: I'm fine with the deploy-on-release; I was just explaining why I think it's worth adding this as an alternative as a comment to an example.

I think it allows a list of strings or a mapping, so the original was fine, and this is reworked one is fine, but the list of one-mapping-each was probably incorrect.

YannickJadoul: Like this?

henryiii: What you have is fine, I would probably have done it like this:

```yaml
on:
  push:
  pull_request:
# alternatively, to publish when a (published) GitHub Release is created, add the following:
# release:
#   types:
#     - published
```

YannickJadoul: Yeah, this keeps more similarity to `example/github-minimal.yml`. That's why I went for this.

YannickJadoul:
OK, @joerick, ready if you're happy, then.

AppVeyor failing is related to current problems PyPy has with portably supporting macOS, it seems. See e.g. https://foss.heptapod.net/pypy/pypy/issues/3229.

I currently don't really have the time to investigate this, though. Not sure if there's a quick and dirty solution. And also not sure why this is suddenly breaking: we're still using the same PyPy 7.3.1 release and binary, so AppVeyor's image must have changed?

henryiii:
Since the other macOS builds are working, I would vote for an Appveyor change being the culprit.

YannickJadoul:
> Since the other macOS builds are working, I would vote for an Appveyor change being the culprit.

Yes, but it seems it's just exposing an issue in PyPy's linking to HomeBrew's `libffi`?

henryiii:
Yes, that's what it looks like.

YannickJadoul:
Let's see: #371

YannickJadoul:
Also, unrelated, but something weird seems to be happening to the `delocate` installation:

```
ERROR: delocate 0.8.0 requires bindepend, which is not installed.
ERROR: delocate 0.8.0 requires machomachomangler, which is not installed.
```

examples/github-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==1.4.2
+
+      - name: Install Visual C++ for Python 2.7
+        if: startsWith(matrix.os, 'windows')
```
webknjaz, line 31:
```suggestion
        if: runner.os == 'Windows'
```

examples/github-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Build sdist
+        run: python setup.py sdist
```
webknjaz, line 55: Instead of having a setuptools lock-in, this could use PEP 517 interface:
```suggestion
        run: |
          python -m pip install pep517
          python -m pep517.build --source --out-dir dist/ .
```

examples/github-deploy.yml
```diff
@@ -0,0 +1,78 @@
...
+    needs: [build_wheels, build_sdist]
+    runs-on: ubuntu-latest
+    # upload to PyPI on every tag starting with 'v'
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+    # alternatively, to publish when a GitHub Release is created, use the following rule:
+    # if: github.event_name == 'release' && github.event.action == 'published'
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
```
webknjaz, line 71: Since there's no way to download things using a wildcard, you could omit `name` arg and it'll download all the things. Then, it's pretty straightforward to mv dists into the `dist/` folder.

henryiii: If you add the name, it unpacks the folder, whereas if you don't, then it's in a nested folder. `upload-artifact@v2` combines automatically, so all the uploads go into the one artifact.

henryiii:
> Also, unrelated, but something weird seems to be happening to the `delocate` installation:
> 
> ```
> ERROR: delocate 0.8.0 requires bindepend, which is not installed.
> ERROR: delocate 0.8.0 requires machomachomangler, which is not installed.
> ```

This should only happen if you have delocate on Windows. Which I don't think you should be doing, since it is for relocating macOS wheels (unless it also supports Windows wheels, which I don't think it does)...

https://github.com/matthew-brett/delocate/blob/ef52b98b9f5e753a07cf5ce7495ff734187be3ad/setup.py#L31-L32

examples/github-deploy.yml
henryiii: Could we replace pip wheel builds with pep517? (does it work for classic setuptools packages, is the PEP 518 environment setup fine, would it fix the extra installed wheels in the wheelhouse issue)

I've seen it, and know that pip vendors it internally, but not sure how well it works standalone.

examples/github-deploy.yml
henryiii: Looks better to me.

YannickJadoul: Thanks, I'll still change that!

examples/github-deploy.yml
YannickJadoul: For just the minimal example, this feels a lot heavier though. Won't users also find the setuptools approach sooner? E.g., here: https://packaging.python.org/tutorials/packaging-projects/#packaging-your-project

And I suppose that once someone is not using basic `setuptools`, it's obvious enough this line needs to be replaced?

Or is this the new default way to build a source distribution that's meant to be spread and replace `setup.py sdist`?

examples/github-deploy.yml
YannickJadoul: Might be better to keep this like it is, then?

YannickJadoul:
> > Also, unrelated, but something weird seems to be happening to the `delocate` installation:
> > ```
> > ERROR: delocate 0.8.0 requires bindepend, which is not installed.
> > ERROR: delocate 0.8.0 requires machomachomangler, which is not installed.
> > ```
> 
> This should only happen if you have delocate on Windows. Which I don't think you should be doing, since it is for relocating macOS wheels (unless it also supports Windows wheels, which I don't think it does)...
>
> https://github.com/matthew-brett/delocate/blob/ef52b98b9f5e753a07cf5ce7495ff734187be3ad/setup.py#L31-L32

Yes, good point. But it is happening on the macOS build (that was failing, which is why I noticed this in the output).

But the lines

```
  Ignoring machomachomangler: markers 'sys_platform == "win32"' don't match your environment
  Ignoring bindepend: markers 'sys_platform == "win32"' don't match your environment
```

do not show for Python 2.7. So maybe something with an old `setuptools` version there?

examples/github-deploy.yml
henryiii: Yes, it won't work if you change it unless you add a copy step. It's better to avoid a step, IMO. Though you could add a nested folder location in the next step, I suppose, but I think this is simpler.

examples/github-deploy.yml
henryiii: I didn't think pep517 was really intended for users, at least yet, but as a backend for other tools, like pip. It's used internally to some extent by pip, I believe, and for the build command, there still is this note on the package readme:

> This ‘build’ module should be considered experimental while the PyPA decides on the best place for this functionality.

I don't think it's work putting into an example quite yet.

henryiii: Currently, setuptools is the only system I know of that produces binaries, and cibuildwheel builds binaries, so lock-in is not yet so important to avoid. ;)

YannickJadoul: Thanks, @henryiii. Is that OK, @webknjaz?
We can still easily change this later, once the Python docs update best practices, etc.

examples/github-deploy.yml
webknjaz: Oh, I guess when I implemented a similar thing previous time, it required the mv step because of v1. Will have to verify the new behaviour with v2.
OTOH I'm currently using different names in different workflows so I have a sync job for combining stuff anyway...

@henryiii do you know if v2 has any protection for race conditions when uploading with the same name?

Issue #357, opened 2020-05-28
manylinux build failing with unable to find image


davidism:
MarkupSafe Linux builds stopped working today with the error:

```
+ docker create --env CIBUILDWHEEL --name cibuildwheel-29afbfcb-daa8-4d72-af75-47fc97b09662 -i -v /:/host quay.io/pypa/manylinux2010_x86_64:2020-05-20-c7bbd11 /bin/bash
Unable to find image 'quay.io/pypa/manylinux2010_x86_64:2020-05-20-c7bbd11' locally
Error response from daemon: Get https://quay.io/v2/: EOF
+ docker rm --force -v cibuildwheel-29afbfcb-daa8-4d72-af75-47fc97b09662
Error: No such container: cibuildwheel-29afbfcb-daa8-4d72-af75-47fc97b09662
```

Config: https://github.com/pallets/markupsafe/blob/49ba2ffe23ee3a15ddab9e5c108d0b0dbbb0ca1e/.github/workflows/build.yaml
Failing run: https://github.com/pallets/markupsafe/runs/717426335#step:5:1

This sounds like it might be a temporary availability issue with Quay or manylinux, but I wanted to check if you had seen this or if this was something I needed to fix.

<details><summary>full log</summary>

```
2020-05-28T16:00:05.8693319Z 
2020-05-28T16:00:05.8695792Z      _ _       _ _   _       _           _
2020-05-28T16:00:05.8697930Z  ___|_| |_ _ _|_| |_| |_ _ _| |_ ___ ___| |
2020-05-28T16:00:05.8700121Z |  _| | . | | | | | . | | | |   | -_| -_| |
2020-05-28T16:00:05.8700251Z |___|_|___|___|_|_|___|_____|_|_|___|___|_|
2020-05-28T16:00:05.8700339Z 
2020-05-28T16:00:05.8700443Z cibuildwheel version 1.4.2
2020-05-28T16:00:05.8700566Z 
2020-05-28T16:00:05.8700685Z Build options:
2020-05-28T16:00:05.8700958Z   platform: 'linux'
2020-05-28T16:00:05.8703262Z   before_build: None
2020-05-28T16:00:05.8703698Z   before_test: ''
2020-05-28T16:00:05.8704167Z   build_selector: BuildSelector('*' - 'cp27-* pp27-* cp35-*')
2020-05-28T16:00:05.8704318Z   build_verbosity: 0
[...]

PR #366, opened 2020-06-02
Ignore errors when cleaning up docker container in linux.py

YannickJadoul:
See discussion in #357 and https://github.com/joerick/cibuildwheel/issues/357#issuecomment-635613056

Not entirely convinced this is the perfect solution, though, as we would also be ignoring other errors and mistakes in future commits (when e.g. a wrong container name is passed or so). One solution would be to still print the error message.

We could also make this more slightly more complex by setting `docker_image_created = True` after the `docker create` call, and only ignore the error if it's not created?

Tagging @Czaki after the discussion in #357.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -279,7 +279,10 @@ def build(options: BuildOptions) -> None:
             exit(1)
         finally:
             # Still gets executed, even when 'exit(1)' gets called
-            call(['docker', 'rm', '--force', '-v', container_name])
+            try:
+                call(['docker', 'rm', '--force', '-v', container_name])
+            except subprocess.CalledProcessError:
+                pass
 
 
 def troubleshoot(package_dir: str, error: Exception) -> None:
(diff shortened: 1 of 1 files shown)


Czaki:
I'm not sure if it is good solution as you mentioned
> (when e.g. a wrong container name is passed or so). 

Maybe change should be in troubleshoot with some text "check your docker images provides"?

joerick:
Another strategy (I think more idiomatic) would be to start the `try` block after the `docker create` - if `docker create` fails, it shouldn't have created anything that needs removing.

Czaki:
> it **shouldn't** have created anything that needs removing

If we are sure...

YannickJadoul:
Yes, that was my first idea, but @Czaki wasn't a big fan. On the other hand, we create unique names, so if something goes wrong, and half a container is created, I still don't think there's any problem?

Czaki:
Good point.

YannickJadoul:
Jup, all in all, none of this is a major issue; but if we notice it, we can just as well fix it.
What a huuuuuge PR, btw :-D

PR #369, opened 2020-06-06
Ensure dir name

Czaki:
If only directory name is passed to `pip wheel` then it try to find package with this name instead of of build wheel form directory. this path add "./" on begin of path if path is not absolute and not starts with dot.

### DIFF
--- a/test/test_subdir_package.py
+++ b/test/test_subdir_package.py
@@ -35,7 +35,7 @@ def test(capfd, tmp_path):
     project_dir = tmp_path / 'project'
     subdir_package_project.generate(project_dir)
 
-    package_dir = os.path.join(project_dir, 'src', 'spam')
+    package_dir = os.path.join('src', 'spam')
     # build the wheels
     actual_wheels = utils.cibuildwheel_run(project_dir, package_dir=package_dir, add_env={
         'CIBW_BEFORE_BUILD': 'python {project}/bin/before_build.py',
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -118,7 +118,7 @@ def main() -> None:
     test_command = get_option_from_environment('CIBW_TEST_COMMAND', platform=platform)
     test_requires = get_option_from_environment('CIBW_TEST_REQUIRES', platform=platform, default='').split()
     test_extras = get_option_from_environment('CIBW_TEST_EXTRAS', platform=platform, default='')
-    package_dir = os.path.abspath(args.package_dir)
+    package_dir = args.package_dir
     before_build = get_option_from_environment('CIBW_BEFORE_BUILD', platform=platform)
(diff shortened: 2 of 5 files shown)


YannickJadoul:
Auwtch. Why has the never been a problem before? Because we almost always just `pip wheel .`?

Could we maybe just call `pip wheel` with the absolute path (just calling `os.path.absolute` before `pip wheel`), rather than the current prepending of `./`? It seems nicer, somehow. Or is there a problem with that?

Is it easy to add a test for this? Or change another test to make sure it's being built rather than downloaded?

Czaki:
> Is it easy to add a test for this? Or change another test to make sure it's being built rather than downloaded?

> Could we maybe just call `pip wheel` with the absolute path (just calling `os.path.absolute` before `pip wheel`), rather than the current prepending of `./`? It seems nicer, somehow. Or is there a problem with that?

It looks better. I change it.

test/test_subdir_package.py
```diff
@@ -35,7 +35,7 @@ def test(capfd, tmp_path):
     project_dir = tmp_path / 'project'
     subdir_package_project.generate(project_dir)
 
-    package_dir = os.path.join(project_dir, 'src', 'spam')
+    package_dir = os.path.join('src', 'spam')
```
joerick, line 38: Not sure I agree with the spirit of this change - package_dir is relative to cwd, but cwd is changed within cibuildwheel_run. It should work either way, right? If so, the absolute path is more correct in my eyes

Czaki:
> I can't recall - is this a potential problem on Linux, too?

On linux there is such line which solve this problem. 
https://github.com/joerick/cibuildwheel/blob/31c9247921a1585b3b6dc78eb150327ff9c883dc/cibuildwheel/linux.py#L110

YannickJadoul:
> On linux there is such line which solve this problem.

Then we should make the solution as consistent as possible, no?

Czaki:
> Then we should make the solution as consistent as possible, no?

I do not understand? Linux has own path logic because it should be path in docker, not on machine. And path in docker is absolute path.

test/test_subdir_package.py
joerick: `cwd` is changed in the subprocess before the cibuildwheel is called. I hadn't realised that the default is relative, but the full path just makes more sense to me... not a big deal though if this change is important for some other reason.

YannickJadoul:
> I do not understand? Linux has own path logic because it should be path in docker, not on machine. And path in docker is absolute path.

I just meant it's nice if we can follow roughly the same order of getting the absolute path on all platforms. But it's fine if there's a good reason why not, of course.

Czaki:
> I just meant it's nice if we can follow roughly the same order of getting the absolute path on all platforms. But it's fine if there's a good reason why not, of course.

without moving control flow from shell script to python I do not see how to do this.

joerick:
Thanks @Czaki !

PR #371, opened 2020-06-08
Fix AppVeyor's PyPy failure on finding libffi

YannickJadoul:
Cfr. https://foss.heptapod.net/pypy/pypy/issues/3229 and https://foss.heptapod.net/pypy/pypy/merge_requests/726

PR #374, opened 2020-06-13
Simplify virtualenv test in test_testing

YannickJadoul:
I'm curious if this will allow us to get rid of the `__CIBW_VIRTUALENV_PATH__` environment variable we always set in the test virtualenv, with the only purpose of using it in our tests.

It's not brilliantly documented, but it seems a reasonable assumption and if anything changes in future `virtualenv` releases, this is much more likely to give false negative detections (and thus fail tests when they should not, rather than vice versa).

See #283

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -200,7 +200,6 @@ def build(options: BuildOptions) -> None:
                             pip install {dependency_install_flags} virtualenv
                             venv_dir=`mktemp -d`/venv
                             python -m virtualenv --no-download "$venv_dir"
-                            export __CIBW_VIRTUALENV_PATH__=$venv_dir
 
                             # run the tests in a subshell to keep that `activate`
                             # script from polluting the env
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -224,7 +224,6 @@ def build(options: BuildOptions) -> None:
                 os.path.join(venv_dir, 'bin'),
                 virtualenv_env['PATH'],
             ])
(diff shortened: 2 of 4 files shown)


YannickJadoul:
Ah, yes, if the CI passes: @Czaki, what do you think of this? Could this replace #283, or do you have objections?

Czaki:
@YannickJadoul I remember that I try with `base_prefix`, etc and do not find working one. Maybe some upgrades in virtualenv will fix it. 

Other option which i see Is that `PATH` variable is controlled by `cibuildwheel`. So maybe:

```
base_path = os.environ.get("PATH").split(os.path.pathsep)[0]
```

YannickJadoul:
> @YannickJadoul I remember that I try with `base_prefix`, etc and do not find working one. Maybe some upgrades in virtualenv will fix it.

@Czaki, yup, it seems like it now does work in virtualenv? The only failure is the problem with the PyPy linking (see #371).
 
> Other option which i see Is that `PATH` variable is controlled by `cibuildwheel`. So maybe:
> 
> ```
> base_path = os.environ.get("PATH").split(os.path.pathsep)[0]
> ```

No, I think your original solution was better then. Here you're just testing the same as what the code does, so I don't think it would make a great test.

But anyway, now this solution with `base_prefix` works, do you think we should merge it to replace the old test?

Czaki:
I think this could be replaced.

YannickJadoul:
Alright, thanks, @Czaki!

Up to @joerick to check then, but this could be merged.

joerick:
Nice bit of cleanup! :)

PR #375, opened 2020-06-14
Allow failures on Appveyor macos

joerick:
Temporarily allow this build to fail to get the tests green again


YannickJadoul:
Alternatively, we could disable the PyPy builds on AppVeyor, for now?

joerick:
this'll do - just a temporary thing

Issue #370, opened 2020-06-08
Support for building python3.9


fruch:
it's almost going out of beta, but we can't build for it using cibuildwheel

YannickJadoul:
Would be nice to have, yes! Is there a manylinux image containing it? And a nuget package?

YannickJadoul:
Alright, I can already answer the second part of my own question: https://www.nuget.org/packages/python/3.9.0-b1

YannickJadoul:
But manylinux doesn't seem to be available?

https://github.com/pypa/manylinux/blob/ac5df8a424d3f8743b8177be6722e609fd728d55/docker/build_scripts/build_env.sh#L5

See also pypa/manylinux#482

I think that's something to ask about upstream, if you want this in `cibuildwheel`? Once we have that, updating `cibuildwheel` should be reasonably painless.

joerick:
Beta support has been added! Check out #382 for instructions for how to use.

PR #377, opened 2020-06-16
Add Python 3.9 (beta 3) to builds

YannickJadoul:
Addresses #370; let's see if this already works, before changing docs, etc


YannickJadoul:
(Partially) blocked by https://github.com/pypa/manylinux/issues/629

Apart from that, you can start thinking if and when you'd like to make a release, @joerick: do we craft a cibuildwheel with a beta version of 3.9 such that users can start testing on 3.9, or do we wait until we at least have a rc?
(Apart from that, it would be nice if the PyPy troubles get solved as well :-/)

YannickJadoul:
Alright, Python 3.9 support!

Some caveat, though: s390x seems to be broken, also for pypa/manylinux, so it didn't get an updated manylinux2014 docker image, and so we can't update `pinned_docker_images.cfg`. So if in some hypothetical future, Travis CI's s390x would start working as expected, cibuildwheel will break (also if a new manylinux2014 image gets uploaded but users do not use `CIBW_MANYLINUX_S390X_IMAGE=latest`).

YannickJadoul:
On a second thought, it might be better to disable [the s390x Python 3.9 build identifier line](https://github.com/joerick/cibuildwheel/pull/377/files#diff-22075e43d4ff5efd0d10f3dcffdb93aaR85)? But then users might never use the release to build 3.9; also not when setting `CIBW_MANYLINUX_S390X_IMAGE` to the latest image. So what is the best solution?

joerick:
Hey, nice! I think we could release this, even in beta form... curious- does PyPI accept 3.9 wheels before release?

> On a second thought, it might be better to disable [the s390x Python 3.9 build identifier line](https://github.com/joerick/cibuildwheel/pull/377/files#diff-22075e43d4ff5efd0d10f3dcffdb93aaR85)? But then users might never use the release to build 3.9; also not when setting `CIBW_MANYLINUX_S390X_IMAGE` to the latest image. So what is the best solution?

Yeah, I guess disabling that build identifier line would be good. Though, I'm tempted to just remove s390x support entirely - we're not even testing that it works on s390x these days since the CI is so flakey. If it's important to somebody they can use an old version of cibuildwheel or help us add it back once manylinux has that bit working. But judging by the lack of progress we've seen so far from TravisCI, I'm not too hopeful.

I think I've talked myself round - I'd be in favour of removing s390x support for now.

Czaki:
> Hey, nice! I think we could release this, even in beta form... curious- does PyPI accept 3.9 wheels before release?

Hey, nice! I think we could release this, even in beta form... curious- does PyPI accept 3.9 wheels before release?
It looks like it accept

Czaki:
I forget to put link https://pypi.org/project/coverage/#files

joerick:
Then yeah, a release that includes 3.9b3 support would be great!

YannickJadoul:
Are we certain the ABI won't change anymore? If not, waiting for a rc might be a good idea?

Also, if we're making a release, we should think about what to do with the failing PyPy build?

YannickJadoul:
> I think I've talked myself round - I'd be in favour of removing s390x support for now.

Could we somehow make it optional? Or just comment it out until things get better?

See also: https://travis-ci.community/t/s390-xenial-builds-fail-with-an-error-occurred-while-generating-the-build-script/9039

YannickJadoul:
Funnily, it's now working, sometimes: https://travis-ci.org/github/joerick/cibuildwheel/builds/700019273

joerick:
> Are we certain the ABI won't change anymore? If not, waiting for a rc might be a good idea?

Good point - it seems we can't say it's ABI stable [until the first rc release](https://www.python.org/dev/peps/pep-0605/#example-announcement-text-for-3-9-0rc1). Then I don't think we should be building these by default, because it might cause users to upload ABI-incompatible wheels to PyPI, if the ABI changes between betas. Hiding behind a config seems to make sense.

Options here:

- `CIBW_INCLUDE_PRERELEASE_PYTHON=True` a new option that enables prerelease versions of Python. 

  Explicit, but requires the use of a new option, that only does something during Python release season.

- Changing the default of `CIBW_BUILD` to something like `cp27-* cp35-* cp36-* cp37-* cp38-* pp??-*`. Then just adding the beta Python, and not selecting it in this expression. 

    3.9 requires opt-in by changing CIBW_BUILD. But we might surprise some users that set CIBW_BUILD to `*` or `cp3*`.

- Publishing prerelease versions of cibuildwheel (e.g. v1.5.0b1) that include prerelease Python. Then users can choose to opt-in to that by specifying that version explicitly when they install using `pip`.

I'm thinking that published prerelease versions would make the most sense. What do you guys think?

Czaki:
> I'm thinking that published prerelease versions would make the most sense. What do you guys think?

agree

YannickJadoul:
Hmmm, interesting. `CIBW_BUILD` seems a bit tricky, though; I've thought about this before, maybe changing it to not build PyPy by default, but then you get into problems with people that do override the current default of `CIBW_BUILD=""`. Something like `CIBW_INCLUDE_PRERELEASE_PYTHON=True` seems like a better idea, then.

But yes, I think I would prefer the last option. It would make sense.

Another option/suggestion: we actually don't want to stimulate users to start building and uploading wheel with a non-final ABI to PyPI, I'd say? The main reason to add this now is to enable users to test building/running their code on Python 3.9 and to prepare ourselves to be there when a release candidate is there.
What if we only release `cibuildwheel` once we have a release candidate, but for now have an official `python3.9` branch that users can `pip install` if they want (based off of master or the latest release commit; I don't really mind either way)? Once again, it would be an explicit opt-in that we can mention in the docs, and it would be quite clear? Meanwhile, it's easy to later include in a following release?

Alternatively, I'm wondering whether we could disable giving users the wheel (unless they maybe opt-in) but still test on 3.9? Again, we want to avoid users publishing wheels to PyPI for now. But it's counter-intuitive to build 3.9 on `cibuildwheel` and nót get the built wheel, ofc.

joerick:
> have an official `python3.9` branch that users can `pip install` if they want

I like this idea! I'll create the branch

YannickJadoul:
> I like this idea! I'll create the branch

We can rebase this whenever there's a new release, I suppose?

Will you write a few lines in the README/docs, as well?

joerick:
Cool, so merge this when you're happy @YannickJadoul, and because I've changed the 'base' of this PR it'll go to a python3.9 branch in this repo.

> We can rebase this whenever there's a new release, I suppose?

I'd prefer to merge rather than rebase (multiple people might have the commits checked out), but we can keep it up to date with master, yes.

> Will you write a few lines in the README/docs, as well?

Yep!

YannickJadoul:
> I'd prefer to merge rather than rebase (multiple people might have the commits checked out), but we can keep it up to date with master, yes.

Alright, good point!

`pip install git+https://github.com/joerick/cibuildwheel.git@python3.9` is probably what we should tell users to run, btw.

But I did notice: it might be nice to have a different version number, in there? Now when I run this, it tells me it installed `cibuildwheel-1.4.2`, so I'm not sure how `pip` will react to these duplicate version numbers?
(On a second though: maybe it's actually fine, though, since you're giving `pip` an explicit version of the repository to install.)

YannickJadoul:
The first rc is supposed to be there in 2 months, btw: https://www.python.org/dev/peps/pep-0596/

joerick:
> But I did notice: it might be nice to have a different version number, in there? Now when I run this, it tells me it installed `cibuildwheel-1.4.2`, so I'm not sure how `pip` will react to these duplicate version numbers?

Hm. I am also not sure, but let's see if we get any error reports because of it. Luckily it's a clean install on CI so probably not a problem.

joerick:
I've added a note to the README 2b22ffda9bd6b8433b90e3dd1183c09a5632143c pointing to the tracking PR for python3.9 - #382

PR #378, opened 2020-06-16
Suppress warnings about pytest trying to collect TestProject

### LINKED ISSUE #372: PytestCollectionWarning: cannot collect test class 'TestProject' because it has a __init__ constructor
Shouldn't be too hard, but there's a warning in our test suite we should probably fix:

https://github.com/joerick/cibuildwheel/runs/732376574#step:6:40

```
 test/test_projects/base.py:10
  /home/runner/work/cibuildwheel/cibuildwheel/test/test_projects/base.py:10: PytestCollectionWarning: cannot collect test class 'TestProject' because it has a __init__ constructor (from: test/test_cpp_standards.py)
    class TestProject:

test/test_projects/base.py:10
  /home/runner/work/cibuildwheel/cibuildwheel/test/test_projects/base.py:10: PytestCollectionWarning: cannot collect test class 'TestProject' because it has a __init__ constructor (from: test/test_subdir_package.py)
    class TestProject:
```

If I understand correctly, pytest is trying to pick up `test_projects`/`TestProject` because it starts with test. We should probably just rename the class/directory or tell pytest to ignore it. Not sure what's the best way to do it, but I'll leave this here until someone has time to look into it, explore the different options, and make a PR (possibly myself, if I ever have some time and some of the other backlog gets cleaned up).

YannickJadoul:
Fixes #372

We (probably) need to use this `__test__ = False` attribute, because it's `from .test_projects import TestProject` that causes the warning. The `test_projects` subfolder is not the issue, since `base.py` does not have the `test_` prefix.

Alternatively, we could have a `new_base_project` function, similar to `new_c_project`?

### DIFF
--- a/test/test_projects/base.py
+++ b/test/test_projects/base.py
@@ -14,6 +14,8 @@ class TestProject:
 
     Write out to the filesystem using `generate`.
     '''
+    __test__ = False  # Have pytest ignore this class on `from .test_projects import TestProject`
+
     files: FilesDict
     template_context: TemplateContext
 


YannickJadoul:
@joerick The warnings are gone, so if this 4a335f4 fixes the unrelated Travis CI-Python 3.6-SSL error, this is ready for review (?) and merging.

YannickJadoul:
So Travis CI still fails, even with the 62f88a2 commit (amended 4a335f4, where I had forgotten to change `C:\Python36` into `C:\Python37`). Let's hope Travis fixes this, and roll back that second commit?

YannickJadoul:
@mayeut Since you keep on providing us with SSL patches for macOS, do you know anything about issues on Windows with SSL? Travis CI apparently uses Windows Server 1809, so maybe something is too old there?

This seems to be relevant, but I cannot get it to work on Travis: https://bugs.python.org/issue36011

mayeut:
@YannickJadoul,
the issue on Windows seems indeed to be the one you linked to.
It's unrelated to why a patch was needed on macOS for CPython 3.4/3.5 and I was unaware of the Windows one.

The method from http://www.chawn.com/RootCerts.htm should work. (also for reference, http://woshub.com/updating-trusted-root-certificates-in-windows-10/)

YannickJadoul:
Amazing! That works!
Thank you very much, @mayeut! Would have taken me a lot longer to figure out :-)

PR #382, opened 2020-06-22
Add python3.9 support

joerick:
Python 3.9 is in beta, and not yet ABI stable, so you shouldn't publish wheels with it yet.

But if you want to check that your wheels build on Python 3.9, try this `python3.9` branch!

To try this branch, in your CI config, replace 

    pip install cibuildwheel==?.?.?

with

    pip install git+https://github.com/joerick/cibuildwheel.git@python3.9


fruch:
@joerick thanks a lot, I'm taking this one for a test drive, when is the plan to merge this ?, only in October when python3.9 would be released formally ?

joerick:
The credit's @YannickJadoul's! We'll be merging this once Python3.9 is ABI stable, which is the first release candidate I believe.

YannickJadoul:
Might be worth rebasing onto 1.5.5, though?

joerick:
> Might be worth rebasing onto 1.5.5, though?

Good point, I'll get on that

joerick:
Also, Python 3.9.0b5 is out, we should update to that once manylinux gets it https://github.com/pypa/manylinux/pull/682

YannickJadoul:
Hurraaaaay, 3.9 rc1! Almost there, except for pypa/manylinux#710!

If this also works, what's the plan, then, @joerick? Do you want to push a release soon and then later have another patch release when 3.9 officially gets release?

YannickJadoul:
Only manylinux2014 on ppc64le is missing. Apparently Travis's ppc64le build is acting up.

joerick:
from the announcement post on python-dev https://mail.python.org/archives/list/python-dev@python.org/thread/FFVG5DAE7RHE5ZYUVVJW5TPAZ2ALRFHB/

> We strongly encourage maintainers of third-party Python projects to prepare
their projects for 3.9 compatibility during this phase. As always, report any issues to
the Python bug tracker https://bugs.python.org/.

Based on the above, I think we should be merging this PR and releasing a version of cibuildwheel that builds packages against 3.9.0rc1. It does make me wonder what happens if a bug is found, but I guess that's what the release candidate phase is for :) 

Does anyone know if PyPI is accepting 3.9 wheels yet?

YannickJadoul:
> Does anyone know if PyPI is accepting 3.9 wheels yet?
> 
> EDIT: it appears PyPI is accepting them - see https://pypi.org/project/scylla-driver/#files for an example

Ah, great, I tried looking, but couldn't immediately find one :-)

Anyhow, sure, as you think is best. ABI should be stable now, I suppose, so ... all that's missing is that `ppc64le` docker image, I suppose. But we could always update later?
(I only updated the docker images, btw, so maybe the constraints files should still be given an update as well?)

joerick:
Okie doke, I'll merge this soon and make a release. Do we need to worry about the missing `ppc64le` image, since it's still building with 3.9b3? I'm leaning towards just merging it, any potential issues would be unlikely.

YannickJadoul:
With a bit of luck, we'll soon enough have a PyPy release that fixes all our problems there, but I guess we shouldn't wait for that.

Worrying about `ppc64le`? I don't know. It's still not there, I think: https://quay.io/repository/pypa/manylinux2014_ppc64le?tab=tags
But I also don't know if Travis CI works again on that image. It seems it is, so I could try if it would build and make an empty PR to manylinux, if you want?

YannickJadoul:
> But I also don't know if Travis CI works again on that image. It seems it is, so I could try if it would build and make an empty PR to manylinux, if you want?

Scratch that. I am actually trying: https://travis-ci.org/github/YannickJadoul/manylinux/builds/724689320
I'll create a PR if this works. Otherwise, we can just merge :-)

YannickJadoul:
OK, nevermind. Not such great news, from the ppc64le side: see pypa/manylinux#758. You can probably release, and we can add a patch when this gets fixed?

YannickJadoul:
@joerick Fixed some more things in the README. Tried to match #419 as closely as possible. Hopefully that will reduce the number of conflicts.

joerick:
Fantastic, thank you @YannickJadoul for updating those. Including a note in the readme about ppc64le is a good idea. I'll also note it in the release notes.

PR #383, opened 2020-06-23
Debug CircleCI failures

joerick:
Debugging recent linux circleCI failures

### DIFF
--- a/.circleci/prepare.sh
+++ b/.circleci/prepare.sh
@@ -3,8 +3,6 @@ set -o xtrace
 
 $PYTHON --version
 $PYTHON -m pip --version
-sudo $PYTHON -m pip uninstall -y poetry
-$PYTHON -m pip install --user --ignore-installed virtualenv==20.0.21
 $PYTHON -m virtualenv -p $PYTHON venv
 venv/bin/python -m pip install -r requirements-dev.txt
 venv/bin/python -m pip freeze
--- a/.circleci/prepare.sh
+++ b/.circleci/prepare.sh
@@ -1,6 +1,9 @@
+set -o errexit
+set -o xtrace
+
 $PYTHON --version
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Haha, you beat me to it. I just wanted to start looking at this.

Seems @Czaki might be right, though, about `poetry` being installed and messing up the installation of `virtualenv`? https://github.com/joerick/cibuildwheel/pull/342#issuecomment-647478067

joerick:
> Seems @Czaki might be right, though, about `poetry` being installed and messing up the installation of `virtualenv`? [#342 (comment)](https://github.com/joerick/cibuildwheel/pull/342#issuecomment-647478067)

Could be... but the install command actually succeeds, despite this error message. Quite confusing...

Czaki:
maybe add `set -e`  to .circleci/prepare.sh

joerick:
> maybe add `set -e` to .circleci/prepare.sh

`set -o errexit` is the same as `set -e`, that's already there.

Czaki:
Ok. I do not notice that it is changed in this PR. Sorry.

joerick:
Alright, think I got it. `virtualenv` is already system-installed on the Docker image, so installing a `--user` version and having multiple versions installed was causing problems. I removed the `--user` version.

YannickJadoul:
> Alright, think I got it. `virtualenv` is already system-installed on the Docker image, so installing a `--user` version and having multiple versions installed was causing problems. I removed the `--user` version.

Pfffffffffffffffffffff, good catch! But horrible error message.

Anyway, it's just our own CI config and nothing in `cibuildwheel`, so whatever makes things green :-)

joerick:
Just testing out a fire-and-forget "merge when green" app.

PR #387, opened 2020-06-23
Remove str() calls in call/shell functions using destructuring syntax

joerick:
While working on #386 I came across a way to remove those str() calls when combining Path and str objects. The trick is to use array destructuring syntax, e.g. `['pip', 'wheel', path_object, *flags]`, rather than `+`. This way the typechecker assumes you intended to mix the types and gives you the type `List[Union[str, PathLike]]`.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -201,16 +201,14 @@ def build(options: BuildOptions) -> None:
 
         # Path.resolve() is needed. Without it pip wheel may try to fetch package from pypi.org
         # see https://github.com/joerick/cibuildwheel/pull/369
-        call(
-            [
-                'pip', 'wheel',
-                options.package_dir.resolve(),
-                '-w', built_wheel_dir,
-                '--no-deps',
-                *get_build_verbosity_extra_flags(options.build_verbosity)
-            ],
-            env=env,
-        )
+        call([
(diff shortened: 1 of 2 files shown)


joerick:
> * I had tried to keep the `typing` imports separate from the other imports, since they do not really import any functionality (and I'm still slightly annoyed we have to import those ;-) ).

These changes were by VScode's 'sort imports' command. I'd guess `isort` would do something similar. Happy to change it back if you like? But I think we just embrace it - Python types are kinda weird, but we're in for the ride! :) also, `from os import PathLike` - this is just a type - should that be separate? gets kinda fuzzy.

> * The exploded `call(['pip', 'wheel', ...])` lines are spread over many lines. Is there any way to make this more compact? E.g., have something like

In retrospect, I agree. I've reduced it down.

YannickJadoul:
> These changes were by VScode's 'sort imports' command. I'd guess `isort` would do something similar. Happy to change it back if you like? But I think we just embrace it - Python types are kinda weird, but we're in for the ride! :) also, `from os import PathLike` - this is just a type - should that be separate? gets kinda fuzzy.

Yeah, I don't really mind. Just noticed. And since no one made remarks when I added them like that, I had assumed it was appreciated. But if it goes against code formatters, I don't feel that strongly ;-)

PR #395, opened 2020-07-04
Update dependencies to get python 3.9 support on s390x

mayeut:
Update dependencies to get python 3.9 support on s390x
Only updating manylinux images would be enough but this also tests other updates.

### DIFF
--- a/cibuildwheel/resources/pinned_docker_images.cfg
+++ b/cibuildwheel/resources/pinned_docker_images.cfg
@@ -1,22 +1,22 @@
 [x86_64]
-manylinux1 = quay.io/pypa/manylinux1_x86_64:2020-06-18-0eb93b9
-manylinux2010 = quay.io/pypa/manylinux2010_x86_64:2020-06-16-1ee6b68
-manylinux2014 = quay.io/pypa/manylinux2014_x86_64:2020-06-18-f5da004
+manylinux1 = quay.io/pypa/manylinux1_x86_64:2020-07-04-283458f
+manylinux2010 = quay.io/pypa/manylinux2010_x86_64:2020-07-04-10a3c30
+manylinux2014 = quay.io/pypa/manylinux2014_x86_64:2020-07-04-bb5f087
 
 [i686]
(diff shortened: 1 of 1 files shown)


mayeut:
This won't get built on travis-ci because only master branch gets built (this includes target branch for PR)

YannickJadoul:
Huh, why do we only build the master branch? Seems a leftover from @joerick trying to stop his own PRs building twice (because it being a PR and it being a commit): https://github.com/joerick/cibuildwheel/commit/97628673d2bee686993d7f9e0b7a3c783f73a8ab. But this seems unintended, no?

Apart from that, thanks for noticing that s390x is now working!

Any reason to update the other versions, if we're not making a release? I'd somehow say it would be nice to have these the same as the last true release, such that trying out Python 3.9 and our branch wouldn't get any unexpected consequences/errors?

joerick:
Can you not see a Travis build? I see it - https://travis-ci.org/github/joerick/cibuildwheel/builds/704961809

YannickJadoul:
I think it's triggered by #396, and Travis just confuses them because they are the same commit, no?

mayeut:
> I think it's triggered by #396, and Travis just confuses them because they are the same commit, no?

Yes exactly.

> Any reason to update the other versions, if we're not making a release? I'd somehow say it would be nice to have these the same as the last true release, such that trying out Python 3.9 and our branch wouldn't get any unexpected consequences/errors?

I wanted to check for unexpected errors with other updates at first (and because running `python bin/update_dependencies.py ` is easy and updates all). I now dropped the other updates and just update the docker images.

joerick:
Ah, I see - #396 is just to trigger CI. Looks like it's green, so feel free to merge into `python3.9`!

YannickJadoul:
> Edit: updating other dependencies had for effect what seems to be a deadlock on macOS on travis-ci...

Huh, good to know, thanks! We shouldn't do this update too last minute before the next update, then, @joerick!

Meanwhile, let's get this updated. Thanks again! :-)

PR #398, opened 2020-07-06
fix: Support environment markers in PEP 518 workaround

### LINKED ISSUE #392: Windows pyproject.toml cannot handle versioned dependencies
If pyproject.toml specifies a versioned dependency, such as...

```toml
[build-system]
requires = [
  "setuptools >= 30.3.0",
]
```

then in windows.py, the code current parses the requirements and calls...

```python
return subprocess.check_call(' '.join(args), env=env, cwd=cwd, shell=True)
```

which means that the result of `' '.join(args)` is...

```
pip install setuptools >= 30.3.0
```

which breaks the shell parser since that needs to be quoted to be not be a redirect.

`shell=True` is usually bad mojo. It should work to do `subprocess.check_call(args, env=env, cwd=cwd)`.

henryiii:
The original version didn't handle environment markers, making the common idiom for supporting all NumPy versions: 

```ini
requires = [
    "numpy==1.13.3; python_version<='3.6'",
    "numpy==1.14.5; python_version=='3.7'",
    "numpy==1.17.3; python_version>='3.8'",
    ...
```

fail. Will fix #392.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -172,8 +172,7 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
             else []
         )
         if requirements:
-            escaped_requirements = [f'{s!r}' for s in requirements]
-            shell(['pip', 'install'] + escaped_requirements, env=env)
+            shell(['pip', 'install'] + requirements, env=env)
 
 
 def build(options: BuildOptions) -> None:
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -8,7 +8,6 @@ from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Sequence, Union
 from zipfile import ZipFile
 import toml
-import shlex
 
 from .environment import ParsedEnvironment
 from .util import (BuildOptions, BuildSelector, download,
@@ -173,7 +172,7 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
             else []
         )
         if requirements:
-            escaped_requirements = [shlex.quote(s) for s in requirements]
+            escaped_requirements = [f'{s!r}' for s in requirements]
             shell(['pip', 'install'] + escaped_requirements, env=env)
 
 
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Sequence, Union
 from zipfile import ZipFile
 import toml
+import shlex
 
 from .environment import ParsedEnvironment
 from .util import (BuildOptions, BuildSelector, download,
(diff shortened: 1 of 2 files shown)


henryiii:
Assuming the tests pass, how hard would it be to push it into a 1.5.2 release?

YannickJadoul:
What about just `[f'"r"' for r in requirements]` (as suggested by https://github.com/joerick/cibuildwheel/issues/392#issuecomment-654076067) ?

henryiii:
I'll try that to see if it works, but then I'll take a stab at removing the shell + join. Really this should have worked in the original form, it just is tripping up because of the join hack, IIUC, so its worth testing to see if that's easy to remove.

YannickJadoul:
Great, that would be even better! Thanks for taking up this annoying job! :-)

Czaki:
I try some time ago to remove `shell=True` bui it is forced by bug in python https://bugs.python.org/issue8557
If I good understand it can be changed after drop of python 3.6 support by `cibuildwheel`

henryiii:
I think that says it's still affecting at least Python 3.7, and it's marked as open and languishing so I bet it is true in 3.8 and 3.9. I'm playing around with it, and will see if I can come up with a solution. Using the correct environment is important, so if that's really completely missing when shell=False, then the solution above is probably the only reasonable one.

test/test_pep518.py
```diff
@@ -24,7 +28,8 @@
 requires = [
     "setuptools>=42",
     "wheel",
-    "requests==2.23.0"
+    "requests==2.22.0; python_version<'3.6'",
```
YannickJadoul, line 31: Maybe add some extra spaces to the test? I'm not sure if it matter a lot, but that's what's reported in #392.

cibuildwheel/windows.py
```diff
@@ -172,7 +172,8 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
             else []
         )
         if requirements:
-            shell(['pip', 'install'] + requirements, env=env)
+            escaped_requirements = [f'"{s}"' for s in requirements]
```
YannickJadoul, line 175: Thinking out loud: is this something we should do here, or more broadly inside of `shell`?

henryiii: We don't have any protection for quotes inside the string, so I was a little hesitant to add it higher up. If someone wrote `"requests==2.22.0; python_version<\"3.6\"",`, in fact, this might not work.

test/test_pep518.py
henryiii: There's a space (and even worse, a semicolon) in the string above. But I can add spaces around the setuptools too, just to be sure.

YannickJadoul: Ah, yeah, sorry. Missed that!

cibuildwheel/windows.py
YannickJadoul: Good point. I suppose we also control most of the other things to not contain spaces?
Though .. what about spaces in `CIBW_TEST_EXTRAS` or so?

henryiii:
> PermissionError: [WinError 5] Access is denied: 'C:\\Users\\travis\\AppData\\Local\\Temp\\tmpusgvk8k0\\Scripts\\nosetests.exe'

Don't think this is related, only on Travis.

test/test_pep518.py
henryiii: Added spaces above. :)

henryiii:
I think this would be a good candidate for 1.5.2, but #399 would be more of a candidate for 1.6...

cibuildwheel/windows.py
henryiii: That can be changed by the cibw user, if something odd happens. You can't change `pyproject.toml` (at least you shouldn't in CI).

PR #399, opened 2020-07-06
WIP: Alternate version of #398

henryiii:
Testing removing shell=True for a better solution to #398. Not expected to work right away.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -172,8 +172,7 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
             else []
         )
         if requirements:
-            escaped_requirements = [f'{s!r}' for s in requirements]
-            shell(['pip', 'install'] + escaped_requirements, env=env)
+            shell(['pip', 'install'] + requirements, env=env)
 
 
 def build(options: BuildOptions) -> None:
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -1,3 +1,4 @@
+import copy
 import os
 import shutil
 import subprocess
@@ -29,6 +30,9 @@ def shell(args: Sequence[Union[str, PathLike]], env: Optional[Dict[str, str]] = 
           cwd: Optional[str] = None) -> int:
     command = ' '.join(normalize(a, cwd) for a in args)
     print(f'+ {command}')
+    if env is not None:
+        env = copy.copy(env)
+        env["PATH"] = os.environ["PATH"]
     return subprocess.check_call(list(normalize(a, cwd) for a in args), env=env, cwd=cwd)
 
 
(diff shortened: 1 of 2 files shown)


henryiii:
I don't see why this change causes `shutil.Error: Destination path 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpk68gtxaq\spam-0.1.0-cp37-cp37m-win_amd64.whl' already exists`...

Czaki:
because without `shell =True` PATH variable is ignored and `cibuildwheel` control which python interpreter is used using `PATH` environment variable. 
You can see that it fails when building wheel for python 2.7. (second version) but in wheel name there is python 3. 

From log:
```
Successfully installed 'python2 2.7.18' to C:\cibw\python
Executing nuget actions took 11.22 sec

+ where python

C:\cibw\python\python2.2.7.18\tools\python.exe
C:\hostedtoolcache\windows\Python\3.6.8\x64\python.exe
C:\hostedtoolcache\windows\Python\3.7.7\x64\python.exe
C:\ProgramData\Chocolatey\bin\python.exe
+ python --version
Python 3.6.8
```

henryiii:
Ah, yes. This hasn't been changed in 3.8, so I don't think it's likely to be easily fixable. Going to try 1-2 things.

YannickJadoul:
Resolving `PATH` ourselves and turning the first element of the list into an absolute path is over the top, I suppose? It should be possible, but it's probably not worth the effort, I'm guessing.

Czaki:
> Resolving `PATH` ourselves and turning the first element of the list into an absolute path is over the top, I suppose? It should be possible, but it's probably not worth the effort, I'm guessing.

And may be problematic if some of tool call other in subprocess.

YannickJadoul:
> And may be problematic if some of tool call other in subprocess.

Ah, yes, if in that subprocess, the correct environment /`PATH` isn't there inside the subprocess itself, either, there's no use. Scratch that, then.

Another thing, though. It's possible to pass a list of arguments, and still have `shell=True` (demonstrated in the https://bugs.python.org/issue8557 issue mentioned). Does that fix the quoting of the arguments, or does it also do a stupid `' '.join` inside?

henryiii:
> Resolving PATH ourselves and turning the first element of the list into an absolute path is over the top, I suppose?

I was thinking about trying exactly this. Might be too much, though.

> And may be problematic if some of tool call other in subprocess.

I am injecting the PATH into the environment, so the subprocess should be just fine (and might be already, without the PATH added). It's just that the command itself does not repeat the PATH when resolving.

henryiii:
Why is this being joined, actually? Shouldn't `shell=True` still take a list, and shouldn't that list be properly quoted?

YannickJadoul:
> Why is this being joined, actually? Shouldn't `shell=True` still take a list, and shouldn't that list be properly quoted?

Yes, that's what I meant, in my previous comment. I don't know the reasons for joining the strings..

Oh, btw, I've noticed that there are some calls that do need to have `shell` enabled, because we want to be able to chain commands, etc (see macOS's `call`).

YannickJadoul:
( Or, let's just ditch Windows; I've had enough of its quirks and outright bullshit? ;-) )

PR #401, opened 2020-07-08
Windows - distinguishing between call and shell for better arg quoting

joerick:
Alternate version of #399  :)

I had a slightly different idea to @henryiii on this one and wanted to try it. Let's see what CI makes of it.

### DIFF
--- a/test/test_pep518.py
+++ b/test/test_pep518.py
@@ -6,8 +6,12 @@ basic_project = test_projects.new_c_project(
     setup_py_add=textwrap.dedent(
         """
         # Will fail if PEP 518 does work
+        import sys
         import requests
-        assert requests.__version__ == "2.23.0", "Requests found but wrong version ({0})".format(requests.__version__)
+        if sys.version_info < (3, 6, 0):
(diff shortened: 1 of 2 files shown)


joerick:
Cool, looks like this is working. Does this look okay to you @YannickJadoul @henryiii ? I can get it merged and in a release pretty soon if so.

henryiii:
I think it's good!

YannickJadoul:
A bit sad this doesn't work the same as macos, then:

https://github.com/joerick/cibuildwheel/blob/d0b6401129ead1164d012835ea6366c8486a562e/cibuildwheel/macos.py#L17-L24

But a similar `call([repair_command_prepared], env=env)` doesn't work, right?

test/test_pep518.py
```diff
@@ -24,7 +28,8 @@
 requires = [
     "setuptools>=42",
```
henryiii, line 29:
```suggestion
    "setuptools >= 42",
```

We could add this, just to ensure this covers #392

YannickJadoul:
(+1 on the newly-coined terminology of "the shell/call debacle", btw! ;-) )

PR #403, opened 2020-07-12
Debug Appveyor errors

joerick:


joerick:
Turns out the Appveyor PATH variable was highlighting a particularly fiddly bug in our environment variable handling. Backslashes in variables (very common in Windows paths) were able to 'escape' the outer quote. The new solution in `bashlex_eval.py` certainly smells better, and I've added a unit test to test the new solution to its limits :)

YannickJadoul:
Wow, sounds like a horrible bug! Kudos for catching that! :-)

So, `context.input` does not contain the same as `node.word`, then? But `part.pos` still indexes into `context.input` and not into `node.word`? That seems quite counterintuitive.

joerick:
> So, `context.input` does not contain the same as `node.word`, then? But `part.pos` still indexes into `context.input` and not into `node.word`? That seems quite counterintuitive.

Yep! The indexes refer to the input string, but node.word is the token that the lexer found. I think. Anyway, it seems to work!

PR #408, opened 2020-07-19
Fix quotes in command in bashlex_eval

### LINKED ISSUE #406: CIBW_ENVIRONMENT broken in 1.5.3
Somehow, my line of `export CIBW_ENVIRONMENT='... CCACHE_BASEDIR=`python -c "import tempfile; import os; print(os.path.realpath(tempfile.gettempdir()))"` ...'` (see https://github.com/YannickJadoul/Parselmouth/blob/1f3dd9abc63afdf4a1d26db40d210bb4187118a8/.travis.yml#L143; yes, I know I ought to clean this up) breaks:

```
    + /bin/true
    + mkdir -p /project
    + uname -i
x86_64
    + python -c import 'tempfile;' import 'os;' 'print(os.path.realpath(tempfile.gettempdir()))'
  File "<string>", line 1
    import
```

See https://travis-ci.org/github/YannickJadoul/Parselmouth/jobs/709679114#L392

I'll start investigating, but is this perhaps related to #403, @joerick?

YannickJadoul:
Closes #406

### DIFF
--- a/cibuildwheel/bashlex_eval.py
+++ b/cibuildwheel/bashlex_eval.py
@@ -97,7 +97,7 @@ def evaluate_nodes_as_compound_command(nodes: Sequence[bashlex.ast.node], contex
 
 
 def evaluate_nodes_as_simple_command(nodes: List[bashlex.ast.node], context: NodeExecutionContext):
-    words = [evaluate_node(part, context=context) for part in nodes]
+    words = [shlex.quote(evaluate_node(part, context=context)) for part in nodes]
     command = ' '.join(words)
     return context.executor(command, context.environment)
 
--- a/cibuildwheel/bashlex_eval.py
+++ b/cibuildwheel/bashlex_eval.py
@@ -1,15 +1,14 @@
-import shlex
 import subprocess
 from typing import Callable, Dict, List, NamedTuple, Optional, Sequence
 
 import bashlex  # type: ignore
 
 # a function that takes a shell command and the environment, and returns the result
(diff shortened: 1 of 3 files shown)


cibuildwheel/bashlex_eval.py
```diff
@@ -1,15 +1,14 @@
-import shlex
 import subprocess
 from typing import Callable, Dict, List, NamedTuple, Optional, Sequence
 
 import bashlex  # type: ignore
 
 # a function that takes a shell command and the environment, and returns the result
```
joerick, line 6:
```suggestion
# a function that takes a command and the environment, and returns the result
```

joerick:
I'm gonna get this merged and another release out

YannickJadoul:
Great, thanks! :-)

All in all just a debugging run, once I had the example of what goes wrong. But yes, it was confusing enough. I first thought replacing `node.word` with `context.input[part.pos[0]:part.pos[1]]` would be enough, but then the tests showed me that gives another problem (the one you had actually just solved), so .. yes, I noticed :-D

joerick:
This kind of code is so subtle that it's better defined by the tests than the code itself, I think. anywho, released as v1.5.4 :)

YannickJadoul:
Hurray! (Almost too quick to have had pinned cibuildwheel==1.5.2 ;-) But anyway, reverting it: https://github.com/YannickJadoul/Parselmouth/commit/a1e674b43c76d31ef2187bc6017045ccb5fa2793)

PR #410, opened 2020-07-19
Fix cwd in Linux docker images

### LINKED ISSUE #409: The "/project/" path isn't there anymore for Linux in version 1.5.3
Upgraded from 1.5.1 to 1.5.3 in a branch here: https://github.com/Chia-Network/bls-signatures/tree/up-cibuildwheel

In 1.5.1 we would install cmake into `/project/cmake-3.17.3-Linux-x86_64` and build various dependencies from there as we add `/project/cmake-3.17.3-Linux-x86_64` to the front of PATH:
https://github.com/Chia-Network/bls-signatures/runs/880546025?check_suite_focus=true#step:10:299

In 1.5.3 cmake gets installed into `//cmake-3.17.3-Linux-x86_64` so the PATH we preset doesn't work and the build fails.
https://github.com/Chia-Network/bls-signatures/runs/887691441?check_suite_focus=true#step:10:284

Operative workflow here: https://github.com/Chia-Network/bls-signatures/blob/up-cibuildwheel/.github/workflows/build.yml

I can work around but seems like either an issue that needs a fix or some mention in documentation. I'm sure it came out of the move to docker - which otherwise looks cool.

YannickJadoul:
Closes #409

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -110,8 +110,8 @@ def build(options: BuildOptions) -> None:
             continue
 
         try:
-            with DockerContainer(docker_image, simulate_32_bit=platform_tag.endswith('i686')) as docker:
-                docker.copy_into(Path.cwd(), Path('/project'))
+            with DockerContainer(docker_image, simulate_32_bit=platform_tag.endswith('i686'), cwd='/project') as docker:
+                docker.copy_into(Path.cwd(), PurePath('/project'))
 
                 if options.before_all:
                     env = docker.get_environment()
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -38,18 +38,16 @@ def test(tmp_path):
     project_dir = tmp_path / 'project'
     project_with_before_build_asserts.generate(project_dir)
 
-    before_build = textwrap.dedent('''
-        python -c "import sys; open('{output_dir}pythonversion.txt', 'w').write(sys.version)" &&
-        python -c "import sys; open('{output_dir}pythonexecutable.txt', 'w').write(sys.executable)" &&
-        python -c "import os; open('{output_dir}cwd.txt', 'w').write(os.getcwd())"
-    ''')
+    before_build = ('''python -c "import sys; open('{output_dir}pythonversion.txt', 'w').write(sys.version)" && '''
+                    '''python -c "import sys; open('{output_dir}pythonexecutable.txt', 'w').write(sys.executable)" && '''
+                    '''python -c "import os; open('{output_dir}cwd.txt', 'w').write(os.getcwd())"''')
 
     # build the wheels
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env={
         # write python version information to a temporary file, this is
         # checked in setup.py
-        'CIBW_BEFORE_BUILD': before_build.format(output_dir="/tmp/"),
-        'CIBW_BEFORE_BUILD_WINDOWS': before_build.format(output_dir="c:\\"),
+        'CIBW_BEFORE_BUILD': before_build.format(output_dir='/tmp/'),
+        'CIBW_BEFORE_BUILD_WINDOWS': before_build.format(output_dir=r'c:\\'),
     })
 
     # also check that we got the right wheels
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -24,6 +24,12 @@ project_with_before_build_asserts = test_projects.new_c_project(
         print('sys.executable', sys.executable)
         # windows/mac are case insensitive
         assert os.path.realpath(stored_executable).lower() == os.path.realpath(sys.executable).lower()
+
+        if sys.platform == 'linux':
+            cwd_file = '/tmp/cwd.txt'
+            with open(cwd_file) as f:
+                stored_cwd = f.read()
+            assert stored_cwd == '/project'
     ''')
 )
 
@@ -32,12 +38,18 @@ def test(tmp_path):
     project_dir = tmp_path / 'project'
     project_with_before_build_asserts.generate(project_dir)
 
+    before_build = textwrap.dedent('''
+        python -c "import sys; open('{output_dir}pythonversion.txt', 'w').write(sys.version)" &&
+        python -c "import sys; open('{output_dir}pythonexecutable.txt', 'w').write(sys.executable)" &&
+        python -c "import os; open('{output_dir}cwd.txt', 'w').write(os.getcwd())"
+    ''')
+
     # build the wheels
     actual_wheels = utils.cibuildwheel_run(project_dir, add_env={
         # write python version information to a temporary file, this is
         # checked in setup.py
-        'CIBW_BEFORE_BUILD': '''python -c "import sys; open('/tmp/pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('/tmp/pythonexecutable.txt', 'w').write(sys.executable)"''',
-        'CIBW_BEFORE_BUILD_WINDOWS': '''python -c "import sys; open('c:\\pythonversion.txt', 'w').write(sys.version)" && python -c "import sys; open('c:\\pythonexecutable.txt', 'w').write(sys.executable)"''',
+        'CIBW_BEFORE_BUILD': before_build.format(output_dir="/tmp/"),
+        'CIBW_BEFORE_BUILD_WINDOWS': before_build.format(output_dir="c:\\"),
     })
 
     # also check that we got the right wheels
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -1,3 +1,5 @@
+import pytest
+import subprocess
 import textwrap
 
 from . import utils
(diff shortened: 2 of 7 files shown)


YannickJadoul:
Think we're almost there. Questions left:

- Is this fine, adding `cwd` to the constructor, or should we add `cwd` to each `docker.call`?
- More tests? Maybe `CIBW_BEFORE_ALL` as well? And anything else to test?

YannickJadoul:
Broken because this issue is fixed, but #411 isn't.

`test_overridden_path`'s ` 'CIBW_BEFORE_ALL': 'mkdir new_path && touch new_path/python && chmod +x new_path/python'` is now also evaluated from `/project`, but `$(pwd)` in ``'CIBW_ENVIRONMENT': '''PATH="$(pwd)/new_path:$PATH"'''` is not interpolated correctly, so `/new_path` does not contain `python`, so the Python executable on `PATH` is not overwritten.

I'll rebase onto #411.

YannickJadoul:
(rebased after merge of #411)

cibuildwheel/linux.py
```diff
@@ -110,8 +110,8 @@ def build(options: BuildOptions) -> None:
             continue
 
         try:
-            with DockerContainer(docker_image, simulate_32_bit=platform_tag.endswith('i686')) as docker:
-                docker.copy_into(Path.cwd(), Path('/project'))
+            with DockerContainer(docker_image, simulate_32_bit=platform_tag.endswith('i686'), cwd='/project') as docker:
+                docker.copy_into(Path.cwd(), PurePath('/project'))
```
joerick, line 114: Setting cwd to `/project` in the constructor feels a little strange, because I don't think `/project` has been created yet. I suppose Docker must create it if it's not there. But doesn't feel right to me. I think a more familar approach might be to have `docker.chdir(PurePath('/project'))` after the project is copied over. 

The cwd can be stored in DockerContainer as an ivar, and used whenever `call()` is called without a `cwd`. What do you think?

test/test_before_build.py
```diff
@@ -24,6 +24,12 @@
         print('sys.executable', sys.executable)
         # windows/mac are case insensitive
         assert os.path.realpath(stored_executable).lower() == os.path.realpath(sys.executable).lower()
+
+        if sys.platform == 'linux':
+            cwd_file = '/tmp/cwd.txt'
+            with open(cwd_file) as f:
+                stored_cwd = f.read()
+            assert stored_cwd == '/project'
```
joerick, line 32: Looks like this is missing a mac and windows test. Also, it might be a bit Rube Goldberg - could you instead `assert os.getcwd() == '/project'`? And in BEFORE_ALL, do the same?

cibuildwheel/linux.py
YannickJadoul: Docker does document that passing `-w` will create the directory: https://docs.docker.com/engine/reference/commandline/create/#options

For the `DockerContainer`, I thought of it as just setting the "default directory" during the with-statement, and just adding an exception if necessary. And also that we then couldn't forget it, later.

But yes, `chdir` would also work, if you prefer that. I hadn't thought of that option. I guess we don't even need to store it, but could just `cd`, rather than storing it as an attribute?

test/test_before_build.py
YannickJadoul: > Looks like this is missing a mac and windows test.

Well, yes, but I didn't know what directory they're supposed to be executed from? As far as I know, it will depend on the CI; probably somewhere in the home directory or a subdirectory?

> Also, it might be a bit Rube Goldberg - could you instead `assert os.getcwd() == '/project'`?

Nope, that's not working. The tests are actually executed from a different directory, and we want to test `CIBW_BEFORE_BUILD` is executed from the right directory. Or do you mean to `python -c "import os;  os.getcwd() == '/project'"` in `CIBW_BEFORE_BUILD`?

> And in BEFORE_ALL, do the same?

Yes, will do. But I'll wait until the above gets settled.

joerick: > Or do you mean to `python -c "import os; os.getcwd() == '/project'"` in `CIBW_BEFORE_BUILD`?

Yeah, that's what I was thinking, I suppose it'd be something like...
```
{
   'CIBW_BEFORE_BUILD': f'''python -c "import os; assert os.getcwd() == {os.getcwd()!r}"''',
   'CIBW_BEFORE_ALL': f'''python -c "import os; assert os.getcwd() == {os.getcwd()!r}"''',
   'CIBW_BEFORE_BUILD_LINUX': '''python -c "import os; assert os.getcwd() == '/project'"''',
   'CIBW_BEFORE_ALL_LINUX': '''python -c "import os; assert os.getcwd() == '/project'"''',
}
```

cibuildwheel/linux.py
joerick: Ah, that's good that's it's documented. Then I think either approach is valid, actually. Happy to leave it as-is.

test/test_before_build.py
hoffmang9: Windows on GH Actions uses `/d/a/bls-signatures/bls-signatures` which is the repo name twice: https://github.com/Chia-Network/bls-signatures/runs/887873992?check_suite_focus=true#step:10:60

Mac is `/Users/runner/work/bls-signatures/bls-signatures/` see: https://github.com/Chia-Network/bls-signatures/runs/887873983?check_suite_focus=true#step:11:535

YannickJadoul: > Windows on GH Actions uses `/d/a/bls-signatures/bls-signatures` which is the repo name twice: https://github.com/Chia-Network/bls-signatures/runs/887873992?check_suite_focus=true#step:10:60
> 
> Mac is `/Users/runner/work/bls-signatures/bls-signatures/` see: https://github.com/Chia-Network/bls-signatures/runs/887873983?check_suite_focus=true#step:11:535

Thanks, but that's different from every CI and every project, even (since it contains the project name), so that doesn't seem feasible to distinguish. Let me try @joerick's plan (maybe with `__file__` rather than `getcwd`?) :-)

YannickJadoul: OK, like this?

I added a test that checks whether a failing `CIBW_BEFORE_BUILD` and `CIBW_BEFORE_ALL` also fail the whole run.

And I (temporarily) reverted the actual fix, to see if these tests catch the errors. Can be dropped in a rebase, later.

cibuildwheel/linux.py
YannickJadoul: Given that we don't have a `os.chdir` for macOS or Windows (and just use the working directory that's there at the start of cibuildwheel/`build`), I think the current approach might match better. But I'm happy changing it as well, if you prefer. I like the suggested `chdir` method as well (and we might end up needing it at some later point?). Your call!

@Czaki, any input on this? Do you prefer `with DockerContainer(..., cwd='/project'):` or `with DockerContainer(...) as docker: docker.chdir('/project')`?

YannickJadoul:
Alright. Next round of review! :-)

Slightly worrying note: we're now at 58 and 59 minutes on Azure Pipelines, for the tests on Windows (one of the two jobs actually timed out before I restarted it) :-/ Not sure why Windows (or the Windows infrastructure at Azure) is twice as slow as Linux...

cibuildwheel/linux.py
Czaki: I prefer `with DockerContainer(..., cwd='/project'):` and change second argument of     
```
copy_into(self, from_path: Path, to_path: Optional[PurePath]=none)
    if to_path is None:
        to_path = self.cwd
    ...
```

YannickJadoul: Oh, that's an interesting suggestion. I'll add that a as extra commit. If @joerick doesn't like it, it can just be removed before merging!

Czaki:
> Slightly worrying note: we're now at 58 and 59 minutes on Azure Pipelines, for the tests on Windows (one of the two jobs actually timed out before I restarted it) :-/ Not sure why Windows (or the Windows infrastructure at Azure) is twice as slow as Linux...

Maybe use `timeoutInMinutes: 180` ?

I also observe in my projects that windows is much slower.

cibuildwheel/linux.py
YannickJadoul: Actually, that's not possible, @Czaki, because `self.cwd` can also be `None`. Do you suggest any way to fix it, or should we just keep `to_path` in `copy_into` mandatory?

Czaki: Raise exception with information that `cwd` should be set or `to_path`. 
I suggest this to reduce number of places where path is provided as literal.

YannickJadoul:
> Maybe use `timeoutInMinutes: 180` ?
> 
> I also observe in my projects that windows is much slower.

Alright, thanks! But it's still a bit worrying and very slow to wait for, so ... maybe only a temporary solution :-/

Czaki:
on github acrions and in azure pipelines there is big amount of available workers. Cannot we split tests on two workers?

YannickJadoul:
Yes, maybe. But that feels like a huge mess, as well :-(

cibuildwheel/linux.py
YannickJadoul: Yeah, possible, but ... it's still used in a lot of places, like `format(..., project='/project', ...)`, so doing this won't solve that all?

Czaki:
How it looks in my project:
obraz

https://stackoverflow.com/questions/10150881/why-is-python-so-much-slower-on-windows

Looks like problems with creating multiple process, antivirus, etc, which create bad performance in this case because of multiple spawning of new python using subprocess. 

Maybe `pytest-xdist`?

YannickJadoul:
> Maybe `pytest-xdist`?

I don't know anything about that, but let's discuss in #412 :-)

PR #411, opened 2020-07-20
Fix command substitution in CIBW_ENVIRONMENT in Linux docker images

YannickJadoul:
See #409

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -182,7 +182,7 @@ class DockerContainer:
 
     def environment_executor(self, command: List[str], environment: Dict[str, str]) -> str:
         # used as an EnvironmentExecutor to evaluate commands and capture output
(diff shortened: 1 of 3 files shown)


YannickJadoul:
This one is quite scare to have missed in our tests, though :-|

joerick:
Yikes. A bit embarrassed to have missed this one! Also, a tighter type on `docker.call` would have found it. Maybe those ugly overloads are worth it...

joerick:
Thank you for getting this done so quickly!

YannickJadoul:
> Yikes. A bit embarrassed to have missed this one! Also, a tighter type on `docker.call` would have found it. Maybe those ugly overloads are worth it...

Ah, yes, I hadn't considered types here to provide another way of avoiding this. At any rate, should've also been caught during review ... sorry :-/

PR #413, opened 2020-07-22
use variable instead of '/project' literal

Czaki:
Due to discussion in #410 I suggest to use variable instead of repeat `/project` 7 times.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -101,8 +101,8 @@ def build(options: BuildOptions) -> None:
     if cwd != abs_package_dir and cwd not in abs_package_dir.parents:
         raise Exception('package_dir must be inside the working directory')
 
-    project_path = PurePath('/project')
-    container_package_dir = project_path / abs_package_dir.relative_to(cwd)
+    container_project_path = PurePath('/project')
+    container_package_dir = container_project_path / abs_package_dir.relative_to(cwd)
     container_output_dir = PurePath('/output')
 
     for implementation, platform_tag, docker_image in platforms:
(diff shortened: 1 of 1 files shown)


joerick:
I understand where you're coming from, but I don't really see the improvement in readability - is this just indirection for the sake of it? i.e. what's the difference between `'/project'` and `project_path`? (except that I would recognise `'/project' in logs, and the other I wouldn't)

joerick:
(I realise I'm being a bit of a jerk, but could be an interesting discussion!)

YannickJadoul:
To be honest, I tried the exact same thing in #410 (even before @Czaki mentioned it), seeing the duplication of all the literals, but then had the same reflection on "Is there any difference between writing this variable and writing the string?". The thing is, it's not a complex or weird string that we'll change.

Then again, I don't have an issue with actually changing it. The one thing I would like, if we do this, is turning it into a `PurePath`, since it semantically is a path (so `PurePath` makes more sense as type, right?).

Czaki:
In general when using constant string in multiple places it is easier to introduce bug in code because of some typo. When there is typo in variable name IDE or flake8 may help. 

> Then again, I don't have an issue with actually changing it. The one thing I would like, if we do this, is turning it into a `PurePath`, since it semantically is a path (so `PurePath` makes more sense as type, right?).

In most place it  is used as string. 
And this line https://github.com/joerick/cibuildwheel/pull/413/files#diff-22075e43d4ff5efd0d10f3dcffdb93aaR105 verify if it is proper path

YannickJadoul:
> In most place it is used as string.

Still, it is a path, so unless it gives a lot of overhead, I'd try it as `PurePath`. Wouldn't that work?

YannickJadoul:
> because of some typo.

That's a good point, though. Just an accidental typo wouldn't be caught!

joerick:
Ah, making it a `PurePath` seems a good reason. Could we rename it to `container_project_path` - to match the other variables that refer to paths inside the container?

Czaki:
And what you think @joerick about introduce same changes to windows and macos?

joerick:
> And what you think @joerick about introduce same changes to windows and macos?

Not all that interested, actually. I quite like the path being `'.'`, and relative, because it makes the logs simpler to read, avoiding CI-specific paths in there.

Issue #417, opened 2020-08-10
auditwheel: error: cannot repair "...cp37m-linux_x86_64.whl" to "manylinux2010_x86_64"


horta:
The relevant bit:

```
    + rm -rf /tmp/cibuildwheel/built_wheel
    + mkdir -p /tmp/cibuildwheel/built_wheel
    + pip wheel /project -w /tmp/cibuildwheel/built_wheel --no-deps
WARNING: You are using pip version 20.1.1; however, version 20.2.1 is available.
You should consider upgrading via the '/opt/_internal/cpython-3.7.7/bin/python3.7 -m pip install --upgrade pip' command.
Processing /project
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
    Preparing wheel metadata: started
    Preparing wheel metadata: finished with status 'done'
Building wheels for collected packages: bgen-reader
  Building wheel for bgen-reader (PEP 517): started
  Building wheel for bgen-reader (PEP 517): finished with status 'done'
  Created wheel for bgen-reader: filename=bgen_reader-4.1.0-cp37-cp37m-linux_x86_64.whl size=103211 sha256=fee846b9b912bee0451c8e6804a0f2e984e5a44cea8398559ab4ed0b348f9e7b
  Stored in directory: /tmp/pip-ephem-wheel-cache-dqhl06it/wheels/84/ce/ef/3768e7d143f298e3f79dda4c505f7ffa005f429720ee7533a3
Successfully built bgen-reader
    + /opt/python/cp38-cp38/bin/python -c 'import sys, json, glob; json.dump(glob.glob('"'"'/tmp/cibuildwheel/built_wheel/*.whl'"'"'), sys.stdout)'
    + rm -rf /tmp/cibuildwheel/repaired_wheel
    + mkdir -p /tmp/cibuildwheel/repaired_wheel
    + sh -c 'auditwheel repair -w /tmp/cibuildwheel/repaired_wheel /tmp/cibuildwheel/built_wheel/bgen_reader-4.1.0-cp37-cp37m-linux_x86_64.whl'
INFO:auditwheel.main_repair:Repairing bgen_reader-4.1.0-cp37-cp37m-linux_x86_64.whl
usage: auditwheel [-h] [-V] [-v] command ...
[...]

PR #418, opened 2020-08-13
Add troubleshoot test

joerick:
Adds a test for the '.so' detection and fixes a regression. See #417.

### DIFF
--- a/test/test_troubleshooting.py
+++ b/test/test_troubleshooting.py
@@ -1,0 +1,28 @@
+import subprocess
+import pytest
+from .test_projects import TestProject
+from . import utils
+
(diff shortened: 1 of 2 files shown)


test/test_troubleshooting.py
```diff
@@ -0,0 +1,28 @@
+import subprocess
+import pytest
+from .test_projects import TestProject
+from . import utils
+
+so_file_project = TestProject()
+
+so_file_project.files['libnothing.so'] = ''
```
YannickJadoul, line 8: Oh, wow. I hadn't realized it would be so easy to test, but of course! That's amazing!

PR #420, opened 2020-08-14
fix: Windows extra file

### LINKED ISSUE #421: [BUG] Pyproject.toml workaround can make a file
The following pyproject.toml on Windows:

```toml
requires = [
  "setuptools_scm[toml]>=4.1.2"
]
```

produces a file

```
4.1.2
```

which breaks the setuptools_scm versioning, since it's now a dirty directory. Illustrated in #420. I think it used to work originally, as I made a few releases after the Windows 3.5 patch initially went in.

This _only_ breaks if there is a bracketed requirement.

henryiii:
Closes #421

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -177,8 +177,11 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
             if 'build-system' in data
             else []
         )
+
         if requirements:
-            shell(['pip', 'install'] + requirements, env=env)
+            # Workaround for bug when ]>= is present, #421
+            requirements = [r.replace("]", "] ") for r in requirements]
+            call(['pip', 'install'] + requirements, env=env)
 
 
 def build(options: BuildOptions) -> None:
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -179,9 +179,12 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
         )
 
         if requirements:
-            # Workaround for bug when ]>= is present, #421
-            requirements = [r.replace("]", "] ") for r in requirements]
-            call(['pip', 'install'] + requirements, env=env)
+            with tempfile.TemporaryDirectory() as d:
+                reqfile = Path(d) / "requirements.txt"
+                with reqfile.open("w") as f:
+                    for r in requirements:
+                        print(r, file=f)
+                call(['pip', 'install', '-r', str(reqfile)], env=env)
 
 
 def build(options: BuildOptions) -> None:
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -184,7 +184,7 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
                 with reqfile.open("w") as f:
                     for r in requirements:
                         print(r, file=f)
-                call(['pip', 'install', '-r', str(reqfile)], env=env)
+                call(['pip', 'install', '-r', reqfile], env=env)
 
 
 def build(options: BuildOptions) -> None:
--- a/test/test_pep518.py
+++ b/test/test_pep518.py
@@ -27,6 +27,7 @@ basic_project.files[
 [build-system]
 requires = [
     "setuptools >= 42",
+    "setuptools_scm[toml]>=4.1.2",
     "wheel",
     "requests==2.22.0; python_version<'3.6'",
     "requests==2.23.0; python_version>='3.6'"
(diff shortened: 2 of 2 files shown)


joerick:
Looks to me that `pip install` line should be a `call`, not a `shell`. The subprocess module should then take care of the escaping.

henryiii:
Okay, I'll try that! FYI, `]>=` seems to trigger the issue, while `] >=`doesn't.

henryiii:
Call seems to do the same thing. Replacing `"]"` with `"] "` fixed the issue. I'm not familiar enough with cmd to know why this is breaking/working. One thing I do know; you do have to explicitly add the quotes on some shells (zsh, most famously) if you use square brackets. Don't know why call is affected, though.

Czaki:
Did you check content of this strange file?

henryiii:
Nope, it's always in CI, so I don't know what's in it. I could cat it, but not sure I know how to do that on Windows.

Czaki:
> Nope, it's always in CI, so I don't know what's in it. I could cat it, but not sure I know how to do that on Windows.

I thinkt that you could do this with conditional step  https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

in this PR you could add code which will check if such file

Maybe double quotes may help: https://www.robvanderwoude.com/escapechars.php

YannickJadoul:
Would there be a chance to run it on a more sane shell, like powershell? But I guess that's maybe not included on ever CI service :-(

henryiii:
I don't think you can avoid having powershell.

henryiii:
At the same time, I'm not sure how you might change subprocess's calling system.

YannickJadoul:
> At the same time, I'm not sure how you might change subprocess's calling system.

I thought you could pass a string as argument to the `powershell` command?

joerick:
I think perhaps the most robust solution here would be to remove `shell=True` from `call()`. Perhaps we could figure out a solution to the PATH resolution problem - resolve programs on PATH ourselves? Or maybe using Windows `where` command?

joerick:
ooh- another idea, sidestep the command line completely - write those pep518 requirements to a temp file and do `pip install -r tempfile`

henryiii:
Would it be okay to call git inside the unit test? Ideally, `git diff --exit-code` would actually be best here for the check.

Czaki:
> Would it be okay to call git inside the unit test? Ideally, `git diff --exit-code` would actually be best here for the check.

Why? This will broke tests without creating commit. And I do not see option how it will help? Unit test does not call `build` function so will not create any files.

henryiii:
@YannickJadoul or @joerick - I've generated a rather large number of runs, if you could cancel all bu the latest ones (at least in GHA), that would speed things up, I don't have permission to cancel a run (even if I started it, which I always thought would have been nice to be able to do)

@Czaki  The problem that I am trying to fix in this PR is that "build" *is* creating a file - causing a dirty git directory and breaking tools like `setuptools_scm`. It's not supposed to be creating any files *inside the git source directory*.

henryiii:
@Czaki Ah, you are right, this is making a project that is not in git. Maybe we could compare the total number of files in the directory to some known total, then? Ignore my git check in the last commit, it will go away later ;)

Czaki:
> @Czaki Ah, you are right, this is making a project that is not in git. Maybe we could compare the total number of files in the directory to some known total, then? Ignore my git check in the last commit, it will go away later ;)

class `TestProject` from `test/test_project/base.py` contains list of expected files in `files` variable. So it could be use to check list of files if all are listed in this var.

cibuildwheel/windows.py
```diff
@@ -177,8 +177,14 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
...
         )
+
         if requirements:
-            shell(['pip', 'install'] + requirements, env=env)
+            with tempfile.TemporaryDirectory() as d:
+                reqfile = Path(d) / "requirements.txt"
+                with reqfile.open("w") as f:
+                    for r in requirements:
+                        print(r, file=f)
+                call(['pip', 'install', '-r', str(reqfile)], env=env)
```
joerick, line 187: `call()` should be able to take Path objects directly - this should work:

```suggestion
                call(['pip', 'install', '-r', reqfile], env=env)
```

test/test_pep518.py
```diff
@@ -47,3 +49,8 @@ def test_pep518(tmp_path):
     # check that the expected wheels are produced
     expected_wheels = utils.expected_wheels("spam", "0.1.0")
     assert set(actual_wheels) == set(expected_wheels)
+
+    assert not (project_dir / "42").exists()
```
joerick, line 53: Might be worth adding a comment why these asserts are here. Feel free to link to #421 if it's easier

YannickJadoul:
Also looks good to me, and approved by @joerick, so merging!

PR #422, opened 2020-08-14
ci: format updates 

henryiii:
The pre-commit mypy support is currently broken - I've fixed it here, and used pre-commit's GitHub Action to do the check, replacing the old CircleCI check. I've also run `pre-commit autoupdate` to bump to the latest version of pre-commit hooks. I've added a few more checks (bugbear and a few more pre-commit default checkers)

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -117,11 +117,12 @@ class DockerContainer:
 
         return [PurePath(p) for p in path_strs]
 
-    def call(self, args: Sequence[Union[str, PathLike]], env: Dict[str, str] = {},
+    def call(self, args: Sequence[Union[str, PathLike]], env: Optional[Dict[str, str]] = None,
              capture_output=False, cwd: Optional[Union[str, PathLike]] = None) -> str:
+
         chdir = f'cd {cwd}' if cwd else ''
         env_assignments = ' '.join(f'{shlex.quote(k)}={shlex.quote(v)}'
-                                   for k, v in env.items())
+                                   for k, v in env.items()) if env is not None else ''
         command = ' '.join(shlex.quote(str(a)) for a in args)
         end_of_message = str(uuid.uuid4())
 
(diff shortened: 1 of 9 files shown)


cibuildwheel/docker_container.py
```diff
@@ -117,11 +117,12 @@ def glob(self, path: PurePath, pattern: str) -> List[PurePath]:
 
         return [PurePath(p) for p in path_strs]
 
-    def call(self, args: Sequence[Union[str, PathLike]], env: Dict[str, str] = {},
+    def call(self, args: Sequence[Union[str, PathLike]], env: Optional[Dict[str, str]] = None,
```
henryiii, line 120: This is the one manual change - using a mutable default argument was caught by bugbear.

YannickJadoul:
Is there a reason to 1) move the flake & mypy to pre-commit (I thought the conclusion in #347 was that pre-commit was mostly provided optionally for convenience, but not really part of the toolset?), and 2) to move that job to CircleCI?
Not saying I'm per se against this, but just pointing this out. Probably up to @joerick to decide.

henryiii:
1. It's not moving anything; I explicitly made sure flake8 works properly when running it directly too. You can still run either check without pre-commit. 
2. It moves the job _from_ CircleCI to GHA, and runs it via pre-commit instead of directly, because a) pre-commit has great GHA tooling, b) GHA has nicer, more visible reports in the GH UI, c) pre-commit runs more checks than the old job, making it better for CI, and d) pre-commit sets up deterministic environments for checks, meaning you can exactly match CI locally.

henryiii:
I do highly recommend trying pre-commit out at some point (either manually with `pre-commit run --all-files`), or with the namesake pre-commit mode using `pre-commit install`. It saves CI cycles, and it provides a uniform environment for running and controlling checks. It's basically the same idea as `cibuildwheel`; really; make a python package that sets up a uniform self-contained environment for running things across all CI's and local machines, too.

YannickJadoul:
Alright, let's merge this, then! :-)

YannickJadoul:
Thanks again, @henryiii!

PR #424, opened 2020-08-27
Add section to the docs' FAQ about VC++ runtimes and VCRUNTIME140_1.dll on VS2019

### LINKED ISSUE #423: Picking specific VC runtime to use?
For Matplotlib 3.3.1, we started using cibuildwheel to produce wheels in [GitHub Actions](https://github.com/matplotlib/matplotlib/blob/master/.github/workflows/cibuildwheel.yml). Previously, they were built by [Christoph Gohlke](https://www.lfd.uci.edu/~gohlke/pythonlibs/#matplotlib) and then posted by us. 

We got [a report on 3.3.1](https://github.com/matplotlib/matplotlib/issues/18292) that the runtime requirements changed from `VCRUNTIME140.dll` to `VCRUNTIME140_1.dll`. In the cibuildwheel docs, I see something about Python 2 for Windows, but nothing specific about VC runtimes. Note, as we do not build for Python 2, I did _not_ install Visual C++ for Python 2.7.

Is there some way we can configure which runtime gets linked?

YannickJadoul:
A first effort to write up the Windows knowledge ﻿gathered over time, and in particular in relation to #423.
Might still need a bit of polishing, so get me some remarks! :-)

Closes #423

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -61,3 +61,11 @@ python3 -m cibuildwheel --output-dir wheelhouse
 python3 -m pip install twine
 python3 -m twine upload wheelhouse/*.whl
 ```
+
+### 'ImportError: DLL load failed: The specific module could not be found' error on Windows
+
+Visual Studio and MSVC link the compiled binary wheels to the Microsoft Visual C++ Runtime. Normally, these are included with Python, but when compiling with a newer version of Visual Studio, it is possible users run into problems on system that do not have these runtime libraries installed. The solution is to download the corresponding Visual C++ Redistributable from the [Microsoft website](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads).
+
+Additionally, Visual Studio 2019 started linking to an even newer DLL, `VCRUNTIME140_1.dll`, besides the `VCRUNTIME140.dll` that is included with recent Python versions (starting from Python 3.5; see [here](https://wiki.python.org/moin/WindowsCompilers) for more details on the corresponding Visual Studio & MSVC versions used to compile the diffent Python versions). To avoid this extra dependency on `VCRUNTIME140_1.dll`, the [`/d2FH4-` flag](https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/) can be added to the MSVC invocations (check out [this issue](https://github.com/joerick/cibuildwheel/issues/423) for details and references).
+
+To investigate the dependecies of a C extension (i.e., the `.pyd` file, a DLL in disguise) on Windows, [Dependency Walker](http://www.dependencywalker.com/) is a great tool.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -64,8 +64,20 @@ python3 -m twine upload wheelhouse/*.whl
 
 ### 'ImportError: DLL load failed: The specific module could not be found' error on Windows
 
-Visual Studio and MSVC link the compiled binary wheels to the Microsoft Visual C++ Runtime. Normally, these are included with Python, but when compiling with a newer version of Visual Studio, it is possible users run into problems on system that do not have these runtime libraries installed. The solution is to download the corresponding Visual C++ Redistributable from the [Microsoft website](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads).
+Visual Studio and MSVC link the compiled binary wheels to the Microsoft Visual C++ Runtime. Normally, these are included with Python, but when compiling with a newer version of Visual Studio, it is possible users will run into problems on systems that do not have these runtime libraries installed. The solution is to ask users to download the corresponding Visual C++ Redistributable from the [Microsoft website](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads) and install it. Since a Python installation normally includes these VC++ Redistributable files for [the version of the MSVC compiler used to compile Python](https://wiki.python.org/moin/WindowsCompilers), this is typically only a problem when compiling a Python 2.7 C extension with a newer compiler, e.g. to support a modern C++ standard (see [the section on modern C++ standards for Python 2.7](cpp_standards.md#windows-and-python-27) for more details).
 
 Additionally, Visual Studio 2019 started linking to an even newer DLL, `VCRUNTIME140_1.dll`, besides the `VCRUNTIME140.dll` that is included with recent Python versions (starting from Python 3.5; see [here](https://wiki.python.org/moin/WindowsCompilers) for more details on the corresponding Visual Studio & MSVC versions used to compile the diffent Python versions). To avoid this extra dependency on `VCRUNTIME140_1.dll`, the [`/d2FH4-` flag](https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/) can be added to the MSVC invocations (check out [this issue](https://github.com/joerick/cibuildwheel/issues/423) for details and references).
+
+To add the `/d2FH4-` flag to a standard `setup.py` using `setuptools`, the `extra_compile_args` option can be used:
+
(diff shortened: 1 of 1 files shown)


YannickJadoul:
@henryiii, if you have 5 minutes to spare at some point, and feel like giving this a read and terminology and sanity check, that'd be highly appreciated! :-)

docs/faq.md
```diff
@@ -61,3 +61,11 @@ python3 -m cibuildwheel --output-dir wheelhouse
 python3 -m pip install twine
 python3 -m twine upload wheelhouse/*.whl
 ```
+
+### 'ImportError: DLL load failed: The specific module could not be found' error on Windows
+
+Visual Studio and MSVC link the compiled binary wheels to the Microsoft Visual C++ Runtime. Normally, these are included with Python, but when compiling with a newer version of Visual Studio, it is possible users run into problems on system that do not have these runtime libraries installed. The solution is to download the corresponding Visual C++ Redistributable from the [Microsoft website](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads).
```
henryiii, line 67:
```suggestion
Visual Studio and MSVC link the compiled binary wheels to the Microsoft Visual C++ Runtime. Normally, these are included with Python, but when compiling with a newer version of Visual Studio, it is possible users will run into problems on systems that do not have these runtime libraries installed. The solution is to download the corresponding Visual C++ Redistributable from the [Microsoft website](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads).
```

henryiii: This is only a problem with 2.7, right? Python 3.5 was compiled with 14.x, so 3.5-3.9 do not have an issue; Python 3.5+ always includes `VCRUNTIME140.dll`? (see table https://wiki.python.org/moin/WindowsCompilers for reference)

Edit: yes, this is said below; I would reword this a little to clarify that this is only a problem for Python 2.7 on Windows when compiled with a non 2008 compiler. Otherwise, I think the MSVC redistributable must be present to run Python itself. (and of course, the problem below)

Czaki: No. Problem is that VisualStudio 2019 Some times links against `VCRUNTIME140_1.dll` next to `VCRUNTIME140.dll` Which is only in modern version of Microsoft Visual C++ Redistributable. older ons contains only  `VCRUNTIME140.dll`

henryiii: That's what I meant by "the problem below". This seems to indicate that `VCRUNTIME140.dll` will sometimes be missing; whereas it will only possibly be missing in the single case where we use the Python 2.7 hack (like in pybind11). Python 3.5+ require `VCRUNTIME140.dll` to run, so it can't be missing.

The paragraph below is fine.

docs/faq.md
```diff
@@ -61,3 +61,11 @@ python3 -m cibuildwheel --output-dir wheelhouse
...
+
+Additionally, Visual Studio 2019 started linking to an even newer DLL, `VCRUNTIME140_1.dll`, besides the `VCRUNTIME140.dll` that is included with recent Python versions (starting from Python 3.5; see [here](https://wiki.python.org/moin/WindowsCompilers) for more details on the corresponding Visual Studio & MSVC versions used to compile the diffent Python versions). To avoid this extra dependency on `VCRUNTIME140_1.dll`, the [`/d2FH4-` flag](https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/) can be added to the MSVC invocations (check out [this issue](https://github.com/joerick/cibuildwheel/issues/423) for details and references).
+
```
joerick, line 70: Could we add a simple example of how you might use the '/d2FH4-' flag? Perhaps something like

~~~markdown
To use the `/d2FH4-` flag you could do something like this in your `setup.py`,

```python
    ext_modules=[
        Extension(
            'c_module',
            sources=['extension.c'],
            extra_compile_args=['/d2FH4-'] if sys.platform == 'win32' else []
        )
    ],
```

~~~

(some people don't know about the `extra_compile_args` argument)

docs/faq.md
YannickJadoul: @henryiii Thanks, yes. I meant to give a generic overview of the situation first. Somehow, this is not always clear when looking for issues. I'll make a link to the Python 2.7 section, though (we already have this, but it's in the "modern C++" part of the docs, because that's where a modern VS is needed).

docs/faq.md
YannickJadoul: Done!

docs/faq.md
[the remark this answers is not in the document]
YannickJadoul: Yup, done!

YannickJadoul: Added an extra sentence and some additional links. I do feel like I'm repeating myself a bit, now? Any suggestions on how to collapse sentences, or is this fine?

docs/faq.md
```diff
@@ -61,3 +61,23 @@ python3 -m cibuildwheel --output-dir wheelhouse
...
+    ext_modules=[
+        Extension(
+            'c_module',
+            sources=['extension.c'],
+            extra_compile_args=['/d2FH4-'] if sys.platform == 'win32' else []
+        )
+    ],
+```
+
+To investigate the dependecies of a C extension (i.e., the `.pyd` file, a DLL in disguise) on Windows, [Dependency Walker](http://www.dependencywalker.com/) is a great tool.
```
joerick, line 83: spelling
```suggestion
To investigate the dependencies of a C extension (i.e., the `.pyd` file, a DLL in disguise) on Windows, [Dependency Walker](http://www.dependencywalker.com/) is a great tool.
```

YannickJadoul: :+1: :-)

docs/faq.md
```diff
@@ -61,3 +61,23 @@ python3 -m cibuildwheel --output-dir wheelhouse
...
+
+Additionally, Visual Studio 2019 started linking to an even newer DLL, `VCRUNTIME140_1.dll`, besides the `VCRUNTIME140.dll` that is included with recent Python versions (starting from Python 3.5; see [here](https://wiki.python.org/moin/WindowsCompilers) for more details on the corresponding Visual Studio & MSVC versions used to compile the diffent Python versions). To avoid this extra dependency on `VCRUNTIME140_1.dll`, the [`/d2FH4-` flag](https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/) can be added to the MSVC invocations (check out [this issue](https://github.com/joerick/cibuildwheel/issues/423) for details and references).
```
joerick, line 69: typo 'different'
```suggestion
Additionally, Visual Studio 2019 started linking to an even newer DLL, `VCRUNTIME140_1.dll`, besides the `VCRUNTIME140.dll` that is included with recent Python versions (starting from Python 3.5; see [here](https://wiki.python.org/moin/WindowsCompilers) for more details on the corresponding Visual Studio & MSVC versions used to compile the different Python versions). To avoid this extra dependency on `VCRUNTIME140_1.dll`, the [`/d2FH4-` flag](https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/) can be added to the MSVC invocations (check out [this issue](https://github.com/joerick/cibuildwheel/issues/423) for details and references).
```

joerick:
I think this looks good, it's better to be explicit in these matters!

docs/faq.md
YannickJadoul: Ugh, sorry :-|

YannickJadoul:
OK, fixed all than (and insisted Travis CI to become green)

joerick (approved):

YannickJadoul:
Thanks, @QuLogic!

Last check with @henryiii, who suggested/requested some changes. Otherwise, I'll merge this :-)

YannickJadoul:
Hurray. Thanks for the feedback, all! :-)

PR #428, opened 2020-09-09
Add project_urls to setup.py

adamchainz:
These links make it easier to jump from PyPI to the respective page.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,10 @@ setup(
     author="Joe Rickerby",
     author_email='joerick@mac.com',
     url='https://github.com/joerick/cibuildwheel',
+    project_urls={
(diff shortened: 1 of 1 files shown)


adamchainz:
Thanks for this amazing tool btw.

You can see an example of what `project_urls` looks like on my packages, e.g. https://pypi.org/project/apig-wsgi/ - see the left sidebar.

I just hunted around for a minute before finding the cibuildwheel changelog so this would at least help future me 😅

YannickJadoul:
Ah, just too late; those links will have to wait for 1.6.1, then :-)
But thanks, good catch!

YannickJadoul (approved):

joerick:
Thanks @adamchainz !

PR #432, opened 2020-09-16
README: add etebase-py to the list of users

tasn:


Czaki (approved):

joerick:
nice, a rust example!

tasn:
Yeah, rust. :)
Thanks for making this project, it made building Python packages a lot easier! I don't know if you looked at the `.travis.yml` file or its history, but it was quite the challenge to get things working. Essentially, because of the way cibuildwheel works, it was rebuilding the project from scratch every time. I now set up sccache (essentially ccache alternative that supports Rust) but it's still not terribly fast.

joerick:
Oh, nice, that's super interesting. I wonder if that would improve more with an external cache backend for sccache?

Czaki:
>  Thanks for making this project, it made building Python packages a lot easier! I don't know if you looked at the `.travis.yml` file or its history, but it was quite the challenge to get things working. Essentially, because of the way cibuildwheel works, it was rebuilding the project from scratch every time. I now set up sccache (essentially ccache alternative that supports Rust) but it's still not terribly fast.

This decision was made to reduce pollution of build with binaries for wrong versions of python. 
In my case I build non glue part using `CIBW_BEFORE_BUILD` I do not understand how `setuptools-rust` work (there is nothing in `etebase/etebase_python` but maybe `build.rs` could perform partial build.

tasn:
@joerick, yeah, probably. Though it's a pity to setup an external cache backend rather than use travis. I think it would be better to just mount a host directory inside the docker image. I vaguely remember that was possible (though maybe it was just a suggestion in a ticket). That will let me share these nicely by just setting the sccache directory to a different location.

@Czaki, I totally get the rationale. More than that, I fell victim to this pollution before, so I know it's a real issue. That's why I'm using sccache. Building non-glue part could make sense I guess, though that's essentially what sccache does, and I'm not sure how to do it correctly because cibuildwheel sets the build directory to a tmp one for every build.

Anyhow, I managed to mostly overcome this issue by just limiting the amount of Python versions I build for and just waiting for the long build to finish. I don't currently have time to speed things up further. I just wanted to flag this for you all so you are aware.

Thanks again for this project!

Czaki:
> Anyhow, I managed to mostly overcome this issue by just limiting the amount of Python versions I build for and just waiting for the long build to finish. I don't currently have time to speed things up further. I just wanted to flag this for you all so you are aware.

As I see in comments you have problems with speed of build on macos and windows. My experience is that they work faster on Github Actions or Azure Pipelines. So maybe switch service may be solution for You. 



> Building non-glue part could make sense I guess, though that's essentially what sccache does, and I'm not sure how to do it correctly because cibuildwheel sets the build directory to a tmp one for every build

`CIBW_BEFORE_BUILD` is run on repository dir, before switch to `tmp`. But I agree. If You have working solution it is not worth of time to switch it.

YannickJadoul:
If it helps, I have some hacks to actually cache `ccache` between different builds [here](https://github.com/YannickJadoul/Parselmouth/blob/081afbb73703019b7c899638a4b6670a65fc8f9e/.travis.yml):

https://github.com/YannickJadoul/Parselmouth/blob/081afbb73703019b7c899638a4b6670a65fc8f9e/.travis.yml#L150
https://github.com/YannickJadoul/Parselmouth/blob/081afbb73703019b7c899638a4b6670a65fc8f9e/.travis.yml#L153-L156
https://github.com/YannickJadoul/Parselmouth/blob/081afbb73703019b7c899638a4b6670a65fc8f9e/.travis.yml#L160

As far as I can remember, especially the `CCACHE_BASEDIR` is important, because `pip wheel` creates a temporary directory to build in.

tasn:
> As I see in comments you have problems with speed of build on macos and windows. My experience is that they work faster on Github Actions or Azure Pipelines. So maybe switch service may be solution for You.

Thanks for the tip! I'll have to take a look at enabling those.

> CIBW_BEFORE_BUILD is run on repository dir, before switch to tmp. But I agree. If You have working solution it is not worth of time to switch it.

Ahh right. I'll take a look at that the next time I'm taking a jab at speeding things up. Maybe it won't even be needed if I end up using Github actions.

> If it helps, I have some hacks to actually cache ccache between different builds here:

Thanks! I'll take a look the next time I'm trying to speed things up.

PR #436, opened 2020-09-20
Update pinned_docker_images.cfg after the ppc64le manylinux2014 image is built again on Travis CI

YannickJadoul:
See also pypa/manylinux#758 and pypa/manylinux#766


YannickJadoul:
Not sure if this is worth a patch release, actually; up to you @joerick!

joerick:
Great, thank you! Yeah I'll cut a new release with this.

YannickJadoul:
> Great, thank you! Yeah I'll cut a new release with this.

Potentially, we could also wait until PyPy 7.3.2 is released?

joerick:
Ah, just pushed it! I can do another when that lands.

PR #438, opened 2020-09-26
Drop Python 3.5

YannickJadoul:
Python 3.5 is [nearing its EOL](https://www.python.org/dev/peps/pep-0478/):

> There are no specific plans for the next release of Python 3.5. Note that Python 3.5 will no longer be supported after September 2020. Therefore, 3.5.10 may be the final release of Python 3.5.﻿

This is preparing to drop it. Though I'm not saying we should per se do so as soon as possible; this might also require a new minor version bump?


Czaki:
Nice clean

YannickJadoul:
> Nice clean

Yes! A nice bunch of workarounds to lose! :-)

YannickJadoul:
Btw, @Czaki, I just did a quick search for `3.5` and `35`, but I might still have missed anything, if you want to check.

henryiii:
What about dropping it when the manylinux images drop it? That, or possibly wait until most upstream projects drop it.

Currently, manylinux2010 has:

```
cp27-cp27m  cp27-cp27mu  cp35-cp35m  cp36-cp36m  cp37-cp37m  cp38-cp38  cp39-cp39
```

Czaki:
Maybe first make release with Python 3.9 stable release? 
This PR is really nice clean of code base. Any of my project which need wheel support python 3.5 but I understand @henryiii point.

YannickJadoul:
> What about dropping it when the manylinux images drop it? That, or possibly wait until most upstream projects drop it.

The problem is that we'll have a (hopefully short) period where cibuildwheel would be broken, then (unless `CIBW_SKIP=cp35-*` is added), if the manylinux images drop it before we do.
But some coordination with manylinux would be nice, indeed.

joerick:
> The problem is that we'll have a (hopefully short) period where cibuildwheel would be broken, then (unless CIBW_SKIP=cp35-* is added), if the manylinux images drop it before we do.

I don't think that's true, unless the user is manually specifying to use the manylinux*:latest image. Reliance on pinned images prevents that.

I like the code simplifications this provides.

Here's the current approximate version use on PyPI:

```sql
SELECT t0.num_downloads, t0.python_version FROM (SELECT
  COUNT(*) AS num_downloads,
  REGEXP_EXTRACT(details.python, r"^(\d+\.\d+)") as python_version
WHERE
  _TABLE_SUFFIX
    BETWEEN FORMAT_DATE(
GROUP BY python_version
ORDER BY num_downloads DESC) AS t0 LIMIT 100; 
```

image

So 3.5 is still at around 5%.  But I think there's a good argument to pull the plaster, because, unlike 2.7, there's a very easy upgrade path for 3.5 users - they should probably just update their interpreter.

> Another possibility is dropping it at the Ubuntu 16.04 EOL (April 30, 2021); that's still on some CIs

April 2021 seems too long further to wait, to me. I'd guess if you're on that platform and using Python seriously you should probably find a way to upgrade too.

YannickJadoul:
> I don't think that's true, unless the user is manually specifying to use the manylinux*:latest image. Reliance on pinned images prevents that.

Oh, yes, I completely forgot about that; good point!

Anyway, I just made this because the EOL was on its way. I don't urgently need to have this PR in, but it's ready when we want :-)

henryiii:
Given that cibuildwheel is a foundational package, dropping 3.5 support would mean that it would force all libraries that depend on it to drop 3.5 support, or to stop upgrading. Right now, upgrading to get 3.9 final support is important, and in the long run, since cibuildwheel sits on top of CI, if there's a change in CI that would require an upgrade, this could force a dependent library.

In this way, it's a little different than the "average" package, or even a library like NumPy - using an older version is just fine in most cases.

For example, in boost-histogram, I want to drop 2.7 and 3.5 support as soon as we release version 1.0, which is still ~1-2 months away, (due heavily to other obligations on my part, and that it's hard to get a consensus on a few design points). I would prefer not to be forced into dropping 3.5 right now because cibuildwheel does.

My minimal recommendation would be to wait at least a few weeks to get 3.9 solidly in before dropping 3.5. And really, I think it's a good idea to go with PyPA and provide exactly what is provided in the manylinux image. 3.4 _was_ in that image, but it was eventually removed.

YannickJadoul:
> My minimal recommendation would be to wait at least a few weeks to get 3.9 solidly in before dropping 3.5. And really, I think it's a good idea to go with PyPA and provide exactly what is provided in the manylinux image. 3.4 _was_ in that image, but it was eventually removed.

OK, yes, I agree with that :-)
I was also planning to get out a final release supporting 3.5, but thought it would be dangerous. I overlooked the effect of all the work on the deterministic docker images and package versions in cibuildwheel. That's really paying off now! :-)

henryiii:
Scared to ask, but what's in that purple sliver, and how large is it? :)

I recently guessed that 3.7 was the most popular version, followed by 3.6, then 2.7, then 3.8. Wow. :)

I wouldn't be surprised if CERN is responsible for a fraction of that 2.6 sliver; experiments don't upgrade OS's after starting to collect data, so there are Scientific Linux 6 (RHEL 6 base) systems that run Python 2.6 still around. Not much is done in 2.6, as stacks generally have 2.7, but some basic config and setup of the stacks is in 2.6, and users who don't know how to enable the stack get stuck in 2.6.

henryiii:
FYI, numba / llvmlite completely messed up `python_requires` when dropping 3.5, so you can't type `pip install numba` on 3.5 without going through pinning hoops (llvmlite must be explicitly pinned too). I think they've learned their lesson for next time, and "yanking" on PyPI helps.

The idea behind the `python_requires` system, though, is that old versions of Python just keep working with regular pip installs after libraries move on (which is entirely why NEP 29 exists).

Complete tangent, not really related.

henryiii:
@joerick Mind if I borrow that image to post in a couple of Slack channels? :)

YannickJadoul:
> Not much is done in 2.6, as stacks generally have 2.7, but some basic config and setup of the stacks is in 2.6

Slightly relaxing (or scary? I actually don't know) thought: 2.6 is also still the default when you run `python` inside `manylinux` (at least `manylinux1` and `manylinux2010`). So from that perspective, let's hope a reasonable part of those percentages comes from mistakenly using `python` in such scenarios, when actually, newer versions are installed?

> Scared to ask, but what's in that purple sliver, and how large is it? :)

Hopefully 3.9, no? Some CI runs etc, trying to prepare for that release? :-)

joerick:
> @joerick Mind if I borrow that image to post in a couple of Slack channels? :)

Please do!

Also here's the raw data if you're curious-

```
num_downloads	python_version
1777310885	3.7
1078476452	3.6
782712455	2.7
665109537	3.8
235199405	3.5
69539508	2.6
22927295	3.4
1362920	3.9
359998	3.10
33438	3.3
8639	3.2
278	2.4
267	2.5
13	2.8
2	3.1
2	2.1
1	3.0
```

henryiii:
Still 200M downloads of Python 3.5 packages.

I want to find the 13 people using 2.8. :) (Guessing it's Tauthon, perhaps?)

And there's one poor person stuck on Python 3.0...

joerick:
> Given that cibuildwheel is a foundational package, dropping 3.5 support would mean that it would force all libraries that depend on it to drop 3.5 support, or to stop upgrading. Right now, upgrading to get 3.9 final support is important, and in the long run, since cibuildwheel sits on top of CI, if there's a change in CI that would require an upgrade, this could force a dependent library to upgrade and lose 3.5.

> In this way, it's a little different than the "average" package, or even a library like NumPy - using an older version is just fine in most cases. Also, since cibuildwheel is pretty heavily used, I bet some fraction of that 5% are cibuildwheel packages.

> For example, in boost-histogram, I want to drop 2.7 and 3.5 support as soon as we release version 1.0, which is still ~1-2 months away (due heavily to other obligations on my part, and that it's hard to get a consensus on a few design points). I would prefer not to be forced into dropping 3.5 right now because cibuildwheel does.

Thanks for this feedback, it's a valuable perspective. I guess we have to trade-off these sorts of issues with how much it costs maintainers and the community to keep support moving forward. But right now that's not a lot of work :)

> My minimal recommendation would be to wait at least a few weeks to get 3.9 solidly in before dropping 3.5. And really, I think it's a good idea to go with PyPA and provide exactly what is provided in the manylinux image. 3.4 was in that image, but it was eventually removed.

I think waiting for 3.9 full release makes a lot of sense. From there, if we still think it's important to users, I'm happy to be a little pragmatic - we can continue supporting 3.5 until it starts costing us, in effort or increased complexity.

henryiii:
FYI, Manylinux should be dropping 3.5 today. https://github.com/pypa/manylinux/issues/995 (Not for manylinux1, irritatingly). I'd be fine with dropping it either immediately after our next release (should be soonish?), or now.

mayeut:
I updated #596 to also drop python 3.5 since the secret target date of @henryiii for dropping python 2.7 is the same as for python 3.5
It might be worth doing a last release with 2.7/3.5 support before that happens.
The removal of Python 3.5 will probably happen next week (I'm planning to add 3.10.0b1 when I remove 3.5 and 3.10.0b1 is planned on Monday).

PR #439, opened 2020-10-04
Add xmlstarlet to Working examples.

dimitern:
Hey @joerick!

First, thank you for the amazing project! I'd like to suggest adding my package to the examples, which uses `cibuildwheel` with GitHub Actions to build binary wheels for Python 3.6+ and Linux, MacOS, and Windows.

Some notes:
 - The easiest way to install and use the tool on Windows (from Python), also builds a truly native MSVC win_amd64 binary (not MinGW64).
 - I couldn't find a good example of a usable CFFI cross-platform binary wheel build setup (esp. when the original C sources only build under MinGW32).
 - I hope it's useful to somebody having difficulty with native Windows wheel builds. 

Cheers,
Dimiter

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Here are some repos that use cibuildwheel.
 - [pikepdf](https://github.com/pikepdf/pikepdf)
 - [fathon](https://github.com/stfbnc/fathon)
 - [etebase-py](https://github.com/etesync/etebase-py) - Python bindings to a Rust library using `setuptools-rust`, and `sccache` for improved speed, built on Travis CI.
-
+- [xmlstarlet](https://github.com/dimitern/xmlstarlet) - Python CFFI bindings to XMLStarlet XML processing CLI.
 > Add your repo here! Send a PR.
 >
 > <sup>I'd like to include notes here to indicate why an example might be interesting to cibuildwheel users - the styles/technologies/techniques used in each. Please include that in future additions!</sup>
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Here are some repos that use cibuildwheel.
 - [pikepdf](https://github.com/pikepdf/pikepdf)
 - [fathon](https://github.com/stfbnc/fathon)
 - [etebase-py](https://github.com/etesync/etebase-py) - Python bindings to a Rust library using `setuptools-rust`, and `sccache` for improved speed, built on Travis CI.
-- [xmlstarlet](https://github.com/dimitern/xmlstarlet) - Python CFFI bindings to XMLStarlet XML processing CLI.
+- [xmlstarlet](https://github.com/dimitern/xmlstarlet) - Python 3.6+ CFFI bindings with true MSVC build and GitHub Actions.
 > Add your repo here! Send a PR.
 >
 > <sup>I'd like to include notes here to indicate why an example might be interesting to cibuildwheel users - the styles/technologies/techniques used in each. Please include that in future additions!</sup>


README.md
```diff
@@ -142,7 +142,7 @@ Here are some repos that use cibuildwheel.
 - [pikepdf](https://github.com/pikepdf/pikepdf)
 - [fathon](https://github.com/stfbnc/fathon)
 - [etebase-py](https://github.com/etesync/etebase-py) - Python bindings to a Rust library using `setuptools-rust`, and `sccache` for improved speed, built on Travis CI.
-
+- [xmlstarlet](https://github.com/dimitern/xmlstarlet) - Python CFFI bindings to XMLStarlet XML processing CLI.
```
henryiii, line 145: Could you explain in an abridged version why these are interesting here? (You've already done that in this PR description) "XMLStarlet XML processing CLI" isn't interesting, IMO, but "Python CFFI bindings with true MSVC build" or something like that is.

henryiii:
> truly native MSVC win_amd64

Personal education: We don't use MinGW in cibuildwheel, do we?

YannickJadoul:
> Personal education: We don't use MinGW in cibuildwheel, do we? This was specific to it being required by the library being bound?

No, we don't. Python is built with MSVC, so it's adviced to also build extensions with MSVC.

Did you do anything special here, @dimitern, to enforce this? Because @henryiii has a point, that all cibuildwheel wheels are built with MSVC and not with MinGW.

henryiii:
I have a feeling it has something to do with:

> esp. when the original C sources only build under MinGW32

But let's see.

README.md
dimitern: Sure, I'll try to rephrase it better. Thanks!

dimitern:
> > Personal education: We don't use MinGW in cibuildwheel, do we? This was specific to it being required by the library being bound?
> 
> No, we don't. Python is built with MSVC, so it's adviced to also build extensions with MSVC.
> 
> Did you do anything special here, @dimitern, to enforce this? Because @henryiii has a point, that all cibuildwheel wheels are built with MSVC and not with MinGW.

I started from the original GNUmakefile with MinGW and got it to work (by mostly rewriting it) under the windows-latest image and a combination of msys2_shell action and some hacks. All the C dependencies built OK and even the EXE was produced, but with msys-2.0 runtime DLL dependency. Then I realized it won't work with the usual cibuildwheel Windows build as there was no mingw target. I managed to get a binary mingw wheel in a local VM without cibuildwheel. Perhaps it might be useful to think about this edge case. Not sure how useful mingw wheels will be in general.

Czaki:
> Not sure how useful mingw wheels will be in general

What Is version tag for mingw wheel? Did this tag is accepted by pypi? Some times ago we have talks about cygwin wheel, but we foud that it cannot be uploaded to pypi so there is no profit from add this to cibuildwheel.

dimitern:
> > Not sure how useful mingw wheels will be in general
> 
> What Is version tag for mingw wheel? Did this tag is accepted by pypi? Some times ago we have talks about cygwin wheel, but we foud that it cannot be uploaded to pypi so there is no profit from add this to cibuildwheel.

I couldn't find a reference that mingw wheels for Windows are allowed on PyPI. So no need to support mingw wheels.

joerick:
Sorry, it appears I forgot to merge this one! Thanks!

myselfhimself:
> > > Personal education: We don't use MinGW in cibuildwheel, do we? This was specific to it being required by the library being bound?
> > 
> > 
> > No, we don't. Python is built with MSVC, so it's adviced to also build extensions with MSVC.
> > Did you do anything special here, @dimitern, to enforce this? Because @henryiii has a point, that all cibuildwheel wheels are built with MSVC and not with MinGW.
> 
> I started from the original GNUmakefile with MinGW and got it to work (by mostly rewriting it) under the windows-latest image and a combination of msys2_shell action and some hacks. All the C dependencies built OK and even the EXE was produced, but with msys-2.0 runtime DLL dependency. Then I realized it won't work with the usual cibuildwheel Windows build as there was no mingw target. I managed to get a binary mingw wheel in a local VM without cibuildwheel. Perhaps it might be useful to think about this edge case. Not sure how useful mingw wheels will be in general.

@dimitern 

I am also on this track of  the Github action https://github.com/marketplace/actions/setup-msys2 hack to have a proper msys2 shell for the gmic-py project.. though just starting and looking for a Github Workflow file to imitate.. Will follow anything related to Mingw/msys2 building using cibuildwheel very closely here...

My related WIP msys2 workflow file is here: https://github.com/myselfhimself/gmic-py/blob/windows-support/.github/workflows/cpythonwindowsbuild.yml

PR #445, opened 2020-10-08
README: Mention which CIs support ARM builds

jbarlow83:


README.md
```diff
@@ -40,14 +40,16 @@ Usage
...
 
-|                 | Linux | macOS | Windows |
-|-----------------|-------|-------|---------|
-| Azure Pipelines | ✅    | ✅    | ✅      |
-| Travis CI       | ✅    | ✅    | ✅      |
-| AppVeyor        | ✅    | ✅    | ✅      |
-| GitHub Actions  | ✅    | ✅    | ✅      |
-| CircleCI        | ✅    | ✅    |         |
-| Gitlab CI       | ✅    |       |         |
+|                 | Linux | macOS | Windows | Supports ARM |
```
Czaki, line 43: Linux ARM

README.md
```diff
@@ -40,14 +40,16 @@ Usage
...
+|-----------------|-------|-------|---------|--------------|
+| Azure Pipelines | ✅    | ✅    | ✅      | ✴️¹           |
+| Travis CI       | ✅    | ✅    | ✅      | ✅           |
+| AppVeyor        | ✅    | ✅    | ✅      |              |
+| GitHub Actions  | ✅    | ✅    | ✅      |              |
```
Czaki, line 48: also support self hosted ARM https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/

But We have no possibilities to test if cibuilwheel works properly on such machines.

README.md
henryiii:
```suggestion
|                 | Linux | macOS | Windows | Linux ARM |
```

README.md
henryiii:
```suggestion
| GitHub Actions  | ✅    | ✅    | ✅      | ✴️¹           |
```

henryiii: Ahh, I just added this, assuming it would work, but maybe we should not if we can't test it? I guess Azure was tested at one point by someone? Though I really think it would work on GHA if it works on Azure, they are basically the same. And "third-party" already indicates a lot of work needs to be done to set it up.

jbarlow83: I suppose we could dial back the footnote to explain the situation: "These CIs require self-hosted runners for ARM. cibuildwheel probably works but this has not been confirmed."

README.md
```diff
@@ -40,14 +40,16 @@ Usage
...
+|                 | Linux | macOS | Windows | Linux ARM |
+|-----------------|-------|-------|---------|--------------|
+| Azure Pipelines | ✅    | ✅    | ✅      | ✴️¹           |
+| Travis CI       | ✅    | ✅    | ✅      | ✅           |
+| AppVeyor        | ✅    | ✅    | ✅      |              |
+| GitHub Actions  | ✅    | ✅    | ✅      | ✴️¹           |
+| CircleCI        | ✅    | ✅    |         |              |
+| Gitlab CI       | ✅    |       |         |              |
+
+<sup>¹ Supported with a "third-party build host".</sup><br>
```
henryiii, line 52:
```suggestion
<sup>¹  Requires a "third-party build host"; expected to work with cibuildwheel but not directly tested by our CI.</sup><br>
```

PR #447, opened 2020-10-08
Macos certif

### LINKED ISSUE #443: SSL issues with OSX on Azure
A contributor to my project recently enabled a test that attempts to download a resource using HTTPS. It seems that running these tests on Azure fails for the OSX environments we have with errors like:

```
 (<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:841)>)
```

Which leads to issues on github like https://github.com/scikit-learn/scikit-learn/issues/10201#issuecomment-365734582. It seems the "proper" solution is to properly install the SSL certificates that come with the python installation: https://bugs.python.org/issue29480 which can apparently be fixed with:

```
sudo /Applications/Python\ 3.6/Install\ Certificates.command
```

I'm not really familiar with how cibuildwheel is using the base image from the build environment, but is this something I should be able to control in my Azure environments? Is this something that can and should be fixed in cibuildwheel? Thanks for any help.

I see there are issues already related to SSL stuff, but they all seem a little different than this or say that this should already be fixed. I must be missing something.

Czaki:
This PR is for fix #443 
Because try of using `/Applications/Python 2.7/Install Certificates.command` fail I decide to vendor in script base on one present on my machine.

I'm not sure what to do with python 3.5 so I skip this step on this version. @mayeut any suggestion?  

The test for this case could be hard. We could download something from github.com, but any break in this functionality may produce error long time after change happen.


henryiii:
Another solution I've seen mentioned that might include Python 3.5 is:

```bash
pip install certifi
CERT_PATH=$(python -m certifi)
export SSL_CERT_FILE=${CERT_PATH}
export REQUESTS_CA_BUNDLE=${CERT_PATH}
```

But I think the problem is that Python 3.6+ have their own copy of OpenSSL, while 3.5 does not and uses the system copy, so does this even show up at all for 3.5?

joerick:
Thanks for looking at this @Czaki. But due to the existing `test_ssl` I'm still unsure exactly what we're fixing here. As I noted in #443, I'd like to see a failing test that recreates the SSL error before we attempt any fixes using `Install Certificates.command` or similar.

cibuildwheel/macos.py
```diff
@@ -11,7 +11,7 @@
 from .environment import ParsedEnvironment
 from .util import (BuildOptions, BuildSelector, NonPlatformWheelError, download,
                    get_build_verbosity_extra_flags, get_pip_script,
-                   prepare_command)
+                   prepare_command, instal_certifi_script)
```
djhoese, line 14: Typo in this import "instal" versus "install"

djhoese: Ah nevermind. That's how it is in the other module too.

YannickJadoul: Still a typo, though ;-)

Czaki: fixed

YannickJadoul: Great, thanks! :-)

mayeut:
> But I think the problem is that Python 3.6+ have their own copy of OpenSSL, while 3.5 does not and uses the system copy, so does this even show up at all for 3.5?

The system copy was so old on python 3.5 that it did not support TLS 1.2 which caused some issue. Its openssl version has been patched in #71 not to use the system one.

> I'm not sure what to do with python 3.5 so I skip this step on this version. @mayeut any suggestion?

It shall be patched like the other versions.

djhoese:
I just tried forcing the TLS protocol version in my branch for vispy and it **worked**! I made sure not to use the dev version of cibuildwheel and I don't think I was skipping the test or anything. I think this is the main cause. Can someone summarize all of the issues related to why cibuildwheel started doing that and why it works?

mayeut:
@joerick, I can reproduce the issue: https://github.com/mayeut/cibuildwheel/runs/1233247122?check_suite_focus=true#step:6:158

Czaki:
my test run does not finish yet, but basic on @mayeut I expand the test set. 

But I think that we should test more ssl context cases.

Czaki:
@mayeut  I got this for python 3.5
```
FileNotFoundError: [Errno 2] No such file or directory: '/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl'
```

Did you have any idea? Or omit python 3.5 as there is plan to drop its support?

Czaki:
@mayeut When omit `test_ssl` for python 3.5 then everything pass. But for python 3.5 its fail. Could You review my changes? Have You any idea why python 3.5 does not use certificates from `certifi`?

mayeut:
> @mayeut When omit test_ssl for python 3.5 then everything pass. But for python 3.5 its fail. Could You review my changes? Have You any idea why python 3.5 does not use certificates from certifi?

I just tested "manually" the patch on my machine (i.e., just following the logic):
before:
```
Matt$ /Library/Frameworks/Python.framework/Versions/3.5/bin/python3
Python 3.5.4 (v3.5.4:3f56838976, Aug  7 2017, 12:56:33) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import ssl
>>> ssl.OPENSSL_VERSION
'OpenSSL 1.0.2u  20 Dec 2019'
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(cafile=None, capath=None, openssl_cafile_env='SSL_CERT_FILE', openssl_cafile='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/cert.pem', openssl_capath_env='SSL_CERT_DIR', openssl_capath='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/certs')
>>> from urllib.request import urlopen
>>> data = urlopen("https://raw.githubusercontent.com/joerick/cibuildwheel/master/README.md")
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/urllib/request.py", line 1254, in do_open
    h.request(req.get_method(), req.selector, req.data, headers)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 1107, in request
    self._send_request(method, url, body, headers)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 1152, in _send_request
    self.endheaders(body)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 1103, in endheaders
    self._send_output(message_body)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 934, in _send_output
    self.send(msg)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 877, in send
    self.connect()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py", line 1261, in connect
    server_hostname=server_hostname)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/ssl.py", line 385, in wrap_socket
    _context=self)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/ssl.py", line 760, in __init__
    self.do_handshake()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/ssl.py", line 996, in do_handshake
    self._sslobj.do_handshake()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/ssl.py", line 641, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:719)
```
patch logic:
```
Matt$ mkdir -p /Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/
Matt$ cp /Library/Frameworks/Python.framework/Versions/3.6/etc/openssl/cert.pem /Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/
```
after:
```
Matt$ /Library/Frameworks/Python.framework/Versions/3.5/bin/python3
Python 3.5.4 (v3.5.4:3f56838976, Aug  7 2017, 12:56:33) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import ssl
>>> ssl.get_default_verify_paths()
DefaultVerifyPaths(cafile='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/cert.pem', capath=None, openssl_cafile_env='SSL_CERT_FILE', openssl_cafile='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/cert.pem', openssl_capath_env='SSL_CERT_DIR', openssl_capath='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/certs')
>>> from urllib.request import urlopen
>>> data = urlopen("https://raw.githubusercontent.com/joerick/cibuildwheel/master/README.md")
>>> data.read()
b'cibuildwheel\n============\n\n[![PyPI](https://img.shie................
```

So it should be working. Now investigating why it's failing in CI.

mayeut:
With an updated test case to print `ssl.get_default_verify_paths()`, I'm still getting the following
```
DefaultVerifyPaths(cafile=None, capath=None, openssl_cafile_env='SSL_CERT_FILE', openssl_cafile='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/cert.pem', openssl_capath_env='SSL_CERT_DIR', openssl_capath='/Library/Frameworks/Python.framework/Versions/3.5/etc/openssl/certs')
```

`cafile` should not be `None` here.

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
...
+def main():
+    openssl_dir, openssl_cafile = os.path.split(
+        ssl.get_default_verify_paths().openssl_cafile)
+    print(" -- pip install --upgrade certifi")
+    subprocess.check_call([sys.executable,
+        "-E", "-s", "-m", "pip", "install", "--upgrade", "certifi"])
+    import certifi
+    # change working directory to the default SSL directory
+    if sys.version_info[0:2] == (3, 5):
+        os.makedirs(openssl_dir, exist_ok=True, mode=775)
```
mayeut, line 30:
```suggestion
        os.makedirs(openssl_dir, exist_ok=True, mode=0o775)
```

mayeut (changes requested):
With `mode` changed with octal `0o775`, tests are passing as expected. c.f. https://github.com/mayeut/cibuildwheel/runs/1235348121?check_suite_focus=true

YannickJadoul:
> With `mode` changed with octal `0o775`, tests are passing as expected.

Great! Thanks you so much for checking and finding that, @mayeut! :-)

@Czaki, I've taken the liberty to apply @mayeut's suggestion, and (hopefully) get a green build.

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
+# Based on: https://github.com/python/cpython/blob/master/Mac/BuildScript/resources/install_certificates.command
+
+# install_certifi.py
+#
+# sample script to install or update a set of default Root Certificates
+# for the ssl module.  Uses the certificates provided by the certifi package:
+#       https://pypi.org/project/certifi/
+import os
```
YannickJadoul, line 8: New line, before?

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
...
+import os.path
+import ssl
+import stat
+import subprocess
+import sys
+STAT_0o775 = ( stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR
```
YannickJadoul, line 14: New line, before?

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
...
+             | stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP
+             | stat.S_IROTH |                stat.S_IXOTH )
+
+if sys.version_info[0] == 2:
+    FileNotFoundError = OSError
+
+def main():
```
YannickJadoul, line 21: Two new lines before?

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        os.remove(openssl_cafile)
+    except FileNotFoundError:
+        pass
+    print(" -- creating symlink to certifi certificate bundle")
+    os.symlink(relpath_to_certifi_cafile, openssl_cafile)
+    print(" -- setting permissions")
+    os.chmod(openssl_cafile, STAT_0o775)
+    print(" -- update complete")
+if __name__ == '__main__':
```
YannickJadoul, line 43: One or two new lines beofore?

cibuildwheel/resources/install_certifi.py
```diff
@@ -0,0 +1,44 @@
...
+if sys.version_info[0] == 2:
+    FileNotFoundError = OSError
+
+def main():
+    openssl_dir, openssl_cafile = os.path.split(
+        ssl.get_default_verify_paths().openssl_cafile)
+    print(" -- pip install --upgrade certifi")
+    subprocess.check_call([sys.executable,
+        "-E", "-s", "-m", "pip", "install", "--upgrade", "certifi"])
+    import certifi
```
YannickJadoul, line 27: A few empty lines in this body of this function, to split off different logical parts, perhaps, if possible. For example here, to have a break between installing `certifi` and using it?

test/test_1_ssl.py
```diff
@@ -13,11 +13,10 @@
         else:
             from urllib.request import urlopen
 
-        if sys.version_info[0:2] == (3, 3):
-            data = urlopen("https://www.nist.gov")
-        else:
-            context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
-            data = urlopen("https://www.nist.gov", context=context)
+        context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
```
YannickJadoul, line 16: Why was this file renamed?

test/test_1_ssl.py
```diff
@@ -13,11 +13,10 @@
...
+        data = urlopen("https://www.nist.gov", context=context)
+        data = urlopen("https://raw.githubusercontent.com/joerick/cibuildwheel/master/CI.md", context=context)
```
YannickJadoul, line 18: Do we need to assert something for the received data? E.g. just that it's not empty, or so?

cibuildwheel/macos.py
```diff
@@ -72,6 +72,9 @@ def install_cpython(version: str, url: str) -> Path:
 
     # if this version of python isn't installed, get it from python.org and install
     python_package_identifier = f'org.python.Python.PythonFramework-{version}'
+    python_executable = 'python3' if version[0] == '3' else 'python'
+    installation_bin_path = Path(f'/Library/Frameworks/Python.framework/Versions/{version}/bin')
```
YannickJadoul, line 76: Minor nitpicking, but why are these moved? I think the reason they were only set later, is because they are only installed in that block. So logically, I think I'd prefer keeping it as it was, and setting them only after the package is installed (or checked that it is installed)

YannickJadoul (commented):
I haven't really followed the whole discussion or problem, so these are mostly minor style-related comments.

Apart from those, it seems like @djhoese, @Czaki, and @mayeut together figured out what went wrong and how to best fix this, so not a lot of comments there.

mayeut (approved):

test/test_1_ssl.py
Czaki: To move on begin, to get error information faster (on `master` branch `test_ssl` is last executed)

cibuildwheel/macos.py
Czaki: because I use this values in line 89

cibuildwheel/resources/install_certifi.py
Czaki: I copy file content, I forget to reformat.

cibuildwheel/resources/install_certifi.py
Czaki: Done

test/test_1_ssl.py
Czaki: maybe we should, but then we should have some file with some fixed content.

Czaki:
Before merge I have one question. 
If we should test multiple SSL version `ssl.PROTOCOL_*` or only this two tests which we already have.

mayeut:
The test with `ssl.PROTOCOL_TLSv1_2` was only there to check support was indeed there for this protocol a couple years back. I don't think we should have any specific tests other than that until web servers start to heavily shift to newer protocols and disable TLS v1.2 (that was the case for PyPI where the minimum supported protocol shifted to TLS v1.2, hence this test).

test/test_1_ssl.py
YannickJadoul: OK, right! I see it's reverted, now :-)

cibuildwheel/macos.py
YannickJadoul: Oh, I didn't see the indentation in the diff, but it's still inside the if-statement. Pity, but OK.

YannickJadoul (approved):
Thanks, @Czaki! Looks good to me, now!

test/test_1_ssl.py
YannickJadoul: Not per se. You could just test if something non-empty was downloaded? Or take away the `data =`, maybe? It just feels weird, all these useless assignments.

mayeut:
Thanks @Czaki !

Czaki:
@joerick 1.6.3?

PR #448, opened 2020-10-09
Add str representation to DependencyConstraints

Czaki:
I found that current preamble contains:

```
  dependency_constraints: <cibuildwheel.util.DependencyConstraints object at 0x10d4ca640>
```

So I add `__str__` method to DependencyConstraints class.


joerick:
Thanks @Czaki , I've been meaning to do this for a while!

PR #449, opened 2020-10-10
Update openssl patch from 1.0.2u to 1.1.1h on macOS

mayeut:
- Update openssl patch from 1.0.2u to 1.1.1h on macOS
- Remove python 3.5 workaround in install_certifi.py (now included in patch)

This will probably be the last update now that python 3.5 has reached End-Of-Life.


joerick:
thanks, @mayeut !

PR #454, opened 2020-10-29
Update repos using cibuildwheel

nsoranzo:
I can add https://github.com/bxlab/bx-python if that's of interest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,6 @@ Here are some repos that use cibuildwheel.
 - [AutoPy](https://github.com/autopilot-rs/autopy)
 - [apriltags2-ethz](https://github.com/safijari/apriltags2_ethz)
 - [TgCrypto](https://github.com/pyrogram/tgcrypto)
-- [Twisted](https://github.com/twisted/twisted)
 - [gmic-py](https://github.com/dtschump/gmic-py)
 - [creme](https://github.com/creme-ml/creme)
 - [PyAV](https://github.com/PyAV-Org/PyAV)
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Working examples
 
 Here are some repos that use cibuildwheel.
 
-- [Matplotlib](https://github.com/matplotlib/matplotlib)
+- [Matplotlib](https://github.com/matplotlib/matplotlib) - The venerable Matplotlib, a Python library with C++ portions, built for Linux, Mac, and Windows on Github Actions.
 - [pyinstrument_cext](https://github.com/joerick/pyinstrument_cext) - A simple C extension, without external dependencies, built on Travis CI/Appveyor.
 - [websockets](https://github.com/aaugustin/websockets)
 - [Parselmouth](https://github.com/YannickJadoul/Parselmouth)
@@ -137,7 +137,7 @@ Here are some repos that use cibuildwheel.
 - [AutoPy](https://github.com/autopilot-rs/autopy)
 - [apriltags2-ethz](https://github.com/safijari/apriltags2_ethz)
 - [TgCrypto](https://github.com/pyrogram/tgcrypto)
-- [twisted-iocpsupport](https://github.com/twisted/twisted-iocpsupport)
+- [twisted-iocpsupport](https://github.com/twisted/twisted-iocpsupport) - A submodule of Twisted that hooks into native C APIs using Cython. Built on Github CI for Windows.
 - [gmic-py](https://github.com/dtschump/gmic-py)
 - [creme](https://github.com/creme-ml/creme)
 - [PyAV](https://github.com/PyAV-Org/PyAV)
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Ow :-( Just curious: why? :-)

Czaki:
It looks like they extracted c-code to separate package. (current setup.py does not contain Extension call)

henryiii:
They do use it, they just moved the backend C++ code. https://github.com/twisted/twisted-iocpsupport/blob/default/.github/workflows/github-deploy.yml 

We should probably add at least matplotlib as well. ;)

YannickJadoul:
Ah, right! :-) Thanks, @nsoranzo!

README.md
```diff
@@ -127,6 +127,7 @@ Working examples
 
 Here are some repos that use cibuildwheel.
 
+- [Matplotlib](https://github.com/matplotlib/matplotlib)
```
joerick, line 130:
```suggestion
- [Matplotlib](https://github.com/matplotlib/matplotlib) - The venerable Matplotlib, a Python library with C++ portions, built for Linux, Mac, and Windows on Github Actions.
```

joerick (commented):
Thanks for updating this @nsoranzo ! I'm trying to add a little context to these repos, so I've taken the opportunity to add little descriptions for the ones you've added here. Could you incorporate them if they sound right to you?

README.md
```diff
@@ -127,6 +127,7 @@ Working examples
 
 Here are some repos that use cibuildwheel.
 
+- [Matplotlib](https://github.com/matplotlib/matplotlib) - The venerable Matplotlib, a Python library with C++ portions, built for Linux, Mac, and Windows on Github Actions.
 - [pyinstrument_cext](https://github.com/joerick/pyinstrument_cext) - A simple C extension, without external dependencies, built on Travis CI/Appveyor.
 - [websockets](https://github.com/aaugustin/websockets)
 - [Parselmouth](https://github.com/YannickJadoul/Parselmouth)
```
YannickJadoul, line 133:
```suggestion
- [Parselmouth](https://github.com/YannickJadoul/Parselmouth) - A pybind11-powered Python interface to the pre-existing Praat software, built with C++17 on GitHub Actions and sharing a build of the Python-independent core between different Python wheels in the same `cibuildwheel` run.
```

YannickJadoul (commented):
While we're at it, we could still add something for Parselmouth as well, if you want. Then I don't need to make a PR for a single line?

@joerick, is this more or less what you're looking for? It feels slightly too elaborate, to me. Core things that I think might be interesting for others are "pybind11", "C++17", and the fact that the thing I'm wrapping is only built once as a static library, and then shared/linked to all individual wheel builds (since most of the code doesn't import `Python.h`).

... but as you can see, I don't really manage to succinctly summarize that, so maybe it's too much to mention.

README.md
joerick: Sure! Maybe also worth getting Cmake in there? :)

Here's my attempt-

> A Python interface to the Praat library, using pybind11, C++17 and Cmake, with the core Praat library shared between each wheel. All platforms built on Github Actions.

YannickJadoul: OK, that's already closer, yes! Thanks! :-)

Let's just make sure it's clear I don't ship a shared library?

> A Python interface to the Praat library, using pybind11, C++17 and CMake, with the core Praat static library built only once and shared between wheels; all platforms built on Github Actions.

joerick: sounds good

YannickJadoul: Thanks! Let me push that and merge @nsoranzo's PR, then :-)

YannickJadoul: @joerick, it still doesn't fit onto one line, when redering the markdown file:

https://github.com/joerick/cibuildwheel/blob/7f6812b13216dfd37912431036a0bd7e11af17f0/README.md

If you want to shrink the description go ahead; otherwise, I'm fine with this being merged, as well. I'll leave the decision up to you!

henryiii:
I've been intending to suggest boost-histogram too.  All cibuildwheel wheels supported, pure setuptools, using setuptools-helpers and multithreaded building, PEP 517, c++14 with modified manylinux1 and normal manylinux 2010.

YannickJadoul:
@henryiii, bring it on, while we're at it! A full PR feels like such a waste, somehow?

README.md
joerick: Looks fine to me!

YannickJadoul:
Rebased again onto `master`; tests should pass now.

joerick:
May as well get things moving. Feel free to make another PR for your boost-histogram, @henryiii

henryiii:
I'm going to make a push very soon (by early December, but maybe at least one today) to move the reset of the Scikit-HEP repos using my old azure-wheel-helpers over to cibuildwheel. I just found that iMinuit has already moved :tada:, so that leaves pyjet (which is currently broken, so I might move it to cibuildwheel instead of patching the helpers), Awkward-1, and one other package, I believe. Then I'll likely make a PR with all the ones I know of.

Issue #349, opened 2020-05-24
Folding output (at least in GHA/travis)


henryiii:
Feature request:

The current output shows up as one long section, but recently GitHub Actions have started showing up with folding support, which is really quite nice. (See https://github.com/peaceiris/actions-gh-pages/blob/master/CHANGELOG.md) I'm sure I first saw it in the docs, where they list both the echo method and a javascript method for custom Actions, but can't find it now; here's a page that comes up from a google search: https://github.community/t5/GitHub-Actions/Has-github-action-somthing-like-travis-fold/td-p/37715. Travis also has a fold feature, and other systems might have them as well.

I would like to see the logs folded per wheel, maybe wheel and test separated. Just thinking about a possible design idea: there could be a set of variables, such as FOLD_START and FOLD_END, that would be printed (with templating, so at the very least the name of the current arch being built is available. Something like this would then be the setup for GitHub Actions:

```bash
CIBW_FOLD_START: "::group::{action} {wheel}"
CIBW_FOLD_END: "::groupend::"
```

Then the logs would look like:

```
> Setup all
> Building cp37-cp37m-macosx_10_9_x84_64
> Testing cp37-cp37m-macosx_10_9_x84_64
```

Maybe the defaults for GitHub Actions could be applied if it detected GHA (don't know how much customization you want per service), or this should just be something you set in examples.

What do you think? I can probably help with the implementation if you think it is a good idea.

PS: This may be called subfold, since each action folds.

joerick:
Love this idea! I've been thinking for a while that it would be lovely to clean up our logs and folding might be just be the perfect way to do it.

My thinking on logs (before considering folding) was something like this: 

![Artboard](https://user-images.githubusercontent.com/1244307/82752720-f5678680-9db7-11ea-9ab6-0710ca225ee4.png)

But the obvious criticism of the above is 'what if I want to see the output of my before_build...', so maybe we could put that output into folding groups instead of hiding entirely.

In terms of how we'd get to the above, I can see something like the above be quite tricky to pull off with our current Docker build in bash. So something that fancy might have to wait until we have more flow control in the host Python (`docker exec` commands into a running container).

So maybe an interim solution would be good, which just starts a fold group at the start of a build - we already have this level of flow control in Python today. That would look something like:

![Artboard Copy 2](https://user-images.githubusercontent.com/1244307/82753048-8e979c80-9dba-11ea-8711-d492dd99e923.png)

[...]

PR #458, opened 2020-11-02
Improve CI logs using fold groups

joerick:
Hello all. Lately I've been working on #349, an improvement to how cibuildwheel writes its logs. It uses fold groups, which are supported on Github Actions, Travis and Azure. But more generally, it's a tidy up and reconsideration about what users want to see.

image

So far, I've only implemented the changes on the Linux build, but I'd love to get some feedback so far before replicating that across the platforms. And testing with the other CIs. [Here is an example build](https://github.com/joerick/cibuildwheel/runs/1339394918?check_suite_focus=true). Mostly I'm interested in how easy it is to browse and get to relevant information, so I also created an [erroring build](https://github.com/joerick/cibuildwheel/runs/1339394935?check_suite_focus=true) to see how that looks. Let me know what you think! 

cc @henryiii was interested in this i think!


cibuildwheel/linux.py
```diff
@@ -111,24 +112,37 @@ def build(options: BuildOptions) -> None:
...
         platform_configs = [c for c in python_configurations if c.identifier.startswith(implementation) and c.identifier.endswith(platform_tag)]
         if not platform_configs:
             continue
 
         try:
+            log.step(f'Starting Docker image {docker_image}...')
             with DockerContainer(docker_image, simulate_32_bit=platform_tag.endswith('i686'), cwd=container_project_path) as docker:
+                log.step_end()
+
+                log.step('Copying project into Docker...')
```
mayeut, line 127: It would be great to also be able to use a context manager no ? It clearly shows the intent and prevents forgetting to call `log.step_end()` at the end of that intent (even though I'm seeing now that steps are closed automatically when a new one starts)
e.g.

```
with log.step('Copying project into Docker...'):
    docker.copy_into(Path.cwd(), container_project_path)
```
even more for longer process:
```
with log.step('Running before_all...'):
    env = docker.get_environment()
    env['PATH'] = f'/opt/python/cp38-cp38:{env["PATH"]}'
    env = options.environment.as_dictionary(env, executor=docker.environment_executor)

    before_all_prepared = prepare_command(options.before_all, project=container_project_path, package=container_package_dir)
    docker.call(['sh', '-c', before_all_prepared], env=env)
```

mayeut:
Thanks @joerick , that seems really nice & useful (checking everything we think cibuild should do is now just a glance away).

I wonder if you could make the failing test to actually fail (check wether it's already expanded, showing said error, maybe [using](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) `fail-fast: false` and/or `continue-on-error` ).
Also, just a test with pytest would be real nice (also one passing, one failing).

cibuildwheel/logger.py
```diff
@@ -0,0 +1,192 @@
...
+            c = self.colors
+            print(f'{c.bright_red}Error{c.end} {error}')
+
+    def _start_fold_group(self, name: str):
+        self._end_fold_group()
+        self.active_fold_group_name = name
+        fold_start_pattern = FOLD_PATTERNS.get(self.fold_mode, DEFAULT_FOLD_PATTERN)[0]
+
+        print(fold_start_pattern.format(name=self.active_fold_group_name))
+        print()
```
YannickJadoul, line 111: Just curious: why this extra empty line in the middle after the start of a group (and not after the end)?

cibuildwheel/logger.py
```diff
@@ -0,0 +1,192 @@
...
+        self.step_start_time = time.time()
+        self._start_fold_group(step_description)
+
+    def step_end(self, success=True):
+        if self.step_start_time is not None:
+            self._end_fold_group()
+            c = self.colors
+            duration = time.time() - self.step_start_time
+            if success:
+                print(f'{c.green}✓ {c.end}{duration:.2f}s'.rjust(78))
```
YannickJadoul, line 89: Open question: do we always want to print times (ánd not fold those?)

YannickJadoul (commented):
Amazing! When can I start using it? :-)

I guess we still need the same for `macos.py` and `windows.py`, once this is reviewed?

cibuildwheel/linux.py
joerick: I originally built the whole thing using context managers in fact, you can see that here:
https://github.com/joerick/cibuildwheel/blob/946306b21dee8411cc27807a9f47a7e91ecf1b5d/cibuildwheel/linux.py

joerick: [whoops, hit submit somehow!]

I concluded that adding an extra two levels of indentation to an already very indented function was too much, and yes, I couldn't wrap the Docker start, so split it into a `build_one` function, at which point it felt like the tail wagging the dog - the logging functionality is mandating the program structure? So I decided to switch to the 'imperative' style as seen in https://github.com/joerick/cibuildwheel/commit/38e7b5a91ded109ae44947b754cd47d28e3a26d5 .

This seemed a little more lightweight.

But I also don't really like those `step_end()` method calls. One option (I think @mayeut suggests) is that I return a context manager from step(), whose `__exit__` calls step_end, but it's usage is optional. I might give that a go to see how it feels.

cibuildwheel/linux.py
```diff
@@ -230,14 +252,24 @@ def build(options: BuildOptions) -> None:
...
-            print(f'Command {error.cmd} failed with code {error.returncode}. {error.stdout}')
+            log.error(f'Command {error.cmd} failed with code {error.returncode}. {error.stdout}')
             troubleshoot(options.package_dir, error)
             exit(1)
 
 
+def build_one(options: BuildOptions, config: PythonConfiguration, docker: DockerContainer,
+              container_project_path: PurePath, container_package_dir: PurePath,
+              container_output_dir: PurePath, log: Logger):
+    pass
```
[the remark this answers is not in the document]
joerick: whoops!

cibuildwheel/linux.py
joerick: > Why do we need the succeeding step_end, though, if it's automatically closed anyway?

Also a good point, I could remove that one.

.github/workflows/sample_build.yml
```diff
@@ -0,0 +1,41 @@
+name: Sample build
+
+on: [push, pull_request]
```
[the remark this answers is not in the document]
joerick: It's quite nice to be able to keep an eye on the output, after all, it's part of the user interface of cibuildwheel, and I don't actually see it all that often! Also it's pretty quick to run and I don't think we're short of concurrency on GHA linux containers. So I think I'd like to keep it around. But maybe once I get the other hosting providers working, I'll remove them and just rely on GHA for the sample build though.

(another option could be to run the sample build on the same job as the unit tests, before they start, because most of the time is spent downloading images/pythons, so it wouldn't increase the test runtime much. )

cibuildwheel/logger.py
```diff
@@ -0,0 +1,192 @@
...
+
+DEFAULT_FOLD_PATTERN = ('{name}', '')
+FOLD_PATTERNS = {
+    'azure': ('##[group]{name}', '##[endgroup]'),
+    'travis': ('travis_fold:start:{name}', 'travis_fold:end:{name}'),
+    'github': ('::group::{name}', '::endgroup::{name}'),
+}
+
+PLATFORM_IDENTIFIER_DESCIPTIONS = {
+    'manylinux_x86_64': 'Manylinux x86_64',
```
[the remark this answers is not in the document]
joerick: Yeh, you're right

cibuildwheel/logger.py
```diff
@@ -0,0 +1,192 @@
...
+
+    bg_grey = '\033[48;5;235m'
+
+    bold = '\033[1m'
+    faint = '\033[2m'
+
+    end = '\033[0m'
+
+    class Disabled:
+        def __getattr__(self, attr: str):
```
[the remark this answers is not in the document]
joerick: > and if it's actually necessary

I'm using it so I can do `f'{c.red}some text{c.end}'` and have those colors fall away when colored output is disabled. Admittedly it looks pretty weird though - suggestions welcome. Curious what you're thinking when you say 'not necessary'?

cibuildwheel/logger.py
joerick: I guess I thought it looks nicer :) the newline appears inside the group, spaces it down a bit

e.g.
this:
image

versus this:
image

dont feel strongly about it though!

cibuildwheel/logger.py
joerick: Hmm. Well I liked having some confirmation that the step completed (the green tick) - it's reassuring. The time was kinda added opportunistically but could be very helpful for debugging slow builds (or preventing people from trying to optimise the wrong bit!). 

But I'll try rolling all that inside the fold group to see how it looks.

.github/workflows/sample_build.yml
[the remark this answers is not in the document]
YannickJadoul: OK, yes, that makes sense. So it's not really meant as a true test, but this could be useful :-)

cibuildwheel/logger.py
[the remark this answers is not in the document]
YannickJadoul: Looking at this again, my suggestion doesn't really feel cleaner in any way, though :-/ So yeah, no worries.

cibuildwheel/linux.py
mayeut: > I concluded that adding an extra two levels of indentation to an already very indented function was too much

Yeah, now that you're saying this, I can see that you're right.

cibuildwheel/logger.py
joerick: > Is the indentation automatically done by GHA? It looks kind of funky in combination with output from those commands:

No, the indentation is from our code. that's why the auditwheel stuff is not indented. Maybe there's a way to indent the subcommand output, but probably not worth the extra complexity.

cibuildwheel/util.py
```diff
@@ -113,8 +116,8 @@ def get_for_python_version(self, version: str) -> Path:
         else:
             return self.base_file_path
 
-    def __str__(self):
-        return f"File '{self.base_file_path}'"
+    def __repr__(self):
+        return f'DependencyConstraints({self.base_file_path!r})'
```
henryiii, line 120:
```suggestion
        return f'{self.__class__.__name__}{self.base_file_path!r})'
```

cibuildwheel/linux.py
henryiii: The nice thing about a context manager is that it always closes, even if an error is thrown. It seems to me there's at least one path with where the close may not happen - I assume it's not critical, but maybe you could have the first step be a context manager, and remove the final end. Then there would just be a `.step` which would close and start a new step. You can use `contextlib.ExitStack` to keep from indenting too much. 

I'm a huge fan of context managers; I added finally context managers to my LCD code in my vent monitor project and suddenly a long standing bug went away - there was a way the Rπ's resources were not being cleaned up, and now they were.

This looks good as is, though, just a general suggestion.

henryiii:
Sorry for being so slow to review, it's looking great, though!

joerick:
Weird. It seems Python won't let me output unicode characters using `print()` on Windows 🤔  Here are some debug output (this is on GHA, CPython 3.7):

```
sys.stdout.encoding: cp1252
sys.stdout.isatty(): False
locale.getpreferredencoding(): cp1252
sys.getfilesystemencoding(): utf-8
os.environ["PYTHONIOENCODING"]: None
```

The error I get is:

    UnicodeEncodeError: 'charmap' codec can't encode character '\u263a' in position 0: character maps to <undefined>

I think it's baulking on the unicode tick/cross symbols. I can render the symbols in ascii instead, but I thought we were past this sort of nonsense... especially after [PEP 528](https://www.python.org/dev/peps/pep-0528/)... anyone have ideas on workarounds?

joerick:
Found I can just set the stdout encoding in Python 3.7+ :) . Python 3.6 will have ascii fallbacks

joerick:
Alrighty! This is ready for another review. It works on Mac and Windows now, as you can see in the Github Action runs. Rather than a separate workflow for sample builds, I've put it as a step in the existing run, but it'll only run when a certain label `CI: Sample build` is added to the PR. So when we're interested in the output we can enable it, but normally it won't be there cluttering up the output.

bin/sample_build.py
```diff
@@ -0,0 +1,23 @@
...
+import argparse
+import tempfile
+from pathlib import Path
+
+if __name__ == '__main__':
+    # move cwd to the project root
+    os.chdir(Path(__file__).resolve().parents[1])
+
+    parser = argparse.ArgumentParser(description='Runs a sample build')
+    parser.add_argument('PROJECT_PYTHON_PATH', nargs='?', default='test.test_0_basic.basic_project')
```
YannickJadoul, line 15: All-caps is not too common for command line arguments?

cibuildwheel/docker_container.py
```diff
@@ -57,6 +57,9 @@ def __enter__(self) -> 'DockerContainer':
             ],
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
+            # output the subcommand stderr to our stdout. Avoids ordering
+            # problems between log messages on stderr and stdout
+            stderr=sys.stdout,
```
YannickJadoul, line 62: What about users piping stdout/stderr separately, and expecting stderr to contain text if things go wrong? Does that happen? Is that something to worry about?

cibuildwheel/logger.py
```diff
@@ -0,0 +1,259 @@
...
+
+        elif 'TRAVIS' in os.environ:
+            self.fold_mode = 'travis'
+            self.colors_enabled = True
+
+        elif 'APPVEYOR' in os.environ:
+            self.fold_mode = 'disabled'
+            self.colors_enabled = True
+
+        else:
```
YannickJadoul, line 60: Don't we already detect the CI systems somewhere else in `cibuildwheel`? Can we share this detection?

cibuildwheel/logger.py
```diff
@@ -0,0 +1,259 @@
...
+            if success:
+                print(f'{c.green}{s.done} {c.end}{duration:.2f}s'.rjust(78))
+            else:
+                print(f'{c.red}{s.error} {c.end}{duration:.2f}s'.rjust(78))
+
+            self.step_start_time = None
+
+    def error(self, error: Union[Exception, str]):
+        self.step_end(success=False)
+        print()
```
YannickJadoul, line 110: Same here, about the stdout vs. stderr, and redirection.

YannickJadoul (approved):
Looks pretty good to me! :-) The few remarks about stdout/stderr are mostly to check if these have been considered.

I think this is good to merge soonish. It seems pretty solid, and even if there are still minor things, it might be easier to create fixes once it get released and we get feedback, rather than now crafting everything to perfection?

bin/sample_build.py
joerick: oh, yeah, good catch. not sure what I was thinking of!

joerick:
thanks for the review @YannickJadoul !

cibuildwheel/docker_container.py
joerick: Yeah, so this change was motivated by Github Actions, where some output on stderr would end up on the wrong side of the fold group and break the formatting. 

> What about users piping stdout/stderr separately, and expecting stderr to contain text if things go wrong? Does that happen? Is that something to worry about?

The output of cibuildwheel isn't intended to be machine readable at all, so I don't expect users to pipe the output of cibuildwheel to anywhere, really. Maybe to their own log file? either way, I see keeping output interleaved as more important than the separate streams. (Also, in tests, the way the pytest splits them up makes it harder to follow execution when something breaks)

YannickJadoul: Fair enough, yes! One thing I just thought of: we do have `sys.stdout = Unbuffered(sys.stdout)  # type: ignore`. Would it be worth trying to add the same for `sys.stderr`?

Apart from that, yes, I think you're right. That's more important indeed.

cibuildwheel/logger.py
joerick: good point, I'll look at pulling that into a util

cibuildwheel/logger.py
joerick: Are you suggesting stderr might be more appropriate for an error message? I can see the argument, but pragmatically speaking I don't think users will see any difference, and because of the potential ordering issues I'd rather just stick to stdout since it's all going to a CI log anyway. Does that sound okay to you?

YannickJadoul: Yes, but yes :-)

cibuildwheel/docker_container.py
joerick: I think stderr is unbuffered by default, but it's a good idea, I'll give it a shot. Weirdly it only seems to be a problem on Github Actions.

joerick: Hm! So I tried this, and removing the stdout-stderr redirect, and saw no problem! Then I tried it without any workarounds, and it still worked fine. So I guess I'll leave it without any workarounds for now, if we see output bleeding out of the fold groups we can add it back. :)

YannickJadoul: Huh, OK, then. I'll keep this in the back of mind as well! If I (we) catch things getting mixed up gaain, we can get back to it!
Thanks for trying! :-)

joerick:
Thanks for the reviews all!

joerick:
Time for a release I think...

YannickJadoul:
@joerick, could it be that you made a GitHub release, but did not upload to PyPI yet?

joerick:
arg, that serves me right for doing a release late at night. Thank you

joerick:
okay, submitted :)

YannickJadoul:
No worries. But thanks, I was waiting for it; eager to try it out :-)

PR #462, opened 2020-11-13
Improve Github triggers to prevent multiple runs on origin branches

joerick:
Minor thing, but this change prevents the multiple pre-commit CI runs for origin branches and prevents the 'skipped' jobs from appearing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,14 +1,14 @@
 name: Test
 
 on:
-  push: {}
+  push:
+    branches:
+      - master
   pull_request:
     paths-ignore:
       - 'docs/**'
-    branches:
-      # PRs from branches on the origin repo get a build from the 'push'
-      # trigger. So the pull_request trigger is limited to fork branches
-      - '**:**'
+  workflow_dispatch:
+    # allow manual runs on branches without a PR
 
 jobs:
   pre-commit:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
     paths-ignore:
       - 'docs/**'
+    branches:
+      # PRs from branches on the origin repo get a build from the 'push'
+      # trigger. So the pull_request trigger is limited to fork branches
+      - '**:**'
 
 jobs:
   pre-commit:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -5,6 +5,10 @@ on:
   pull_request:
     paths-ignore:
       - 'docs/**'
+    branches:
+      # PRs from branches on the origin repo get a build from the 'push'
+      # trigger. So the pull_request trigger is limited to fork branches
+      - '**:**'
```
henryiii, line 11: I usually put master here - does `**:**` match master?

henryiii: Oh, I see, you are not building the PR trigger but only the push one, instead of the other way around.

joerick: Yeah. It can be quite handy for me to push a branch and see the results of CI without making a PR.

henryiii: But then if you fork, GHA builds on your fork (because there it is a "normal" branch) and on the PR when you make a PR? Not that it shows up in the PR, but it doubles the building needed.

henryiii: Contrasting with my current system in most projects, which is to only build on the main (master, develop) branches and then to use workflow_dispatch to manually trigger builds when testing without a PR.

joerick: I guess that's true, yeah. Not sure this change affects that though?

joerick: Ooh what is workflow_dispatch?

henryiii: No, I'm just contrasting against having "master" listed in the push:. 

And: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

henryiii: I also use it empty (no release) to trigger wheels whenever I think they may be broken. It's a little slower for development, though, since you have to go to the UI to trigger a build.

joerick: ah okay, that's nice. Maybe I'll give the `branches: [master]` version a shot then. Seems like that's more what the designers had in mind when designing these triggers anyway.

PR #464, opened 2020-11-22
Ensure modern version of Python in CIBW_BEFORE_ALL on manylinux

YannickJadoul:
I think this line is missing `bin/`: https://github.com/joerick/cibuildwheel/blob/2f4ed1fafc6661b1bcfef10c28b47d593e53dfa5/cibuildwheel/linux.py#L125

First a commit to make sure I'm correct, though.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -122,7 +122,7 @@ def build(options: BuildOptions) -> None:
 
                 if options.before_all:
                     env = docker.get_environment()
-                    env['PATH'] = f'/opt/python/cp38-cp38:{env["PATH"]}'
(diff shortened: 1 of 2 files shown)


YannickJadoul:
> First a commit to make sure I'm correct, though.

Took a while, but seems I am :-)

This can be merged, I believe, unless there's a better way to test this?

henryiii (approved):

joerick:
Good catch @YannickJadoul !

YannickJadoul:
> Good catch @YannickJadoul !

Not so much a catch, as well as being bitten by it ;-) But luckily I noticed reasonably quickly why.

PR #471, opened 2020-11-29
Add coverage.py to the repos list

nedbat:
I wish I could say I was doing something interesting with cibuildwheel, but the thing I love about it is that it just worked right out of the box! Thanks so much for building it!!


joerick:
That's great, thanks @nedbat! Pleased it can be helpful 🙌

PR #478, opened 2020-12-08
docs: minor wording change for multibuild

henryiii:
@joerick I noticed the working here was a little out of date, "It can do a lot more than this project" is not very accurate anymore. I touched it up a bit, what do you think?


joerick:
Nice one, thanks @henryiii !

joerick:
Nice one, thanks @henryiii !

Issue #468, opened 2020-11-27
Ideas for improving the first-run experience


robbmcleod:
Using the minimal example (https://github.com/joerick/cibuildwheel/blob/master/examples/github-minimal.yml) for GitHub Actions the Ubuntu build fails as it appears to be grabbing the system Python 2.7 instead of the specified Python 3.7:

https://github.com/robbmcleod/cpufeature/runs/1461364117

I tried explicitly using `python3` instead but that fails on Windows. An I assume `pip` is still `pip2` rather than `pip3`. 

I'm trying to learn how to use this tool so I can build the wheels for `numexpr` (which I maintain), as that's going to hold up the  release of `pandas` among other things. So any guidance you can provide here would be greatly appreciated.

Czaki:
You should use [`CIBW_BUILD`](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip) to limit python versions for which wheels will be built.

From your log `  build_selector: BuildSelector('*' - '')` means tat cibuildwheel will try build wheel for python 2.7, 3.5, 3.6, 3.7, 3.8 and 3.9

henryiii:
It requires a recent Python to _run_ - it builds all Python versions (2.7, 2.5-3.9). You need to set `CIBW_BUILD` or `CIBW_SKIP` to describe what your project supports.

henryiii:
Any remaining questions? Can we close?

robbmcleod:
It's up to you, regarding closing or not. Personally I would have found it helpful if there was a minimal explanation on what `cibuildwheel` does. The documentation goes straight from minimalist examples to an explanation of the build options. When I read it, I didn't have the order of operations in my head. So some flowchart would be helpful.

The other problem I ran into was the directories defined by `{project}` and `{package}` are not the install directory. Instead the package is installed somewhere in `/tmp`, which makes running pytest potentially an irritation when the package has a c-extension. Setting `CIBW_TEST_COMMAND: pytest {package}` as suggested by the documentation definitely did not work for me as there was name collision and the c-extensions were not found. I'm unclear on what the practical difference between `{project}` and `{package}` is, since they both seem to be source dists.

Czaki:
Linux wheels are built inside docker images. So it cannot use the same path as in GitHub worker. 

[...]

PR #479, opened 2020-12-09
docs: use GitHub example in readme

henryiii:
Swapping the readme example with GHA. Should we do deploy (properly) in the readme? I would rather think not, since it will make it a bit longer if we do it properly (cleaner and more modular). I've also approximately changed the Travis CI example to the correct deployment scheme. I didn't add mention of the trick I have to use to stop it from adding the SDist (important if you combine GHA + Travis for ARM, for example). Hopefully this is at least a starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,16 +83,16 @@ jobs:
         run: python -m pip install cibuildwheel==1.7.1
 
       - name: Build wheels
-        env:
-          CIBW_SKIP: "?p25-*" # Skip building Python 2.7 wheels on all platforms
         run: python -m cibuildwheel --output-dir wheelhouse
+        env:
+          CIBW_SKIP: "cp27-* pp27-*"  # skip Python 2.7 wheels
 
       - uses: actions/upload-artifact@v2
         with:
           path: ./wheelhouse/*.whl
 ```
 
-You can add a deploy job using [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish). For more information, including a full example of building on GitHub Actions with PyPI deploy, Travis CI, Appveyor, Azure Pipelines, or CircleCI, check out the [documentation](https://cibuildwheel.readthedocs.org) and the [examples](https://github.com/joerick/cibuildwheel/tree/master/examples).
+For more information, including building on Python 2, PyPI deployment, and the use of other CI services like Travis CI, Appveyor, Azure Pipelines, or CircleCI, check out the [documentation](https://cibuildwheel.readthedocs.org) and the [examples](https://github.com/joerick/cibuildwheel/tree/master/examples).
 
 Options
 -------
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Usage
 Example setup
 -------------
 
-To build manylinux, macOS, and Windows wheels on Github and make them availble for download after the job runs, you could use this `.github/workflows/wheels.yml`:
+To build manylinux, macOS, and Windows wheels on Github Actions, you could use this `.github/workflows/wheels.yml`:
 
 ```yaml
 name: Build
@@ -71,7 +71,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-20.04, windows-2019, macOS-10.15]
 
     steps:
       - uses: actions/checkout@v2
@@ -82,11 +82,9 @@ jobs:
       - name: Install cibuildwheel
         run: python -m pip install cibuildwheel==1.7.1
 
-      - name: Install Visual C++ for Python 2.7
-        if: runner.os == 'Windows'
-        run: choco install vcpython27 -f -y
-
       - name: Build wheels
+        env:
+          CIBW_SKIP: "?p25-*" # Skip building Python 2.7 wheels on all platforms
         run: python -m cibuildwheel --output-dir wheelhouse
 
       - uses: actions/upload-artifact@v2
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ cibuildwheel
 
 Python wheels are great. Building them across **Mac, Linux, Windows**, on **multiple versions of Python**, is not.
 
-`cibuildwheel` is here to help. `cibuildwheel` runs on your CI server - currently it supports Azure Pipelines, Travis CI, AppVeyor, GitHub Actions and CircleCI - and it builds and tests your wheels across all of your platforms.
+`cibuildwheel` is here to help. `cibuildwheel` runs on your CI server - currently it supports GitHub Actions, Azure Pipelines, Travis CI, AppVeyor, and CircleCI - and it builds and tests your wheels across all of your platforms.
 
 
 What does it do?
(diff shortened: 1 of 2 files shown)


YannickJadoul (approved):
Last call for @joerick, ofc, since he originally added the deployment version here, but I'm happy just referring to it and keeping it short :-)

examples/travis-ci-deploy.yml
```diff
@@ -19,22 +19,18 @@ jobs:
...
-      python3 -m twine upload wheelhouse/*.whl
-    fi
+deploy:
+  provider: pypi
+  username: "__token__"
+  skip_cleanup: true
+  on:
+    tags: true
+  password:
+    secure: "" # Set here
```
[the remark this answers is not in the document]
henryiii: Will do that, just waiting to see if I get any further feedback from @joerick or @mayeut. Or @Czaki.

README.md
```diff
@@ -58,49 +58,43 @@ Usage
...
-    if [[ $TRAVIS_TAG ]]; then
-      python3 -m pip install twine
-      python3 -m twine upload wheelhouse/*.whl
-    fi
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
```
Czaki, line 74: why `ubuntu-18.04` instead of `ubuntu-latest`?

Czaki (changes requested):
I think that it should be nice to shou usage of environment variables in examples. Maybe something with `CIBW_BUILD` or `CIBW_SKIP`?

README.md
henryiii: I have no idea, that's what's in the example. Mayne because `ubuntu-latest` generates a warning that it will change to `ubuntu-20.04` soon, but that example predated that warning.

I think for a 2020 example, skipping Python 2 might be useful in the example? `CIBW_SKIP: "cp27-*" # Skip building Python 2.7 on all platforms` for example?

Czaki: > I think for a 2020 example, skipping Python 2 might be useful in the example? `CIBW_SKIP: "cp27-*" # Skip building Python 2.7 on all platforms` for example?

This does not skip pypy27. But I agree that short example may skip python 2.7 Also this allow to remove "Install Visual C++ for Python 2.7" step.

joerick: In general, I prefer to lock to specific versions when writing CI configs, so in that case, it should be`os: [ubuntu-20.04, windows-2019, macos-10.15]`. I think previously it was only possible to lock the ubuntu image, but now you can lock windows and macos too.

I also like the idea of adding a single CIBW setting.

README.md
```diff
@@ -58,49 +58,43 @@ Usage
...
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==1.7.1
+
+      - name: Install Visual C++ for Python 2.7
+        if: runner.os == 'Windows'
+        run: choco install vcpython27 -f -y
+
+      - name: Build wheels
+        run: python -m cibuildwheel --output-dir wheelhouse
```
joerick, line 90: for the option, how about:
```yaml
        env:
          # add options here, for example,
          # CIBW_SKIP: "cp27-*" # Skip building CPython 2.7 on all platforms
```

README.md
```diff
@@ -58,49 +58,43 @@ Usage
 Example setup
 -------------
 
-To build manylinux, macOS, and Windows wheels on Travis CI and upload them to PyPI whenever you tag a version, you could use this `.travis.yml`:
+To build manylinux, macOS, and Windows wheels on Github and make them availble for download after the job runs, you could use this `.github/workflows/wheels.yml`:
```
joerick, line 61: minor copy tweak, more succinct-
```suggestion
To build manylinux, macOS, and Windows wheels on Github Actions, you could use this `.github/workflows/wheels.yml`:
```

examples/travis-ci-deploy.yml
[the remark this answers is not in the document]
joerick: Seems good to me, maybe something like 
```suggestion
    secure: "" # Set to an encrypted API token. See https://docs.travis-ci.com/user/deployment/pypi/
```

README.md
henryiii:
```suggestion
        os: [ubuntu-20.04, windows-2019, macOS-10.15]
```

Assuming we want 10.15 instead of 11.0, as it's still listed as preview?

README.md
henryiii, line 90:
```suggestion
      - name: Build wheels
        env:
          CIBW_SKIP: "?p25-*" # Skip building Python 2.7 wheels on all platforms
        run: python -m cibuildwheel --output-dir wheelhouse
```

README.md
joerick: Yeah that's what I had assumed, because `latest` pointed to 10.15

README.md
```diff
@@ -58,49 +58,41 @@ Usage
...
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==1.7.1
+
+      - name: Build wheels
+        env:
+          CIBW_SKIP: "?p25-*" # Skip building Python 2.7 wheels on all platforms
```
joerick, line 87: There's a typo here, it should be 27, not 25. But also I don't think the setting should be written in the config like this, as users might assume that it's required for this setup to work. Also skipping 2.7 would also mandate the removal of the choco install in windows. If the line is there but commented, users can assume it's educational, not required.

henryiii: Oops. How about:

```suggestion
          CIBW_SKIP: "?p27-*" # Skipping Python 2.7 for this example. See docs and examples for details on Python 2.7.
```

This has several benefits:

1. It avoids having to show choco install, which is very verbose for a simple readme example and is wrong for a _large_ portion of Python 2.7 users, since it doesn't support anything above C++03 (so no pybind11, etc).
2. cibuildhwheel will need to support Python 2 as long as the  manylinux wheels have it, but we don't have to include it in the "simplest possible" example in the readme. It's already been tripping up users like in #468 
3. It can remain in the bigger example in the examples folder.

I'd hope most users are past 2.7 support, so this is the new "default", with <50% now needing the Python 2 part, making an example that skips it more useful for a minimal first example. I'll restore it if you want it back, though.

YannickJadoul: Do we really want to do this? I thought the idea was that for this example to be as minimal/general as possible?

Czaki: As I see our general example does not show how to use config variables. Also, we got bug reports like #468. I`m ok with disabling python 2.7 in the initial configuration from readme and does not do this in the example yaml file.

henryiii: Dropping 2.7 via a variable is more minimal than adding a whole step that adds Python 2.7 but only works for C/C++<11. And does removing 2.7 from the readme example really make it less general at the end of 2020? If anything it still shows that we support it (since we remove it via a variable), and highlights that there are caveats to using it (the extra step or steps for C++11+ support).

YannickJadoul: > As I see our general example does not show how to use config variables.

There's a whole section "Options", right underneath the minimal example, so it should be reasonably clear if users read the docs?

> adding a whole step that adds Python 2.7 but only works for C/C++<11

Is that still an issue on GitHub Actions (compared to Travis)? I thought everything was much more streamlined on GHA?
But yes, if that's the case, I can probably agree.

YannickJadoul: > Is that still an issue on GitHub Actions (compared to Travis)?

Oh, I see, the installation step of the ancient Python 2.7 compiler is still there.

Czaki: > There's a whole section "Options", right underneath the minimal example, so it should be reasonably clear if users read the docs?

There is rather a problem that you need to found how to set an environment variable in yaml file. 
I think that it may be unclear that cibuildwheel build wheel on all python version if the environment variable is not set. I see examples with hole os times python version matrix.

henryiii:
As I see it, this is just a demo of "simple" usage and the examples still exist for more realistic use cases. I think having a Python 3+ example on the main page, with a mention of how to find Python 2.7+ examples, would be fine. If you have a Python 3+ package, then this example is pretty close to complete. If you have a 2.7+ package, then the other example wrong 50% of the time, since there are two choices for setting up the compiler on Windows. Plus it gives a real usage of the env variable setting, which 95% of all projects will likely need.

A full example would probably also show an upload job - this just makes the wheels available for download via the GitHub GUI (or for a later publish job). But it should be done as a separate job; that's easy to show in an example, but too verbose for a README.

Happy to go with whatever @joerick wants, though - just need confirmation of what to change.

joerick:
I see there's a lot of discussion here over something quite small, which smells a little bikesheddy so I apologise for not jumping in sooner. I've thought about it, and I agree with @henryiii that building for Python 2.7 shouldn't be the default any longer:

- Python 2 has been deprecated for nearly a year
- supporting Python 2 makes the sample config longer and slower, because of the need to `choco install` the compiler.

So let's leave it out of this config, here on the README. It can remain in the minimal examples in the docs for now, but I'm also considering if it should go from there as well, since I'd bet that most people setting up cibuildwheel these days don't want to build for 2.7 either. If we remove it from the minimal configs, we could have a page in the docs "Building Python 2 wheels" that shows the `choco install` command

I've made a couple tweaks to the PR, hope that's okay @henryiii  - just things that help make this more readable.

PR #482, opened 2020-12-12
Support qemu in GitHub Actions

### LINKED ISSUE #364: Seamless ARM builds using emulation through qemu docker images

asfaltboy:
Looks like it works generally, tests are failing as we need to update expected versions, and some more

We should also discuss if we want to enable all architectures by default, or set the default to only x86/64 architectures, as this may considered a type of "breaking change".

fixes #364

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,8 +24,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
-        # os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python_version: ['3.7']
     timeout-minutes: 180
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,7 @@ jobs:
 
   test-emulated:
     name: Test emulated cibuildwheel using qemu
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     timeout-minutes: 180
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,13 @@ jobs:
       id: pr-labels
       uses: joerick/pr-labels-action@v1.0.6
 
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: all
+      if: runner.os == "Linux"
+
     - name: Sample build
       if: contains(steps.pr-labels.outputs.labels, ' ci-sample-build ')
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-18.04]
+        # os: [ubuntu-18.04, windows-latest, macos-latest]
         python_version: ['3.7']
     timeout-minutes: 180
     steps:
@@ -52,7 +53,7 @@ jobs:
       uses: docker/setup-qemu-action@v1
       with:
         platforms: all
-      if: runner.os == "Linux"
+      if: runner.os == 'Linux'
 
     - name: Sample build
       if: contains(steps.pr-labels.outputs.labels, ' ci-sample-build ')
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,13 +48,6 @@ jobs:
       id: pr-labels
       uses: joerick/pr-labels-action@v1.0.6
 
-    - name: Set up QEMU
-      id: qemu
-      uses: docker/setup-qemu-action@v1
-      with:
-        platforms: all
-      if: runner.os == 'Linux'
-
     - name: Sample build
       if: contains(steps.pr-labels.outputs.labels, ' ci-sample-build ')
       run: |
@@ -63,3 +56,24 @@ jobs:
     - name: Test cibuildwheel
       run: |
         python ./bin/run_tests.py
+
+  test-emulated:
+    name: Test emulated cibuildwheel using qemu
+    runs-on: ubuntu-20.04
+    timeout-minutes: 180
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - name: Install dependencies
+      run: |
+        python -m pip install -r requirements-dev.txt
+
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: all
+
+    - name: Run the emulation tests
+      run: |
+        pytest --run-emulation test/test_emulation.py
--- a/bin/run_example_ci_configs.py
+++ b/bin/run_example_ci_configs.py
@@ -8,6 +8,7 @@ import sys
 import textwrap
 import time
 import click
+from glob import glob
 from collections import namedtuple
 from urllib.parse import quote
 
@@ -30,49 +31,67 @@ def generate_basic_project(path):
     project.generate(path)
 
 
-CIService = namedtuple('CIService', 'name src_config_path dst_config_path badge_md')
+CIService = namedtuple('CIService', 'name dst_config_path badge_md')
 services = [
     CIService(
         name='appveyor',
-        src_config_path='examples/appveyor-minimal.yml',
         dst_config_path='appveyor.yml',
         badge_md='[![Build status](https://ci.appveyor.com/api/projects/status/wbsgxshp05tt1tif/branch/{branch}?svg=true)](https://ci.appveyor.com/project/joerick/cibuildwheel/branch/{branch})',
     ),
     CIService(
         name='azure-pipelines',
-        src_config_path='examples/azure-pipelines-minimal.yml',
         dst_config_path='azure-pipelines.yml',
         badge_md='[![Build Status](https://dev.azure.com/joerick0429/cibuildwheel/_apis/build/status/joerick.cibuildwheel?branchName={branch})](https://dev.azure.com/joerick0429/cibuildwheel/_build/latest?definitionId=2&branchName={branch})',
     ),
     CIService(
-        name='circle-ci',
-        src_config_path='examples/circleci-minimal.yml',
+        name='circleci',
         dst_config_path='.circleci/config.yml',
         badge_md='[![CircleCI](https://circleci.com/gh/joerick/cibuildwheel/tree/{branch_escaped}.svg?style=svg)](https://circleci.com/gh/joerick/cibuildwheel/tree/{branch})',
     ),
     CIService(
         name='github',
-        src_config_path='examples/github-minimal.yml',
         dst_config_path='.github/workflows/example.yml',
         badge_md='[![Build](https://github.com/joerick/cibuildwheel/workflows/Build/badge.svg?branch={branch})](https://github.com/joerick/cibuildwheel/actions)',
     ),
     CIService(
         name='travis-ci',
-        src_config_path='examples/travis-ci-minimal.yml',
         dst_config_path='.travis.yml',
         badge_md='[![Build Status](https://travis-ci.org/joerick/cibuildwheel.svg?branch={branch})](https://travis-ci.org/joerick/cibuildwheel)',
     ),
     CIService(
         name='gitlab',
-        src_config_path='examples/gitlab-minimal.yml',
         dst_config_path='.gitlab-ci.yml',
         badge_md='[![Gitlab](https://gitlab.com/joerick/cibuildwheel/badges/{branch}/pipeline.svg)](https://gitlab.com/joerick/cibuildwheel/-/commits/{branch})'
     ),
 ]
 
 
+def ci_service_for_config_file(config_file):
+    for service in services:
+        if Path(config_file).name.startswith(service.name+'-'):
+            return service
+    raise ValueError(f'unknown ci service for config file {config_file}')
+
+
 @click.command()
-def run_example_ci_configs():
+@click.argument('config_files', nargs=-1, type=click.Path())
+def run_example_ci_configs(config_files=None):
+    '''
+        Test the example configs. If no files are specified, will test
+        examples/*-minimal.yml
+    '''
+
+    if len(config_files) == 0:
+        config_files = glob('examples/*-minimal.yml')
+
+    # check each CI service has at most 1 config file
+    configs_by_service = {}
+    for config_file in config_files:
+        service = ci_service_for_config_file(config_file)
+        if service.name in configs_by_service:
+            raise Exception('You cannot specify more than one config per CI service')
+        configs_by_service[service.name] = config_file
+
     if git_repo_has_changes():
         print('Your git repo has uncommitted changes. Commit or stash before continuing.')
         exit(1)
@@ -91,18 +110,19 @@ def run_example_ci_configs():
         example_project = Path('example_root')
         generate_basic_project(example_project)
 
-        for service in services:
-            src_config_file = Path(service.src_config_path)
+        for config_file in config_files:
+            service = ci_service_for_config_file(config_file)
+            src_config_file = Path(config_file)
             dst_config_file = example_project / service.dst_config_path
 
             dst_config_file.parent.mkdir(parents=True, exist_ok=True)
             shutil.copyfile(src_config_file, dst_config_file)
 
         run(['git', 'add', example_project], check=True)
-        run(['git', 'commit', '-m', textwrap.dedent(f'''
+        run(['git', 'commit', '--no-verify', '-m', textwrap.dedent(f'''
             Test example minimal configs
 
-            Testing files: {[s.src_config_path for s in services]}
+            Testing files: {config_files}
             Generated from branch: {previous_branch}
             Time: {timestamp}
         ''')], check=True)
@@ -116,9 +136,10 @@ def run_example_ci_configs():
         print('> ')
         print('> | Service | Config | Status |')
         print('> |---|---|---|')
-        for service in services:
+        for config_file in config_files:
+            service = ci_service_for_config_file(config_file)
             badge = service.badge_md.format(branch=branch_name, branch_escaped=quote(branch_name, safe=''))
-            print(f'> | {service.name} | `{service.src_config_path}` | {badge} |')
+            print(f'> | {service.name} | `{config_file}` | {badge} |')
         print('> ')
         print('> Generated by `bin/run_example_ci_config.py`')
         print()
--- a/bin/run_example_ci_configs.py
+++ b/bin/run_example_ci_configs.py
@@ -67,9 +67,12 @@ services = [
 
 
 def ci_service_for_config_file(config_file):
+    service_name = Path(config_file).name.rsplit('-', 1)[0]
+
     for service in services:
-        if Path(config_file).name.startswith(service.name+'-'):
+        if service.name == service_name:
             return service
+
     raise ValueError(f'unknown ci service for config file {config_file}')
 
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -70,19 +70,16 @@ def main() -> None:
                         ''')
 
     parser.add_argument(
-        '-a',
-        '--architectures',
-        choices=[arch.value for arch in Architecture],
-        default=[],
+        '--archs',
+        default=os.environ.get("CIBW_ARCHS", 'auto'),
         help='''
             Comma-separated list of CPU architectures to build for.
             If unspecified, builds the architectures natively supported
             on this machine. Set this option to build an architecture
             via emulation, for example, using binfmt_misc and qemu.
-        ''',
-        action="extend",
-        nargs="+",
-        type=Architecture,
+            Default: auto
+            Choices: auto, {}
+        '''.format(", ".join(a.name for a in Architecture)),
     )
     parser.add_argument('--output-dir',
                         default=os.environ.get('CIBW_OUTPUT_DIR', 'wheelhouse'),
@@ -185,8 +182,10 @@ def main() -> None:
         print('cibuildwheel: Could not find setup.py, setup.cfg or pyproject.toml at root of package', file=sys.stderr)
         exit(2)
 
+    archs = [Architecture(a) for a in args.archs.split(",")]
+
     if args.print_build_identifiers:
-        print_build_identifiers(platform, build_selector, args.architectures)
+        print_build_identifiers(platform, build_selector, archs)
         exit(0)
 
     manylinux_images: Optional[Dict[str, str]] = None
@@ -219,7 +218,7 @@ def main() -> None:
             manylinux_images[build_platform] = image
 
     build_options = BuildOptions(
-        architectures=args.architectures,
+        architectures=archs,
         package_dir=package_dir,
         output_dir=output_dir,
         test_command=test_command,
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -69,18 +69,16 @@ def main() -> None:
                             Python. Default: auto.
                         ''')
 
-    parser.add_argument(
-        '--archs',
-        default=None,
-        help='''
-            Comma-separated list of CPU architectures to build for.
-            If unspecified, builds the architectures natively supported
-            on this machine. Set this option to build an architecture
-            via emulation, for example, using binfmt_misc and qemu.
-            Default: auto
-            Choices: auto, {}
-        '''.format(", ".join(a.name for a in Architecture)),
-    )
+    parser.add_argument('--archs',
+                        default=None,
+                        help='''
+                            Comma-separated list of CPU architectures to build for.
+                            When set to 'auto', builds the architectures natively supported
+                            on this machine. Set this option to build an architecture
+                            via emulation, for example, using binfmt_misc and QEMU.
+                            Default: auto.
+                            Choices: auto, {}
+                        '''.format(", ".join(a.name for a in Architecture)))
     parser.add_argument('--output-dir',
                         default=os.environ.get('CIBW_OUTPUT_DIR', 'wheelhouse'),
                         help='Destination folder for the wheels.')
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -17,6 +17,7 @@ from cibuildwheel.environment import (
     parse_environment,
 )
 from cibuildwheel.util import (
+    Architecture,
     BuildOptions,
     BuildSelector,
     DependencyConstraints,
@@ -67,6 +68,22 @@ def main() -> None:
                             run in Windows, and it will build and test for all versions of
                             Python. Default: auto.
                         ''')
+
+    parser.add_argument(
+        '-a',
+        '--architectures',
+        choices=[arch.value for arch in Architecture],
+        default=[],
+        help='''
+            Comma-separated list of CPU architectures to build for.
+            If unspecified, builds the architectures natively supported
+            on this machine. Set this option to build an architecture
+            via emulation, for example, using binfmt_misc and qemu.
+        ''',
+        action="extend",
+        nargs="+",
+        type=Architecture,
+    )
     parser.add_argument('--output-dir',
                         default=os.environ.get('CIBW_OUTPUT_DIR', 'wheelhouse'),
                         help='Destination folder for the wheels.')
@@ -169,7 +186,7 @@ def main() -> None:
         exit(2)
 
     if args.print_build_identifiers:
-        print_build_identifiers(platform, build_selector)
+        print_build_identifiers(platform, build_selector, args.architectures)
         exit(0)
 
     manylinux_images: Optional[Dict[str, str]] = None
@@ -202,6 +219,7 @@ def main() -> None:
             manylinux_images[build_platform] = image
 
     build_options = BuildOptions(
+        architectures=args.architectures,
         package_dir=package_dir,
         output_dir=output_dir,
         test_command=test_command,
@@ -284,10 +302,12 @@ def print_preamble(platform: str, build_options: BuildOptions) -> None:
     print('\nHere we go!\n')
 
 
-def print_build_identifiers(platform: str, build_selector: BuildSelector) -> None:
+def print_build_identifiers(
+    platform: str, build_selector: BuildSelector, architectures: List[Architecture]
+) -> None:
     python_configurations: List[Any] = []
     if platform == 'linux':
-        python_configurations = cibuildwheel.linux.get_python_configurations(build_selector)
+        python_configurations = cibuildwheel.linux.get_python_configurations(build_selector, architectures)
     elif platform == 'windows':
         python_configurations = cibuildwheel.windows.get_python_configurations(build_selector)
     elif platform == 'macos':
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -71,7 +71,7 @@ def main() -> None:
 
     parser.add_argument(
         '--archs',
-        default=os.environ.get("CIBW_ARCHS", 'auto'),
+        default=None,
         help='''
             Comma-separated list of CPU architectures to build for.
             If unspecified, builds the architectures natively supported
@@ -182,7 +182,11 @@ def main() -> None:
         print('cibuildwheel: Could not find setup.py, setup.cfg or pyproject.toml at root of package', file=sys.stderr)
         exit(2)
 
-    archs = [Architecture(a) for a in args.archs.split(",")]
+    if args.archs is not None:
+        archs_config_str = args.archs
+    else:
+        archs_config_str = get_option_from_environment('CIBW_ARCHS', platform=platform, default='auto')
+    archs = Architecture.parse_config(archs_config_str, platform=platform)
 
     if args.print_build_identifiers:
         print_build_identifiers(platform, build_selector, archs)
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -18,10 +18,10 @@ from .util import (
 re_pattern = re.compile(r'[cp]p\d{2}-manylinux_(\w*)')
 
 
-def matches_platform(identifier: str, supported_platforms: List[Architecture]) -> bool:
+def matches_platform(identifier: str, architectures: List[Architecture]) -> bool:
     matched_architecture = re_pattern.search(identifier)
     id_architecture = matched_architecture.group(1) if matched_architecture else ''
-    return id_architecture in supported_platforms
+    return id_architecture in architectures
 
 
 class PythonConfiguration(NamedTuple):
@@ -73,10 +73,16 @@ def get_python_configurations(
     ]
 
     # skip builds as required
-    target_archs = architectures or [Architecture(platform.machine())]
-    # x86_64 machines can run i686 docker containers
-    if Architecture.i686 not in target_archs and Architecture.x86_64 in target_archs:
-        target_archs.append(Architecture.i686)
+    target_archs = architectures
+
+    if Architecture.auto in architectures:
+        target_archs.remove(Architecture.auto)
+        native_architecture = Architecture(platform.machine())
+        target_archs.append(native_architecture)
+        # x86_64 machines can run i686 docker containers
+        if native_architecture == Architecture.x86_64:
+            target_archs.append(Architecture.i686)
+
     print("matching platforms to architectures: ", target_archs)
     return [
         c for c in python_configurations
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -1,4 +1,3 @@
-import re
 import subprocess
 import sys
 import textwrap
@@ -12,15 +11,6 @@ from .util import (
     Architecture, BuildOptions, BuildSelector, NonPlatformWheelError,
     get_build_verbosity_extra_flags, prepare_command,
 )
-
-
-re_pattern = re.compile(r'[cp]p\d{2}-manylinux_(\w*)')
-
-
-def matches_platform(identifier: str, architectures: List[Architecture]) -> bool:
-    matched_architecture = re_pattern.search(identifier)
-    id_architecture = matched_architecture.group(1) if matched_architecture else ''
-    return id_architecture in architectures
 
 
 class PythonConfiguration(NamedTuple):
@@ -71,12 +61,11 @@ def get_python_configurations(
         PythonConfiguration(version='3.9', identifier='cp39-manylinux_s390x', path_str='/opt/python/cp39-cp39'),
     ]
 
-    # skip builds as required
-    target_archs = architectures
-
+    # return all configurations whose arch is in our `architectures` list,
+    # and match the build/skip rules
     return [
         c for c in python_configurations
-        if matches_platform(c.identifier, target_archs)
+        if any(c.identifier.endswith(arch) for arch in architectures)
         and build_selector(c.identifier)
     ]
 
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -9,18 +9,18 @@ from typing import List, NamedTuple, Union
 
 from .docker_container import DockerContainer
 from .logger import log
-from .util import (BuildOptions, BuildSelector, NonPlatformWheelError,
-                   get_build_verbosity_extra_flags, prepare_command)
+from .util import (
+    Architecture, BuildOptions, BuildSelector, NonPlatformWheelError,
+    get_build_verbosity_extra_flags, prepare_command,
+)
 
 
 re_pattern = re.compile(r'[cp]p\d{2}-manylinux_(\w*)')
 
 
-def matches_platform(identifier: str, supported_platforms: List[str]) -> bool:
+def matches_platform(identifier: str, supported_platforms: List[Architecture]) -> bool:
     matched_architecture = re_pattern.search(identifier)
     id_architecture = matched_architecture.group(1) if matched_architecture else ''
-    # x86_64 machines can run i686 docker containers
-    id_architecture = id_architecture if id_architecture == 'i686' else 'x86_64'
     return id_architecture in supported_platforms
 
 
@@ -34,7 +34,9 @@ class PythonConfiguration(NamedTuple):
         return PurePath(self.path_str)
 
 
-def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfiguration]:
+def get_python_configurations(
+    build_selector: BuildSelector, architectures: List[Architecture]
+) -> List[PythonConfiguration]:
     python_configurations = [
         PythonConfiguration(version='2.7', identifier='cp27-manylinux_x86_64', path_str='/opt/python/cp27-cp27m'),
         PythonConfiguration(version='2.7', identifier='cp27-manylinux_x86_64', path_str='/opt/python/cp27-cp27mu'),
@@ -71,7 +73,11 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
     ]
 
     # skip builds as required
-    target_archs = [platform.machine()]
+    target_archs = architectures or [Architecture(platform.machine())]
+    # x86_64 machines can run i686 docker containers
+    if Architecture.i686 not in target_archs and Architecture.x86_64 in target_archs:
+        target_archs.append(Architecture.i686)
+    print("matching platforms to architectures: ", target_archs)
     return [
         c for c in python_configurations
         if matches_platform(c.identifier, target_archs)
@@ -90,7 +96,7 @@ def build(options: BuildOptions) -> None:
         exit(2)
 
     assert options.manylinux_images is not None
-    python_configurations = get_python_configurations(options.build_selector)
+    python_configurations = get_python_configurations(options.build_selector, options.architectures)
     platforms = [
         ('cp', 'manylinux_x86_64', options.manylinux_images['x86_64']),
         ('cp', 'manylinux_i686', options.manylinux_images['i686']),
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -65,7 +65,7 @@ def get_python_configurations(
     # and match the build/skip rules
     return [
         c for c in python_configurations
-        if any(c.identifier.endswith(arch) for arch in architectures)
+        if any(c.identifier.endswith(arch.value) for arch in architectures)
         and build_selector(c.identifier)
     ]
 
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -1,4 +1,3 @@
-import platform
 import re
 import subprocess
 import sys
@@ -74,14 +73,6 @@ def get_python_configurations(
 
     # skip builds as required
     target_archs = architectures
-
-    if Architecture.auto in architectures:
-        target_archs.remove(Architecture.auto)
-        native_architecture = Architecture(platform.machine())
-        target_archs.append(native_architecture)
-        # x86_64 machines can run i686 docker containers
-        if native_architecture == Architecture.x86_64:
-            target_archs.append(Architecture.i686)
 
     return [
         c for c in python_configurations
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -13,6 +13,8 @@ from .util import (BuildOptions, BuildSelector, NonPlatformWheelError,
 
 
 def matches_platform(identifier: str) -> bool:
+    output = subprocess.check_output(['docker', '--version'])
+    print("We found support for these platforms:", output)
     pm = platform.machine()
     if pm == "x86_64":
         # x86_64 machines can run i686 docker containers
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -1,4 +1,5 @@
 import platform
+import re
 import subprocess
 import sys
 import textwrap
@@ -12,27 +13,15 @@ from .util import (BuildOptions, BuildSelector, NonPlatformWheelError,
                    get_build_verbosity_extra_flags, prepare_command)
 
 
-def matches_platform(identifier: str) -> bool:
-    output = subprocess.check_output(['docker', '--version'])
-    print("We found support for these platforms:", output)
-    pm = platform.machine()
-    if pm == "x86_64":
-        # x86_64 machines can run i686 docker containers
-        if identifier.endswith('x86_64') or identifier.endswith('i686'):
-            return True
-    elif pm == "i686":
-        if identifier.endswith('i686'):
-            return True
-    elif pm == "aarch64":
-        if identifier.endswith('aarch64'):
-            return True
-    elif pm == "ppc64le":
-        if identifier.endswith('ppc64le'):
-            return True
-    elif pm == "s390x":
-        if identifier.endswith('s390x'):
-            return True
-    return False
+re_pattern = re.compile(r'[cp]p\d{2}-manylinux_(\w*)')
+
+
+def matches_platform(identifier: str, supported_platforms: List[str]) -> bool:
+    matched_architecture = re_pattern.search(identifier)
+    id_architecture = matched_architecture.group(1) if matched_architecture else ''
+    # x86_64 machines can run i686 docker containers
+    id_architecture = id_architecture if id_architecture == 'i686' else 'x86_64'
+    return id_architecture in supported_platforms
 
 
 class PythonConfiguration(NamedTuple):
@@ -80,8 +69,14 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
         PythonConfiguration(version='3.8', identifier='cp38-manylinux_s390x', path_str='/opt/python/cp38-cp38'),
         PythonConfiguration(version='3.9', identifier='cp39-manylinux_s390x', path_str='/opt/python/cp39-cp39'),
     ]
+
     # skip builds as required
-    return [c for c in python_configurations if matches_platform(c.identifier) and build_selector(c.identifier)]
+    target_archs = [platform.machine()]
+    return [
+        c for c in python_configurations
+        if matches_platform(c.identifier, target_archs)
+        and build_selector(c.identifier)
+    ]
 
 
 def build(options: BuildOptions) -> None:
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -83,7 +83,6 @@ def get_python_configurations(
         if native_architecture == Architecture.x86_64:
             target_archs.append(Architecture.i686)
 
-    print("matching platforms to architectures: ", target_archs)
     return [
         c for c in python_configurations
         if matches_platform(c.identifier, target_archs)
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -4,13 +4,14 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import textwrap
 from os import PathLike
 from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Sequence, Union
 
 from .environment import ParsedEnvironment
 from .logger import log
-from .util import (BuildOptions, BuildSelector, NonPlatformWheelError,
+from .util import (Architecture, BuildOptions, BuildSelector, NonPlatformWheelError,
                    download, get_build_verbosity_extra_flags, get_pip_script,
                    install_certifi_script, prepare_command)
 
@@ -182,6 +183,12 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
+    if options.architectures != [Architecture.x86_64]:
+        raise ValueError(textwrap.dedent(f'''
+            Invalid archs option {options.architectures}. macOS only supports x86_64 for the moment.
+            If you want to set emulation architectures on Linux, use CIBW_ARCHS_LINUX instead.
+        '''))
+
     temp_dir = Path(tempfile.mkdtemp(prefix='cibuildwheel'))
     built_wheel_dir = temp_dir / 'built_wheel'
     repaired_wheel_dir = temp_dir / 'repaired_wheel'
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -124,13 +124,16 @@ class DependencyConstraints:
 
 
 class Architecture(str, Enum):
+    # mac/linux archs
     x86_64 = 'x86_64'
     i686 = 'i686'
     aarch64 = 'aarch64'
     ppc64le = 'ppc64le'
     s390x = 's390x'
-    win32 = 'win32'
-    win_amd64 = 'win_amd64'
+
+    # windows archs
+    x86 = 'x86'
+    amd64 = 'AMD64'
 
     @staticmethod
     def parse_config(config: str, platform: str) -> 'List[Architecture]':
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -122,6 +122,7 @@ class DependencyConstraints:
 
 
 class Architecture(str, Enum):
+    auto = 'auto'  # gets expanded to the native arch, possibly 32+64 bit
     x86_64 = 'x86_64'
     i686 = 'i686'
     aarch64 = 'aarch64'
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -123,7 +123,7 @@ class DependencyConstraints:
         return f'{self.__class__.__name__}{self.base_file_path!r})'
 
 
-class Architecture(str, Enum):
+class Architecture(Enum):
     # mac/linux archs
     x86_64 = 'x86_64'
     i686 = 'i686'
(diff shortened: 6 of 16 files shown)


henryiii:
Isn’t this just #469, but without the setup step? I am liking the idea of letting users activate this manually, then just expanding the set of arch’s. Can it be detected when enabled?

asfaltboy:
Actually missed that that work exists 😅  ...

Yes, we can detect supported architectures and installed emulators by running [Tonis' latest `binfmt`][1]:

```bash
docker run --privileged --rm tonistiigi/binfmt
```

In the [latest version][2] allows to install support for all or some architectures, as well as uninstalling, and showing.

This is what I did in this PR, i.e without first adding the support (whether via the github action or directly installing support), we would only build wheels for existing platforms on our linux machine.

[1]: https://hub.docker.com/r/tonistiigi/binfmt
[2]: https://github.com/tonistiigi/binfmt

.github/workflows/test.yml
```diff
@@ -47,6 +48,13 @@ jobs:
       id: pr-labels
       uses: joerick/pr-labels-action@v1.0.6
 
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: all
+      if: runner.os == 'Linux'
```
asfaltboy, line 56: The user can select which (additional linux) platforms they want to support with the above action.

In other CIs, they can simply execute `docker run --privileged --rm tonistiigi/binfmt --install all` (or any other platform selection), as per [`binfmt` docs](https://hub.docker.com/r/tonistiigi/binfmt)

joerick:
Hello @asfaltboy! Thanks for this approach! It is similar to #469, but yes, a little lighter. 

Due to the existence of Github Actions like `docker/setup-qemu-action` and Docker images like `binfmt`, I'm thinking that it perhaps does make sense for us to keep things simple here, and offload the qemu setup onto the CI config. 

If we do that, all that cibuildwheel needs to do is to provide a way to change what archs are built. As discussed in #469, I'm in favour of an explicit control here, so we can raise errors when things are misconfigured. I'm thinking something like:

```
$ cibuildwheel --help
...
optional arguments:
  --archs {x86_64,i686,aarch64,ppc64le,s390x}
                        Comma-separated list of CPU architectures to build for.
                        If unspecified, builds the architectures natively supported
                        on this machine. Set this option to build an architecture
                        via emulation, for example, using binfmt_misc and qemu.
...
```

I think the changes to our code would be pretty minimal then, but we could still have e.g. `aarch64` manylinux builds in our CI using the `docker/setup-qemu-action` action, and we can provide example configs in our docs to help people set it up.

asfaltboy:
I tend to agree, in that case we would be able to  remove [the lines here](https://github.com/joerick/cibuildwheel/pull/482/files#diff-847d1844bec880db6a75e5180243ce84d5482c21e99d2caf43666299b952e659R74-R81), and replace them with updated documentation explaining how to set qemu up.

Then the current tests should pass as is, although I'd probably like to add some additional tests for the new `--archs` arg.

If you're in favour of this, I can start tackling this later today perhaps

joerick:
Yes that sounds good! As long as @henryiii or @YannickJadoul don't have any issues with the above approach?

henryiii:
I think this is basically what I suggested in https://github.com/joerick/cibuildwheel/pull/469#issuecomment-743932191 under 2) - yes, I think this is best. Seems to be the simplest and most flexible solution.

cibuildwheel/linux.py
```diff
@@ -71,7 +73,11 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
     ]
 
     # skip builds as required
-    target_archs = [platform.machine()]
+    target_archs = architectures or [Architecture(platform.machine())]
+    # x86_64 machines can run i686 docker containers
+    if Architecture.i686 not in target_archs and Architecture.x86_64 in target_archs:
+        target_archs.append(Architecture.i686)
```
henryiii, line 79: Shouldn't this only be appended if this is not specified manually?

asfaltboy: Yes, the `Architecture.i686 not in target_archs` part of the condition checks that it wasn't specified manually. I found this way is more explicit and allows to specify both, or i686 (without x86), or x86 (which includes both) on their own.

joerick: If I do `cibuildwheel .` then I get the native archs (x86_64 and i686) but if I specify manually `cibuildwheel --archs "x86_64" .`, I should only get the x86_64 build. So adding the i686 build should only happen if the user doesn't specify archs in their options.

joerick (commented):
Looking good so far... 

I'm also wondering, for completeness, how this `--archs` option would relate to macOS/Windows, too. BUILD/SKIP has that pretty well covered, right now, but it might be easier to understand if --archs works the same across all platforms?

If we start with Windows, since the macOS situation is still evolving (#484). I think a sensible approach would be to keep x86_64 and i686 as the canonical descriptions of the archs, and map them to win_amd64 and win32 inside the script. 

(of course, I don't really imagine anyone would use --archs on macos/windows, it's mostly just for neatness i'm thinking about it)

The other option would just be that we document that this option only works on linux.

Czaki:
> I'm also wondering, for completeness, how this `--archs` option would relate to macOS/Windows, too. BUILD/SKIP has that pretty well covered, right now, but it might be easier to understand if --archs works the same across all platforms?

As I read comments in this thread I start thinking if it is possible to go back to an environment variable. Environment variables are simpler to use in the matrix of jobs. 

Because of my preference to use a job matrix for a simple package, I also prefer to use a colon separate list of arch instead of multiple add arguments like `cibuildwheel --arch x86_64 --arch i686 --arch ppc64le`.

I also like `CIBW_ARCHITECTURES` idea with the default value `686 x86_64 universal2`.

cibuildwheel/__main__.py
```diff
@@ -67,6 +68,22 @@ def main() -> None:
...
+        help='''
+            Comma-separated list of CPU architectures to build for.
+            If unspecified, builds the architectures natively supported
+            on this machine. Set this option to build an architecture
+            via emulation, for example, using binfmt_misc and qemu.
+        ''',
+        action="extend",
+        nargs="+",
+        type=Architecture,
+    )
```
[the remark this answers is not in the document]
henryiii: Can't we just call it `--arch` ? That's short enough and clear enough, I think. And then maybe it would be `CIBW_ARCH`?

henryiii:
> I don't really imagine anyone would use --archs on macos/windows

In an ideal world, if we could control this, it would be nice to have `--arch= x86_64 universal2 arm64` options to enable these builds, with auto being the native binaries.

Would it make sense to have `auto` or maybe better yet `native` be the default? Then you could do `--arch=native,ppc64le` to combine native (32 & 64 bit on non-macOS) with an emulated build without having to list out both.

cibuildwheel/linux.py
henryiii: I think "native", which is both the default and allowed in the list would be better. `native` would expand to `i686` and `x86_64`. Then you can do native + something else, or you can specify just one of these and not get the other.

.github/workflows/test.yml
henryiii: We aren't running a test that uses this yet, though. Should we add a build with ARM or PowerPC? Possibly trying to not add too much, since it's probably 5-10x slower?

henryiii:
@asfaltboy let me know if you need help. :) Looks like #484 will end up building on this.

ogrisel:
If `--arch` means building in an emulated environment, what flag would be used for cross-compiling without emulation?

For information, conda-forge has been using crossenv successfully to cross-compile conda binary packages to target the macos/arm64 platform on macos/ix86_64 host:

- https://pypi.org/project/crossenv/
- https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/

If cibuildwheel was to add support for cross-compiling, one option would be to use another dedicated flag such as `--xarch` for instance to make it possible to switch between emulation and crossenv.

Pros and cons of emulation  vs cross-compiling:

- cross compiling makes it possible to support architectures for which there is no free CI service (e.g. macos/arm64 from a macos/x86_64 host on Azure Pipelines / Github Actions)
- cross compiling is fast but cannot execute the test suite
- QEMU on works on linux hosts and AFAIK not possible to build windows or macos wheels with it (because one does not have access to the runtime libs for those)
- emulation is slow but can execute the test suite (assuming it is not too long to avoid a timeout)

Czaki:
> If `--arch` means building in an emulated environment, what flag would be used for cross-compiling without emulation?

What is the difference for user if compilation is done with, or without emulation? Emulation is needed rather for testing. 

I do not think that there is a need to create a separate method to distinguish method to achieve build wheel for a given architecture.

ogrisel:
But we need a way for the maintainer whether to use cross-compiling or QEMU emulation when running cibuildwheel on a CI host, no?

henryiii:
The speed here is _horrible_ for compiling (a 5 minute compile becomes 50 minute). If we could cross compile and then emulate only for testing, that would be a huge win. Also would be clearer with macOS - we can cross-compile, but not emulate, so that's why testing is unavailable. For Apple Silicon, we can cross compile and emulate, so that's why testing can be used.

henryiii:
But I think that's orthogonal to this; here we are only providing a way to override the listed supported architectures, and then assuming you can run the docker containers. Currently, we only support native docker containers. I assume you'd have to have to use the normal manylinux docker container for cross compiling, since you are running the builder's arch not the host arch.  Then you'd need to "trick" it inside the docker container with something like the package linked above. If you give cibuildwheel a docker image for `CIBW_MANYLINUX_AARCH64_IMAGE` that is x86, force the `--arch` as seen above, then all you'd need is the new build procedure with the trick above (which might be something you could "burn-in" to the docker image itself, possibly? Or you'd need a cross compile flag/var and a new procedure.).

It's more relevant with the macOS change - as that's not tricking the system into running a different docker container, so is really much more inline with the "cross-compile" change than this one.

henryiii:
```
>       assert len(expected_wheels) == len(build_identifiers)
E       assert 17 == 18
E        +  where 17 = len(['spam-0.1.0-cp35-cp35m-manylinux1_x86_64.whl', 'spam-0.1.0-cp35-cp35m-manylinux1_i686.whl', 'spam-0.1.0-cp36-cp36m-ma...-manylinux1_i686.whl', 'spam-0.1.0-cp37-cp37m-manylinux1_x86_64.whl', 'spam-0.1.0-cp37-cp37m-manylinux1_i686.whl', ...])
E        +  and   18 = len(["matching platforms to architectures:  [<Architecture.x86_64: 'x86_64'>, <Architecture.i686: 'i686'>]", 'cp27-manylinux_x86_64', 'cp27-manylinux_x86_64', 'cp35-manylinux_x86_64', 'cp36-manylinux_x86_64', 'cp37-manylinux_x86_64', ...])
```
Why is the print output in here?

henryiii:
? Apologies! Your build didn’t run because you haven’t selected a plan that includes macOS yet. Please change to a Performance Plan. If you have any questions please feel free to contact support.

Looks like it runs on the "users's" plan, not cibuildwheel, so that's why CircleCI macOS is out.

henryiii:
@joerick, any ideas for testing this? And everyone, review? Other than adding a test and enabling the non-linux GHA testing (not sure why that one is slimmed down but not any of the others), I think this is ready for a review?

.github/workflows/test.yml
```diff
@@ -24,7 +24,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-18.04]
+        # os: [ubuntu-18.04, windows-latest, macos-latest]
```
henryiii, line 28:
```suggestion
        os: [ubuntu-latest, windows-latest, macos-latest]
```

@joerick Not sure what you want to do for our own testing - latest or pinned.

fruch:
I was in the middle of moving to github actions (and about to forget about building arm)
Taking it for for a ride:
https://github.com/scylladb/python-driver/pull/80/checks?check_run_id=1588381001

cibuildwheel/__main__.py
```diff
@@ -67,6 +68,19 @@ def main() -> None:
...
+    parser.add_argument(
+        '--archs',
+        default=os.environ.get("CIBW_ARCHS", 'auto'),
+        help='''
+            Comma-separated list of CPU architectures to build for.
+            If unspecified, builds the architectures natively supported
+            on this machine. Set this option to build an architecture
+            via emulation, for example, using binfmt_misc and qemu.
+            Default: auto
+            Choices: auto, {}
```
asfaltboy, line 81: `auto` choice is already in the Architectures choices

.github/workflows/test.yml
asfaltboy: sure, let's commit this

cibuildwheel/linux.py
```diff
@@ -78,8 +71,23 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
...
     # skip builds as required
-    return [c for c in python_configurations if matches_platform(c.identifier) and build_selector(c.identifier)]
+    target_archs = architectures
+
+    if Architecture.auto in architectures:
+        target_archs.remove(Architecture.auto)
+        native_architecture = Architecture(platform.machine())
+        target_archs.append(native_architecture)
+        # x86_64 machines can run i686 docker containers
+        if native_architecture == Architecture.x86_64:
```
asfaltboy: if i686 is already selected (+x86+64), it will duplicate it. I think it's harmless though

asfaltboy (commented):
Looks good @henryiii thanks for picking it up 🙇

joerick:
> @joerick, any ideas for testing this? And everyone, review? Other than adding a test and enabling the non-linux GHA testing (not sure why that one is slimmed down but not any of the others), I think this is ready for a review?

I've added some unit tests and an integration test. Because the integration test requires the QEMU action I've moved it into a separate runner.

We still need some docs, and maybe an example config for emulation.

Just realised, I added platform-specific options, but currently this doesn't actually do anything on Mac/Windows yet. Should we add preliminary support for completeness sake?

joerick:
We've had this CircleCI failure before... I think it happens when a fork gets associated with the forker's billing account rather than cibuildwheel's. Anyway, we can ignore that failure.

cibuildwheel/__main__.py
joerick: Ah, dang I think you reviewed while I was making a change! I've actually removed `auto` from the Architecture choices, it's now resolved into an Architecture at parse time.

.github/workflows/test.yml
joerick: I think latest for our own CI, we want to catch incompatibilities I think

asfaltboy:
> Just realised, I added platform-specific options, but currently this doesn't actually do anything on Mac/Windows yet. Should we add preliminary support for completeness sake?

I was pondering whether, in addition to mentioning it in the docs, we should raise an exception perhaps, to make it clear archs can only be used on Linux right now?

joerick:
> I was pondering whether, in addition to mentioning it in the docs, we should raise an exception perhaps, to make it clear --archs can only be used on Linux right now?

Good idea. And seeing as we might be using CIBW_ARCH for universal2, we should ensure that it explicitly isn't set on macOS, for now. So I'll add an assert that `config.architectures` is just x86_64 on macOS. Otherwise we might see linux archs polluting that option, which might cause an error in future versions.

cibuildwheel/util.py
```diff
@@ -121,7 +123,42 @@ def __repr__(self):
...
+    # mac/linux archs
+    x86_64 = 'x86_64'
+    i686 = 'i686'
+    aarch64 = 'aarch64'
+    ppc64le = 'ppc64le'
+    s390x = 's390x'
+
+    # windows archs
+    x86 = 'x86'
+    amd64 = 'AMD64'
```
henryiii, line 136: Shouldn’t the caps match here?

henryiii (approved):
Having a string enum not match its value seems potentially confusing, calling .upper() on windows might be better?

cibuildwheel/util.py
henryiii: Though maybe it’s fine. Just want med to point out a potential confusing point.

cibuildwheel/macos.py
```diff
@@ -182,6 +183,12 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
+    if options.architectures != [Architecture.x86_64]:
+        raise ValueError(textwrap.dedent(f'''
+            Invalid archs option {options.architectures}. macOS only supports x86_64 for the moment.
+            If you want to set emulation architectures on Linux, use CIBW_ARCHS_LINUX instead.
```
asfaltboy, line 189: Just had a thought, should we update the CI configurations where setting the env var in a "global" level will make non-Linux builds in the matrix fail because of this assertion?

cibuildwheel/util.py
joerick: I think I agree with you, actually. Seems like it's asking for trouble, as written here

cibuildwheel/macos.py
joerick: I'm not sure what you mean, "the CI configurations"?

joerick:
Okay, I've added some docs, I think this is ready for final review now!

cibuildwheel/linux.py
```diff
@@ -8,29 +8,19 @@
...
+)
+
+
+re_pattern = re.compile(r'[cp]p\d{2}-manylinux_(\w*)')
+
+
+def matches_platform(identifier: str, architectures: List[Architecture]) -> bool:
+    matched_architecture = re_pattern.search(identifier)
+    id_architecture = matched_architecture.group(1) if matched_architecture else ''
+    return id_architecture in architectures
```
joerick, line 23: I had a comment on this earlier, which seems to have been lost - do people think the following might be simpler?

```suggestion
    return any(identifier.endswith(arch) for arch in supported_platforms)```

**Comment by joerick** [comment]
> **Examples test run**
> 
> Branch: [example-config-test---support-quemu-on-github-2020-12-23T21-09-59](https://github.com/joerick/cibuildwheel/tree/example-config-test---support-quemu-on-github-2020-12-23T21-09-59)
> 
> | Service | Config | Status |
> |---|---|---|
> | github | `examples/github-with-qemu.yml` | [Build](https://github.com/joerick/cibuildwheel/actions) |
> 
> Generated by `bin/run_example_ci_config.py`

File: cibuildwheel/macos.py (hunk 8 above)
Comment (asfaltboy) [reply cibuildwheel/macos.py:189]: (I think) I meant the documentation on "how to configure CI to use cibuildwheel" , so for example in the sample file below:

https://github.com/joerick/cibuildwheel/pull/482/files?short_path=2e04679#diff-fcf87e56f37482d09dcba260e0c81bf1817b22e1bd964091931002ea67918fabR42

If I'm not mistaken, specifying the env var there without os restriction, will set `options.architerctures` for MacOS as well, which raises an error here on line 187 right ?

File: cibuildwheel/linux.py (hunk 9 above)
Comment (asfaltboy) [reply cibuildwheel/linux.py:23]: The only reason I went with a regex pattern was because I wasn't sure we will always match the architecture at the end of each identifier. Though thats probably premature optimization if we don't expect that to change any time soon.

**Comment by fruch** [comment]
> I was in the middle of moving to github actions (and about to forget about building arm)
> Taking it for for a ride:
> https://github.com/scylladb/python-driver/pull/80/checks?check_run_id=1588381001

I'm so happy with that, that I have it merged as experimental workflow, and wheels out of it are already published into pypi.
https://github.com/scylladb/python-driver/actions/runs/446668338

:1st_place_medal:

File: cibuildwheel/macos.py (hunk 8 above)
Comment (joerick) [reply cibuildwheel/macos.py:189]: Here I'm setting `CIBW_ARCHS_LINUX`, which won't affect the value of `options.architectures` on macOS/Windows.

**Comment by joerick** [comment]
Any further comments @henryiii @YannickJadoul before we merge this one?

File: cibuildwheel/macos.py (hunk 8 above)
Comment (henryiii) [reply cibuildwheel/macos.py:189]: @asfaltboy means (I think) that there is no test that sets `CIBW_ARCHS` and verifies that an exception is thrown. I think it's something we could add later, better to get this in for now? Personally, I'm pretty bad about forgetting to add tests for expected failures, and it has bit me a couple of times.

**Review by henryiii — APPROVED** [review]
I'm in favor of merging it when you are ready to do so!

**Comment by henryiii** [comment]
Just a thought; should we do a 1.7.2 release then merge this to give it a little time in master? And maybe so we can work on Apple Silicon as well for 1.8.0, while we still have a little flexibility to adjust this if needed for consistency?

**Comment by joerick** [comment]
Good idea @henryiii 👍

**Comment by YannickJadoul** [comment]
Sorry for being late to the party! I'll give this a quick look, now.

File: cibuildwheel/util.py [hunk 10]
```diff
@@ -121,7 +123,42 @@ def __repr__(self):
         return f'{self.__class__.__name__}{self.base_file_path!r})'
 
 
+class Architecture(str, Enum):
```
YannickJadoul, line 126: Subclassing `str` is to satisfy `mypy`?

cibuildwheel/__main__.py
```diff
@@ -67,6 +68,19 @@ def main() -> None:
...
+        '--archs',
+        default=None,
+        help='''
+            Comma-separated list of CPU architectures to build for.
+            If unspecified, builds the architectures natively supported
+            on this machine. Set this option to build an architecture
+            via emulation, for example, using binfmt_misc and qemu.
+            Default: auto
+            Choices: auto, {}
+        '''.format(", ".join(a.name for a in Architecture)),
```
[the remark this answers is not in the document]
YannickJadoul: As expected, an enum.

That's horrible syntax IMO, looping over a type, but apparently [how Python envisions you doing things](https://docs.python.org/3/library/enum.html), so why not, then.

cibuildwheel/util.py
```diff
@@ -121,7 +123,42 @@ def __repr__(self):
...
+    AMD64 = 'AMD64'
+
+    @staticmethod
+    def parse_config(config: str, platform: str) -> 'List[Architecture]':
+        result = []
+        for arch_str in re.split(r'[\s,]+', config):
+            if arch_str == 'auto':
+                result += Architecture.auto_archs(platform=platform)
+            else:
+                result.append(Architecture(arch_str))
```
YannickJadoul, line 145: Do we need better error messages, either here or where it's used? Or is some stack trace + `'foo' is not a valid Architecture` good enough?

cibuildwheel/util.py
```diff
@@ -121,7 +123,42 @@ def __repr__(self):
...
+        return result
+
+    @staticmethod
+    def auto_archs(platform: str) -> 'List[Architecture]':
```
YannickJadoul, line 149: Same here about `native`? The variable names inside are `native`, so why the function not as well?
(depends on `--arch auto` vs. `--arch native`, though)

cibuildwheel/__main__.py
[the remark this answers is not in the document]
YannickJadoul: This should be `a.value` rather than `a.name`, though, to match the logic from `Architecture.parse_config()` (even though, yes, names and values are the same) ?

cibuildwheel/util.py
```diff
@@ -121,7 +123,42 @@ def __repr__(self):
...
+        if platform == 'linux' and native_architecture == Architecture.x86_64:
+            # x86_64 machines can run i686 docker containers
+            result.append(Architecture.i686)
+        if platform == 'windows' and native_architecture == Architecture.AMD64:
+            result.append(Architecture.x86)
+        return result
+
+
 class BuildOptions(NamedTuple):
+    architectures: List[Architecture]
```
YannickJadoul, line 161: Nitpick: these aren't alphabetically sorted, but rather chronologically (when it's needed by `cibuildwheel`) or vaguely by topic or something like that. So let's move it between `build_selector` and `environment`?

docs/faq.md
```diff
@@ -24,6 +24,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
 
 - Alternative dockers images can be specified with the `CIBW_MANYLINUX_X86_64_IMAGE`, `CIBW_MANYLINUX_I686_IMAGE`, and `CIBW_MANYLINUX_PYPY_X86_64_IMAGE` options to allow for a custom, preconfigured build environment for the Linux builds. See [options](options.md#manylinux-image) for more details.
 
+### Building non-native architectures using emulation  {: #emulation}
+
+cibuildwheel supports building non-native architectures on Linux, via
+emulation through the binfmt_misc kernel feature. The easiest way to use this
```
YannickJadoul, line 30: To be pedantically correct: "emulation through QEMU and integration through the binfmt_misc kernel feature"

As far as I understood, `binfmt` is really just the feature that links an executable file to the right way of executing it, but QEMU is doing all the hard emulation work here, right?

docs/faq.md
```diff
@@ -24,6 +24,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
+is via the [docker/setup-qemu-action][setup-qemu-action] on Github Actions or
+[tonistiigi/binfmt][binfmt].
+
+[setup-qemu-action]: https://github.com/docker/setup-qemu-action
+[binfmt]: https://hub.docker.com/r/tonistiigi/binfmt
+
+Check out the following config for an example of how to set it up on Github
+Actions. Once qemu is set up and registered, you just need to set the
```
YannickJadoul, line 38: "QEMU"

YannickJadoul: Do we also need a non-GHA example on how to set this up with the Docker image?

docs/faq.md
```diff
@@ -24,6 +24,26 @@ Linux wheels are built in the [`manylinux` docker images](https://github.com/pyp
...
+`CIBW_ARCHS_LINUX` variable, and the other architectures are emulated
```
YannickJadoul, line 39: Or pass `--archs`

docs/options.md
```diff
@@ -510,6 +533,12 @@ optional arguments:
                         don't run on your development machine. For "windows",
                         you need to run in Windows, and it will build and test
                         for all versions of Python. Default: auto.
+  --archs ARCHS         Comma-separated list of CPU architectures to build
+                        for. If unspecified, builds the architectures natively
```
YannickJadoul, line 537: Unspecified is shown underneath by "default". It's more important to link it to the fact that that's what `auto` means.

Then again, that should be reasonably clear...

unit_test/main_tests/conftest.py
```diff
@@ -70,5 +70,14 @@ def platform(request, monkeypatch):
 @pytest.fixture
 def intercepted_build_args(platform, monkeypatch):
     intercepted = ArgsInterceptor()
-    monkeypatch.setattr(globals()[platform], 'build', intercepted)
+
+    if platform == 'linux':
```
YannickJadoul, line 74: What happened here? Teah, it's probably cleaner this way, but just curious if the old approach broke something?

YannickJadoul (approved):
Nice! Great job, @asfaltboy, @henryiii, and @joerick! (and @russkel as well, in #469, ofc!)

I have a few nitpicks here and there, but not much more.

Three main remarks:
- The `auto` default value bugs me a bit, because everywhere, there's an explanation that it's the "native architectures". So I don't really see why it wouldn't just be `native` and be more transparant?
- Why do we actually still need the command line argument `--archs` if you (probably) can't set it on the command line anyway (since it'll shout at you that you need `CIBW_ARCH_LINUX`) ? I'm perfectly fine to keep it, though, but taking a step backwards, I'm slightly confused :-)
- Not really blocking to this PR (I quite like this approach, now that I see it fully!), but I'm really not convinced macOS's cross-compilation should be lumped together with these architectures. But maybe once this gets merged, we can take this as a starting point for that discussion.

cibuildwheel/__main__.py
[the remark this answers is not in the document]
henryiii: That's not a "type", it's an enum. How else would you loop over all the possible enums? C++ doesn't have this capability (yet, at least, until reflection is available?). Keep in mind adding any methods to Enum would remove possible Enum values! And yes, I think it should be `name.value`.

cibuildwheel/__main__.py
```diff
@@ -67,6 +68,19 @@ def main() -> None:
...
+
+    parser.add_argument(
+        '--archs',
+        default=None,
+        help='''
+            Comma-separated list of CPU architectures to build for.
+            If unspecified, builds the architectures natively supported
+            on this machine. Set this option to build an architecture
+            via emulation, for example, using binfmt_misc and qemu.
+            Default: auto
```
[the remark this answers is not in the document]
henryiii: I discussed both names then picked `auto` because it is used by `--plat`, just above. It might still be better to make this `native`, and not match `--plat`, though.

cibuildwheel/util.py
henryiii: No, this makes this a "string enum". That is, `Architecture.X` _is_ the value string X. This has a few odd repercussions (string methods are no longer valid Enum names), but makes the enum a bit more natural as it also works naturally as a string. `Architecture.X == "X"` is true, for example.

henryiii: This is required for cross-library enums, which is where I've used it before.

henryiii:
Summary:
* I'm just as happy with `native` as `auto`, I just chose the latter since it was already used by `--plat/CIBW_PLATFORM` in a similar (but not quite the same?) way.
* I'm a little unsure of `--arch`. I feel this is actually better as a command-line argument, since it is affecting the search parameters just like `--plat` (and keeps it clearly distinct from BUILD/SKIP), but the need to do this per-platform forces most users to use the environment variable version. I'm rather hoping this evolves a bit as we add Universal2, and maybe Linux cross-compilation (which is part of why I'd like this to sit in master for a little bit while we work on that).
* Python Enums are not as nice as Ruby Symbol's, but otherwise aren't bad for enums in an interpreted language. :) Do keep in mind they really aren't normal classes, they somewhat hijack the class syntax. Not quite as much as TypedDict or dataclasses do.

henryiii:
I was just looking up the definition of IntEnum to prove it is just `int, Enum` ([and it is](https://github.com/python/cpython/blob/ba0e49a4648e727d1a16b3ce479499eb39f22311/Lib/enum.py#L806-L809)), and found StrEnum was [recently proposed and merged]() and will be in 3.10. Nice. :)

There are three extra components to [the implementation](https://github.com/python/cpython/blob/ba0e49a4648e727d1a16b3ce479499eb39f22311/Lib/enum.py#L812-L843); some correctness checking on new (not needed), support for `auto()` (not needed), and `__str__ = str.__str__`, which is a very nice and natural way to drop the `.value` requirements when joining above.

mayeut:
Sorry for being  late to the party too  !
I made some quick experiment and I'll provide a more complete feedback later today.
For the quick feedback, overall very nice, thanks for all the work, I could build different ways as can be seen there:
- "Single build per os": This is straightforward, `cibuildwheel` does its job as always, https://github.com/mayeut/pybase64/actions/runs/452500462
- "Split per arch (almost) on linux": the previous one was a bit too slow for my taste so I tried too split a bit more, works like a charm, https://github.com/mayeut/pybase64/actions/runs/452542463
- "Also split per arch on windows": this one is failing and I guess shouldn't be, https://github.com/mayeut/pybase64/actions/runs/452571598

cibuildwheel/__main__.py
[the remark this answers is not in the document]
YannickJadoul: Well, an enum is a type, no? Both in the broad sense (we're using it to tell mypy what's passed) as well as in the Python sense where `enum.EnumMeta` is a subclass of `type`.
But yeah, I can see it's a special one. Just haven't seen this that often and the capital/syntax highlighting is quite confusing; nothing this PR should do anything about, though!

Reiterating that `a.name` -> `a.value` is still a change to be made, just to make sure the serious part of my comments stands out ;-)

cibuildwheel/__main__.py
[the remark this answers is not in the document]
YannickJadoul: Up to @joerick, as far as I'm concerned. I just found it striking, reading this PR's code for the first time, that every time there's `auto`, there's some kind of explanation that "auto means the native architectures, here".

cibuildwheel/util.py
YannickJadoul: Ah, cool! I didn't know about that; thanks!

YannickJadoul:
Thanks, @henryiii :-)

>    * I'm just as happy with `native` as `auto`, I just chose the latter since it was already used by `--plat/CIBW_PLATFORM` in a similar (but not quite the same?) way.

As reacted somewhere deep in the comments: 

> Up to @joerick, as far as I'm concerned. I just found it striking, reading this PR's code for the first time, that every time there's auto, there's some kind of explanation that "auto means the native architectures, here".
 
>    * I'm a little unsure of `--arch` too. I feel this is actually better as a command-line argument, since it is affecting the search parameters just like `--plat` (and keeps it clearly distinct from BUILD/SKIP), but the need to do this per-platform forces most users to use the environment variable version.

We're fine to keep `--arch`. I am just doubtful how often it's going to be used, as you say.

> I'm rather hoping this evolves a bit as we add Universal2, and maybe Linux cross-compilation (which is part of why I'd like this to sit in master for a little bit while we work on that).

I'm still very skeptical about mixing these two, though. "Universal2" is not really an architecture, and no one is emulating it. But that's a discussion for the other PR, so I agree we can see how things fit in there.

The only repercussion that could have on this PR is if we decide to go with `--emulate-archs` and `--cross-archs`, or so, but ... by itself, I think this PR looks great and has the right approach!

>    * Python Enums are not as nice as Ruby Symbol's, but otherwise aren't bad for enums in an interpreted language. :) Do keep in mind they really aren't normal classes, they somewhat hijack the class syntax. Not quite as much as TypedDict or dataclasses do.

No worries about this; let's blame it on me not having used enums too often in Python! `StrEnum` seems cool, though!

cibuildwheel/__main__.py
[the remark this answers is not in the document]
mayeut: I find `auto` a bit better than `native` because, imho, `native` implies a single architecture whereas `auto` suggests something more.

cibuildwheel/util.py
mayeut: This part of the code reflects what I commented earlier `auto` and `native` are not the same. It's very clear with this function.

cibuildwheel/windows.py
```diff
@@ -201,6 +202,13 @@ def pep_518_cp35_workaround(package_dir: Path, env: Dict[str, str]) -> None:
 
 
 def build(options: BuildOptions) -> None:
+    if options.architectures != [Architecture.AMD64, Architecture.x86]:
```
mayeut, line 205: Here's the source of the error I'm seeing when trying to split the windows build using `--archs`.
I know it could be done other ways but it might be good to be able to build individual architectures the same way on all OS. I know it was not the focus of the PR but would make configuration even easier (at least for my use case).

YannickJadoul: What about `not set(option.architectures) <= {Architecture.AMD64, Architecture.x86}`?

cibuildwheel/__main__.py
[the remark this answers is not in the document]
YannickJadoul: My issue with `auto` is that it sounds like "you detect which architectures I can build".

If `native` only implies one architecture (so... no `i686` on `x86_64` Linux, I assume?), then "If unspecified, builds the architectures natively supported on this machine." is an incorrect description, though.

But yes, I was just bikeshedding here; there's nothing else to complain about, so ... ;-)

mayeut:
Summary:
- I think `auto` is better than `native` in the proposed scoped. `native` could be added to get another, single architecture, level of abstraction.
- I was  a bit skeptical about `--archs`:
  - command-line vs environment variable: after misconfiguring the `cibuildwheel` to use when testing the PR, I find it has a significant advantage to environment variable. It will tell you right away if the command-line switch is unsupported (e.g. typo, wrong `cibuildwheel` version) rather than silently ignoring the unused/misspelled environment variable. This might apply to other options as well (though they might have their own advantage with the suffix part being probably easier on some CIs ?)
  - `ARCHS` vs `BUILD/SKIP`: while all of this should be doable with `BUILD/SKIP`, I really think it's much clearer keeping those separate.
- I think we should allow aliasing between some architectures or just remove some. For windows, `x86` and `AMD64` are expected. It might be easier to allow aliasing with `i686` and `x86_64`. At least, on linux, the architectures name end-up in the wheel so, even without reading the docs thoroughly, you can get this right if you know what to expect. This does not hold on windows where you need to read the docs thoroughly (or the code).

cibuildwheel/windows.py
mayeut: I will try to get this working in a branch. For the specific test, I guess you're right but I've not read the rest of the code which will probably require some adjustments.

YannickJadoul: Meanwhile, the alternative would be to say "don't touch `--archs` on Windows, and just use `CIBW_BUILD`/`CIBW_SKIP`" (cfr. only one way to do things?)

cibuildwheel/__main__.py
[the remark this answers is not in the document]
henryiii: Looking at the code, `CIProvider` really should not be a StrEnum - there's no need for that. I'm rather thinking `Architecture` shouldn't be either. It is _not_ needed to subclass `str` to _use_ strings as values, but only adds implicit conversion. I don't see any reason that we can't use explicit conversion here.

Though the implicit form does allow this beauty:

```python
", ".join(Architecture),
```

How's that for a "type"? :)

BTW, an enum should be thought of as both a type and a container of all the instances of that type. The fact that is is a type is actually a bit of a workaround in leu of true dedicated language support, for example `Architecture.count` is the `str.count` method; you can't add a "count" member to a StrEnum (at least without breaking its stringiness).

@joerick what do you think? Should I remove `str, ` from both of these and make them regular enums with string-based values, or do you want me to add:
```python
# Identical to StrEnum in Python 3.10 except for less checking on construction
# that we really are providing strings, and no support for enum.auto().
class StrEnum(str, Enum):
    __str__ = str.__str__
```
and then use that, making these full StrEnums?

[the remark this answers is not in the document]
YannickJadoul: > How's that for a "type"? :)

Hah, still horrible, but better than before? ;-)

henryiii:
> ARCHS vs BUILD/SKIP: while all of this should be doable with BUILD/SKIP, I really think it's much clearer keeping those separate.

What I meant was that `--archs` and `--plat` basically control which selectors are presented to `BUILD`/`SKIP`; and setting anything other than `"auto"` could be _much_ slower via docker or emulated docker. So it seems natural that they are primarily command line variables, while `BUILD`/`SKIP` are only env variables.

cibuildwheel/windows.py
mayeut: Using https://github.com/mayeut/cibuildwheel/commit/b5e620741f42d2ed3a62ae4c9d1a0fd301a15929 on top of this PR, I can achieve what I want: https://github.com/mayeut/pybase64/actions/runs/452918655

> Meanwhile, the alternative would be to say "don't touch --archs on Windows, and just use CIBW_BUILD/CIBW_SKIP" (cfr. only one way to do things?)

@YannickJadoul, "only one way to do things" makes sense however, although I understand why it would be specific for `qemu`, this PR now gives 2 ways of doing things on Linux (even if you're not using `qemu` specific arch) and if you're using `qemu`, then it easier to write your workflow configuration if this also works on Windows (at least on GHA). Not only it's easier, I find it conveys a very clear message on what is being achieved.

With `--archs` for all builds: 
```
    name: Build ${{ matrix.archs }} wheels on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        include:
          - os: ubuntu-18.04
            archs: "x86_64,i686"
          - os: ubuntu-18.04
            archs: "aarch64"
          - os: ubuntu-18.04
            archs: "ppc64le"
          - os: ubuntu-18.04
            archs: "s390x"
          - os: windows-2019
            archs: "AMD64"
          - os: windows-2019
            archs: "x86"
          - os: macos-10.15
            archs: "x86_64"

    steps:
      ...
      - name: Build wheels
        env:
          CIBW_TEST_REQUIRES: "pytest"
          CIBW_TEST_COMMAND: "pytest --pyargs pybase64"
          CIBW_SKIP: "?p27* pp*"
        run: python -m cibuildwheel --archs ${{ matrix.archs }} --output-dir wheelhouse
```

vs
With `--archs` for all builds except Windows which uses `skip`:
```
name: Build ${{ matrix.archs }} wheels on ${{ matrix.os }} with skip ${{ matrix.skip }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        include:
          - os: ubuntu-18.04
            archs: "x86_64,i686"
            skip: ""
          - os: ubuntu-18.04
            archs: "aarch64"
            skip: ""
          - os: ubuntu-18.04
            archs: "ppc64le"
            skip: ""
          - os: ubuntu-18.04
            archs: "s390x"
            skip: ""
          - os: windows-2019
            archs: "auto"
            skip: "*win32*"
          - os: windows-2019
            archs: "auto"
            skip: "*win_amd64*"
          - os: macos-10.15
            archs: "x86_64"
            skip: ""

    steps:
      ...
      - name: Build wheels
        env:
          CIBW_TEST_REQUIRES: "pytest"
          CIBW_TEST_COMMAND: "pytest --pyargs pybase64"
          CIBW_SKIP: "?p27* pp* ${{ matrix.skip }}"
        run: python -m cibuildwheel --archs ${{ matrix.archs }} --output-dir wheelhouse
```

mayeut:
>> ARCHS vs BUILD/SKIP: while all of this should be doable with BUILD/SKIP, I really think it's much clearer keeping those separate.
>
> What I meant was that --archs and --plat basically control which selectors are presented to BUILD/SKIP; and setting anything other than "auto" could be much slower via docker or emulated docker. So it seems natural that they are primarily command line variables, while BUILD/SKIP are only env variables.

@henryiii, sorry I misunderstood your comment.
In my opinion, wether it's available on command-line, environment variable or both shall only be a usability concern and I don't think that the fact it could be much slower when running emulation falls in that category. 
`--platform` will never be used on CI (unless you're using an unsupported CI provider). Thus, from a usability standpoint, it does not make much sense to add support for an environment variable for this parameter which could only cause confusion when running locally (at your own risks on macOS) for debugging.

mayeut:
Following-up on @henryiii's comment for clear warning that emulation is slow and @YannickJadoul's that there should be only one way of doing things:
Why not just an `--allow-emulation` flag and leave `BUILD/SKIP` filtering the architectures ?
If we're worried about warning the user about speed a bit more, `--allow-emulation` could take the same list of architecture as `--archs` with the exception that `auto`/`native` wouldn't be allowed (i.e. clear statement of what is allowed to run slow, `all` would be nice in this case). This wouldn't impact `windows.py` or `macos.py` but only `linux.py` to filter what's being fed into the build selector which will almost be the same as in this PR i.e `auto` + `--allow-emulation`

PS: I'm alright merging this PR as-is.

e.g. in GHA
```
jobs:
  build_wheels:
    name: Build ${{ matrix.build }} wheels on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        include:
          - os: ubuntu-18.04
            build: "*x86_64* *i686*"
          - os: ubuntu-18.04
            build: "*aarch64*"
          - os: ubuntu-18.04
            build: "*ppc64le*"
          - os: ubuntu-18.04
            build: "*s390x*"
          - os: windows-2019
            build: "*win_amd64*"
          - os: windows-2019
            build: "*win32*"
          - os: macos-10.15
            build: "*x86_64*"

    steps:
      ...
      - name: Build wheels
        env:
          CIBW_BUILD: "${{ matrix.build }}"
          CIBW_SKIP: "?p27* pp*"
        run: python -m cibuildwheel --allow-emulation all --output-dir wheelhouse
```

YannickJadoul:
> Following-up on @henryiii's comment for clear warning that emulation is slow and @YannickJadoul's that there should be only one way of doing things:
> Why not just an `--allow-emulation` flag and leave `BUILD/SKIP` filtering the architectures ?

I like that idea, yes.

Something tells me we visited it and rejected it straightaway, though? Was it because we definitely want to fail if the user specifies to build for an architecture when that architecture is not available in QEMU? We can still do that, though, since the combination of `--allow-emulation` and `CIBW_BUILD`/`CIBW_SKIP` still allows us to determine which architectures exactly the user expects to be built?

So I like the idea of trying this, but I can also see that @joerick, @henryiii, & @asfaltboy maybe just want to merge this PR and move on :-)

.github/workflows/test.yml
```diff
@@ -55,3 +55,24 @@ jobs:
     - name: Test cibuildwheel
       run: |
         python ./bin/run_tests.py
+
+  test-emulated:
+    name: Test emulated cibuildwheel using qemu
+    runs-on: ubuntu-latest
+    timeout-minutes: 180
```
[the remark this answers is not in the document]
joerick: This runs in parallel with the other runners, and takes less time (only 19m, because it's only  doing the emulation tests), so I don't think it costs us much

mayeut:
> Something tells me we visited it and rejected it straightaway, though? Was it because we definitely want to fail if the user specifies to build for an architecture when that architecture is not available in QEMU? We can still do that, though, since the combination of --allow-emulation and CIBW_BUILD/CIBW_SKIP still allows us to determine which architectures exactly the user expects to be built?

> So I like the idea of trying this, but I can also see that @joerick, @henryiii, & @asfaltboy maybe just want to merge this PR and move on :-)

I don't know if we visited that idea or not (too much catch-up to do in order to check this) so I'll let @joerick, @henryiii & @asfaltboy who probably know this just discard it right away if need be and/or, as you said, they might just want to merge this and be done with it :-)

bin/run_example_ci_configs.py
```diff
@@ -30,49 +31,67 @@ def generate_basic_project(path):
...
-        src_config_path='examples/gitlab-minimal.yml',
         dst_config_path='.gitlab-ci.yml',
         badge_md='[![Gitlab](https://gitlab.com/joerick/cibuildwheel/badges/{branch}/pipeline.svg)](https://gitlab.com/joerick/cibuildwheel/-/commits/{branch})'
     ),
 ]
 
 
+def ci_service_for_config_file(config_file):
+    for service in services:
+        if Path(config_file).name.startswith(service.name+'-'):
```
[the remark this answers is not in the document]
joerick: good idea 👍

YannickJadoul:
@mayeut 

I found the comments I was thinking of: https://github.com/joerick/cibuildwheel/pull/469#issuecomment-739547612, https://github.com/joerick/cibuildwheel/pull/469#issuecomment-739577500, https://github.com/joerick/cibuildwheel/pull/469#issuecomment-739582756, https://github.com/joerick/cibuildwheel/pull/469#issuecomment-740490633

Especially that latter one is relevant, I believe, with @joerick's argumentation.

cibuildwheel/__main__.py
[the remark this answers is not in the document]
joerick: I'm thinking the code might actually be clearer if we used plain old enums, rather than the StrEnum thing. Now we have type-checking, we end up having to be strict with them anyway. And `enum.value` is clearer and more explicit.

cibuildwheel/linux.py
```diff
@@ -93,7 +81,7 @@ def build(options: BuildOptions) -> None:
         exit(2)
 
     assert options.manylinux_images is not None
-    python_configurations = get_python_configurations(options.build_selector)
+    python_configurations = get_python_configurations(options.build_selector, options.architectures)
```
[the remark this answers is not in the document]
joerick: Yes, worth keeping in mind for future, but this works for the moment I think

cibuildwheel/macos.py
joerick: We have the option parsing unit tests, which verify what gets passed through to the build functions, so maybe that's covered? But you're right @asfaltboy , the exception isn't tested at all. Maybe I should add something.

> Personally, I'm pretty bad about forgetting to add tests for expected failures, and it has bit me a couple of times.

me too :)

joerick: that test added in 5ee663f

cibuildwheel/util.py
joerick: I considered this, but the enum error message says it all, to be honest

```
(env) joerick@joerick4 ~/P/cibuildwheel> CIBW_ARCHS=x8686 cibuildwheel --platform linux          
ValueError: 'x8686' is not a valid Architecture

During handling of the above exception, another exception occurred:

  File "/Users/joerick/Projects/cibuildwheel/env/bin/cibuildwheel", line 11, in <module>
    load_entry_point('cibuildwheel', 'console_scripts', 'cibuildwheel')()
  File "/Users/joerick/Projects/cibuildwheel/cibuildwheel/__main__.py", line 189, in main
    archs = Architecture.parse_config(archs_config_str, platform=platform)
  File "/Users/joerick/Projects/cibuildwheel/cibuildwheel/util.py", line 145, in parse_config
    result.append(Architecture(arch_str))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/enum.py", line 304, in __call__
    return cls.__new__(cls, value)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/enum.py", line 595, in __new__
    raise exc
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/enum.py", line 579, in __new__
    result = cls._missing_(value)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/enum.py", line 608, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 'x8686' is not a valid Architecture
```

cibuildwheel/util.py
joerick: i'm happy with either, but I agree with @mayeut's argument that `i686` isn't 'native' on an x86_64 machine, so `auto` makes more sense to me

cibuildwheel/util.py
joerick: yep

docs/faq.md
joerick: I think another emulator that integrated via `binfmt_misc` would work here too, though? (I don't know if such a thing exists, however!)

docs/faq.md
joerick: I don't think it's critical, but could be supplied in a follow-up PR if somebody wanted to add it?

docs/faq.md
YannickJadoul: Oh, that's true. Maybe if you look at it that way, then yes :-)

docs/options.md
joerick: Good point. I'm changing it to 

```
Comma-separated list of CPU architectures to build for.
When set to 'auto', builds the architectures natively supported
on this machine. Set this option to build an architecture
via emulation, for example, using binfmt_misc and QEMU. Default: auto. Choices: ...
```

henryiii:
What about changing `if options.architectures != [Architecture.AMD64, Architecture.x86]:` to `if set(options.architectures) <= {Architecture.AMD64, Architecture.x86}`? That does not seem to have been done, and the former is really not the right condition anyway, and it depends on the order. Maybe this should always be a set, in fact?

unit_test/main_tests/conftest.py
joerick: I just found the old approach a bit surprising, when I was using it! So I made it more explicit.

YannickJadoul:
> What about changing `if options.architectures != [Architecture.AMD64, Architecture.x86]:` to `if set(options.architectures) <= {Architecture.AMD64, Architecture.x86}`? That does not seem to have been done, and the former is really not the right condition anyway, and it depends on the order. Maybe this should always be a set, in fact?

That's what I suggested here: https://github.com/joerick/cibuildwheel/pull/482#discussion_r550220181. @mayeut was trying that out, I believe?

joerick:
> > Something tells me we visited it and rejected it straightaway, though? Was it because we definitely want to fail if the user specifies to build for an architecture when that architecture is not available in QEMU? We can still do that, though, since the combination of --allow-emulation and CIBW_BUILD/CIBW_SKIP still allows us to determine which architectures exactly the user expects to be built?
> 
> > So I like the idea of trying this, but I can also see that @joerick, @henryiii, & @asfaltboy maybe just want to merge this PR and move on :-)
> 
> I don't know if we visited that idea or not (too much catch-up to do in order to check this) so I'll let @joerick, @henryiii & @asfaltboy who probably know this just discard it right away if need be and/or, as you said, they might just want to merge this and be done with it :-)

The exact version of an `--emulate` flag you're describing (to _require_ all archs to be built, and the user deselects via CIBW_SKIP) hasn't quite been considered before, but I do think the ergonomics of it are a little tricky, and as @YannickJadoul says, were discussed in https://github.com/joerick/cibuildwheel/pull/469#issuecomment-740490633 .

(i'd quite like to be done with this too! 😄 😛)

mayeut:
>> What about changing if options.architectures != [Architecture.AMD64, Architecture.x86]: to if set(options.architectures) <= {Architecture.AMD64, Architecture.x86}? That does not seem to have been done, and the former is really not the right condition anyway, and it depends on the order. Maybe this should always be a set, in fact?
>
>That's what I suggested here: #482 (comment). @mayeut was trying that out, I believe?
>EDIT: +1 on the set, btw, though not very crucial. 

I have the changes for using a `set` on windows in a branch in my fork:
mayeut@b5e6207 + mayeut@3958a51

I can push those commits in the PR if required.

joerick:
> I have the changes for using a `set` on windows in a branch in my fork:
> [mayeut/cibuildwheel@b5e6207](https://github.com/mayeut/cibuildwheel/commit/b5e6207) + [mayeut/cibuildwheel@3958a51](https://github.com/mayeut/cibuildwheel/commit/3958a51)
> 
> I can push those commits in the PR if required.

I don't know about anyone else, but the discussion around this PR is getting a little confusing (and there is a lot of it!). Would you be amenable to merging this just supporting Linux, and we could open a new PR to extend support for Windows? We'd have docs updates to do for that too.

henryiii:
I'd be fine with that, but let's get #502 in and make a new release before we hit merge on this one!

mayeut:
> Would you be amenable to merging this just supporting Linux, and we could open a new PR to extend support for Windows? We'd have docs updates to do for that too.

@joerick, I'm fine with that option.

henryiii:
I'm ready for a merge when you are, @joerick :)

janaknat:
@henryiii Is this available in the latest PyPI cibuildwheel release?

henryiii:
Nope, only in master for the moment. We are working on Apple Silicon support and minor adjustments to the interface that affect `--archs` slightly. I don't think it's horribly far off, but for now, you can use a recent commit?

janaknat:
Ok. I'm looking to add aarch64 support for markupsafe using GHA and PyPI cibuildwheel. I'll give master a shot. Any ideas on when it'll be available in the PyPI version?

joerick:
It's a good question. #484 needs a bit more work - mostly docs - but is also waiting on a few upstream releases (packaging, pip and virtualenv), which could be two weeks or more away. Seems a bit of a shame to hold this for so long... options here-

1. wait for upstream and merge #484 without workarounds
2. merge #484 with workarounds and release soon (in the next week, perhaps)
3. release now without merging #484, since we pretty much know what's happening design-wise with CIBW_ARCHS now, and the way it works in this PR seems compatible with #484.

I'd be happy with (3), I think, and it would be the quickest route to getting this on PyPI. Any thoughts @YannickJadoul @henryiii ?

henryiii:
I think we can aim for emulation in 1.8, and Universal2 in 1.9, that sounds good to me. Leads up perfectly to 2.0 :)

I'd like to get the change to ARCH we discussed (auto, native, all) in before we release. I think #496 is ready, though doesn't really affect the release much, it would keep the diff down for #484, as it will be affected.

The other two changes I would have liked, but could be done later: the requires-python flag (since the sooner users start specifying that, the fewer changes will be needed when we up the "default" minimum), and the addition of `{}` syntax.

My wife is in the hospital for high blood pressure due to pregnancy; so any time in the next 7 weeks I might mostly drop out for a while.

YannickJadoul:
> * release now without merging #484, since we pretty much know what's happening design-wise with CIBW_ARCHS now, and the way it works in this PR seems compatible with #484.

2 notes on this:
- Does that include docs?
- Do we want to include a caveat about there being a (reasonably small) chance the archs part will still change?

henryiii:
Just pushed #535.

joerick:
> the change to ARCH we discussed (auto, native, all) in before we release

Ah yes, excellent point. I had forgotten about that.

> I think #496 is ready, though doesn't really affect the release much, it would keep the diff down for #484, as it will be affected. (See #528)

I'm just reviewing #496 now.

janaknat:
@joerick Any consensus on when the CIBW_ARCH option will be available on PyPI?

henryiii:
Current plan is "this week", I think we are still on target for that.

joerick:
Yes. Probably tomorrow :)

russkel:
:partying_face:

brettlangdon:
I just set this up on our project to try out before it is released. It appears to be working as intended!

One thing I noticed is that `aarch64` jobs take considerably longer than `x86_64`/`i686` builds.

https://github.com/DataDog/dd-trace-py/actions/runs/502345587

aarch64 on linux taking 24m
x86_64/i686 on linux taking 5m
x86_64/i686 on mac-os or windows about 10-15m


I do not have a ton of experience building on aarch64, is this expected or is there something I may have misconfigured?

Our GitHub action if it helps: https://github.com/DataDog/dd-trace-py/blob/4a456dd839d1c5bdb052a17f669561f479c5c3c9/.github/workflows/build_deploy.yml#L76-L110

We build on Python 2.7, 3.5, 3.6, 3.7, 3.8, 3.9

russkel:
Emulating another architecture will be quite slow, I don't imagine there's a huge amount of CPU power behind the GHA runners.

brettlangdon:
@russkel yeah, exactly what I was thinking as well. Didn't seem alarming, but figured I'd share 👍🏻

henryiii:
Emulating an arch is slow. For boost-histogram, it takes about 50 mins per Python version instead of 5-10 mins (unless I accidentally build NumPy, which takes more than 50 mins by itself), pushing me close to the 6 hour limit.

Cross compiling is fast, but support for that is pretty spotty in the Python Linux ecosystem currently. Maybe in the future cibuildwheel will be able to support that too, though it will always probably take a little work from the package as well. Setuptools burns in the wrong python executable for the console entry points, for example.

brettlangdon:
@henryiii good insights, it seems at least for our package 25m for all python versions on aarch64 isn't too bad then. Thanks!

kwankyu:
This is part of our github actions workflow:
```
  ...
  build_wheels:
    name: wheels ${{ matrix.build }}*_${{ matrix.arch }}
    runs-on: ${{ matrix.os }}
    needs: sdists_for_pypi
    strategy:
      fail-fast: false
      matrix:
        os:      [ubuntu-latest]
        arch:    [x86_64, i686, aarch64]
        build:   [manylinux, musllinux]
        include:
          - os: macos-13
            arch: x86_64
            build: macosx
          - os: macos-14
            arch: arm64
            build: macosx
    env:
      ...
      CIBW_BUILD: "*${{ matrix.build }}*"
      # Disable building PyPy wheels on all platforms
      CIBW_SKIP: "pp*"
      #
      CIBW_ARCHS: ${{ matrix.arch }}
      # https://cibuildwheel.readthedocs.io/en/stable/options/#requires-python
      CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9, <3.13"
      # Environment during wheel build
      CIBW_ENVIRONMENT: "PATH=$(pwd)/prefix/bin:$PATH ..."
      # Use 'build', not 'pip wheel'
      CIBW_BUILD_FRONTEND: build
    steps:
      - uses: actions/checkout@v4

      - name: Set up QEMU
        if: runner.os == 'Linux' && matrix.arch != 'x86_64' && matrix.arch != 'i686'
        uses: docker/setup-qemu-action@v3
        with:
          platforms: all
```
In our test run, this is seen:
```
  ...
  981e5eedeceb: Pull complete
  Digest: sha256:bf6f23641e14f3813b17e872a1d31ca6b1e9fd47be5be5f7a89cbbbe0051f0ee
  Status: Downloaded newer image for quay.io/pypa/musllinux_1_2_aarch64:2024-05-10-7415d48
  WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested
```
The "WARNING" seems harmless, and arm64 wheels are correctly built. But is there a way to avoid the warning? I also wonder if the warning is from docker or cibuildwheel...

PR #484, opened 2020-12-16
Universal2 wheels on macOS

### LINKED ISSUE #473: Support for macOS universal2 builds for ARM-based Macs
There is a plan to change installer from x86_64 to universal. Please see:
https://github.com/pypa/wheel/issues/387#issuecomment-738285687

joerick:
Add support for building `arm64` and `universal2` wheels on macOS 11+.

- [x] Building arm64 and universal2 on x86_64 runners using `CIBW_ARCHS`
- [x] Testing universal2 wheels
- [x] Docs update
   - [x] CIBW_ARCHS option
   - [x] 'How to' build for Apple Silicon, (probably including the nuances around testing)
- [x] Fix up CIBW_TEST_SKIP support for specifying parts of universal2 wheels
- [x] Examples update?

Closes #473.


joerick:
Hmm, due to https://github.com/actions/virtual-environments/issues/2247 this is gonna be slow going... it might be better to start work on this next week...

henryiii:
We will have to download the new images for [Python 3.9]((https://www.python.org/downloads/release/python-391/)) specifically `https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg` (which, regardless of the name, does support 10.9+). I think 3.8 may get a release as well.

I think the builds are a _bit_ slower, but still faster than Travis!

henryiii:
Do the 10.6 32/64 fat binaries not work on macOS 11? Is that perhaps the issue with 3.5? 2.7 does seem to be working, but not 3.5, if I read the log correctly?

Czaki:
> Do the 10.6 32/64 fat binaries not work on macOS 11? Is that perhaps the issue with 3.5? 2.7 does seem to be working, but not 3.5, if I read the log correctly?

https://discourse.jupyter.org/t/jupyter-notebook-error-dyld-library-not-loaded-corefoundation-after-macos-big-sur-upgrade/6838
https://stackoverflow.com/questions/64916693/jupyter-notebook-error-dyld-library-not-loaded-corefoundation-after-macos-big

So it looks like the bigger problem with all not patched python versions..

henryiii:
I'm not sure. It's not affecting 2.7, and Big Sur comes with 3.8.2 (CLT), so I think it's worth checking to see if 3.6 builds, anyway - maybe just reverse the order and do 3.9 first, then 3.8, etc. :) It might be that the Universal2 3.9 wheel will need to be in a separate job for now, anyway - but that shouldn't stop us from working out the 11.0 issues. @joerick let me know if / when it's okay to push to this branch. It's better if you do it for now, as I have to prepare to record some CMake training videos, but I'd like to help this along when you aren't working on it.

joerick:
> We will have to download the new images for Python 3.9 specifically `https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg` (which, regardless of the name, does support 10.9+ and Intel too). I think 3.8 may get a release as well.
> 
> I think the builds are a _bit_ slower, but still faster than Travis!

Hm, I wonder if you're talking about native ARM support, or the ability to build universal2 wheels? I was under the impression that `wheel` and the Xcode toolchain was cross-compiling both x86_64 and arm by default, regardless of the Python version?

I'll disable the 3.5 macOS build for now. Feel free to hack on this branch @henryiii, it just takes about half an hour for each macOS build to queue so it's not a lot of fun :)

henryiii:
PyPy is not macOS 11 ready either? At least all the normal wheels built. Though are they univesal2? Don't see a difference in naming.

joerick:
```
Building wheel for spam (setup.py): started
316
    Building wheel for spam (setup.py): finished with status 'done'
317
    Created wheel for spam: filename=spam-0.1.0-cp37-cp37m-macosx_10_9_x86_64.whl size=2382 sha256=efb4f386693223f8a943c5b60d7571318f0258eda5a7339992c62e957e40d493
318
    Stored in directory: 
```

Yeah, maybe you're right that we need the new Python macos11 builds

Czaki:
> PyPy is not macOS 11 ready either? At least all the normal wheels built. Though are they univesal2? Don't see a difference in naming.

In the current release, of the wheel depends on `get_platform` from `distutils.util`. 
Its output depends on the installer (it is hardcoded, not calculated). 

So without changing the installer platform tag will be x86. 

As I do not understand, there is no plan for changing this behavior.

henryiii:
```
  + delocate-listdeps /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheel_4pal_gg/built_wheel/spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl && delocate-wheel --require-archs x86_64 -w /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheel_4pal_gg/repaired_wheel /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibuildwheel_4pal_gg/built_wheel/spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl
```

Looks like that did the trick. :) How do we want to switch to using the new download? Always, with a flag, etc? Can you build an Intel (or Apple Silicon) wheel from the universal Python binary? If so, I don't see a reason to not always download it?

Also, PyPy and Python 3.5 don't support macOS 11, but I think that's okay - we could either recommend they be skipped for now until 11.0 becomes the default for macOS-latest on one of the CI systems, and by then, hopefully PyPy will have fixed their issues. @mattip is this a known issue with PyPy? Also will it be able to make Universal2 wheels too eventually? I feel like I have seen this in homebrew too for the 11.0 (but could be mixing it with Apple Silicon support - that I know is missing currently for PyPy). GitHub is pretty fast, so maybe macOS-latest will change sooner than I think, but I bet we have some time.

henryiii:
More wheel related, if you need a really new pip to use a universal2 wheel, I expect most projects will provide both for the near future - in which case, it seems much more sensible to provide a Intel and an Apple Silicon wheel instead of an Intel and a Universal2 wheel. So the question remains, can you set a target?

henryiii:
Also, can we silence all the silly warnings from pip? The Python 2 warning, the newer version available warning? We pin pip, so these warnings just get in the way when trying to find the real error.

henryiii:
And, final message tonight - do we have any 10.x builds left? Azure and GHA ran on 11, and I can't see CircleCI, so can't tell if this still works on 10.15 or if using the new Python broke it.

mattip:
> Also, PyPy and Python 3.5 don't support macOS 11

Is that this issue with `_ctypes` and `SCDynamicStoreCopyProxies` or are there more problems?

Czaki:
> Looks like that did the trick. :) How do we want to switch to using the new download? Always, with a flag, etc? Can you build an Intel (or Apple Silicon) wheel from the universal Python binary? If so, I don't see a reason to not always download it?

I do not know the method to do this. As far as I know, there is a need to check the Xcode version. Maybe @ronaldoussoren will know more. 

Maybe the best workaround will be updated delocate.

joerick:
perhaps you quoted the wrong text there, @Czaki ?

> Also, can we silence all the silly warnings from pip? The Python 2 warning, the newer version available warning? We pin pip, so these warnings just get in the way when trying to find the real error.

If we can be selective enough about the warnings, yes please :)

> I can't see CircleCI, so can't tell if this still works on 10.15 or if using the new Python broke it.

Wait, you can't see the CircleCI logs? I thought it was open-source. I do have to login with Github, maybe you need to do that?

To answer your question, Circle's mac builds are still on xcode:9.4.1, macOS 10.13. Which makes this test failure all the more interesting:

```
>       assert set(actual_wheels) == set(expected_wheels)
E       AssertionError: assert {'spam-0.1.0-..._64.whl', ...} == {'spam-0.1.0-..._64.whl', ...}
E         Extra items in the left set:
E         'spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl'
E         Extra items in the right set:
E         'spam-0.1.0-cp39-cp39-macosx_10_9_x86_64.whl'
E         'spam-0.1.0-cp35-cp35m-macosx_10_9_x86_64.whl'
E         'spam-0.1.0-pp27-pypy_73-macosx_10_9_x86_64.whl'
E         Use -v to get the full diff
```

To me, that looks like it produced a universal2 wheel, even though it was built on xcode9 which almost certainly doesn't support Apple Silicon! Wat.

> More wheel related, if you need a really new pip to use a universal2 wheel, I expect most projects will provide both for the near future - in which case, it seems much more sensible to provide a Intel and an Apple Silicon wheel instead of an Intel and a Universal2 wheel. So the question remains, can you set a target?

Did a little searching though the CPython source and found `_PYTHON_HOST_PLATFORM`: https://github.com/python/cpython/blob/master/Lib/distutils/util.py#L47 (thanks @Czaki for the pointer). And I remembered that [we already do this on CPython 3.5](https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/macos.py#L171), I think @mayeut figured it out. So I guess the same approach would work for us.

If that works, we might as well select the universal2 installer where it's available.

Czaki:
@joerick I think I quote too long text. I comment mainly:

> Can you build an Intel (or Apple Silicon) wheel from the universal Python binary?

cibuildwheel/macos.py
```diff
@@ -35,13 +35,18 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
...
         PythonConfiguration(version='3.6', identifier='cp36-macosx_x86_64', url='https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.9.pkg'),
         PythonConfiguration(version='3.7', identifier='cp37-macosx_x86_64', url='https://www.python.org/ftp/python/3.7.8/python-3.7.8-macosx10.9.pkg'),
         PythonConfiguration(version='3.8', identifier='cp38-macosx_x86_64', url='https://www.python.org/ftp/python/3.8.4/python-3.8.4-macosx10.9.pkg'),
-        PythonConfiguration(version='3.9', identifier='cp39-macosx_x86_64', url='https://www.python.org/ftp/python/3.9.0/python-3.9.0-macosx10.9.pkg'),
+        # TODO: Find some better way to select Universal2 vs. regular (note that this works on 10.9+, regardless of the name)
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_x86_64', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_universal2', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
```
henryiii, line 45:
```suggestion
        PythonConfiguration(version='3.9', identifier='cp39-macos_universal2', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
```

henryiii: Maybe the same for ARM?

cibuildwheel/macos.py
```diff
@@ -35,13 +35,18 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
...
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_arm64', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
```
henryiii, line 46:
```suggestion
        PythonConfiguration(version='3.9', identifier='cp39-macos_arm64', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
```

cibuildwheel/macos.py
```diff
@@ -175,6 +180,14 @@ def setup_python(python_configuration: PythonConfiguration,
         # https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260
         env.setdefault('ARCHFLAGS', '-arch x86_64')
 
+    if python_configuration.version == '3.9':
+        if python_configuration.identifier.endswith('x86_64'):
+            env.setdefault('_PYTHON_HOST_PLATFORM', 'macosx-10.9-x86_64')
+            env.setdefault('ARCHFLAGS', '-arch x86_64')
+        elif python_configuration.identifier.endswith('arm64'):
+            env.setdefault('_PYTHON_HOST_PLATFORM', 'macosx-11.0-arm64')
```
henryiii, line 188:
```suggestion
            env.setdefault('_PYTHON_HOST_PLATFORM', 'macos-11.0-arm64')
```

henryiii:
Looks like the x is not removed everywhere: https://github.com/pypa/packaging/blob/353f270a4e49283c8e98ed8c801b457665774525/packaging/tags.py#L457

henryiii:
So can we download the wheels to try them? I can find someone with a Apple Silicon machine to test them. Looks like we'll need to turn off all PyPy, not just PyPy2.

cibuildwheel/macos.py
```diff
@@ -35,13 +35,18 @@ def get_python_configurations(build_selector: BuildSelector) -> List[PythonConfi
...
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_x86_64', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_universal2', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
+        PythonConfiguration(version='3.9', identifier='cp39-macosx_arm64', url='https://www.python.org/ftp/python/3.9.1/python-3.9.1-macos11.0.pkg'),
         # PyPy
-        PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='https://downloads.python.org/pypy/pypy2.7-v7.3.3-osx64.tar.bz2'),
+        # TODO: may not support 11.0 yet
+        # PythonConfiguration(version='2.7', identifier='pp27-macosx_x86_64', url='https://downloads.python.org/pypy/pypy2.7-v7.3.3-osx64.tar.bz2'),
         PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='https://downloads.python.org/pypy/pypy3.6-v7.3.3-osx64.tar.bz2'),
         PythonConfiguration(version='3.7', identifier='pp37-macosx_x86_64', url='https://downloads.python.org/pypy/pypy3.7-v7.3.3-osx64.tar.bz2'),
```
henryiii, line 51:
```suggestion
        # PythonConfiguration(version='3.6', identifier='pp36-macosx_x86_64', url='https://downloads.python.org/pypy/pypy3.6-v7.3.3-osx64.tar.bz2'),
        # PythonConfiguration(version='3.7', identifier='pp37-macosx_x86_64', url='https://downloads.python.org/pypy/pypy3.7-v7.3.3-osx64.tar.bz2'),
```

joerick:
Looks like this is working as we expect so far...

```
>       assert set(actual_wheels) == set(expected_wheels)
E       AssertionError: assert {'spam-0.1.0-..._64.whl', ...} == {'spam-0.1.0-..._64.whl', ...}
E         Extra items in the left set:
E         'spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl'
E         'spam-0.1.0-cp39-cp39-macosx_11_0_arm64.whl'
E         Extra items in the right set:
E         'spam-0.1.0-pp27-pypy_73-macosx_10_9_x86_64.whl'
E         'spam-0.1.0-pp37-pypy37_pp73-macosx_10_9_x86_64.whl'
E         'spam-0.1.0-pp36-pypy36_pp73-macosx_10_9_x86_64.whl'
E         'spam-0.1.0-cp35-cp35m-macosx_10_9_x86_64.whl'
E         Use -v to get the full diff
```

So we have built wheels that claim to be universal2 and arm64, and we've built an x86_64 wheel from the universal2 installer.

henryiii:
Can the wheels be posted as artifacts? I know someone with a Apple Silicon machine and I _might_ be able to get them to test it if we have something I can send him.

mayeut:
> Looks like this is working as we expect so far...

Nice.

I had a quick look at the logs and if you look at the `delocate` command line, you'll always see `--require-archs x86_64`. It only checks dependencies that are grafted and since there are none, there's no error for `arm64`. `delocate` has no knowledge of `universal2` so `x86_64,arm64` should probably be used in that case. This probably requires a new test with a grafted library (probably both on linux and macOS) as I think there's none at the moment.

henryiii:
[Interesting](https://pythoninsider.blogspot.com/2020/12/python-387-is-now-available.html) - under notes it mentions a few things that do not work through Rosetta 2 (finding shared libraries, mainly).

joerick:
I've actually received my M1 Macbook, so I can test these wheels on my machine! Hopefully this build should upload something...

joerick:
Okay, so I have some results. It wasn't exactly _easy_ to find the universal2 installer for macOS, but I found it on https://www.python.org/downloads/mac-osx/ .

```shell
$ python -c "import platform; print(platform.machine())"     
arm64
$ pip --version                          
pip 20.3.3 from /private/tmp/env/lib/python3.9/site-packages/pip (python 3.9)
```

So first up, the universal2 wheel.

```shell
$ pip install spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl
ERROR: spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl is not a supported wheel on this platform.
```

Hmm. Strange... but I did wonder why the wheel had a 10_9 platform tag. I tried renaming it...

```shell
$ cp spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl spam-0.1.0-cp39-cp39-macosx_11_0_universal2.whl
$ pip install spam-0.1.0-cp39-cp39-macosx_11_0_universal2.whl 
Processing ./spam-0.1.0-cp39-cp39-macosx_11_0_universal2.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ python -c 'import spam; spam.system("echo it works!")'
it works!
```

And, as we have a universal2 wheel, so I can try the x86_64 part.

```shell
$ arch -x86_64 python -c "import platform; print(platform.machine())"
x86_64
$ arch -x86_64 python -c 'import spam; spam.system("echo it works!")'
it works!
```

Finally, we can try the arm64 wheel.

```shell
$ pip uninstall spam
Found existing installation: spam 0.1.0
Uninstalling spam-0.1.0:
  Would remove:
    /private/tmp/env/lib/python3.9/site-packages/spam-0.1.0.dist-info/*
    /private/tmp/env/lib/python3.9/site-packages/spam.cpython-39-darwin.so
Proceed (y/n)? y
  Successfully uninstalled spam-0.1.0
$ pip install spam-0.1.0-cp39-cp39-macosx_11_0_arm64.whl 
Processing ./spam-0.1.0-cp39-cp39-macosx_11_0_arm64.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ python -c 'import spam; spam.system("echo it works!")'             
it works!
```

But, of course, this one doesn't work in x86_64:

```shell
$ arch -x86_64 python -c 'import spam; spam.system("echo it works!")'
ImportError: dlopen(/private/tmp/env/lib/python3.9/site-packages/spam.cpython-39-darwin.so, 2): no suitable image found.  Did find:
	/private/tmp/env/lib/python3.9/site-packages/spam.cpython-39-darwin.so: mach-o, but wrong architecture
	/private/tmp/env/lib/python3.9/site-packages/spam.cpython-39-darwin.so: mach-o, but wrong architecture
```

So that's good news! But it does make me wonder - why did I need to rename the universal2 wheel to 11_0? @ronaldoussoren, any ideas?

mayeut:
> So that's good news! But it does make me wonder - why did I need to rename the universal2 wheel to 11_0? @ronaldoussoren, any ideas?

Can you try to run `python -m pip debug -v | grep -A 10 'Compatible tags'` with both archs to check what's going on ?

henryiii:
Also, since I use homebrew [exclusively to manage software](https://iscinumpy.gitlab.io/post/setup-a-new-mac/) on my mac, and [homebrew's ARM](https://twitter.com/fxcoudert/status/1342602929585451008) bottles are not universal, so might be a good idea check brew's Python too, just as a sanity check?

henryiii:
You'll need more than 10 lines to see past 11_0/10_16, I believe.

joerick:
I haven't got a homebrew install right now, but here's `python -m pip debug -v` on my current setup, for arm64 and x86_64:

https://gist.github.com/joerick/a6887e118973459fb192528c444457ef

mayeut:
This seems like a "bug/feature" in `pip`/`packaging`...
`pip` expects things that run on `arm64` to start at macOS 11.0
In fact this `universal2` package indeed requires macOS 11.0 on `arm64` but only 10.9 on `x86_64`...
This shall probably be reported to `pip`/`packaging`. In the meantime, what happens if you rename the wheel `spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl` ?

mattip:
The ability to install x86_64 10.9 wheels on 11 is considered a feature, not a bug. See pypa/packaging#319. I don't know if they considered a universal2 wheel that does not contain arm64 binaries.

mayeut:
> The ability to install x86_64 10.9 wheels on 11 is considered a feature, not a bug.

I wasn't clear I guess.

`macosx_10_9_universal2`:
- installs fine on `x86_64` for macOS 10.9+ (including 11+, this is expected)
- fails to install on `arm64` on macOS 11.0 as mentioned by @joerick (it shall probably not fail, here is the bug)

`macosx_11_0_universal2`:
- installs fine on `x86_64` for macOS 11.0+
- installs fine on `arm64` for macOS 11.0+ 

So, as I  see it `universal2` lacks its "universal" purpose because you'll need 2 `universal2` wheels to be able to install on `x86_64` pre-11 and `arm64` at the same time.
Or maybe just renaming the wheel `macosx_10_X_universal2. macosx_11_0_universal2` when you target pre-11 on `x86_64` would be enough (but seems like a hugly hack). Or am I missing something ?

I think the `arm64` tag list from `packaging` has been built taking into consideration only `arm64` (i.e. start at 11) or, more exactly, the list is extended when you run on 11.0+ & `x86_64`:
https://github.com/pypa/packaging/blob/353f270a4e49283c8e98ed8c801b457665774525/packaging/tags.py#L461-L473

While it makes no sense to look for `arm64` before macOS 11, it does make sense to look for `universal2` wheels before that, meaning they'll be `10_X` on `x86_64` and "first version with support" for `arm64`. Anyway, this is a discussion to be held  over at `pip`/`packaging`. As I said, not sure if this one is a bug or a feature...

Czaki:
@ronaldoussoren make a patch for `wheel` that `universal2` wheel should have platform tag from `x86_64` part if platform tag from `arm` part is `11.0`. So for most basic wheel files expected format should be `macosx_10_9_universal2` even if `arm` part needs macOS in version 11.0.

henryiii:
The patch mentioned by @Czaki is here: https://github.com/pypa/packaging/pull/380

henryiii:
A thought for now would be to suggest that libraries might be to provide a `macosx_10_9_x86_64` wheel and a `macosx_11_0_universal2` one. (That would break if https://github.com/pypa/packaging/issues/381 was implemented; zipped apps created on macOS 11 x86 would not work on older macOS's).

Czaki:
I think about this https://github.com/pypa/wheel/pull/390

ronaldoussoren:
Providing "macosx_10_9_universal2" wheels for the univeral2 variant would IMHO be preferable, those can in the long run replace architecture specific wheels and match the set of systems supported by the Python.org installers for macOS.

Currently pip doesn't recognise "macosx_10_9_universal2" as an acceptable alternative for M1 systems, I've filed https://github.com/pypa/packaging/issues/379 about that with a corresponding PR. I'm hopeful that this PR will get merged.

Note that the current plan for Python 3.10 is to have only a "universal2" installer and drop the architecture specific one, that brings us back to having a single installer for macOS which is a lot clearer for our users.   The "universal2" installer for Python 3.9 is experimental for two reasons: (1) teething problems in the wider ecosystem (for various reasons we haven't been as proactive as I would have liked with helping out the various packaging related projects), and (2) we're also changing the way the installer is build and that results in some instabilities (in particular with Tcl/Tk).

ronaldoussoren:
> I think about this [pypa/wheel#390](https://github.com/pypa/wheel/pull/390)

In what way? That change was the right thing to do: A "macosx_10_9_universal2" can be used on every macOS system running macOS 10.9 or later.

ronaldoussoren:
> [Interesting](https://pythoninsider.blogspot.com/2020/12/python-387-is-now-available.html) - under notes it mentions a few things that do not work through Rosetta 2 (finding shared libraries, mainly).

The main problem (and IIRC the only problem with the Python.org installers) is that the find_library function in ctypes cannot find shared libraries on Big Sur (macOS 11) in general. That's not M1 related, but is due to an architecture change in Big Sur: on that system system shared libraries are no longer present at their usual place in the filesystem (no /usr/lib/libSystem.dylib), but are merged into a "shared library cache".   CTypes assumes that shared libraries are actually in the file system.   For 3.9 and 3.10 we have a fix for that (when you build with Xcode 12 or later).

Other than that Python 3.8 (or earlier) will not build ctypes correctly on M1 systems, there are several changes needed to support the calling convention for M1 systems. Again, fixes for this are in 3.9 and 3.10.

These changes will end up in a future release of 3.8 as well, the plan was to have it in the current release but back porting took more time and energy than expected.

joerick:
Thank you @ronaldoussoren for chiming in and confirming what we suspected! Hopefully pypa/packaging#380 will be merged soon, and we'll be able to get an updated `pip` that accepts `10_9_universal2`.

> @mayeut: In the meantime, what happens if you rename the wheel `spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl` ?

```shell
$ pip uninstall spam
Found existing installation: spam 0.1.0
Uninstalling spam-0.1.0:
  Would remove:
    /private/tmp/env/lib/python3.9/site-packages/spam-0.1.0.dist-info/*
    /private/tmp/env/lib/python3.9/site-packages/spam.cpython-39-darwin.so
Proceed (y/n)? y
  Successfully uninstalled spam-0.1.0
$ cp spam-0.1.0-cp39-cp39-macosx_10_9_universal2.whl spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
$ pip install spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Processing ./spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ python -c 'import spam; spam.system("echo it works!")'                            
it works!
```

Indeed, renaming to list both tags in seems to fix the issue. I think we can move forward using this rename hack, for now, so we're not delaying this further waiting for a pip release.

Czaki:
@ronaldoussoren I comment this comment https://github.com/joerick/cibuildwheel/pull/484#issuecomment-753102135, that expected tag will be `macosx_10_9_universal2`

ronaldoussoren:
> Thank you @ronaldoussoren for chiming in and confirming what we suspected! Hopefully [pypa/packaging#380](https://github.com/pypa/packaging/pull/380) will be merged soon, and we'll be able to get an updated `pip` that accepts `10_9_universal2`.

I'd also like to see this because I'm migrating to an M1 laptop for my daily driver :-).    A review by someone in the packaging space could help here, my PR seems to be blocked on review. Not that I'm complaining, we're all volunteers and at least some folks are offline for the holidays.

> 
> > @mayeut: In the meantime, what happens if you rename the wheel `spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl` ?

What I did for now is to patching the copy of packaging in pip._vendored using my PR for this, that way pip does the right thing. The disadvantage is of course that this is a manual step that needs to be repeated every time pip gets installed in virtual environments.

joerick:
Currently not able to work on this, because the Github Actions builds on `macos-11.0` are stalled, saying 'Starting your workflow run...'. Anyone know if there's a way around this?

henryiii:
They might not have an unlimited number of 11.0 machines, and they may work on them a bit more often - it's still provisional. But I think it clears up eventually; you can always cancel and restart. There's a cancel-action I keep meaning to try that will cancel when you push a new commit.

henryiii:
Packaging PR was merged. :)

joerick:
I've done more work here, doing some testing locally. Still not getting much from the GHA `macos-11.0` runner. I've reported it to https://github.community/t/macos-11-0-big-sur-builds-stalling-out/153761 .

.github/workflows/test.yml
```diff
@@ -52,6 +53,16 @@ jobs:
       run: |
         python bin/sample_build.py
 
+    - name: Get some sample wheels
+      run: |
+        python -m test.test_projects test.test_0_basic.basic_project sample_proj
+        cibuildwheel --output-dir wheelhouse sample_proj
+
+    - uses: actions/upload-artifact@v2
```
henryiii, line 61: There shouldn't be any harm in always having this, I'd think.

cibuildwheel/__main__.py
```diff
@@ -308,7 +308,7 @@ def print_build_identifiers(
     elif platform == 'windows':
         python_configurations = cibuildwheel.windows.get_python_configurations(build_selector, architectures)
     elif platform == 'macos':
-        python_configurations = cibuildwheel.macos.get_python_configurations(build_selector)
+        python_configurations = cibuildwheel.macos.get_python_configurations(build_selector, architectures)
```
henryiii, line 311: We _could_ just use `module =getattr(cibuildwheel, platform)`, then `python_configurations = platform.get_python_configurations(build_selector, architectures)` now that they all match, and drop the whole if, if you wanted. (or some more static version of that with a dict or simpler if).

cibuildwheel/macos.py
```diff
@@ -251,60 +311,140 @@ def build(options: BuildOptions) -> None:
...
-                # run the tests from $HOME, with an absolute path in the command
-                # (this ensures that Python runs the tests against the installed wheel
-                # and not the repo code)
-                test_command_prepared = prepare_command(
-                    options.test_command,
-                    project=Path('.').resolve(),
-                    package=options.package_dir.resolve()
-                )
-                call(test_command_prepared, cwd=os.environ['HOME'], env=virtualenv_env, shell=True)
+            if repaired_wheel.stem.endswith('_universal2'):
```
henryiii, line 334: Should we have a flag or something to disable this? Or even disable it by default, with a flag to enable? Seems like this is expected to be a very temporary workaround.

henryiii:
I think we need to have a way to disable emulation testing, maybe even have it off by default. Due to the fact things don't always work correctly in emulation (loading shared libraries might be part of a wheel, and I think that was listed as unsupported, and AVX* is unsupported), you might get a "broken" emulation test when the wheel is just fine. A flag like `CIBW_MACOSARCH_TEST` or something?

I think we could have universal 2 in the "auto" set, since we can test half the wheel, if you want to. It should be (if things work well) really hard to make a universal wheel that is half broken. I think testing the cross-compiled parts should probably always be opt in.

joerick:
Oh wow, the tests finally passed! No docs yet, but I think this is ready for an initial review. @henryiii is quick off the mark :)

joerick:
> I think we need to have a way to disable emulation testing, maybe even have it off by default. Due to the fact things don't always work correctly in emulation (loading shared libraries might be part of a wheel, and I think that was listed as unsupported, and AVX* is unsupported), you might get a "broken" emulation test when the wheel is just fine. A flag like CIBW_MACOSARCH_TEST or something?

Rosetta 2 in my experience is _very_ good, like _I haven't had a single problem_ good, and it's surprisingly fast, so I'm wondering if we just wait and see? The AVX* stuff should be detected at runtime, I think, and I believe the shared library thing you mentioned was a macOS Big Sur problem, not a Rosetta problem. (Might be wrong there though. )

But I can definitely see the logic in something like CIBW_TEST_ARCHS with options like `native`, `auto`, `<specific arch>`, like we've discussed for CIBW_ARCHS in #516. But also,  So it's good to have the design in a back pocket, so if a real use-case is reported where somebody needs to turn it off, we know how to implement it. 

> I think we could have universal 2 in the "auto" set, since we can test half the wheel, if you want to. It should be (if things work well) really hard to make a universal wheel that is half broken. I think testing the cross-compiled parts should probably always be opt in.

I'm not sure I agree... the defaults should do the 'right' thing, and I believe the right thing is to test the entire wheel. What's motivating this - are you concerned about increased build/test durations to run the tests twice?

It will take time for the Python community to move to building arm64-compatible stuff, I expect lots of people to running in Rosetta 2 so some test coverage there might be helpful - even if it's just checking your SIMD detection is good. Of course, all this relies on arm64 CI runners, and we have no idea when they might arrive :)

cibuildwheel/__main__.py
joerick: I think it's mostly good fortune that they match right now, so I think I'd rather spell it out. Plus mypy probably groks it better this way.

cibuildwheel/macos.py
joerick: It's pretty harmless, but even so, I wasn't expecting to merge including this. We're waiting for a pip release that includes https://github.com/pypa/packaging/pull/380 ... so we'd need a packaging/tags release and a pip release it looks like pip is getting close to their 21.0 release, which is slated for some time January 2021. I don't know their policy for updating the vendored dependencies though - do you know if they'll normally bump packaging before an update?

cibuildwheel/macos.py
```diff
@@ -251,60 +311,140 @@ def build(options: BuildOptions) -> None:
...
+                # See https://github.com/pypa/packaging/pull/380 and
+                # https://github.com/pypa/packaging/issues/379
+                renamed_wheel = repaired_wheel.parent / f'{repaired_wheel.stem}.macosx_11_0_universal2.whl'
+                repaired_wheel.rename(renamed_wheel)
+                repaired_wheel = renamed_wheel
 
-                # clean up
-                shutil.rmtree(venv_dir)
+            log.step_end()
+
```
joerick, line 345: p.s. this next bit is easier to read if you 'Hide whitespace changes' in Github

test/test_macos_archs.py
```diff
@@ -0,0 +1,69 @@
...
+from . import test_projects, utils
+
+basic_project = test_projects.new_c_project()
+
+ALL_MACOS_WHEELS = (
+    utils.expected_wheels('spam', '0.1.0', machine_arch='x86_64')
+    + utils.expected_wheels('spam', '0.1.0', machine_arch='arm64')
+)
+
+
```
joerick, line 14: By the way, I've been able to run these tests locally on my MacBook Air, and they pass on arm64 too. But I don't think the next release of cibuildwheel will officially include arm64 host support, because we can't run the entire test suite on arm64 yet. We might need to do a few tweaks once arm64 CI is available before declaring full support. The arm64 parts of this PR are mostly to nail down a design and show the intention.

henryiii:
As it is now, I often have to turn off testing on some jobs due building more wheels than upstream requirements (NumPy, generally). It's not that big of an issue to ship with wheel like NumPy missing further up the chain; many users get NumPy from other locations (package mangers, conda, etc). But it's already irritating that I have to turn off all testing when NumPy does support 1 or 2 in the set (PyPy 3.6, or some specific ARM one). I could split the build even further, so that I can have "testable" vs. "not testable". I almost see the use case for `CIBW_TEST_SKIP=`pp37* ...`. :) But if we do dual-testing on Universal2, that would likely force users to skip all testing if something was not supported, instead of testing half. In fact, if we got ARM runners tomorrow, we might only be able to test on the x86_64 part for some packages! Maybe by the time we get ARM runners, this won't be so common, but some major packages are rather slow to add new wheels. (NumPy. NumPy. NumPy.)

I want to be sure we have enough flexibility to control dual testing if we need to, since a half-tested wheel is 90% better than a completely untested one. Testing wheels often isn't really to ensure everything works, but just to make sure the build worked - normal CI usually tests that everything works.

Since this really only affects ARM runners, it's likely not that big of a deal yet, and may never be. Unless GHA or Azure already support local ARM runners, in which case I might be setting one up when I get my Mini. :)

henryiii:
Since it's cross-testing Intel on ARM, I'm okay with opt-out, or even eventual opt-out if needed. The other direction would be much more important to be opt-in, but it's not available or planned, IFAIK.

henryiii:
For reference, looks like you can make it work with some hacks on local runners: https://github.com/actions/runner/issues/805

joerick:
Thanks for the explanation @henryiii, that helps a lot! I hadn't considered the difficulties of testing when some of your test dependencies are hard to get hold of. So you have projects where your build doesn't depend on e.g. NumPy, but the tests do?

> I want to be sure we have enough flexibility to control dual testing if we need to, since a half-tested wheel is 90% better than a completely untested one. Testing wheels often isn't really to ensure everything works, but just to make sure the build worked - normal CI usually tests that everything works.

Good point.

Taken at face value, this sounds like there's a rationale for both `CIBW_TEST_SKIP` - to skip testing on awkward platforms, and `CIBW_TEST_ARCHS(_MACOS)` - to control the dual-testing of universal2 wheels. Honestly, despite the increase in API surface area, that doesn't sound too crazy to me, given how closely these would mirror the `CIBW_SKIP` and `CIBW_ARCHS`. They wouldn't affect regular users at all, but might come in handy for power users. But I don't want to be too hasty, let's see if @YannickJadoul or @mayeut have ideas. 

Also, so long as we agree on what should happen by default, adding these options could easily come after this PR, especially since (as you say) it's all irrelevant on x86_64 - you can only test the x86_64 bit.

henryiii:
I could easily imaging a custom option just for dual Universal2 testing, as it's quite special - we don't run tests twice anywhere else, and it would be only on the ARM runners.

Yes, pybind11 does not require NumPy headers to build with `pybind11/numpy.py` (which is somewhere between handy and fantastic, depending on how much you like intricate pinning NumPy in pyproject.toml). There are probably other cases, but that's the one that tends to bite me most often (NumPy doesn't build out-of-the-box on macOS, too, just to make it fun).

henryiii:
Regardless of what spelling and structure you decide on for #529, I think "if `TESTING_*` allows the matching arm64 identifier, run the arm universal2 test, and if it allows the x86_64 identifier, run the x86 test" is the right way to go here; it avoids a new special variable for universal2 testing and is likely flexible enough for all reasonable usage. We could still get this in first always testing all archs, then do a PR working on the resolution to `#529`.

henryiii:
Pip 21 just was released with no packaging update. https://github.com/pypa/packaging/issues/385

cibuildwheel/macos.py
```diff
@@ -34,21 +37,63 @@ def call(args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: O
...
+    # skip builds as required by BUILD/SKIP
+    python_configurations = [c for c in python_configurations if build_selector(c.identifier)]
+
+    # When running on macOS 11 and x86_64, the reported OS is '10.16', but
+    # there is no such OS - it really means macOS 11.
+    if get_macos_version() >= (10, 16):
+        if any(c.identifier.startswith('pp') for c in python_configurations):
+            # pypy doesn't work on macOS 11 yet
+            # See https://foss.heptapod.net/pypy/pypy/-/issues/3314
+            log.warning(unwrap('''
```
henryiii, line 80: Should this be a warning or an error? Is someone really going to go the the logs and look for these warnings?

cibuildwheel/macos.py
```diff
@@ -34,21 +37,63 @@ def call(args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: O
...
+                PyPy is currently unsupported when building on macOS 11. To build macOS PyPy wheels,
+                build on an older OS, such as macOS 10.15. To silence this warning, deselect PyPy by
+                adding "pp*-macosx*" to your CIBW_SKIP option.
+            '''))
+            python_configurations = [c for c in python_configurations if not c.identifier.startswith('pp')]
+
+        if any(c.identifier.startswith('cp35') for c in python_configurations):
+            # CPython 3.5 doesn't work on macOS 11
+            log.warning(unwrap('''
```
henryiii, line 89: This one doesn't bother me as a warning as much. I think the only place Python 3.5 is still around is in package managers for Linux.

cibuildwheel/util.py
```diff
@@ -18,13 +18,7 @@
...
 from .typing import PathOrStr, PlatformName
 
-if sys.version_info < (3, 9):
-    from importlib_resources import files
-else:
-    from importlib.resources import files
-
-
-resources_dir = files('cibuildwheel') / 'resources'
+resources_dir = Path(__file__).parent / 'resources'
```
henryiii, line 21: Why this direction? This was the "correct" language to access resources using the stdlib (and backport in <3.9). It follows the standard import mechanisms instead of relying on manual path manipulations. We don't load Python files by path, why load data that way?

And I believe the rewritten version is misleading as it stands, without a `.resolve()` it doesn't work from a main file before 3.{8,9,10}, don't remember exactly which one they started always resolving the `__file__` in to protect against changed directories.

cibuildwheel/macos.py
```diff
@@ -34,21 +37,63 @@ def call(args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: O
...
-    # skip builds as required
-    return [c for c in python_configurations if build_selector(c.identifier)]
+    # filter out configs that don't match any of the selected architectures
+    python_configurations = [c for c in python_configurations
+                             if any(c.identifier.endswith(a.value) for a in architectures)]
+
+    # skip builds as required by BUILD/SKIP
+    python_configurations = [c for c in python_configurations if build_selector(c.identifier)]
+
+    # When running on macOS 11 and x86_64, the reported OS is '10.16', but
```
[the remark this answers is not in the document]
YannickJadoul: That's how version works, right? You pick which version of the version you want? _sigh_

cibuildwheel/util.py
joerick: I couldn't get mypy to accept the importlib_resources version, for some reason. So I put it back to the way it used to be. Btw, I don't think that we can get away from the fact that resources are files, since we get their paths to pass them to subprocesses, or we run them from the filesystem like `get-pip`. I'd rather use simpler concepts like Paths if there's no reason not to.

cibuildwheel/macos.py
joerick: GHA pulls them up to the summary screen as 'annotations', but I take your point. Trouble is, if this is an error then we can't run without an explicit `CIBW_SKIP=pp*-macosx*`, which is a very strange first-run journey / minimal config.

cibuildwheel/architecture.py
```diff
@@ -46,19 +52,26 @@ def parse_config(config: str, platform: PlatformName) -> 'Set[Architecture]':
...
+
         return result
 
     @staticmethod
     def all_archs(platform: PlatformName) -> 'Set[Architecture]':
         if platform == 'linux':
             return {Architecture.x86_64, Architecture.i686, Architecture.aarch64, Architecture.ppc64le, Architecture.s390x}
         elif platform == 'macos':
-            return {Architecture.x86_64}
+            return {Architecture.x86_64, Architecture.arm64, Architecture.universal2}
```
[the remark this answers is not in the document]
joerick: I don't know... my gut is that it shouldn't depend on macOS version, because the new archs also require CPython 3.9, so we should localise that logic in `get_python_configurations`, not spread it out over the codebase.

joerick:
> Pip 21 just was released with no packaging update. [pypa/packaging#385](https://github.com/pypa/packaging/issues/385)

I guess we're shipping the rename hack then! To be honest, it's pretty harmless anyway.

cibuildwheel/util.py
henryiii: > rather use simpler concepts

I still think it's better to use the correct concept, rather than propagating the idea that it's okay to manipulate paths inside a package. I'm find with path manipulations in `bin`, but I don't like them in `cibuildwheel`.

> since we get their paths to pass them to subprocesses, or we run them from the filesystem like get-pip.

importlib.resouces has a method for doing this, though it's one extra step. If you add a context manager, it will promise a real path (to somewhere, either the actual location or a temporary location for some loaders) for the duration. I didn't go through and update the usages of pure pathnames for this - but in theory, if we did, we could then make a cibuldwheel.pyz zip file that could be run from anywhere without an internet connection, for example.

Because we aren't careful with this, it's mostly a conceptual improvement for the moment, so if you wanted to revert it for now, but ideally we could do it correctly some day. cibuildwheel does have the distinction of simply needing to run on CI rather than absolutely everywhere.

> I couldn't get mypy to accept the importlib_resources version, for some reason

The MyPy improvements in #536 touch on this partially. It's generally a little odd for backports, since MyPy assumes the target Python version but usually runs on the latest Python version.

cibuildwheel/macos.py
henryiii: Yes, but it means if you simply upgrade from macOS-10.15 to macOS-11.0, suddenly your PyPy wheels disappear. :/ If you were not building them before, that's not a big deal, but if you were trying to build them and they go away with a green checkmark still, that seems problematic.

cibuildwheel/util.py
joerick: > rather than propagating the idea that it's okay to manipulate paths inside a package

But the resources are files, right? They need to be files, otherwise we'd have to copy them out to a temp dir before we could use them?

henryiii: Opening a resource works just like normal files, even if, for example, are coming from a zip file because Python supports site-packages being a zip file. If you need an actual file name, you do:

```
from importlib.resources import as_file
with as_file(package_dir / "my_file.txt") as file:
    third_party_api_requiring_file_system_path(file)
```
For the duration of that context manager, the resource is actually a real file (either the normal one from the file system, or one extracted to a tmp dir if the original loader (like the zip loader) doesn't provide file names.

https://importlib-resources.readthedocs.io/en/latest/using.html#file-system-or-zip-file

henryiii: So until we add the above, some parts of cibuildwheel do not support running from arbitrary loaders, since we do require file names in a few places. So there's only a conceptual benefit (not showing bad examples of path manipulation in a library) to using importlib.resouces until we do that (at which time, there would be a real benefit).

cibuildwheel/architecture.py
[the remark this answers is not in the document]
henryiii: Especially since this currently only affects 'all', then I think it's okay. Now if we added `universal2` to `auto`, then this be a (much) bigger issue, as `auto` should not break on macOS < 11.

henryiii:
I thought @ronaldoussoren said that even though 10.15 technically could work, the trigger in (packaging, perhaps?) simply did the simple thing and checked macOS version instead of Xcode version. Happy to be wrong though!

joerick:
That does ring a bell, but it looks like it's trying to build... looks like the SDK is wrong though. That might be resolvable... I'm having a play around

joerick:
Hey it looks like that worked! :) so we don't need to use the macOS-11 image just yet. 

The CircleCI error is this one:

> + Download https://www.python.org/ftp/python/3.7.9/python-3.7.9-macosx10.9.pkg to /tmp/Python.pkg
> + sudo installer -pkg /tmp/Python.pkg -target /
installer: Error the package path specified was invalid: '/tmp/Python.pkg'.

Seems unrelated to my changes... are we seeing this in any other PRs?

henryiii:
Yes, randomly in all other PRs. Kicking it fixes it once in a while.

henryiii:
I wonder if Circle clears /tmp at some rate? Could we make a tmpdir with Python here?

joerick:
It's a good job we can build on macOS 10.15, as it looks like the macos-11.0 image is going private again - https://github.com/actions/virtual-environments/issues/2486 .

cibuildwheel/macos.py
joerick: Yeah, fair point. Now that we can recommend people stick with 10.15, I'm happy to upgrade this to an error.

joerick: Ah.. actually this would mean that every integration test would need to add `CIBW_SKIP: pp*-macosx*` so the suite passes on macOS-11.... maybe that's okay, because we're not testing on macOS-11 yet? 

Looks like theres some progress on PyPy Big Sur support, so maybe it'll be resolved before people start trying to run on macOS-11. Otherwise we might have to switch it back to a warning again, just to save our sanity ;)

joerick: Yeah... actually, raising an error here means I can't run unit tests locally, because `get_python_configurations` is called by `main`... I'm leaning towards keeping this a warning, and hopefully PyPy have a macOS 11 release out before macOS 11 is mainstream in CIs.

henryiii: Okay, in that case, for now, it can be a warning. If macos-latest starts looking like it will be 11.0 and PyPy doesn't support it yet, then we'll see about an error?

henryiii:
Let's make sure the rename hack it over internal for tests, and listed in docs so users can do it if they want - cibuildwheel should not be renaming the wheels for users, as hopefully this will be sorted out soon and loading these on PyPI with the wrong names, unexpectedly, seems quite wrong.

One workaround that some projects might want to do is to use a x86_64 10.9 and a Universal2 11.0 wheel. That ruins most of the purpose for Universal2 that @ronaldoussoren wanted, but at least it gets more reasonable over time. If large projects do x86_64 10.9 and arm64 11.0, that works out of the box.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
joerick: Actually the requirement is Xcode 12, so it's not something we should be fiddling with here.

joerick:
> Let's make sure the rename hack it over internal for tests, and listed in docs so users can do it if they want - cibuildwheel should not be renaming the wheels for users, as hopefully this will be sorted out soon and loading these on PyPI with the wrong names, unexpectedly, seems quite wrong.

I'm not sure I understand the need for caution here - the rename hack just adds a more restrictive tag, doesn't it? What unintended consequences could this have?

henryiii:
I don't think cibuildwheel should be renaming your output wheels. Generally, we add notes about how to handle issues (like avoiding an extra dependency on MSVC 2019), rather than try to guess what the user wants to do. Also, this rename will depend on the macOS setting, 11.0 will not need a rename, while anything before does. And once this gets fixed, the rename won't be needed.

We can have a note "currently, universal wheels targeting anything before macOS 11.0 will not load on ARM. This is being fixed, but if you need to support ARM today, add `...` to your wheel name" or something like that.

Having files surprisingly change names [causes bugs](https://github.com/jwlawson/actions-setup-cmake/issues/13).

joerick:
Sigh. I agree that it would be _better_ for the issue to be fixed upstream, but that's why it's a hack. I'd say that 'producing working wheels' is more important than 'not renaming wheels'. Besides, it's a temporary thing, we can remove it once pip fixes the problem on their end.

henryiii:
There is a chance Packaging will update today and Pip on the weekend!

cibuildwheel/macos.py
joerick: Yeah, sounds good, we'll pick up the conversation then if it becomes an issue.

joerick:
Testing out these wheels that were built on macOS-10.15 on my local M1 machine.

```shell
$ /tmp/arm64-venv/bin/python -m pip install spam-0.1.0-cp39-cp39-macosx_11_0_arm64.whl 
Processing ./spam-0.1.0-cp39-cp39-macosx_11_0_arm64.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ /tmp/arm64-venv/bin/python -c 'import platform; print(platform.machine()); import spam; spam.system("echo it works!")'
arm64
it works!
```

So far, so good.

```shell
$ /tmp/arm64-venv/bin/python -m pip uninstall spam                                                        
Found existing installation: spam 0.1.0
Uninstalling spam-0.1.0:
  Would remove:
    /Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/spam-0.1.0.dist-info/*
    /Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/spam.cpython-39-darwin.so
Proceed (y/n)? y
  Successfully uninstalled spam-0.1.0
```

Let's try the universal2 wheel.

```shell
$ /tmp/arm64-venv/bin/python -m pip install spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Processing ./spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ /tmp/arm64-venv/bin/python -c 'import platform; print(platform.machine()); import spam; spam.system("echo it works!")'
arm64
ImportError: dlopen(/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/spam.cpython-39-darwin.so, 2): no suitable image found.  Did find:
	/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/spam.cpython-39-darwin.so: mach-o, but wrong architecture
	/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/spam.cpython-39-darwin.so: mach-o, but wrong architecture
```

Ah. Problem. It's missing the arm64 arch? Weird, because as we saw above, we _can_ build arm64 on macOS 10.15.

```shell
$ lipo -archs spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2/spam.cpython-39-darwin.so
x86_64
```

Yes, that's the issue. The library wasn't built with arm64. Did a bit of Googling, I found [this setuptools issue](https://github.com/pypa/setuptools/issues/2520) and [this CPython issue](https://bugs.python.org/issue42619).

Looking at https://github.com/python/cpython/blob/aee7c99874e5d0960977143c5b8b0ee9462a7ad8/Lib/_osx_support.py , we might be able to work around it by setting ARCHFLAGS, since we know that works for arm64! I'm trying that now.

henryiii:
Did it work? I don't' know when my Apple Silicon Mac mini will arrive; I can't help test until then...

joerick:
```shell
$ /tmp/arm64-venv/bin/python -m pip install spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Processing ./spam-0.1.0-cp39-cp39-macosx_10_9_universal2.macosx_11_0_universal2.whl
Installing collected packages: spam
Successfully installed spam-0.1.0
$ /tmp/arm64-venv/bin/python -c 'import platform; print(platform.machine()); import spam; spam.system("echo it works!")'
arm64
it works!
```

Looks good! So, I'm gonna look at tidying this PR up then we can do a final review.

.github/workflows/test.yml
joerick: Yeah.. maybe let's leave it in, at least until we can test these universal2/arm64 wheels in CI.

cibuildwheel/macos.py
joerick: This is still an open question, I'm happy to merge it as-is, but hopefully we'll get a pip/packaging update very soon and we can remove the hack entirely.

joerick:
Okay, this is ready for review!

The only think that will probably change is the removal of the wheel rename hack -  judging by https://github.com/pypa/pip/pull/9533 it looks like pip will soon support universal2 on arm64, and then we can remove that too. Thank you @henryiii for pushing that forward!

cibuildwheel/architecture.py
```diff
@@ -46,19 +52,26 @@ def parse_config(config: str, platform: PlatformName) -> 'Set[Architecture]':
...
+
         if platform == 'linux' and native_architecture == Architecture.x86_64:
             # x86_64 machines can run i686 docker containers
             result.add(Architecture.i686)
+
         if platform == 'windows' and native_architecture == Architecture.AMD64:
             result.add(Architecture.x86)
+
+        if platform == 'macos' and native_architecture == Architecture.arm64:
+            # arm64 can build and test both archs of a universal2 wheel.
```
mayeut, line 64: It's unlikely we'll be able to leverage this code path anytime soon with the lack of arm64 CI runners but should `x86_64` be added here ? Or is that on purpose and only do `x86_64` on `x86_64` runners with `auto` ?

joerick: Good point, it is on purpose but maybe I haven't explained my thought process fully. Once arm64 runners are commonplace, I'm guessing that the best setup would be to have two macOS runners - one x86_64, and one arm64. This is so we test natively as much as possible, and because it's _probably_ better to build all of cp35-cp38/pp on the x86_64 runner.

So, to avoid doubling-up of wheels, the arm64 runner doesn't build x86_64 by default. 

Ultimately, users would upgrade cibuildwheel and add an arm64 runner in their matrix, and the defaults do all the right things.

cibuildwheel/macos.py
```diff
@@ -1,12 +1,16 @@
...
 import shlex
 import shutil
 import subprocess
 import sys
 import tempfile
 from pathlib import Path
-from typing import Dict, List, NamedTuple, Optional, Sequence
+from typing import Any, Dict, List, NamedTuple, Optional, Sequence, Set, Tuple, cast
 
+from typing_extensions import Literal
```
mayeut, line 11: This import fails when using the GHA action: https://github.com/mayeut/pybase64/runs/1797295461?check_suite_focus=true

joerick: strange.. it's listed in setup.cfg as a requirement... https://github.com/joerick/cibuildwheel/blob/macos-universal2/setup.cfg#L35 @henryiii any ideas?

mayeut: it's listed for python<3.8
GHA uses python 3.9 in the build that's linked.

joerick: oohh i should be importing from `cibuildwheel.typing`. will fix

joerick: Ok, that should work now.

mayeut: All good now !

cibuildwheel/macos.py
```diff
@@ -180,6 +240,35 @@ def setup_python(python_configuration: PythonConfiguration,
...
+        if get_macos_version() < (10, 16):
+            # xcode 12.2 or higher can build arm64 on macos 10.15 or below, but
+            # needs the correct SDK selected
+
+            # however, different versions of Xcode contain different SDK
+            # versions...
+            if get_xcode_version() == (12, 2):
+                env.setdefault('SDKROOT', 'macosx11.0')
+            else:
+                env.setdefault('SDKROOT', 'macosx11.1')
```
mayeut, line 270: @joerick, 
Won't this be failing in the near future ? I guess the only way is to update this each time there's a new version of Xcode / macosx SDK ?

cibuildwheel/macos.py
```diff
@@ -242,60 +331,140 @@ def build(options: BuildOptions) -> None:
...
+                    def call_with_arch(args: Sequence[PathOrStr], **kwargs: Any) -> int:
+                        if isinstance(args, str):
+                            args = ' '.join(arch_prefix) + ' ' + args
+                        else:
+                            args = [*arch_prefix, *args]
+                        return call(args, **kwargs)
+
+                    # Use --no-download to ensure determinism by using seed libraries
+                    # built into virtualenv
+                    call_with_arch(['python', '-m', 'virtualenv', '--no-download', venv_dir], env=env)
```
mayeut, line 430: @joerick,

Have you been able to confirm that virtualenv and all the following calls indeed honor arch on M1 ?
It was not the case for i386 at some point in the past.

henryiii:
Let's bump our pins and dump the hack!!!

https://github.com/pypa/pip/issues/9282#issuecomment-770209600

cibuildwheel/macos.py
joerick: Unfortunately, yes, it's probably gonna need a few updates, if Apple release updated Xcodes with newer SDKs. Normally there are about [4 SDK releases per Xcode major version](https://xcodereleases.com/)...

I suppose another way to do this would be to parse the output of `xcodebuild -showsdks`, which looks like this:

```shell
$ xcodebuild -showsdks 2>/dev/null
iOS SDKs:
	iOS 14.3                      	-sdk iphoneos14.3

iOS Simulator SDKs:
	Simulator - iOS 14.3          	-sdk iphonesimulator14.3

macOS SDKs:
	DriverKit 20.2                	-sdk driverkit.macosx20.2
	macOS 11.1                    	-sdk macosx11.1

tvOS SDKs:
	tvOS 14.3                     	-sdk appletvos14.3

tvOS Simulator SDKs:
	Simulator - tvOS 14.3         	-sdk appletvsimulator14.3

watchOS SDKs:
	watchOS 7.2                   	-sdk watchos7.2

watchOS Simulator SDKs:
	Simulator - watchOS 7.2       	-sdk watchsimulator7.2
```

Kind of messy output, but perhaps we could look for `-sdk macosx(\d+)\.[\.\d]+` and choose anything >11...

cibuildwheel/macos.py
joerick: Yes, the tests are checking that `platform.machine()` is correct inside TEST_COMMAND. They're not checked in CI yet, but I've run these locally on my M1 machine and they pass.

joerick:
I can't actually run `bin/update_dependencies.py` locally anymore - a combination of no Docker on M1 and no CPython 3.5 on macOS 11 - so I've done it in GHA instead. I wonder if this works...

cibuildwheel/macos.py
mayeut: Up to you.
On one hand, doing nothing now will require some updates now & then.
On the other hand, doing something requires adding some parsing code...

> Kind of messy output, but perhaps we could look for -sdk macosx(\d+)\.[\.\d]+ and choose anything >11...

That might not even add that much code. Maybe that's the good option. Up-to-you.

joerick:
Ah, it does not. Hey ho. I'll do it on master instead and merge that in.

henryiii:
You could let it make it on master then redirect it. But not really any different.

henryiii:
Actually, you can just cherry pick the change here and keep going. :)

cibuildwheel/macos.py
henryiii: I think pulling it this way would be more reliable and less work long term than manual updates.

joerick: Yeah, maybe it would be better to detect the SDK. I'm not too worried about the amount of code, more that we're parsing something that doesn't look particularly designed for machine-readability and might change..? maybe there's a better way to list the sdks?

joerick: This would be an alternative, but it's not documented...
```python
import subprocess
import plistlib
from pathlib import Path

xcode_path = Path(subprocess.check_output(['xcode-select', '--print-path'], text=True).strip())
sdk_settings_plist = xcode_path / 'Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/SDKSettings.plist'
sdk_settings = plistlib.loads(sdk_settings_plist.read_bytes())
print(sdk_settings['CanonicalName'])
# prints 'macosx11.1'
```

But, on reflection, looking around the web, `xcodebuild -showsdks` hasn't changed its format in at least 8 years, and it's the documented way to list sdks, so I think I'll do that.

joerick: See 74170ba for implementation.

cibuildwheel/macos.py
```diff
@@ -34,21 +38,84 @@ def call(args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: O
...
+        (10, 14) <= (11, 0) == True
+        (10, 14) <= (10, 16) == True
+        (11, 2) <= (11, 0) != True
+    '''
+    version_str, _, _ = platform.mac_ver()
+    version = tuple(map(int, version_str.split(".")[:2]))
+    return cast(Tuple[int, int], version)
+
+
+def get_xcode_version() -> Tuple[int, int]:
```
mayeut, line 55: This function is not called anymore and can be removed.

joerick: Good catch 👍

mayeut (approved):

cibuildwheel/macos.py
```diff
@@ -180,6 +238,41 @@ def setup_python(python_configuration: PythonConfiguration,
         # https://github.com/python/cpython/blob/a5ed2fe0eedefa1649aa93ee74a0bafc8e628a10/Lib/_osx_support.py#L260
         env.setdefault('ARCHFLAGS', '-arch x86_64')
 
+    if python_configuration.version == '3.9':
```
henryiii, line 241: Once we have access to packaging from #536 , then we should do a `>=` 3.9 here so we don't have to change it when we add 3.10. But this is fine until then.

henryiii (approved):

YannickJadoul:
Wow. I haven't really been following (partially because I didn't feel like I wanted more macOS issues in my life) this intensively, but great job, everyone! :-)

cibuildwheel/macos.py
```diff
@@ -180,6 +238,41 @@ def setup_python(python_configuration: PythonConfiguration,
...
+
+    if python_configuration.identifier.endswith('arm64') or python_configuration.identifier.endswith('universal2'):
+        if get_macos_version() < (10, 16) and 'SDKROOT' not in env:
+            # xcode 12.2 or higher can build arm64 on macos 10.15 or below, but
+            # needs the correct SDK selected.
+            sdks = get_macos_sdks()
+
+            # Different versions of Xcode contain different SDK versions...
+            # we're happy with anything newer than macOS 11.0
+            arm64_compatible_sdks = [s for s in sdks if not s.startswith('macosx10.')]
```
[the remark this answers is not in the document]
joerick: From what I can see on https://xcodereleases.com/ , each version of Xcode only ships with one macOS SDK version. I couldn't find a case where there are multiple. But I wasn't 100% sure, hence the use of lists here.

Also not sure what the selection criteria would be... would we want the newest , or the oldest? I tend to find that binaries are more often forward-compatible than backward-compatible. So then we'd want the oldest? ¯\\\_(ツ)\_/¯

joerick:
Aaaand it's in! Thanks everyone for the input, especially @henryiii , @Czaki , @ronaldoussoren and @mayeut !

ogrisel:
Thanks very much to everybody involved in making this happen.

PR #489, opened 2020-12-20
Update dependencies

henryiii:
Update the versions of our dependencies.

PR generated by `make_dependency_update_pr.py`.

See #487.

PR #490, opened 2020-12-20
Add a workflow to update dependencies

mayeut:
Add a workflow to update dependencies.

Manual dispatch.


bin/update_dependencies.py
```diff
@@ -12,21 +12,39 @@
...
-for python_version in ['27', '35', '36', '37']:
-    subprocess.check_call([
-        f'./env{python_version}/bin/pip-compile',
-        '--allow-unsafe',
-        '--upgrade',
-        'cibuildwheel/resources/constraints.in',
-        '--output-file', f'cibuildwheel/resources/constraints-python{python_version}.txt'
-    ])
+
+if 'GITHUB_ACTIONS' in os.environ:
```
henryiii, line 23: Honestly, a docker based solution for local runs wouldn't be so bad either...

mayeut: I'm +1 on this. I didn't want to change the current behavior at first.

henryiii: How about a `--docker` flag instead?

henryiii: Though, if @joerick is okay with it, for simplicity, I'd be happy to have a docker method as the only method.

mayeut:
The workflow does not appear in "Actions"... I must be doing something wrong but I have no clue at the moment.

henryiii:
It may need to be in master - or it may have had to run once.

mayeut:
> It may need to be in master - or it may have had to run once.

From what I've read around, probably needs to be in master. Adding on push for now (to trigger a sample PR)

mayeut:
And here is the sample PR: #491

bin/update_dependencies.py
joerick: Using a manylinux image is a great idea. But it would be nice if there was still a way to run using local pythons, as I'm about to get an Apple Silicon machine and I'm not sure if i'll be able to run Docker, at least in the short term. `'--local-pythons' in sys.argv` would do the trick

joerick:
Nice!

bin/update_dependencies.py
mayeut: @joerick, updated with your proposed change.

PS: docker is available in preview https://docs.docker.com/docker-for-mac/apple-m1/. It should be able to run x86_64 images through emulation.

joerick:
It's pretty cool that it updates the existing PR if there's already one open! I think once we remove the 'push' trigger this looks good, to me

mayeut:
> It's pretty cool that it updates the existing PR if there's already one open! I think once we remove the 'push' trigger this looks good, to me

I removed the `push` trigger and enforce PR creation to be done only on master (I can revert this last one if that's not wanted).

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,31 @@
...
+      name: Install Python 3.9
+      with:
+        python-version: 3.9
+        architecture: x64
+    - name: Install dependencies
+      run: python -m pip install requests pip-tools
+    - name: Run update
+      run: python ./bin/update_dependencies.py
+    - name: Create Pull Request
+      if: github.ref == 'refs/heads/master' && always()
```
joerick, line 21: this might be some debug code left over?
```suggestion
      if: github.ref == 'refs/heads/master'
```

bin/update_dependencies.py
```diff
@@ -12,21 +13,39 @@
...
-    subprocess.check_call([
-        f'./env{python_version}/bin/pip-compile',
-        '--allow-unsafe',
-        '--upgrade',
-        'cibuildwheel/resources/constraints.in',
-        '--output-file', f'cibuildwheel/resources/constraints-python{python_version}.txt'
-    ])
+
+if '--no-docker' in sys.argv:
+    for python_version in ['27', '35', '36', '37']:
```
joerick, line 25: could the `['27', '35', '36', '37']` part be extracted to a common variable? `PYTHON_VERSIONS` or something?

bin/update_dependencies.py
```diff
@@ -12,21 +13,39 @@
 
 # CUSTOM_COMPILE_COMMAND is a pip-compile option that tells users how to
 # regenerate the constraints files
-os.environ['CUSTOM_COMPILE_COMMAND'] = "bin/update_constraints.py"
+os.environ['CUSTOM_COMPILE_COMMAND'] = "bin/update_dependencies.py"
 subprocess.check_call([
```
joerick, line 17: I just realised... this command is implicitly run in python3.8, because that's what my venv is in on my mac >.< (maybe a little sloppy, sorry!). Could it be made explicit? Perhaps `38` and `39` should be added to the list of versions and the default constraints.txt uses the final entry of this list? Then it should also be generated inside the docker container, so we're sure which version of Python is running it.

joerick (commented):
Apologies, I just realised the implicit Python that's generating the versionless `constraints.txt` should probably be controlled too.

.github/workflows/update-dependencies.yml
henryiii: `always()` refers to the previous step - otherwise, there is an implicit `&& success()` here, IIRC. Not sure if you do want this to make a PR if the script fails, though?

mayeut: The `aways()` is a leftover of a repo where I want to create the PR even if the previous step failed. we don't want to use `always()` here. good catch.

bin/update_dependencies.py
mayeut: Shall we use a default `constraints.txt` at all ?

joerick: That's a great point. Maybe we don't need it at all, in fact!

joerick: Oh, well actually our code does assert that it exists, currently

joerick: It's also maybe useful as a starting point if somebody wanted to customise their constraints. But just copying the latest Python we support should do

mayeut: Done. I had to modify a test so I updated all dependencies. Anyway, I think I'll have to rebase on #489 anyway.

bin/update_dependencies.py
```diff
@@ -12,21 +14,39 @@
...
+    for python_version in PYTHON_VERSIONS:
+        subprocess.check_call([
+            f'./env{python_version}/bin/pip-compile',
+            '--allow-unsafe',
+            '--upgrade',
+            'cibuildwheel/resources/constraints.in',
+            '--output-file', f'cibuildwheel/resources/constraints-python{python_version}.txt'
+        ])
+else:
+    image = 'quay.io/pypa/manylinux2010_x86_64:2020-12-19-8df9e2d'
```
YannickJadoul, line 31: How do we update this? Do we update this, or is an older version OK?

unit_test/dependency_constraints_test.py
```diff
@@ -10,7 +10,8 @@ def test_defaults():
     resources_dir = project_root / 'cibuildwheel' / 'resources'
 
     assert dependency_constraints.base_file_path.samefile(resources_dir / 'constraints.txt')
-    assert dependency_constraints.get_for_python_version('3.8').samefile(resources_dir / 'constraints.txt')
+    assert dependency_constraints.get_for_python_version('3.99').samefile(resources_dir / 'constraints.txt')
```
YannickJadoul, line 13: Is there a reason to still have a default `constraints.txt` file?

I thought the reason it's there is because 3.8 and 3.9 currently have the exact same version file?

bin/update_dependencies.py
mayeut: Any image version that has all python versions is ok. I only expect this one to require updating when adding a new python version so manually would do or maybe use `latest` because anyway, it will update the corresponding image to `latest` (well the properly tagged version of `latest`).

unit_test/dependency_constraints_test.py
mayeut: see @joerick comments https://github.com/joerick/cibuildwheel/pull/490#discussion_r546425734 on his review.

YannickJadoul: Yeah, I saw, but to start customizing your own constraints, you could also start from `constraints-python39.txt`? `constraints.txt` feels like a historical remainder, somehow. But ofc, it's not wrong; I was just thinking if it's not better to get a reminder (in the form of a failing CI), when we add 3.10, and future versions to `cibuildwheel`?

bin/update_dependencies.py
YannickJadoul: Ah, yes. When adding a Python version, we'll have to update the Python versions in `update_dependencies.py` as well, anyway, now.

YannickJadoul: `update_update_dependencies.py`? ;-)

unit_test/dependency_constraints_test.py
henryiii: I think it's simpler just to have one per version. "But just copying the latest Python we support should do" I think is exactly what @YannickJadoul is referring to here.

bin/update_dependencies.py
henryiii: Can’t this be latest? This is a case where pinning absolutely everything might be too much. :)

mayeut: done

unit_test/dependency_constraints_test.py
YannickJadoul: Yeah, we're coming from a situation at the start where we only had one `constraints.txt`. Then 2.7 started acting up, and we had `constraints-python2.7.txt`, then 3.5, then 3.6, ... and now, there seems to be the realization that just having a separate file for all is better.

So why not just ditch the "default", then, if falling back to the default is a mistake?

YannickJadoul: Not that crucial, though. Just wondering if there's a good reason. (There's nothing else to nitpick about, so I need to find _something_? I very much like the idea and rest of the PR :-) )

mayeut: @YannickJadoul, @henryiii, I agree and proposed to do so. Final decision left to @joerick.

YannickJadoul: Right. Sorry; should have added my comment to that thread, then :-)

joerick: I'm happy to go with that - that we don't bundle a constraints.txt. I think as @YannickJadoul says, the version-specific ones appeared to solve specific problems, but version-specific is probably a more robust solution. We'll have to remove [this line](https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/util.py#L98) but that's not a huge deal, I think.

YannickJadoul: Hmmm, looking at that code, I'm now seeing it's because we want to keep supporting a default `constraints.txt` for users.

I'm fine either way, then! As you see fit, @mayeut. Sorry about the confusion!

YannickJadoul (approved):

henryiii (approved):

mayeut:
I merged as-is.
We can still update the bit about `constraints.txt` later if needed.

Thanks for your reviews.

mayeut:
I can, now that it's in master, dispatch the workflow. #495 created (can probably be closed though).
@joerick expressed [at some point](https://github.com/joerick/cibuildwheel/pull/256#issuecomment-580808794) that he was not a big fan of `dependabot` because of the "sea of PR" it was creating. If need be and if this is not seen as "nuisance" PR, the workflow could also be run as a cron job.

henryiii:
Dependabot [(referring to `.github/dependabot.yml`)](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates) is configurable - you can have it update once a week if you like.  There have been a few other things under the name dependabot, including a built-in security updater. But the configurable one is quite nice.

joerick:
I'd be happy for this to run regularly, perhaps weekly? It's always just one PR, as opposed to the dependabot 'sea', and we can dial it back if it becomes irritating.

YannickJadoul:
As noted by @henryiii (somewhere else; I forgot why), it might also be nice to update the Python versions?
The main question is: is there a clean way of getting a list of Python versions?

mayeut:
@YannickJadoul, I don't know about "clean" but I have been experimenting with this because it's also something I'd like to have in `manylinux`. I expended the experiment for `cibuildwheel` and I will post a draft PR showing it can work. I'm not sure I will have time to improve on it though but it might give some ideas/pointers to others.

YannickJadoul:
@mayeut That would be amazing :-) Anything that can help us get started could help, anyway.

joerick:
I agree, it would be nice to automate the Python upgrades as well. Sounds good @mayeut , I look forward to seeing it! The only way I can think of to do it would be to parse https://www.python.org/ftp/python/ ...

YannickJadoul:
Yes, by "clean", I was referring to something like PyPy's https://downloads.python.org/pypy/versions.json, and not having to parse things, or check if trying to download e.g. https://www.python.org/ftp/python/3.9.2/python-3.9.2-macosx10.9.pkg doesn't fail.

mayeut:
Draft PR #496 created.

Issue #431, opened 2020-09-14
Configuration files for download links


YannickJadoul:
Something that crossed my mind, a few days ago: since there needs to be a new release every time download links change or new Python versions get added or ..., should we do something similar as for the docker images and their versions and have them in a separate file that's read by `cibuildwheel`?

I can't make up my mind on whether this is a potentially good idea, or not, so I thought I'd just throw this out there. Curious to hear pros and cons :-)

Czaki:
It is release moved in another place. And such release need some infrastructure for hosting. And still one of cons of cibuildwheel is version pining which gave reproducibility in builds. so it produce additional parameters need. 

As I good understand main problem are pypy download links. (because bump of cpython version are does not break old build system).  Maybe ask pypy maintainers for provide some mapping files on their website?

joerick:
I can see two ways to interpret this-

1. Python download links distributed outside the current release process (this seems to be what @Czaki had assumed).

2. A config file in the repo, distributed in the package like `pinned_docker_images.cfg`. This is then read by the various build scripts to extract the download URLs. 

which one were you thinking @YannickJadoul ? curious also to hear what advantages this might have.

Czaki:
>     2\. A config file in the repo, distributed in the package like `pinned_docker_images.cfg`. This is then read by the various build scripts to extract the download URLs.

Second one still need create new release and @YannickJadoul  wrote

> since there needs to be a new release every time download links change or new Python versions get added

YannickJadoul:
I was indeed thinking 2., which would allow (just like `pinned_docker_images.cfg`) to be (partially) overridden by the user, if necessary. Of course we'd update with new releases.

> curious also to hear what advantages this might have.

I'm not sure, but I was wondering if we could escape the new-release-for-every-Python-or-PyPy-update, this way. E.g., to #426, you could say to just provide a config file with updated download links, instead of forcing an update.

But again, I'm not convinced myself this is a good idea (especially with respect to reproducable builds and using tested Python versions). It's just something that crossed my mind.

YannickJadoul:
>     1. Python download links distributed outside the current release process (this seems to be what @Czaki had assumed).

This could also be interesting, though. We could link to some repository + commit, allowing the user to set the commit to choose newer versions in an old version of cibuildwheel.

[...]

PR #496, opened 2020-12-23
feat: update python versions as part of update_dependencies

mayeut:
As mentioned in #490, a method to update CPython & PyPy on Windows/macOS.
Per @YannickJadoul criteria, the idea behind it is "clean" however, the implementation here is far from being clean and thus is opened as draft to give ideas/pointers to others as I probably won't have time to work on this.

### DIFF
--- a/bin/update_pythons.py
+++ b/bin/update_pythons.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python3
+
+import sys
+from collections import defaultdict
+from itertools import groupby
+from pathlib import Path
+from typing import Dict, List, Tuple
+
+import requests
+import toml
+from packaging.version import Version
+
+if sys.version_info < (3, 8):
+    from typing_extensions import TypedDict
+else:
+    from typing import TypedDict
+
+# Use pretty printing for debugging
+# from rich import print
+
+
+allow_prerelease = False
+
+
+class InlineArrayDictEncoder(toml.encoder.TomlEncoder):
+    def dump_sections(self, o: dict, sup: str):
+        if all(isinstance(a, list) for a in o.values()):
+            val = ""
+            for k, v in o.items():
+                inner = ",\n  ".join(self.dump_inline_table(d_i).strip() for d_i in v)
+                val += f"{k} = [\n  {inner},\n]\n"
+            return val, self._dict()
+        else:
+            return super().dump_sections(o, sup)
+
+
+DIR = Path(__file__).parent.parent.resolve()
+RESOURCES_DIR = DIR / "cibuildwheel/resources"
+
+
+class ConfigWinCP(TypedDict):
+    identifier: str
+    version: str
+    arch: str
+
+
+class ConfigWinPP(TypedDict):
+    identifier: str
+    version: str
+    arch: str
+    url: str
+
+
+def get_cpython_windows() -> Dict[str, List[Version]]:
+    response = requests.get("https://api.nuget.org/v3/index.json")
+    response.raise_for_status()
+    api_info = response.json()
+
+    for resource in api_info["resources"]:
+        if resource["@type"] == "PackageBaseAddress/3.0.0":
+            endpoint = resource["@id"]
+
+    cp_versions: Dict[str, List[Version]] = {"64": [], "32": []}
+    for id, package in [("64", "python"), ("32", "pythonx86")]:
+        response = requests.get(f"{endpoint}{package}/index.json")
+        response.raise_for_status()
+        cp_info = response.json()
+
+        for version_str in cp_info["versions"]:
+            version = Version(version_str)
+            if version.is_devrelease:
+                continue
+            if not allow_prerelease and version.is_prerelease:
+                continue
+            cp_versions[id].append(version)
+        cp_versions[id].sort()
+
+    return cp_versions
+
+
+def get_pypy_windows(
+    plat_arch: str = "win32-x86",
+) -> Dict[str, List[Tuple[Version, str]]]:
+
+    response = requests.get("https://downloads.python.org/pypy/versions.json")
+    response.raise_for_status()
+    pp_realeases = response.json()
+    pp_versions = defaultdict(list)
+
+    for pp_realease in pp_realeases:
+        if pp_realease["pypy_version"] == "nightly":
+            continue
+        version = Version(pp_realease["pypy_version"])
+        python_version = Version(pp_realease["python_version"])
+        python_version_str = f"{python_version.major}.{python_version.minor}"
+        url = None
+        for file in pp_realease["files"]:
+            if f"{file['platform']}-{file['arch']}" == plat_arch:
+                url = file["download_url"]
+                break
+        if url:
+            pp_versions[python_version_str].append((version, url))
+
+    return pp_versions
+
+
+# Debugging printout:
+# print(get_cpython_windows())
+# print()
+# print("[bold]Getting PyPy")
+# print(get_pypy_windows())
+
+ARCH_DICT = {"32": "win32", "64": "win_amd64"}
+
+
+def build_ids_cp(in_dict: Dict[str, List[Version]]) -> List[ConfigWinCP]:
+    items: List[ConfigWinCP] = []
+    for arch in in_dict:
+        for minor, grp in groupby(in_dict[arch], lambda v: v.minor):
+            # Filter pre-releases, unless it's all pre-releases
+            if not all(v.is_devrelease for v in grp):
+                grp = filter(lambda v: not v.is_devrelease, grp)
+
+            version = sorted(grp)[-1]
+            identifier = f"cp3{minor}-{ARCH_DICT[arch]}"
+
+            items.append(
+                ConfigWinCP(
+                    identifier=identifier,
+                    version=str(version),
+                    arch=arch,
+                )
+            )
+
+    return items
+
+
+def build_ids_pp(in_dict: Dict[str, List[Tuple[Version, str]]]) -> List[ConfigWinPP]:
+    items: List[ConfigWinPP] = []
+    for vers, matches in in_dict.items():
+        vers_id = vers.replace(".", "")
+        if not all(v[0].is_devrelease for v in matches):
+            matches = list(filter(lambda v: not v[0].is_devrelease, matches))
+
+        version, url = sorted(matches, key=lambda v: v[0])[-1]
+        identifier = f"pp{vers_id}-win32"
+
+        items.append(
+            ConfigWinPP(
+                identifier=identifier,
+                version=vers,
+                arch="32",
+                url=url,
+            )
+        )
+
+    return items
+
+
+windows_configs = [
+    *build_ids_cp(get_cpython_windows()),
+    *build_ids_pp(get_pypy_windows()),
+]
+
+configs = toml.load(RESOURCES_DIR / "build-platforms.toml")
+origpy2 = list(
+    filter(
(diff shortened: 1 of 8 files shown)


mayeut:
If this helps or gives other ideas, I started to rework a bit the "CPython from python.org" part for source packages in `manylinux`.
It's not committed anywhere yet.

```
from collections import namedtuple

import packaging.version
import requests


_CandidateRelease = namedtuple('_CandidateRelease', [
    'version',
    'release',
])


def _get_id(resource_uri):
    return int(resource_uri.rstrip('/').split('/')[-1])


def _get_os():
    response = requests.get('https://www.python.org/api/v2/downloads/os/?slug=source')
    response.raise_for_status()
    os_info = response.json()
    assert len(os_info) == 1
    return _get_id(os_info[0]['resource_uri'])


def _get_update(bin, os_):
    current_version = bin[0].version
    candidates = list(bin[1:])
    candidates.sort(key=lambda x: x.version, reverse=True)
    for prerelease in {False, current_version.is_prerelease}:
        for candidate in candidates:
            if candidate.version.is_prerelease != prerelease:
                continue
            # manylinux hardcodes URL now. Only the version is mutable.
            expected_url = f'https://www.python.org/ftp/python/{candidate.version.base_version}/Python-{candidate.version}.tgz'
            response = requests.get(f'https://www.python.org/api/v2/downloads/release_file/?os={os_}&release={candidate.release}')
            response.raise_for_status()
            file_info = response.json()
            for file in file_info:
                if file['url'] == expected_url:
                    return candidate.version

    return current_version


def _get_release(current_versions, os_):
    response = requests.get('https://www.python.org/api/v2/downloads/release/?version=3&is_published=true')
    response.raise_for_status()
    release_info = response.json()
    bin = {}
    for current_version in current_versions:
        bin_key = f'{current_version.major}.{current_version.minor}'
        bin[bin_key] = [_CandidateRelease(current_version, None)]
    for release in release_info:
        parts = release['name'].split()
        if parts[0].lower() != 'python':
            continue
        assert len(parts) == 2
        version = packaging.version.parse(parts[1])
        bin_key = f'{version.major}.{version.minor}'
        if bin_key not in bin.keys():
            continue
        current = bin[bin_key][0]
        if version <= current.version:
            continue
        if version.is_postrelease or version.is_devrelease:
            continue
        if version.is_prerelease and not current.version.is_prerelease:
            continue
        bin[bin_key].append(_CandidateRelease(version, _get_id(release['resource_uri'])))
    updates = []
    for current_version in current_versions:
        bin_key = f'{current_version.major}.{current_version.minor}'
        updates.append(_get_update(bin[bin_key], os_))
    return list(zip(current_versions, updates))


os_ = _get_os()
current_versions = list([packaging.version.parse(v) for v in ['3.5.1', '3.6.2', '3.7.5', '3.8.4', '3.9.0', '3.10.0a0']])
print(_get_release(current_versions, os_))
```

It prints:
```
[(<Version('3.5.1')>, <Version('3.5.10')>), (<Version('3.6.2')>, <Version('3.6.12')>), (<Version('3.7.5')>, <Version('3.7.9')>), (<Version('3.8.4')>, <Version('3.8.7')>), (<Version('3.9.0')>, <Version('3.9.1')>), (<Version('3.10.0a0')>, <Version('3.10.0a3')>)]
```

joerick:
Thanks @mayeut ! I didn't know about the python.org downloads API, that certainly simplifies things! The code is here for the curious- https://github.com/python/pythondotorg/blob/master/downloads/api.py , it's a simple Django Rest Framework API.

As for integrating these into the code, I think a human/machine readable TOML config file would be best, similar to the manylinux one at [cibuildwheel/resources/pinned_docker_images.cfg](https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/resources/pinned_docker_images.cfg), instead of modifying the code in this script. @YannickJadoul I remember you had a similar idea in https://github.com/joerick/cibuildwheel/issues/431.

mayeut:
> As for integrating these into the code, I think a human/machine readable TOML config file would be best, similar to the manylinux one at cibuildwheel/resources/pinned_docker_images.cfg, instead of modifying the code in this script. @YannickJadoul I remember you had a similar idea in #431.

I did not saw (or remember seeing) #431 before but that's certainly one option which matches one of the few comment in the commit (`# hugly search pattern, package configuration shall probably done otherwise if we want to do this`)

YannickJadoul:
I already completely forgot about #431. For the record, that was just a loose thought, and nothing concrete; so at any rate, we'd have to discuss a bit on how to best approach that, but yes, I agree the combination of these two ideas could be interesting!

henryiii:
A few comments after I rewrote it a bit:

* Needs a macOS version
* I think we should require cibuildwheel dev install, so we can use things in utils
* Should we have a way to opt-in to dev versions and include them in the list? Just like `CIBW_MIN_PYTHON`, having an upper limit or some other way we could control opt-in to dev would allow users to start testing 3.10a4 now, if they wanted too, and would smooth out having to turn it on in a new release. Just a thought, because we have to filter out dev versions now.

henryiii:
Okay, think it's ready, checks macOS and Windows. Example of output:

Screen Shot 2021-01-12 at 3 52 31 PM

henryiii:
@joerick GitLab CI seems to be both required, and stuck. :/

joerick:
> @joerick GitLab CI seems to be both required, and stuck. :/

I've 'fixed' it by removing that CI from being required. The way Github does the auto merge is pretty odd. I have to choose the CI checks that are 'required' in settings, which are matched by *display name* to the checks being run. So, as we see in #484, because I changed the name of the macos job, it's forever waiting for the old job to report. Sigh.

Maybe something 3rd party would fit better... really I think we just want 'anything that's started, wait for that to be successful' rather than choosing individual checks.

joerick (commented):
Wow, quite an impressive bit of work @henryiii ! 

general points

- I'm not sure how the 'prereleases' flag works, at the moment... should it even be a global flag? Generally we don't want them, but sometimes we want to release a beta version of cibuildwheel with a new version of Python e.g. 3.10 support - then we'd want prerelease support but _only_ on 3.10.x versions.
- There might be a simpler way to write this script, by starting with a read of the `build-platforms.toml`, and then iterating through each configuration, trying to find a patch release that's newer than the one listed. This would allow us to manually edit the `build-platforms.toml` to add e.g. 3.10 support, or something else, and would avoid the hard-coding of 2.7 versions here. What do you think?

bin/update_pythons.py
```diff
@@ -0,0 +1,295 @@
...
+                    break
+
+    return items
+
+
+def _get_id(resource_uri: str) -> int:
+    return int(resource_uri.rstrip("/").split("/")[-1])
+
+
+def get_cpython(
```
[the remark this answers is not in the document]
henryiii: I sort of left this in, but the new version is clearer. But it can read from Win/Linux, we just don't use them from here.

bin/update_pythons.py
```diff
@@ -0,0 +1,295 @@
...
+    response = requests.get("https://www.python.org/api/v2/downloads/release/?is_published=true")
+    response.raise_for_status()
+
+    releases_info = response.json()
+    # Removing the prefix, Python 3.9 would use: release["name"].removeprefix("Python ")
+    known_versions = {Version(release["name"][7:]): _get_id(release["resource_uri"]) for release in releases_info}
+
+    items: List[ConfigMacOS] = []
+
+    sorted_versions = sorted((v for v in known_versions if versions.contains(v) and not v.is_prerelease), reverse=True)
```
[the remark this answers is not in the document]
henryiii: A little work might need to go in to get this working with pre-releases, I've mostly remove it now.

bin/update_pythons.py
```diff
@@ -0,0 +1,295 @@
...
+    *,
+    prereleases: bool = False,
+) -> List[AnyConfig]:
+
+    items: List[AnyConfig] = []
+
+    # Groupby requires pre-grouped input
+    orig_items = sorted(orig_items, key=lambda x: x["identifier"])
+
+    for _, grp in groupby(orig_items, lambda x: x["identifier"]):
```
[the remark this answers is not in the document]
henryiii: grp was an iterator that gave all the items where the key stays the same. But it's gone now.

unit_test/build_ids_test.py
```diff
@@ -1,19 +1,11 @@
+import pytest
 import toml
-from toml.encoder import TomlEncoder
 
 from cibuildwheel.util import resources_dir
 
+Version = pytest.importorskip("packaging.version").Version
```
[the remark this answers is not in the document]
henryiii: We don't (yet) require this to install, so it might be missing - this test is skipped in that case. Now with the CIBW_REQUIRES_PYTHON addition, we would require packaging, so this would go away.

henryiii:
Redesigned based on the idea of having the input file drive the search.

joerick:
Hope you don't mind, I had a little play with this @henryiii, to give nice diff outputs. Just an excuse to play with rich, really!

image

Anyway, so long as you're happy with my changes, LGTM!

henryiii:
I'm making one last style change (removing a comma), and looking to see if there's anything else I missed or any comments that might be useful. Then I'll push and trigger a pending merge!

bin/update_pythons.py
```diff
@@ -213,10 +217,7 @@ def __init__(self) -> None:
 
         self.macos_6 = CPythonVersions(plat_arch="macosx_x86_64", file_ident="macosx10.6.pkg")
         self.macos_9 = CPythonVersions(plat_arch="macosx_x86_64", file_ident="macosx10.9.pkg")
-        self.macos_u2 = CPythonVersions(
-            plat_arch="macosx_universal2",
-            file_ident="macos11.0.pkg",
-        )
+        self.macos_u2 = CPythonVersions(plat_arch="macosx_universal2", file_ident="macos11.0.pkg")
```
henryiii: This is the "magic" comma in Black, which lets you force multiline. I didn't want to do that here, so I removed it.

PS: I originally tried to write this without Black, but the first time a ran I had 20 or so Flake8 errors; I had no interest in waisting time trying to fix them by hand, so I ran `black bin/update_python.py`, and they all went away. So this is our one Black'ed file. :)

bin/update_pythons.py
```diff
@@ -281,7 +283,7 @@ def update_pythons(force: bool, level: str) -> None:
 
     if original_toml == result_toml:
         rich.print("[green]Check complete, Python configurations unchanged.")
-        exit()
+        return
```
henryiii: exit() should never be used in a file. It is only defined if site is imported (same with `quit()`). Either use sys.exit, or in this case, just return, we are in a main function already. :)

bin/update_pythons.py
```diff
@@ -227,6 +228,7 @@ def update_config(self, config: Dict[str, str]) -> None:
         orig_config = copy.copy(config)
         config_update: Optional[AnyConfig]
 
+        # We need to use ** in update due to MyPy (probably a bug)
```
henryiii, line 231: MyPy's handling of TypedDict's is pretty buggy. I had to do this, and I also don't know how to type narrow on a Union of TypedDicts, as `assert isinstance` is not allowed, so this takes a normal Dict for now. 🤷

bin/update_pythons.py
joerick: good to know!

henryiii: I have realized it's not the only offender. :)

henryiii: #539

PR #497, opened 2020-12-23
Run update workflow weekly

mayeut:
Per https://github.com/joerick/cibuildwheel/pull/490#issuecomment-750432177, this updates the update workflow to run weekly (manual dispatch still possible)


joerick (approved):
thanks @mayeut !

YannickJadoul (approved):

henryiii (approved):
Keeping in mind we can also trigger right before release. :)

Issue #501, opened 2020-12-24
PyPy virtualenv headers not available


Czaki:
I meet the next bug in PyPy creating virtualenv. File `Python.h` is not available form virtualenv, which breaks compilation of packages which does not provide PyPy wheels:

https://dev.azure.com/bokota/imagecodecs/_build/results?buildId=565&view=logs&j=9ea3a1cf-027d-5faa-88b4-e0b2b34fabd3&t=77745ba4-e9ed-58b1-0e6d-5824980d22ee&l=249

https://dev.azure.com/bokota/imagecodecs/_build/results?buildId=565&view=logs&j=c436af4d-dd17-5b6b-70ac-85e5b57c9f0c&t=bc0cc3b0-6386-533a-de3c-d01e46ec05cd&l=279

Did You remember who should be called for this? Would we like to add a test that simulates this? I do not want to open heptapod account (which I will forget in short time).

YannickJadoul:
> I do not want to open heptapod account (which I will forget in short time).

Seems you can login by linking to your GitHub account. But if you have a minimal reproducer, I can report it, if you prefer.

Czaki:
```
python -m pip install scikit-build
python -m pip install blosc==1.9.2
```
Works without virtualenv and fail in virtualenv with 

```
 ../../../blosc/blosc_extension.c:12:10: fatal error: Python.h: No such file or directory
     12 | #include "Python.h"
```

Happens on pypy3.6 and pypy3.7.

henryiii:
This seems vaguely familiar... Don't remember where I've seen this before. Pretty sure it had something to do with a PyPy venv there too.

henryiii:
You should poke the blosc authors to add a pyproject.toml to install Scikit-Build instead of having to do it manually.

henryiii:
Scikit-build does get the include directory for PyPy and pass it on, but it might not do that properly for PyPy? It happens to work outside the venv because the includes are in a common spot. Have you compared to CPython's venv and includes location?

Czaki:
I test it. I have no problem with cPython virtualenv. I also install pypy using pyenv, so it is outside the standard location.

henryiii:
What I meant was:

* where is Python.h for CPython venv? In venv or outside?
* If inside, is it missing for PyPy?

henryiii:
[...]

PR #502, opened 2020-12-27
Patching pypy venv 

Czaki:
Fix for the problem reported in #501


Czaki:
@henryiii @mattip
Could you check this? 

I do not understand why windows build pass, maybe my modification of the test is not enough?

mattip:
These changes are only needed for posix.

henryiii:
I think we should have a reminder somewhere that this can be reverted after the next PyPy release. @joerick?

Czaki:
> I think we should have a reminder somewhere that this can be reverted after the next PyPy release. @joerick?

one of the problem of this PR is that this code fails when cannot apply the patch. So It will fail when the user does not use pinned images, but will signal that there is need to remove patch after update. 

Or I should ignore fail of patch.

henryiii:
Can it check PyPy version?

Czaki:
> 
> 
> Can it check PyPy version?

maybe it could parse pypy string.

mattip:
`pypy -c "import sys; print(sys.pypy_version_info < (7,3,4))"` prints `True` for versions needing the patch

joerick:
> I think we should have a reminder somewhere that this can be reverted after the next PyPy release. @joerick?

Let's open a draft PR that reverts these changes once this is merged, that way we won't forget about it

henryiii:
> `pypy -c "import sys; print(sys.pypy_version_info < (7,3,4))"` prints `True` for versions needing the patch

You could use a `sys.exit` return value, but anyway this probably should be the thing the patch triggers on.

> Let's open a draft PR that reverts these changes once this is merged, that way we won't forget about it

Most users will use latest versions or the pinned versions, I think, so there would be no reason to still support 7.3.3 after we have a version pinning 7.3.4?

Czaki:
> Most users will use latest versions or the pinned versions, I think, so there would be no reason to still support 7.3.3 after we have a version pinning 7.3.4?

I one project I use prebuild docker images because build libraries take ~0.5h. Of course, I will know that I need immediately rebuild docker images, but maybe some other users decide for the same workflow.

joerick:
Ah, right. Well maybe we can leave it around until the next minor point release.

YannickJadoul:
Sorry, @Czaki, I was trying something, but by accident pushed to your branch instead of my local one!

YannickJadoul:
Not sure why GHA wasn't triggered, but this seems to work, and is a dramatic simplification: https://github.com/YannickJadoul/cibuildwheel/commit/615e9059ac22e954536df91c2c2bc5f3f5e51e00.
Here's the manual GHA run: https://github.com/YannickJadoul/cibuildwheel/runs/1622585851

Mind if I push to your branch, @Czaki, or do you prefer to pull from my branch yourself?

Czaki:
> Mind if I push to your branch, @Czaki, or do you prefer to pull from my branch yourself?

`-d` allows to patch the whole directory?

YannickJadoul:
> `-d` allows to patch the whole directory?

Officially, it says that flag changes to that directory before applying the patch. Then if you only pass one file, that's the patch file, and it will go looking for which files to patch (since they are in the diff, anyway)

henryiii:
Patches are not per-file, add “.patch” to the url of any pull request and you get a single file that describes the whole PR.

YannickJadoul:
@Czaki, so I'm pushing?

Czaki:
Push. I do not know why I have a problem with patch the whole directory (so I split on two files).

-

YannickJadoul:
I sneaked in another commit, removing the `pypy_patch` function, since it has become so small, and isn't really very informative to read (to me, at least; it doesn't say what patch or anything).

Feel free to discuss/remove, if necessary.

I also added a small test to `test_testing.py`:

```
    def test_time_to_remove_the_pypy_venv_patch(self):
        assert not hasattr(sys, "pypy_version_info") or sys.pypy_version_info < (7,3,4)
```

Not sure if this is a great plan, but it could work as this reminder? (We can't just easily test that in `cibuildwheel`, because we're not running on PyPy, there; at least, we don't know what we're running on, so probably CPython.)

joerick:
> I also added a small test to test_testing.py:

Good idea! But I think we were planning to leave the patch in for a couple releases, even after our pinned version has upgraded, because users might still choose to specify an older version via a docker image. So it's perhaps not appropriate. Anyway, since we can check sys.pypy_version_info before applying the patch, it's not causing much harm being there for a little while.

cibuildwheel/linux.py
```diff
@@ -138,6 +138,13 @@ def build(options: BuildOptions) -> None:
                     log.build_start(config.identifier)
 
                     dependency_constraint_flags: List[Union[str, PathLike]] = []
+                    if config.identifier.startswith("pp"):
+                        # Patch PyPy to make sure headers get installed into a venv
+                        patch_version = '_27' if config.version == '2.7' else ''
+                        patch_path = Path(__file__).absolute().parent / 'resources' / f'pypy_venv{patch_version}.patch'
+                        patch_docker_path = PurePath('/pypy_venv.patch')
+                        docker.copy_into(patch_path, patch_docker_path)
+                        docker.call(['patch', '-N', '-d', config.path, patch_docker_path])
```
joerick, line 147: This is much neater :)

YannickJadoul:
> Anyway, since we can check sys.pypy_version_info before applying the patch, it's not causing much harm being there for a little while.

Well, that's the part that's not obvious to do, because you need to check the version of the "guess" PyPy, rather than of the host Python.

cibuildwheel/linux.py
YannickJadoul: It's only slightly duplicated, now, in `macos.py`, since I removed the `pypy_patch` function. But I think I prefer this clarify.

Czaki:
I update the code to ignore the fail of patch on Linux. 
I also modify the test to fail only on macOS. But it could be committed, as patch operation should fail on the updated version of pypy.

YannickJadoul:
> I update the code to ignore the fail of patch on Linux.
> I also modify the test to fail only on macOS. But it could be committed, as patch operation should fail on the updated version of pypy.

Wait, why the difference between macOS and Linux? I don't like that these would behave differently actually.

Czaki:
> Wait, why the difference between macOS and Linux? I don't like that these would behave differently actually.

On macOS we hardcode PyPy version in code:
https://github.com/joerick/cibuildwheel/blob/4e17c275efff28bae8577d84515e78b62e204f55/cibuildwheel/macos.py#L44-L46

On Linux PyPy version depends on the selected docker container.

YannickJadoul:
> On macOS we hardcode PyPy version in code:
> 
> https://github.com/joerick/cibuildwheel/blob/4e17c275efff28bae8577d84515e78b62e204f55/cibuildwheel/macos.py#L44-L46
> 
> On Linux PyPy version depends on the selected docker container.

Right, but shouldn't we just say that the future version of `cibuildwheel` that support PyPy 7.3.4 doesn't support PyPy 7.3.3 anymore? If we don't suppport it on mac anymore, why would we support the docker image for Linux?

This is my suggestion to everyone, btw; I believe @Czaki is just implementing what's currently been discussed above. I'd just be confused by this, as user. If you update `cibuildwheel`, I don't see why you wouldn't update your pinned docker image...

Czaki:
> Right, but shouldn't we just say that the future version of `cibuildwheel` that support PyPy 7.3.4 doesn't support PyPy 7.3.3 anymore? If we don't support it on mac anymore, why would we support the docker image for Linux?

There are two questions. If version supporting 7.3.3 should support 7.3.4 (someone updates his docker image without the update of cibuildwheel, because of the rebuild of some dependency). And current implementation supports it. 

The second question is what to do when changing the base version of PyPy to 7.3.4. I personally prefer to not break compatibility immediately. Especially in the context of the last Cent Os changes.

YannickJadoul:
> There are two questions. If version supporting 7.3.3 should support 7.3.4 (someone updates his docker image without the update of cibuildwheel, because of the rebuild of some dependency). And current implementation supports it.

That's a good point though; I had only thought about backwards compatibility.

> The second question is what to do when changing the base version of PyPy to 7.3.4. I personally prefer to not break compatibility immediately. Especially in the context of the last Cent Os changes.

Well, but then you should stay on an older version of `cibuildwheel`, I'd say. It's pretty bad practice to build on a different PyPy version on macOS than on Linux, no?
But fair enough.

Czaki:
> Well, but then you should stay on an older version of `cibuildwheel`, I'd say. It's pretty bad practice to build on a different PyPy version on macOS than on Linux, no?
> But fair enough.

Maybe a fast drop is a good decision when adding proper information to the documentation.

joerick:
> It's pretty bad practice to build on a different PyPy version on macOS than on Linux, no?

Good point, I had not considered this. In that case I'd also be in favour of a quick drop once we have the bug fix in pypy.

YannickJadoul:
@Czaki, I think you're looking for

```
assert sys.platform != "darwin" or not hasattr(sys, "pypy_version_info") or sys.pypy_version_info < (7,3,4)
``` 

if you only want the last two parts tested on macOS.

YannickJadoul:
Something seems to be going horribly wrong, though? Tests just stall after printing `test/test_0_basic.py` :-(

Czaki:
> 
> 
> Something seems to be going horribly wrong, though? Tests just stall after printing `test/test_0_basic.py` :-(

I suggest wait. Maybe some download fail. I suggest restart tomorrow morning. Last changes does not affect test_0_basic.

henryiii:
All linux builds on all CIs are stalling out still. :(

Czaki:
The bug comes from @YannickJadoul patch.
```
patch -N -d /opt/python/pp27-pypy_73 -i /pypy_venv.patch 
can't find file to patch at input line 3
Perhaps you should have used the -p or --strip option?
The text leading up to this was:
--------------------------
|--- a/lib-python/2.7/sysconfig.py
|+++ b/lib-python/2.7/sysconfig.py
--------------------------
File to patch: 
Skip this patch? [y] n 
File to patch: 
```

Pure call CLI

```
patch -N -d /opt/python/pp27-pypy_73 /pypy_venv.patch
```

Block execution waiting on stdinput. 

So this is the problem why I split the patch into two files. 

It looks like macOS patch works differently than Linux one. Did we revert @YannickJadoul changes?

Czaki:
I have hope that I fix it. I suggest keeping `pytest-timeout` for better debug.

YannickJadoul:
This is weird. I tried this out, and things passed. What has changed meanwhile?

cibuildwheel/linux.py
```diff
@@ -138,6 +138,16 @@ def build(options: BuildOptions) -> None:
...
 
                     dependency_constraint_flags: List[Union[str, PathLike]] = []
+                    if config.identifier.startswith("pp"):
+                        # Patch PyPy to make sure headers get installed into a venv
+                        patch_version = '_27' if config.version == '2.7' else ''
+                        patch_path = Path(__file__).absolute().parent / 'resources' / f'pypy_venv{patch_version}.patch'
+                        patch_docker_path = PurePath('/pypy_venv.patch')
+                        docker.copy_into(patch_path, patch_docker_path)
+                        try:
+                            docker.call(['patch', '-N', '-p1', '-d', config.path, '-i', patch_docker_path])
```
YannickJadoul, line 148: Why `-N`, why `-i`?

> `-N  --forward  Ignore patches that appear to be reversed or already applied.`

I don't see the point of this? We know our patches should apply forward?

> `-i PATCHFILE  --input=PATCHFILE  Read patch from PATCHFILE instead of stdin.`

This was also not the issue in what you showed. The issue seems to have been the leading `a`/`b`?

Better apply one of these, then?

> `  -t  --batch  Ask no questions; skip bad-Prereq patches; assume reversed.`
> `  -f  --force  Like -t, but ignore bad-Prereq patches, and assume unreversed.`

YannickJadoul:
> I suggest keeping `pytest-timeout` for better debug.

That's going to be annoying for the occasional slow test? Why not just add it when debugging?

henryiii:
> That's going to be annoying for the occasional slow test? Why not just add it when debugging?

If a test suddenly takes 3x longer, isn't that a bug? As long as the timeout isn't too short so that it triggers on reasonable runs, there's no reason we need to wait 1-6 hours (depending on the CI) to fail?

YannickJadoul:
> If a test suddenly takes 3x longer, isn't that a bug? As long as the timeout isn't too short so that it triggers on reasonable runs, there's no reason we need to wait 1-6 hours (depending on the CI) to fail?

Probably. I was just thinking that we now already get the occasional spurious CI failure (just restarted a Travis job), so... myeah

cibuildwheel/linux.py
henryiii: The leading a/b character should be fixed by `-p1`. The `-i` (or `--input`) is correct, we want to specify an input patch, even if it works without it. I think `--force` is probably what you want here, though, since you never want it to ask questions, which I think `-N` still could do if it's not one of {unpatched, already patched}.

henryiii:
```suggestion
                            docker.call(['patch', '--force', '-p1', '-d', config.path, '--input', patch_docker_path])
```

Perhaps?

Czaki:
The main difference is that job killed by CI timeout does not provide any information. We do not know in which phase it stuck. Maybe timeout should be increased, but for me, it is better to get a job killed by pytest-timeout than CI timeout. I got more information.

cibuildwheel/linux.py
Czaki: `-i` is needed to specify the patch file. Patch on Linux assumes that the first file is one to patch. 

I see no difference in behavior between `-N` and `--force`. We have properly formated patch files.

YannickJadoul:
> The main difference is that job killed by CI timeout does not provide any information. We do not know in which phase it stuck. Maybe timeout should be increased, but for me, it is better to get a job killed by pytest-timeout than CI timeout. I got more information.

That's true, that's why I suggested to keep it as way of debugging. But fair enough, if the timeout is long enough, it can stay as well

joerick:
I'm happy with keeping the timeout. I'm not particularly familiar with `patch`, but `--force` seems more appropriate in our situation? Other than that this looks good 👍

joerick (approved):

PR #504, opened 2020-12-31
refactor: setup.cfg and dev requirements

henryiii:
There are two changes in this PR:

First, the setup.py code has been [modernized into setup.cfg](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html), supported by setuptools since 2016. Note this _only_ affects the source; the wheel produced should be identical; neither setup.py nor setup.cfg even make it to the wheel. The declarative nature tends to be simpler and reduce errors, for example with reading files and decoding them.

By the way, you can remove the `setup.py` file entirely instead of leaving the simplest one (`from setuptools import setup; setup()`. I've generally always left that file, but I've been temped to remove it a few times recently because it makes people think it's okay to run `python setup.py <stuff>`, which it's not - that's an implementation detail. But [~~if I recall correctly~~](https://setuptools.readthedocs.io/en/latest/setuptools.html#setup-cfg-only-projects), you lose the ability to do `pip -e .`, since they haven't worked out a general form for that yet. So I left it here as usual for now.

Also by the way, the [PEP](https://www.python.org/dev/peps/pep-0631/) to allow this to be put into pyproject.toml was accepted provisionally, but hasn't been implemented yet in setuptools or other systems (and would require a really new setuptools for a little bit after it's in, of course).

The second change is replacing `-r requirements-dev.txt` with an extra, specifically `-e .[dev]`. This is nicer and more elegant than putting `-e .` inside a requirements file!

I've restructured the package finding around best practices - if we ever add a subpackage for example, it should get picked up now.

TODO:

- [x] Update bump version script
- [x] Update .readthedocs.yml
- [x] Update or add dev docs


YannickJadoul:
> This is nicer and more elegant than putting `-e .` inside a requirements file!

Can't we just keep `requirement-dev.txt` (I kind of prefer to have that file; much easier to look for than somewhere in `setup.cfg`, if you ask me) and have `python -m pip install -e . -r requirements-dev.txt pytest-custom-exit-code`?

setup.cfg
```diff
@@ -1,3 +1,63 @@
+[metadata]
+
+name = cibuildwheel
+version = 1.7.2
```
YannickJadoul, line 4: This needs to change in the `bin/bump_version.py`

henryiii:
The problem with requirements files is that they are randomly named and tend to not be the best solution for any issue. I think they are tending toward being a lock-file, with a set of exact requirements that you can maintain via dependabot or something similar.

Personally, if there is a test requirements file, my preferred name would be `tests/requirements.txt`, as it keeps the main directory cleaner and is not as arbitrarily named as adding or prefixing "dev" or "test". And of course, drop the surprising `-e .` inside it.

But extras are even more standard; simply add `[dev]` and you get all the dev requirements; you don't have to do an editable install or even a source install; you could do this on the PyPI version as well. Also, there are exactly two places to look for extras (setup.cfg and setup.py), vs. having to look throughout the library for a randomly named requirements file. (You can also programmatically list extras with importlib_metadata, but there's not a pip option for it, sadly; it's been discussed but ideally you'd list extras before installing, rather than afterwards).

Czaki:
> (You can also programmatically list extras with importlib_metadata, but there's not a pip option for it, sadly; it's been discussed but ideally you'd list extras before installing, rather than afterwards).

`pip-compile` may allow doing this.

YannickJadoul:
> I think they are tending toward being a lock-file, with a set of exact requirements that you can maintain via dependabot or something similar.

For a tool like these, that's not deploying to some webserver or something, I tend to find these exactly-pinned versions less interesting.

> But extras are even more standard; simply add `[dev]` and you get all the dev requirements; you don't have to do an editable install or even a source install; you could do this on the PyPI version as well.

Also fine, it was just a suggestion; I tend to sometime run `pip install -U -r some_kind_of_requirement.txt`, but maybe it's time to update that workflow as well.

henryiii:
I would not exactly pin, which is one reason I don't think requirements.txt is really the right format.

Run  `pip install -U -e .[dev]` then, less typing and no searching for a ?/requirements-X.txt file. :)

You _can_ do both, but you have to parse and read the file in setup.py, which I don't like doing.

joerick:
Personally, I find the presence of a `requirements-dev.txt` more approachable when looking at a library. You can often expect to be able to do `python3 -m venv env ; . env/bin/activate ; pip install -r requirements-dev.txt ; pytest` or similar, without checking setup.cfg or the docs for how to set up a dev environment. 

But you may well be right that it's becoming an old fashioned practice @henryiii, in which case I'd be open to having both - a `[dev]` extra with the dev tools, and a `requirements-dev.txt` that only lists `-e .[dev]`.

joerick:
But the change to setup.cfg is great!

henryiii:
I've split it into two parts, since mixing the commits was a bad idea. I can add that, yes. Forgot about doing it in the reversed order! Do you want the file to be called `requirements-dev.txt` or `tests/requirements.txt`?

joerick:
`requirements-dev.txt` seems more 'normal', to me.

joerick (approved):

henryiii:
Hmm, that didn't merge when green. :'(

[I _think_ GitHub is adding something like this natively?](https://github.blog/2020-12-08-new-from-universe-2020-dark-mode-github-sponsors-for-companies-and-more/#✨-improved-daily-experience)

YannickJadoul:
@henryiii, this is supposed to do that job: https://github.com/joerick/cibuildwheel/blob/master/.github/auto-merge.yml

PR #506, opened 2021-01-01
Macos patch fix

Czaki:
On MacOS 10.14 current patch syntax does not work (I build code with OpenMP, which needs libomp from `brew`).

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -109,7 +109,7 @@ def install_pypy(version: str, url: str) -> Path:
         # Patch PyPy to make sure headers get installed into a venv
         patch_version = '_27' if version == '2.7' else ''
         patch_path = Path(__file__).absolute().parent / 'resources' / f'pypy_venv{patch_version}.patch'
-        call(['patch', '--force', '-d', installation_path, '-i', patch_path])
+        call(['patch', '--force', '-p1', '-d', installation_path, '-i', patch_path])
 
     installation_bin_path = installation_path / 'bin'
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Cross referencing #502.

YannickJadoul (approved):
This only increases consistency, so sure!

Also, don't we test on macOS 10.14, then?

Czaki:
I found test only on 10.15

joerick:
Once merged I'm happy to do a hotfix 1.7.4 if that's useful @czaki?

Czaki:
> Once merged I'm happy to do a hotfix 1.7.4 if that's useful @Czaki?

It could be useful for any person who build pypy wheels on macos 10.14. Without it build stuck. 

@joerick  What did you think about change one of macos in test on 10.14? I this PR on in others?

henryiii:
Are there CI systems running 10.14? I'm pretty sure there must be (GHA/Azure do not have an official 10.14 advertised runner anymore, but assuming some of the other ones are not?) If so, I think it's a great idea to backport the fix and release a 1.7.4 while we work on archs & Universal2. If not, still wouldn't be a bad idea.

Czaki:
I run 10.14 on Azure Pipelines.

henryiii (approved):

joerick:
10.14 doesn't seem to be advertised on GHA/Azure anymore, so I don't think we need to test with it.

joerick:
Released as v1.7.4 (note that release doesn't include the CIBW_ARCHS work)

PR #507, opened 2021-01-01
feat: Windows filtering and sets

henryiii:
Followup to #482, use sets instead of lists for Archs, and make Windows handling behave more like Linux, where one arch will only use that arch. (Eventually, maybe could include ARM for windows, perhaps?).

No docs changes (yet).

I converted to sets, then added the changes from @mayeut.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -4,15 +4,14 @@ import shutil
 import subprocess
 import sys
 import tempfile
-import textwrap
 from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Sequence
 
 from .environment import ParsedEnvironment
 from .logger import log
-from .util import (Architecture, BuildOptions, BuildSelector, NonPlatformWheelError,
+from .util import (BuildOptions, BuildSelector, NonPlatformWheelError,
                    download, get_build_verbosity_extra_flags, get_pip_script,
-                   install_certifi_script, prepare_command)
+                   install_certifi_script, prepare_command, allowed_architectures_check)
 from .typing import PathOrStr
 
 
@@ -187,11 +186,7 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
-    if not options.architectures <= {Architecture.x86_64}:
-        raise ValueError(textwrap.dedent(f'''
-            Invalid archs option {options.architectures}. macOS only supports x86_64 for the moment.
-            If you want to set emulation architectures on Linux, use CIBW_ARCHS_LINUX instead.
-        '''))
+    allowed_architectures_check("macos", options)
 
     temp_dir = Path(tempfile.mkdtemp(prefix='cibuildwheel'))
     built_wheel_dir = temp_dir / 'built_wheel'
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -187,7 +187,7 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
-    if options.architectures != [Architecture.x86_64]:
+    if not options.architectures <= {Architecture.x86_64}:
         raise ValueError(textwrap.dedent(f'''
             Invalid archs option {options.architectures}. macOS only supports x86_64 for the moment.
             If you want to set emulation architectures on Linux, use CIBW_ARCHS_LINUX instead.
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -2,13 +2,14 @@ import os
 import platform as platform_module
 import re
 import ssl
+import functools
 import textwrap
 import urllib.request
 from enum import Enum
 from fnmatch import fnmatch
 from pathlib import Path
 from time import sleep
-from typing import Dict, List, NamedTuple, Optional
+from typing import Dict, List, NamedTuple, Optional, Set
 
 import certifi
 
@@ -124,7 +125,10 @@ class DependencyConstraints:
         return f'{self.__class__.__name__}{self.base_file_path!r})'
 
 
+@functools.total_ordering
 class Architecture(Enum):
+    value: str
+
     # mac/linux archs
     x86_64 = 'x86_64'
     i686 = 'i686'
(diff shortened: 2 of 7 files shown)


cibuildwheel/macos.py
```diff
@@ -187,7 +187,7 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
-    if options.architectures != [Architecture.x86_64]:
+    if options.architectures <= {Architecture.x86_64}:
```
YannickJadoul, line 190: You've missed the `not`, I think

YannickJadoul: (oh, this is only 8 minutes old; sorry for the impatience!)

henryiii: Yes, I did. 😳 

Working on mypy at the moment, it should have been able to point out the issues in the tests comparing sets and lists before runtime.

henryiii: MyPy support is kind of a mess. There's a typo in the config so everything that's supposed to be set is not being set. :/ Will have to manually fix for now, then work on fixing the MyPy config. Also, mypy support in "test" is harder, since we run the tests on Python 2 :/.

joerick: BTW we don't run `test` on Python 2, only the generated projects run on Python 2.

henryiii:
Any opinions? This unifies and simplifies the handling of architectures a bit, but shouldn't be controversial? Unless we want docs changes?

cibuildwheel/macos.py
```diff
@@ -187,7 +187,7 @@ def setup_python(python_configuration: PythonConfiguration,
 
 
 def build(options: BuildOptions) -> None:
-    if options.architectures != [Architecture.x86_64]:
+    if not options.architectures <= {Architecture.x86_64}:
```
joerick, line 190: Apologies if I missed some discussion related to this, but I'm a little puzzled here - if a user requests an arch that we can't build, it should be an error, I think.

cibuildwheel/util.py
```diff
@@ -136,33 +140,37 @@ class Architecture(Enum):
     x86 = 'x86'
     AMD64 = 'AMD64'
 
+    # Allow this to be sorted
+    def __lt__(self, other: "Architecture") -> bool:
+        return self.value < other.value
+
```
YannickJadoul, line 146: To nitpick on sómething: I'm not sure I like alphabetic ordering, here. Isn't there a way we can keep the same kind of ordering as usual?

YannickJadoul: (I.e., the same as in `get_python_configurations`/the order they're listed here)

cibuildwheel/macos.py
joerick: Ah, I think I understand, this matches the behaviour on other platforms, where unused archs are just ignored.

Hmm. I wonder if it would be better to raise an error in these cases too? e.g. setting `CIBW_ARCHS_LINUX=AMD64,x86` should probably be an error? Because currently that's going to silently build nothing - probably not what you expect.

cibuildwheel/util.py
YannickJadoul: What about keeping `allowed_architectures` a list, as it was?

joerick (commented):
Other than the one note on unsupported archs, this looks good.

YannickJadoul (approved):
Oops, sorry, I thought I had already seen and reviewed this!

cibuildwheel/macos.py
YannickJadoul: Wasn't this what @mayeut suggested in https://github.com/joerick/cibuildwheel/pull/482#discussion_r550216651, though?

henryiii: So we should add a check on all these to watch for an empty set.

cibuildwheel/util.py
henryiii: It’s treated as a set other places, so this is more natural I think. I could probably keep the same order as defined. Though arches don’t have an order? Order by age?

YannickJadoul: I thought of "order by 'this is the order in which we always list them'", i.e. "x86_64 i686 aarch64 ppc64le s390x" (plus Windows at the end AFAIC).

> It’s treated as a set other places, so this is more natural I think.

It's just a local thing, when you're testing/printing in `build` (and sorting to get a list ;-) ), so yes, you're right, but ... eh, that might just make thing easier?

I guess it's not a blocker, but I find it weird that `aarch64` would suddenly suddenly first when printing a message on Linux.

cibuildwheel/macos.py
YannickJadoul: Or have the same check on Linux?

In @joerick's example, `not {Architecture.AMD64,Architecture.x86} <= all_those_linux_archs` would be `True`

cibuildwheel/util.py
henryiii: Let me give it a shot.

cibuildwheel/macos.py
henryiii: If there's no way that `options.architectures` can be an empty set, this should always be fine? It's a subset test, the left side must be a subset of the right side.

cibuildwheel/util.py
henryiii: Done, sorts in definition order now.

YannickJadoul: Thanks, that's great! :-)

cibuildwheel/macos.py
henryiii: https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset

There is a way to spell it, but I think operators are quite elegant with sets. There are some tests (such as `<`, proper subset) that are not available with a method, and `set | set` will be much more common in Python now that it's also supported with dictionaries, and in 3.10 will be supported with Types. The comparisons are rather cute, as you can easily see which is subset of the other. But if you think it's clearer, we can spell it out. I think this form is "fun". :)

Okay, so will add the missing check on Linux and will add checks for empty sets.

joerick: Now that I understand it, it makes total sense. For some reason I was misreading it as 'contains at least x86_64', but I now see that it's 'only contains x86_64'.

henryiii:
Pulled arch checking into one place, and now they all behave the same way - specifying an empty set or an arch that's not allowed is an error.

YannickJadoul:
> Pulled arch checking into one place, and now they all behave the same way

That's great.

Already approved, but ... yep, good to merge, if you ask me!

henryiii:
I didn't order the includes correctly in one place, but didn't want to rebuild after pushing so added #518 instead. I can fix it though.

cibuildwheel/util.py
```diff
@@ -2,19 +2,26 @@
...
-from typing import Dict, List, NamedTuple, Optional
+from typing import Dict, List, NamedTuple, Optional, Set
 
 import certifi
 
 from .environment import ParsedEnvironment
 from .typing import PathOrStr
 
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
```
joerick, line 21: Ah, that's handy

joerick (approved):
Looks good!

cibuildwheel/util.py
henryiii: Everything that goes into typing that can be backported goes here. :)

Issue #508, opened 2021-01-01
[question] Raise minimum required Python to 3.7


henryiii:
I found a typo in the mypy configuration; we are not actually imposing anything on mypy. If we enable what we are trying to check, we have dozens of typing errors, mostly due to missing or incomplete types. It would be nice to expand to checking some of the testing too, now that pytest supports typing and it would help with issues like #507 , where running mypy locally is easy, while running the tests locally is harder (except if CIBW_PLATFORM is set to linux). Fixing the incomplete types would be significantly easier if I can just bump the required Python to 3.7 and use `from __future__ import annotations`; would that be something we could do for the next version? Otherwise, I'll just add TYPE_CHECKING :/. It would also let us remove some stringified annotations.

joerick:
I guess 3.7 is available across all the CIs we support, so I don't see why not! If we're changing the minimum, we could even skip one and go to 3.8?

henryiii:
I don't think we get that much from 3.8 at the moment (though, admittedly, I've built a 3.7+ codebase before, but not a 3.8+ one, so not as familiar with the changes in practice). Though, admittedly, supporting 3.7 if all CI systems have 3.8 might be a little silly.

`from __future__ import annotations` make it possible to keep up with mypy changes instead of Python ones for annotation syntax, which is fantastic. When mypy releases 0.800, we would be able to use Python 3.9 `list[str]` syntax in annotations without requiring 3.9. :) Did you see there's a PR to build mypy with cibuildwheel?

joerick:
> When mypy releases 0.800, we would be able to use Python 3.9 `list[str]` syntax in annotations without requiring 3.9

I'd love this. Also, would we be able to use `str|Path` instead of `Union[str, Path]`? The Union thing always seemed extra verbose.

> Did you see there's a PR to build mypy with cibuildwheel?

Yes :) very cool. 

> I don't think we get that much from 3.8 at the moment

One thing I just came across today, which would be very handy - [literal types](https://mypy.readthedocs.io/en/stable/literal_types.html).

henryiii:
> Also, would we be able to use `str|Path`

I'm thinking that will take a bit longer, personally my goal would be to get a 3.9 compatible version out, then start adding the 3.10 features like this one. But not following it all that closely in MyPy (and I'd think it might be pretty easy, but could be very wrong). They might want to wait till after 3.10 is out to make sure it doesn't get pulled.

[...]

PR #509, opened 2021-01-02
fix: enable and fix full type checking

henryiii:
Discovered by #507; we are currently not actually enabling any of the settings it looks like we are enabling due to a missing `.*`. Fixes the various errors, and reworks the Colors/Symbols to be statically typeable.﻿

Adding pytest does slow down the type checking a bit; we could revert to ignoring pytest if the pre-commit run is too slow.

I've also added a py.typed file (not sure if it's in the SDist without a manifest), but it's nice to say we are typed, even if we don't really expect to be used as a library?

Could be simplified a bit by #508.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,13 @@
 repos:
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.8.3
+  rev: 3.8.4
   hooks:
   - id: flake8
     exclude: cibuildwheel/resources/
     additional_dependencies: [flake8-bugbear]
 
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v3.2.0
+  rev: v3.4.0
   hooks:
   - id: check-case-conflict
   - id: check-merge-conflict
@@ -18,7 +18,7 @@ repos:
   - id: trailing-whitespace
 
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.782
+  rev: v0.790
   hooks:
   - id: mypy
     files: ^(cibuildwheel/|test/|bin/projects.py|unit_test/)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,4 +23,3 @@ repos:
   - id: mypy
     files: ^(cibuildwheel/|test/|bin/projects.py|unit_test/)
     pass_filenames: false
-    additional_dependencies: [pytest>=6.0]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,5 +21,6 @@ repos:
   rev: v0.782
   hooks:
   - id: mypy
-    files: ^(cibuildwheel/|test/|bin/projects.py)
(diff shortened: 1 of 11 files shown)


henryiii:
@joerick Can you see if there's a setting to enable CircleCI to build on forked PRs that's disabled, perhaps? https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/

joerick:
I actually noticed this a couple weeks back, probably should have said something but thought I'd hold back that revelation it until things were a little quieter around here! 

Thank you @henryiii for performing all these fixes. This might be a good opportunity to consider which of these 'strict' rules we actually want? Although, looking at the diff it's not too crazy, I'm mostly wondering about compulsory `-> None` annotations or the changes to the `Colors` `Symbols` objects.

Other that that, maybe we should define our own `PathLike` (`os.PathLike[str]`) and `StrOrPath` (`Union[str, os.PathLike[str]]`) types? I think that might clean up a lot of the existing annotations

henryiii:
Returning `-> None` is useful information; it helps you not `return self` or something like that, it also protects you from subclassing and accidentally changing an output. But most importantly, it lets you immediately tell that a function is not supposed to return something, making the code clearer.

Changing `Colors` and `Symbols` from classes within classes that have magic accessors to simple classes doesn't seem to bad to me - thinking statically promotes simper, clearer design, I think. If we have Python 3.7, we might be able to come up with a dataclass that is simpler (but I rather doubt it). We could also use a color library (I've written one before, `plumbum.colorlib`, but sadly it's not available standalone from plumbum, and you are using my competitor Click instead - looks like they added [color support too at some point](https://click.palletsprojects.com/en/7.x/utils/#ansi-colors), though I'm sure plumbum.colors runs circles around it :P ).

MyPy supports gradual typing, but that's mostly so you can turn it on bit by bit, not that partially typed should be a goal. I usually run `mypy --strict` to see how far off something is from fully typed.

henryiii:
I pulled out the types to a file, makes a nice simplification (especially since we only use `PathOrStr` everywhere). Dropped pytest from the pre-commit environment and allow it to be missing - it changes a 1-2 second type check into a 10 second type check or so, for basically no gain.

joerick (approved):
Thank you @henryiii ! I feel a little guilty for having somebody else tidy up my mess! Happy to leave things on strict settings for now.

It is interesting how typing changes what's considered 'neat' in programming practices. I'm thinking mostly about the Colors/Symbols thing. Generally I do think it's a good thing, but it does sometimes feel that one is being pushed in a more 'uniform'/'standard' direction (less 'fun'?). 

Anyway... it would be good to get this in ASAP as there might be conflicts that would be more easily resolved sooner rather than later.

henryiii:
> more 'uniform'/'standard' direction (less 'fun'?)

Yes. But often, that's better long-term design, easier to read/maintain, less likely to break, easier to use downstream due to the typing. While it often can be more repetitive, you also have more checks verifying that you remembered to change everything.

There are special cases where "fun" is more important, and then you can just "Any"-it and go full dynamic. But that should probably be the point of the library, rather than just part of implementing something else.

There's still plenty to have fun with! Restricting the rules just focuses your efforts on the important parts.

PR #510, opened 2021-01-02
style: pyupgrade

henryiii:
This adds a pyupgrade check, which makes sure old-style constructs are not introduced / cleaned up. When we upgrade Python versions, we could bump the minimum here too.﻿

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,4 @@
 repos:
-- repo: https://gitlab.com/pycqa/flake8
-  rev: 3.8.4
-  hooks:
-  - id: flake8
-    exclude: cibuildwheel/resources/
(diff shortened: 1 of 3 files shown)


YannickJadoul (approved):
Oh, this is cool! I had no idea it existed!!

henryiii:
Everything @asottile writes is cool. ;) It's really fantastic for moving a code base from Python 2+3 to Python 3, though it helps for point updates a bit too, and here, in keeping a code base nice and fresh. :)

joerick (approved):
nice!

henryiii:
The emulation test died after a 2 hour time out (and I think after doing nothing). Not related to the PR, though I can go ahead and reorder the checks - checks that modify should always be before checks that do not.

PR #511, opened 2021-01-02
refactor: identifiers in a file

### LINKED ISSUE #431: Configuration files for download links

henryiii:
Fixes #431, should make #496 easier. I did not provide a way to override the files. I am now using [importlib.resources](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-importlib-resources) here. ~~instead of messing with the backport, we can wait until #508 and then just use the stdlib.﻿~~ Edit: the 3.9 version is much nicer and less intrusive, using that.


YannickJadoul:
To nitpick, `pinned_docker_images.cfg` is read with `configparser.ConfigParser`. Do we have a reason to pick `toml` for this?

henryiii:
See https://github.com/joerick/cibuildwheel/pull/496#issuecomment-750459371

And TOML is a real format, while `.cfg` is a rather poorly defined thing. And we already require toml (YAML would have been _much_ more natural - I'm kind of working around the fact that TOML/cfg has to have keys, while YAML could have just been a list. Really is ugly when you have the dual Python 2.7 builds with matching identifiers.

YannickJadoul:
> See [#496 (comment)](https://github.com/joerick/cibuildwheel/pull/496#issuecomment-750459371)

Right! I also says

> similar to the manylinux one at cibuildwheel/resources/pinned_docker_images.cfg

but fair enough ;-)

Would it then make sense to move `pinned_docker_images.cfg` to `toml` as well (maybe in a follow-up)? These formats might look too similar to both keep around, no?

> Really is ugly when you have the dual Python 2.7 builds with matching identifiers.

"ugly" and "Python 2.7" in the same sentence? You must be joking! ;-)

But yes, the overview is completely gone in these files :-(

cibuildwheel/util.py
```diff
@@ -34,6 +35,18 @@ def get_build_verbosity_extra_flags(level: int) -> List[str]:
...
 
 
+def read_python_configs(input_path: Union[str, os.PathLike]) -> List[Dict[str, str]]:
+    loaded_file = toml.load(input_path)
+    results: List[Dict[str, str]] = []
+
+    for k, v in loaded_file.items():
+        results.append({'identifier': k, **v})
+        if 'cp27-manylinux_' in k:
+            results.append({'identifier': k, 'version': v['version'], 'path_str': f'{v["path_str"]}u'})
```
henryiii, line 45: This is because we use TOML instead of YAML, we can't have a list with dicts.

Would json be all that bad?

henryiii:
> Would it then make sense to move pinned_docker_images.cfg to toml as well

I think we should match them up, yes. If we don't want to add yaml as a dependency, what about using json?

henryiii:
TOML (basically the same as cfg):
```toml
[cp39-manylinux_aarch64]
version="3.9"
path_str="/opt/python/cp39-cp39"
```

YAML:
```yaml
- identifier: cp39-manylinux_aarch64
  version: 3.9
  path_str: /opt/python/cp39-cp39
```

JSON (technically valid YAML too):
```json
{"identifier": "cp39-manylinux_aarch64", "version": "3.9", "path_str": "/opt/python/cp39-cp39"},
```

YannickJadoul:
The JSON looks best to me, indeed, the nice thing being that if you have a list of those, they will immediately show you the versions without scrolling between blocks.

Funnily enough, none of them look as good as our Python `namedtuple`.

(This might of course also just be me, being used to the current format in Python code, and not yet to the TOML version.)

henryiii:
This seems more like data to me, always seems odd to have this sitting inside the .py files, especially having to edit them for updates. And if we add the ability to update these files programmatically, that’s much better as a data file!

YAML would allow:

```yaml
- {identifier: cp39-manylinux_aarch64", version: 3.9, path_str: /opt/python/cp39-cp39}
```

not a fan of mixing yaml and json styles but would probably be cleanest. But adds a dependency.

we could shorten identifier or path_str.

YannickJadoul:
> not a fan of mixing yaml and json styles but would probably be cleanest.

Me neither, but this looks nice.

> But adds a dependency. And likely hard to auto generate for an update.

Well, with something like this we're pretty close, no?

```json
{
"linux": [
    ...
    {"identifier": "cp39-manylinux_x86_64", "version": "3.9", "path_str": "/opt/python/cp39-cp39"},
    {"identifier": "cp39-manylinux_i686", "version": "3.9", "path_str": "/opt/python/cp39-cp39"},
    ...
    {"identifier": "cp39-manylinux_aarch64", "version": "3.9", "path_str": "/opt/python/cp39-cp39"},
],
...
}
```

I'm indeed not sure whether adding `yaml` is worth it (though it should be easy enough); up to @joerick, that one, I'd say.

YannickJadoul:
For the record, I did ask about `json`/`yaml` in #256, but seems we got caught up in another discussion: https://github.com/joerick/cibuildwheel/pull/256/files/8a450b76550d6331de66905fa8e18c7f54eb8e1b#diff-cdf254321017f9a32499135cace65adc9085efa0eb28e54902f12e9968c7dd54

henryiii:
Questions for @joerick:

* One file or three
* Format (JSON, YAML, TOML with current workaround, cfg)
* Should we change `pinned_docker_images.cfg` to match if we pick on of the others?

If you want it to be a vote, I ordered the format list above by my preferences.

joerick:
> * One file or three

The PythonConfiguration objects for each platform have different schemas, so I think they should be stored in different files.

> * Format (JSON, YAML, TOML with current workaround, cfg with same workaround)

I don't think we should be adding another runtime dependency to parse an internal config file. TOML is currently only needed for the cp35 windows workaround, so it might be dropped again after that. I'd be more in favour of configparser or JSON. Generally speaking I don't much enjoy JSON as a user-facing format, it's very brittle to edit directly.

I think we should be using a list of configs, not the dict of 'identifier': {...attributes}. Ordering is important, and as you've found, on linux there are two builds with the same identifier. That rules out configparser, because it doesn't support lists. TOML does with `[[array_name]]`, but it's probably not worth it here.

So, it's probably back to JSON. But, for a bit of fun, I tried out https://github.com/Instagram/LibCST to see if there was a pure-python way to do it. And honestly, it's not too crazy. Here's a script that patches the urls in macos.py:

```python
import libcst as cst


class UpgradeTransformer(cst.CSTTransformer):
    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call) -> cst.BaseExpression:
        node = updated_node

        if not (isinstance(node.func, cst.Name) and node.func.value == 'PythonConfiguration'):
            return node

        for arg in node.args:
            assert arg.keyword is not None

            if arg.keyword.value == 'url':
                assert isinstance(arg.value, cst.SimpleString)
                url = arg.value.evaluated_value

                print(url)
                # patch the url here
                url += '#patched'

                node = node.with_deep_changes(arg.value, value=f"'{url}'")

        return node


macos_cst = cst.parse_module(open('cibuildwheel/macos.py').read())
macos_cst = macos_cst.visit(UpgradeTransformer())
print(macos_cst.code)
```

This could be used to update the NamedTuples in-place, or another option could be to use something like the above with a standalone configs file, like `macos_python_configs.py`.

I'm probably too close to this to tell if it's a good idea, what do you guys think?

henryiii:
I'd rather a config format (JSON) over editing Python files. I'm not a fan of Python files for config; Python generally should be code, and data should be in a data format. I would expect to be able to take a Python file and add arbitrary Python code, which such a parser as you've shown would choke on.

> I think we should be using a list of configs, not the dict of 'identifier': {...attributes}. Ordering is important, and as you've found, on linux there are two builds with the same identifier.

That's what I mean with "the current workaround" - I was referring to the fact that it was a dict rather than a list. TOML allows an array inside a section, yes, but the values of that array are not allowed to be dicts, I believe - only simple values.

YannickJadoul:
I definitely like the "fun" part, but I'm not entirely sure it's the right match here? Also, if you didn't want to introduce another dependency, `libcst` is a bit annoying, there ;-)

Not sure I'm convinced of that myself, but we don't need to per se follow a fancy "everything's possible" kind of standard, ofc. Something like

```
cp39-manylinux_x86_64, 3.9, /opt/python/cp39-cp39
cp39-manylinux_i686, 3.9, /opt/python/cp39-cp39
cp39-manylinux_aarch64, 3.9, /opt/python/cp39-cp39
```

might perfectly server our purpose for now to populate the existing `NamedTuple`, won't be anything hard to parse (some `split()` and `strip()`?), and can be easily replaced by something more fancy once we do suddenly want `separator`s (could be `;` or `\t` or anything else) or spaces in our strings.

henryiii:
This is called `csv`, and it's got a parser in the standard library. ;)

YannickJadoul:
I know, but I didn't want to pin us to `csv` per se ;-)
Also, never been a great fan of the CSV parser in the standard library, since it's still pretty verbose. But yes.

YannickJadoul:
Actually, I stand corrected. It's just a oneliner:

```
list(csv.DictReader(open("test.csv", 'r'), fieldnames=["identifier", "version", "path_str"]))
```

(Except for those spaces behind the comma; we'd have to write `cp39-manylinux_x86_64,3.9, /opt/python/cp39-cp39`)

henryiii:
Actually, maybe toml does support inline dicts, I just noticed this: https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax

YannickJadoul:
Seems this also works with pytoml yes:

```
[x86_64]
cp38 = {identifier = "cp38-manylinux_x86_64", version = "3.8", path_str = "/opt$
cp39 = {identifier = "cp39-manylinux_x86_64", version = "3.9", path_str = "/opt$
```

Just a pity of those `cp38` and `cp39` keys (since toml doesn't have arrays).

henryiii:
Yes it does.

henryiii:
Try:

```toml

[linux]
vals = [
    {identifier = "cp38-manylinux_x86_64", version = "3.8", path_str = "/opt/python/cp38-cp38"},
    {identifier = "cp39-manylinux_x86_64", version = "3.9", path_str = "/opt/python/cp39-cp39"},
]
```

I can check later.

YannickJadoul:
> Yes it does. Just not at top level.

Once again, you're right. I hadn't tired that!

Well, I'd be up for TOML, then, actually, if @joerick doesn't mind the extra dependency. if not, I do kind of like the simplicity of the CSV (or similar) version.

henryiii:
I like JSON, CSV, and this form of TOML pretty equally, with a minor preference on the JSON still.

Currently working on a cookie cutter and the specs for it all all JSON. It's not unusable. :)

I'd probably do something like:

```toml
[tool.cibw.build-platforms]
linux = [
    {...},
```
That way, you could imagine them being specifiable in a single file if someone wanted to.

henryiii:
CSV wouldn't work, because they are variable length. That leaves JSON and TOML. The TOML I have right now is not bad at all, IMO. Though JSON would also work.

cibuildwheel/resources/windows.toml
```diff
@@ -0,0 +1,18 @@
...
+  {identifier="cp37-win32", version="3.7.9", arch="32"},
+  {identifier="cp37-win_amd64", version="3.7.9", arch="64"},
+  {identifier="cp38-win32", version="3.8.7", arch="32"},
+  {identifier="cp38-win_amd64", version="3.8.7", arch="64"},
+  {identifier="cp39-win32", version="3.9.1", arch="32"},
+  {identifier="cp39-win_amd64", version="3.9.1", arch="64"},
+  {identifier="pp27-win32", version="2.7", arch="32", url="https://downloads.python.org/pypy/pypy2.7-v7.3.3-win32.zip"},
+  {identifier="pp36-win32", version="3.6", arch="32", url="https://downloads.python.org/pypy/pypy3.6-v7.3.3-win32.zip"},
+  {identifier="pp37-win32", version="3.7", arch="32", url="https://downloads.python.org/pypy/pypy3.7-v7.3.3-win32.zip"},
+]
```
henryiii, line 18: This is an example of what it looks like.

YannickJadoul:
> CSV wouldn't work, because they are variable length.

It will if you have multiple files per OS, as @joerick said he preferred anyway? The `url` on Windows can just stay empty, no?

But fair enough, all my concerns about readability and getting an overview are gone, indeed, in the current form! So that just leaves the dependency thing.

joerick:
This TOML is looking very nice, now :) I think I might be swayed over. I also really like the way it slots into the namedtuple using the `PythonConfiguration(**items)` syntax. Does this way of writing it survive a round-trip through `toml.load()`, `toml.dump()`?

If those headers are required, (I know this is me changing my mind), perhaps we should put them all in one file? Then the headers actually mean something.

henryiii:
> Does this way of writing it survive a round-trip

Excellent question. With some customization and a few extra spaces, it can be done. Will commit an example tomorrow.

henryiii:
One benefit of json + one file here, by the way, it would make it easy to write a parser for it in javascript that would allow a user to type into a box and see the results live in the docs (never tried that, but should be possible, I think?). TOML or any other format would be harder to parse in JS (though you could always just read it with toml and stick a simple list JSON into the docs during generation, so really, not a big issue).

Going to push my toml version soon.

unit_test/test_build_ids.py
```diff
@@ -0,0 +1,28 @@
+import toml
+from toml.encoder import TomlEncoder
+
+from cibuildwheel.util import resources_dir
+
+
+class InlineArrayDictEncoder(TomlEncoder):
```
YannickJadoul, line 7: Hah, wow!

henryiii: Yeah, the libraries for TOML are not, um, amazing examples of great libraries. There was supposed to be a way to preserve inline status but it doesn't work for inline arrays, the beginning of the work seems to be there from several years back, but unfinished. So we can just handle it here. Or use JSON. Though we'd probably have to control the output a bit there too to keep it nice.

henryiii:
I've put an example roundtrip (the linux one and a test for now). What do you think? Should I go with this, in a single file?

unit_test/test_build_ids.py
YannickJadoul: Still nice that it's possible.

However, seeing this: is it very important that we can roundtrip?

YannickJadoul: Oh, of course, if we want to automate this in the next step, it is :-)

YannickJadoul:
Now that we have this nice formatting, I'm up for a single file. I think it could enhance updates, when now, you need to pass 3 files. Though one file is not a must-have for me, either.

henryiii:
Touched up types a bit in my last comment, and worked around the fact that `toml` doesn't have py.typed or .pyi files in the package, even though it does have them. (https://github.com/uiri/toml/pull/347)

cibuildwheel/__main__.py
```diff
@@ -46,6 +48,8 @@ def get_option_from_environment(option_name: str, platform: Optional[str] = None
 
 
 def main() -> None:
+    platform: PlatStr
```
joerick, line 51: Perhaps `PlatformId` or `PlatformName` would be a better type name.

cibuildwheel/__main__.py
```diff
@@ -120,15 +124,21 @@ def main() -> None:
...
 
     if platform == 'linux':
         repair_command_default = 'auditwheel repair -w {dest_dir} {wheel}'
     elif platform == 'macos':
         repair_command_default = 'delocate-listdeps {wheel} && delocate-wheel --require-archs x86_64 -w {dest_dir} {wheel}'
-    else:
+    elif platform == 'windows':
         repair_command_default = ''
+    else:
+        assert_never(platform)
```
joerick, line 141: I suppose it's there to satisfy mypy(?), but `assert_never(platform)` isn't very descriptive, and requires a trip to the definition to see what it does... could it be `raise UnhandledValueError(platform)` or even `raise ValueError(platform)`?

cibuildwheel/__main__.py
```diff
@@ -231,6 +243,7 @@ def main() -> None:
         environment=environment,
         dependency_constraints=dependency_constraints,
         manylinux_images=manylinux_images,
+        python_configs=python_configs,
```
joerick, line 246: I'm unconvinced that python_configs should be part of the BuildOptions object, because it's not a user-controllable option. IMO it would be better to call `read_python_configs` from the `get_python_configurations` functions in the platform-specific files.

cibuildwheel/__main__.py
```diff
@@ -248,8 +261,7 @@ def main() -> None:
     elif platform == 'macos':
         cibuildwheel.macos.build(build_options)
     else:
-        print(f'cibuildwheel: Unsupported platform: {platform}', file=sys.stderr)
-        exit(2)
+        assert_never(platform)
```
joerick, line 264: See above comment, `raise` is probably better here.

cibuildwheel/resources/build-platforms.toml
```diff
@@ -0,0 +1,63 @@
+[tool.cibw.build-platforms]
+linux = [
```
joerick, line 2: Apologies, I might have been a little unclear on this earlier. I think the config file should be-

```toml
[linux]
python_configurations = [
  { identifier = "cp27-manylinux_x86_64", version = "2.7", path_str = "/opt/python/cp27-cp27m" },
  "...",
  "...",
  "...",
]

[macos]
python_configurations = [
  { identifier = "cp27-macosx_x86_64", version = "2.7", url = "https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg" },
  "...",
  "...",
  "...",
]

[windows]
python_configurations = [
  { identifier = "cp27-win32", version = "2.7.18", arch = "32" },
  "...",
  "...",
  "...",
]
```

I think it looks neater, and might let us extend with more stuff in the future, if we need.

joerick (commented):
Think we're nearly there with this one! Just a few minor comments on naming etc.

cibuildwheel/resources/build-platforms.toml
henryiii: Okay, I like following the pyproject.toml style, but technically, in retrospect, we don't own "cibw" so I don't think we would have been correct in using `tool.cibw`; it needs to be `tool.cibuildwheel`. If we made all configs pyproject.toml compatible, then eventually add the ability for cibuildwheel to pick up pyproject.tom, you could avoid having to pass overrides for these, and instead just use the "one configuration file" as it was meant to be used! :) (Not that I'd expect many/anyone to override this, but pinned docker images specified in pyprjoect.toml could be very useful?)

So is `tool.cibuildwheel.linux`, etc. fine?

cibuildwheel/__main__.py
henryiii: Okay, sounds good.

cibuildwheel/__main__.py
henryiii: This is [a standard name](https://hakibenita.com/python-mypy-exhaustive-checking#the-future), and it's for exhaustiveness checking. If you forget to include all possible platforms, mypy can tell (via type narrowing), that `assert_never` can be reached, and will error out.

It hasn't made it into mypy since the "correct" solution would be pattern matching. Which last I checked was a complete mess. To me it seems like a simple form (only `_` matches, always use `:=` to save) would have been far simpler than that mess they are making now (adding lots of new, only-in-pattern-matching syntax).

cibuildwheel/resources/build-platforms.toml
YannickJadoul: Why does this need to be here? I can see why you need to namespace in `pyproject.toml`, but this is a purely internal file; I'm not sure if/why it should be namespaced?

YannickJadoul: Oh, OK, you're thinking ahead of trying to read from `pyproject.toml`?

cibuildwheel/__main__.py
YannickJadoul: Since it's in `typing`, I don't really mind using `assert_never` (though it's arguably slightly weirdly named; I'd expect `assert_unreachable` or `unreachable` or so).

cibuildwheel/resources/build-platforms.toml
henryiii: Yes, should make the merge easy and be consistent. Though [some tools](https://pycqa.github.io/isort/docs/configuration/config_files/) do keep namepacing out of "custom" files, and only use it in the shared config, so it's not critical. (most tools pre-date toml, as well, so the custom config is .cfg instead). Will wait to update until @joerick weighs in.

cibuildwheel/__main__.py
joerick: Ahhhh, it checks that the condition is unreachable at 'compile' time. That's nice.

Not a great name, though. Not all that widely used, either: https://github.com/search?l=Python&q=assert_never&type=Code . I also saw reference to `assert_exhaustive(variable)`, which I quite like. I'm happy either way - up to you, @henryiii

cibuildwheel/__main__.py
joerick: scratch that!

cibuildwheel/resources/build-platforms.toml
joerick: If we did have `[tool.cibuildwheel]` options in `pyproject.toml`, we'd want those options to be deliberately user-facing - designed, and documented. This seems like it's internal data, not config (and it's very prone to change in schema).

henryiii: Okay, sounds fine to me then. I would expect eventually allowing some things to be set from `[tool.cibuildwheel.*]` to be quite useful, though I don't see a reason to have this ever settable, and certainly not likely in it's current form.

henryiii: @joerick Sorry, one more question. What's a good name for the file? This fine or something else?

henryiii: (Assuming it might depend on what you think could go in here later)

joerick: This seems good to me :)

cibuildwheel/__main__.py
henryiii: It's in PyRight's tests: https://github.com/microsoft/pyright/search?q=assert_never , and 10x more common than assert_exhaustive, so I'd rather go with this and hope there's a chance someone has seen it (can link to the blog post at the definition, if you'd like?), but I'd be perfectly happy with assert_exhaustive if you do have a strong preference.

henryiii: PS: It's not in typing, though, as far as I know.

joerick: Okay, let's go with that.

joerick (approved):

cibuildwheel/util.py
```diff
@@ -13,14 +13,20 @@
...
-if sys.version_info < (3, 8):
-    from typing_extensions import Literal
+if sys.version_info < (3, 9):
+    from importlib_resources import files
 else:
-    from typing import Literal
+    from importlib.resources import files
+
+
+resources_dir = files('cibuildwheel') / 'resources'
```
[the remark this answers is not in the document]
henryiii: It supports directories that are zip files (see https://github.com/scikit-hep/particle/releases/tag/v0.14.0 - that `pyz runs directly without installing, `./particle.pyz`), and namespace packages that are split across directories.

Though, however, I'm using it here because it's more elegant and the "right' thing to do. We are accessing a resource.

PS: There's a few tricks when running from a zip file, you can open the path or access the filename, but cannot use the file name to open the file without a little workaround which I didn't attempt to implement.

cibuildwheel/__main__.py
henryiii: It's now in typing, with this name. :) https://github.com/python/cpython/pull/30842

joerick: :) ohhh, it's assert_never because of the type `Never`. That makes a bit more sense.

PR #519, opened 2021-01-05
style: isort

henryiii:
This actually can do full Black-style includes, I like it. :) Discussed in #518.

- style: isort
- style: run isort

This is Black-style, which I like - short includes are bundled onto one line, longer ones are one-per line. You can get merge conflicts for the short includes, but hopefully they are not changed often? And matching changes are okay.

Sticking configs in setup.cfg instead of pyproject.toml because mypy and flake8 don't support that yet. (Flake8 may never :( )

### DIFF
--- a/test/test_before_all.py
+++ b/test/test_before_all.py
@@ -1,9 +1,9 @@
-import pytest
 import subprocess
 import textwrap
 
-from . import utils
-from . import test_projects
+import pytest
+
+from . import test_projects, utils
 
 project_with_before_build_asserts = test_projects.new_c_project(
     setup_py_add=textwrap.dedent(r'''
--- a/unit_test/main_tests/conftest.py
+++ b/unit_test/main_tests/conftest.py
@@ -4,12 +4,7 @@ from pathlib import Path
 
 import pytest
 
-from cibuildwheel import (
-    linux,
-    macos,
-    util,
-    windows,
-)
+from cibuildwheel import linux, macos, util, windows
 
 
 class ArgsInterceptor:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,11 @@ repos:
   - id: mixed-line-ending
   - id: trailing-whitespace
 
+- repo: https://github.com/PyCQA/isort
+  rev: 5.7.0
+  hooks:
+  - id: isort
+
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.790
   hooks:
--- a/bin/bump_version.py
+++ b/bin/bump_version.py
@@ -1,13 +1,15 @@
 #!/usr/bin/env python3
 
-import click
-from pathlib import Path
-import os
-import cibuildwheel
-from packaging.version import Version, InvalidVersion
-import subprocess
 import glob
+import os
+import subprocess
 import urllib.parse
+from pathlib import Path
+
+import click
+from packaging.version import InvalidVersion, Version
+
+import cibuildwheel
 
 config = [
     # file path, version find/replace format
--- a/bin/make_dependency_update_pr.py
+++ b/bin/make_dependency_update_pr.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python3
 
 import os
+import textwrap
 import time
 from pathlib import Path
 from subprocess import run
 
 import click
-import textwrap
 
 
 def shell(cmd, **kwargs):
--- a/bin/projects.py
+++ b/bin/projects.py
@@ -11,17 +11,16 @@ Suggested usage:
 
 import builtins
 import functools
+import urllib.request
+import xml.dom.minidom
 from datetime import datetime
 from io import StringIO
-from typing import Dict, Any, List, Optional, TextIO
+from pathlib import Path
+from typing import Any, Dict, List, Optional, TextIO
 
 import click
 import yaml
 from github import Github
-import urllib.request
-import xml.dom.minidom
-from pathlib import Path
-
 
 ICONS = (
     "appveyor",
--- a/bin/run_example_ci_configs.py
+++ b/bin/run_example_ci_configs.py
@@ -1,16 +1,17 @@
 #!/usr/bin/env python3
 
 import os
-from pathlib import Path
 import shutil
-from subprocess import run
 import sys
 import textwrap
 import time
-import click
-from glob import glob
 from collections import namedtuple
+from glob import glob
+from pathlib import Path
+from subprocess import run
 from urllib.parse import quote
+
+import click
 
 
 def shell(cmd, **kwargs):
--- a/bin/sample_build.py
+++ b/bin/sample_build.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 
+import argparse
 import os
 import subprocess
 import sys
-import argparse
 import tempfile
 from pathlib import Path
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -5,17 +5,13 @@ import textwrap
 import traceback
 from configparser import ConfigParser
 from pathlib import Path
-
 from typing import Any, Dict, List, Optional, Set, overload
 
 import cibuildwheel
 import cibuildwheel.linux
 import cibuildwheel.macos
 import cibuildwheel.windows
-from cibuildwheel.environment import (
-    EnvironmentParseError,
-    parse_environment,
-)
+from cibuildwheel.environment import EnvironmentParseError, parse_environment
 from cibuildwheel.util import (
     Architecture,
     BuildOptions,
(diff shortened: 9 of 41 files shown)


henryiii:
Second commit is completely generated so this can easily be rebased, no rush.

YannickJadoul:
This might be the first time I more or less agree with Black :-o

Two things that're ugly, IMO:
1. `from . import test_projects, utils` is importing 2 modules on 1 line; why?
2. The multiline `from ... import (...)` is ugly, but arguably not much more ugly than writing less lines with several things on one line.
3. Oh, yeah, in any test file, having `pytest` first makes perfect sense to me. But that's personal taste, I guess :-)

I'll come back complaining when Black starts correcting me when I'm sure I'm right ;-)

Czaki:
https://pycqa.github.io/isort/

What is wrong with grouping imports?

henryiii:
`The multiline from ... import (...) is ugly`

Sorting imports and many of the decisions in Black are not about beauty, but are about reducing merge conflicts. Sorted imports would have removed 1 of 6 merge conflicts when I rebased #511, Black would have removed at least 2 more with function arguments.

1. Might be a way to turn that off? Weird. We can try to fix it. There are lots of options, unlike Black. ;)
2. Multiline imports are great; if I add Set and you add Union, we don't conflict! Same thing for function arguments if there are more than a couple or so.
3. PEP 8 grouping says stdlib first, 3rd party next, internal last. That's nice to help show what is stdlib vs. added, and that can actually catch bugs, if you `import X` and think that it's local (Python 3) or from the stdlib/remote (Python 2), it will try to move it and you'll see the mistake. I had to fix someone who was using `import cmake` for a local file when they also had `pip installed cmake`.

henryiii:
You could put pytest first: https://pycqa.github.io/isort/#custom-sections-and-ordering :)

test/test_before_all.py
```diff
@@ -1,9 +1,9 @@
-import pytest
 import subprocess
 import textwrap
 
-from . import utils
-from . import test_projects
+import pytest
+
+from . import test_projects, utils
```
henryiii, line 6: This is slightly confusing, actually, the thing after a `from x import` ideally shouldn't be a module; if you have an `__init__.py` and it contains a `test_projects` or `utils` object, you'll get that instead of the module. But it's a bit special for `.` imports, since you are required to use `from` with a relative import.

I believe you are guaranteed to have the test directory as the first on the path in PyTest, so this could be just `import test_projects`, etc? Also, we could `from .test_projects import new_c_project`, etc.

unit_test/main_tests/conftest.py
```diff
@@ -4,12 +4,7 @@
 
 import pytest
 
-from cibuildwheel import (
-    linux,
-    macos,
-    util,
-    windows,
-)
+from cibuildwheel import linux, macos, util, windows
```
henryiii, line 7: This is irritating, it didn't respect Black's magic comma.

henryiii: I guess it doesn't. Sad. For things like typing I add the comma because I know they are likely to change. (Not true of this case, of course, but might be useful elsewhere, especially with typing).

test/test_before_all.py
henryiii: Some of the exiting ones are written this way, so I'd probably leave it for now, and eventually rewrite them properly (see `test_0_basic.py`, for example).

unit_test/main_tests/conftest.py
Czaki: > This is irritating, it didn't respect Black's magic comma.

What did you mean? Black put `,` only at the end of the multiline import. When the import is a single line then black does not put a comma on the end.

test/test_before_all.py
YannickJadoul: Yeah, I first wanted to suggest `import .test_projects`, but apparently that only works in the `from ... import ...` construct

unit_test/main_tests/conftest.py
YannickJadoul: True, that would be nice, if we could teach it. And once again (I'm surprised) I would agree with Black that the comma there is a pretty good indicator to say that this is a changing list.

YannickJadoul:
Yeah, don't pay too much attention to my grumbling. These are just the things that I'm used to, so not even sure whether this was `cibuildwheels`'s (implicit) style.
I must say I'm surprised by the merge conflicts. I don't think I ever ran into those too much for includes. Or maybe they are so easily fixed that I just forgot about them :-D

Czaki:
>  I must say I'm surprised by the merge conflicts. I don't think I ever ran into those too much for includes. Or maybe they are so easily fixed that I just forgot about them :-D

After finish the discussion here the simplest thing will be created the next PR which starts from the current master and has copied isort configuration.

test/test_before_all.py
joerick: > the thing after a `from x import` ideally shouldn't be a module

Not sure I agree, e.g. `from numpy import linalg` `from django import models` `from tensorflow import keras`... also I'd rather rely on the `__init__.py` file being empty than rely on the value of `sys.path`. I expect tools like mypy or language servers to be able to read the code better without relying on a `sys.path` set by pytest too.

joerick (approved):
Looks great!

test/test_before_all.py
henryiii: `from numpy import linalg` is an often used shortcut for `import numpy.linalg as linalg`, but it is confusing - a reader can't tell if linalg is a package or an object. We are stuck with it, but I wish the language wasn't ambiguous. Ideally, relative imports probably should import objects, `from .test_projects import new_c_project`.

Relative imports also don't work if you don't execute it as a package; `python test_before_all.py` will fail. (`ImportError: attempted relative import with no known parent package`) (PyTest handles this correctly).

henryiii: A "tests" directory (or worse, "test") isn't really a package - `import test` imports the stdlib `test`. I think PyTest does some shenanigans to make it work.

joerick: We are off on a big ol' tangent here, but-

> a reader can't tell if linalg is a package or an object

I'm not sure it matters, from a users' point of view. It's still an imported namespace, or global state if you mutate it. Probably I've just used different bits of Python to you, but to me, importing a module like this is pretty common.

joerick: Further aside - I often see people talk negatively of relative imports, in general... why is that? I've never had trouble understanding them, and they seem explicit enough ¯\_(ツ)_/¯

henryiii: > I'm not sure it matters, from a users' point of view.

If you try to rewrite it, you have to either run it interactively or look up the source. For example:

```python
from a import b
b.c()
# May or may not be written as
from a.b import c
c()
```

And, if you need to look up the source, is there an `a/b.py` or an `a/__init__.py` file with a b in it, or is it a file? (I often teach Python...)

henryiii: I use relative imports a lot, but they have downsides, especially for new users who do not know how to set up a proper environment. They _have_ to be in a loaded package; it's not only "in the same directory". So `python x.py` inside a package fails if there are relative imports in `x`; I think `python outer/x.py` fails too (I could be wrong). If you stick a `if __name__ == "__main__"` in it for standalone runs or tests, it fails.

henryiii: I also use absolute imports a lot...

henryiii: I do know someone who will not do any import except `import a.b.c`, and always uses fully qualified names. That's it. No shortening. It drive me crazy seeing `numpy.stuff` sometimes. :)

henryiii:
To put @Czaki into code; if you are working on a branch:

```bash
git checkout master -- .pre-commit-config.yaml setup.cfg
pre-commit run -a
git add -u
git commit -m "style: pre-commit run -a"
```

Then you can squash on master if you want. Or `git commit --amend`, etc.

henryiii:
Also, for simplicity, I'll rebase and merge instead of squash and merge. Maybe we could even eventually add a `.git-blame-ignore-revs` file and add these style change runs to it.

test/test_before_all.py
YannickJadoul: FWIW (given we're on this tangent anyway?), I tend to not like `from module import thing` either. Often, `module.thing` is not so much longer to write, should be auto-completed, and we're not trying to save bytes in code anyway? But yes, `numpy` to `np` just becomes a habit, after a while :-)

henryiii:
Permissions error on Windows Travis can be ignored, I think.

YannickJadoul:
> Permissions error on Windows Travis can be ignored, I think.

I restarted that build, but indeed. Go ahead if you want to merge :-)
Wait, I'll still give you that precious checkmark.

YannickJadoul (approved):
(Unless @joerick still wants to fix the `from . import x, y`, but you already got an approval there, so I'm not holding up this PR on that!)

henryiii:
There already was a `from . import x, y` in the codebase, so not important, I think. ;)

henryiii:
Actually, even easier, just rebase delete the merge conflict lines, and then run `pre-commit run -a`, then `git rebase --continue`. Done. :)

PR #521, opened 2021-01-06
feat: support pypa build

### LINKED ISSUE #442: [idea] pypa-build support
Currently, we (along with most of the Python ecosystem) are misusing the `pip wheel` command to build wheels. It was the best available tool, but only because it was the only available tool; it wasn't really designed for wheel building. One side effect is that _all_ wheels are built into the "wheelhouse", and users have to pick out just the correct ones. This is partially mitigated by including "--no-deps", but that is just a hack solution that also requires more hacks (requiring dependencies to be installed separately, etc), and PEP 518 dependencies could possibly trigger more wheels. My recommendation so far has always been to use `<package_name>*.whl` when uploading/moving outputs to avoid grabbing an extra built dependency.

There is a solution; a proper PEP 517 builder instead of using the wheelhouse output. The infrastructure for it has been around (the `pep517` package, which pip vendors and uses, actually), but a proper front end has not been available until quite recently (not counting `pep517.build`, which was intended as an example, not intended for widespread use - though it was picked up by a large number of packages before being pulled).

A very young package, [targeted for PPA inclusion](https://github.com/FFY00/python-build/issues/42), is finally available and is the recommended migration path for `pep517.build`: [`python-build`](https://python-build.readthedocs.io/en/latest/). I think it would be a good idea to support this in cibuildwheel. Ideas:

I think we shouldn't lock in `build`, at least at this early point, partially because this could change in the near future. So how about providing a new environment variable that overrides the default build command, so that you could write `CIBW_BUILD_CMD="python -m build -w -o {output_dir}"`; when combined with `CIBW_BEFORE_BUILD: pip install build`, this should work, I think. Eventually, if this gets accepted into the PyPA and settles down, we could make this a simpler toggle and have it in the constraints file, perhaps?

Thoughts? Just putting it down while thinking about it, no rush.

Edit: It would also be nice to be able to adjust the build command, either to remove `--no-deps` or in **special** cases to add `--no-build-isolation`.

PS: Personally, I've been moving most pure python wheels to `build`, since it can build SDists using pyproject.toml, which pip cannot.

henryiii:
This will close #442.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -317,7 +317,7 @@ def troubleshoot(package_dir: Path, error: Exception) -> None:
         error.cmd[0:4] == ["python", "-m", "pip", "wheel"]
         or error.cmd[0:3] == ["python", "-m", "build"]
     ):
-        # the 'pip wheel' step failed.
+        # the wheel build step failed
         print("Checking for common errors...")
         so_files = list(package_dir.glob("**/*.so"))
 
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -182,7 +182,21 @@ def build(options: BuildOptions) -> None:
 
                     verbosity_flags = get_build_verbosity_extra_flags(options.build_verbosity)
 
-                    if options.pypa_build:
+                    if options.build_frontend == "pip":
+                        docker.call(
+                            [
+                                "python",
+                                "-m",
+                                "pip",
+                                "wheel",
+                                container_package_dir,
+                                f"--wheel-dir={built_wheel_dir}",
+                                "--no-deps",
+                                *verbosity_flags,
+                            ],
+                            env=env,
+                        )
+                    elif options.build_frontend == "build":
                         config_setting = " ".join(verbosity_flags)
                         docker.call(
                             [
@@ -197,18 +211,8 @@ def build(options: BuildOptions) -> None:
                             env=env,
                         )
                     else:
-                        docker.call(
-                            [
-                                "python",
-                                "-m",
-                                "pip",
-                                "wheel",
-                                container_package_dir,
-                                f"--wheel-dir={built_wheel_dir}",
-                                "--no-deps",
-                                *verbosity_flags,
-                            ],
-                            env=env,
+                        raise RuntimeError(
+                            f"build_frontend {options.build_frontend!r} not understood"
                         )
 
                     built_wheel = docker.glob(built_wheel_dir, "*.whl")[0]
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -7,7 +7,7 @@ from typing import List, NamedTuple, Set
 from .architecture import Architecture
 from .docker_container import DockerContainer
 from .logger import log
-from .typing import PathOrStr
+from .typing import PathOrStr, assert_never
 from .util import (
     BuildOptions,
     BuildSelector,
@@ -211,9 +211,7 @@ def build(options: BuildOptions) -> None:
                             env=env,
                         )
                     else:
-                        raise RuntimeError(
-                            f"build_frontend {options.build_frontend!r} not understood"
-                        )
+                        assert_never(options.build_frontend)
 
                     built_wheel = docker.glob(built_wheel_dir, "*.whl")[0]
 
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -145,7 +145,7 @@ def build(options: BuildOptions) -> None:
                         env, executor=docker.environment_executor
                     )
 
-                    # check config python and pip are still on PATH
+                    # check config python is still on PATH
                     which_python = docker.call(
                         ["which", "python"], env=env, capture_output=True
                     ).strip()
@@ -180,18 +180,36 @@ def build(options: BuildOptions) -> None:
                     docker.call(["rm", "-rf", built_wheel_dir])
                     docker.call(["mkdir", "-p", built_wheel_dir])
 
-                    docker.call(
-                        [
-                            "pip",
-                            "wheel",
-                            container_package_dir,
-                            "--wheel-dir",
-                            built_wheel_dir,
-                            "--no-deps",
-                            *get_build_verbosity_extra_flags(options.build_verbosity),
-                        ],
-                        env=env,
-                    )
+                    verbosity_flags = get_build_verbosity_extra_flags(options.build_verbosity)
+
+                    if options.pypa_build:
+                        config_setting = " ".join(verbosity_flags)
+                        docker.call(
+                            [
+                                "python",
+                                "-m",
+                                "build",
+                                container_package_dir,
+                                "--wheel",
+                                f"--outdir={built_wheel_dir}",
+                                f"--config-setting={config_setting}",
+                            ],
+                            env=env,
+                        )
+                    else:
+                        docker.call(
+                            [
+                                "python",
+                                "-m",
+                                "pip",
+                                "wheel",
+                                container_package_dir,
+                                f"--wheel-dir={built_wheel_dir}",
+                                "--no-deps",
+                                *verbosity_flags,
+                            ],
+                            env=env,
+                        )
 
                     built_wheel = docker.glob(built_wheel_dir, "*.whl")[0]
 
@@ -291,7 +309,10 @@ def build(options: BuildOptions) -> None:
 
 
 def troubleshoot(package_dir: Path, error: Exception) -> None:
-    if isinstance(error, subprocess.CalledProcessError) and error.cmd[0:2] == ["pip", "wheel"]:
+    if isinstance(error, subprocess.CalledProcessError) and (
+        error.cmd[0:4] == ["python", "-m", "pip", "wheel"]
+        or error.cmd[0:3] == ["python", "-m", "build"]
+    ):
         # the 'pip wheel' step failed.
         print("Checking for common errors...")
         so_files = list(package_dir.glob("**/*.so"))
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -4,6 +4,8 @@ import itertools
 import os
 import re
 import ssl
+import subprocess
+import sys
 import textwrap
 import time
 import urllib.request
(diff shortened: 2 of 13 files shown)


henryiii:
Might be easier if https://github.com/pypa/manylinux/pull/941 is accepted.

henryiii:
pypa-build builds in a virtual environment, so the path check in setup.py is incorrect. It sees the executable we require in the first step, and the build virtual environment path in the second. Any ideas about how to change the check? Wouldn't an exact version check be enough, perhaps?

henryiii:
This should probably be seen as a proof of concept until pypa/manylinux#941 goes in, because we currently don't touch the contents of the manylinux image.

joerick:
BTW, I wouldn't be opposed to releasing PyPA build support before pypa/manylinux#941, given that this would be advertised as 'experimental' for now anyway.

henryiii:
I've included pypa/build in manylinux now, including manylinux1 for Python 2.7 (!), but we can't use it because we've pinned manylinux2010 to a version before that addition.

Honestly, I don't think dropping 2.7 support for manylinux2010 but keeping it for manylinux1 for now would be that big of an issue for a 1.11 release. We have very few users using manylinux2010 anyway, and even fewer of them making 2.7 wheels. Requires-Python support already helps, and we could check to see what Pythons are in the image and limit to that, producing an error if any are missed that are included by Requires-Python, and a warning if Requires-Python is not set.

henryiii:
Does the PyPy image not contain build?

mayeut:
It seems they are base on an old manylinux2010 image so no pypa/build yet.
https://github.com/pypy/manylinux/pull/16 should resolve this once merged.

mayeut:
pypy/manylinux#16 has been merged. The new  image should have build (& auditwheel 4)

henryiii:
> Actually, it looks like pypy/manylinux is based on an old version of pypa/manylinux; notably it has a copy of requirements.txt internally (which is missing other important things we've added in this repo, including build).

_Originally posted by @henryiii in https://github.com/pypa/manylinux/issues/1055#issuecomment-842672688_

This means no build in pypy/manylinux, which is why this PR is failing.

mayeut:
Switched to quay.io/pypa/manylinux2010_x86_64 for PyPy while waiting for #671 just to see what the tests look like on GHA (need to rebase to get CI green)

henryiii:
Windows is failing on PyPy 7.3.3 + build:

```
Unable to copy 'C:\\cibw\\pypy3.7-v7.3.3-win32\\venvlauncher.exe'
```

henryiii:
`pypa/build` is tested on Windows + PyPy{2,3}, so not sure where this error is coming from.

I also think there are a few tests I need to update the output checking on, will look into that later.

mayeut:
@henryiii, rebased on `main` to get CI again. I fixed the `cibuildwheel.linux.troubleshoot`  function to get all builds passing locally on linux. Let's see if CI agrees, there will only be the Windows PyPy issue left to investigate I think.

henryiii:
Looks like it's checking the exact pip versions, which is not true in the env created by build.

mayeut:
> Looks like it's checking the exact pip versions, which is not true in the env created by build.

This is the issue for macOS.
For Windows PyPy, something else is going on.

mayeut:
I investigated a bit the issue with PyPy on Windows.
It seems the `venv` module does not like to be called from the `python.exe` symlink at all.
Worked around for now by installing `build[virtualenv]`

mayeut:
The only test failing now is `test/test_dependency_versions.py::test_dependency_constraints_file` on macOS & Windows i.e.

> Looks like it's checking the exact pip versions, which is not true in the env created by build.

Should we create/discuss something in pypa/build or should we just ignore this one ?
My 2 cents, it's good to be able to pin things down for reproducibility. We might also want to extend that to linux builds (although it's a different issue) ?

mayeut:
I implemented the hack mentioned  in pypa/build#292 just to see if everything passes with it. Not sure it's a good idea to implement such a hack.

cibuildwheel/util.py
```diff
@@ -274,3 +277,18 @@ def print_new_wheels(msg: str, output_dir: Path) -> Iterator[None]:
     s = time.time() - start_time
     m = s / 60
     print(msg.format(n=n, s=s, m=m), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
+
+
+def get_pip_version(env: Dict[str, str]) -> str:
+    # we use shell=True here for windows, even though we don't need a shell due to a bug
+    # https://bugs.python.org/issue8557
```
henryiii, line 284: This would be simpler with #672?

henryiii:
What's the status of this, @mayeut?

mayeut:
Short status:
- all tests are passing whether build is the default or not (https://github.com/pypa/cibuildwheel/pull/521/commits/2e52bf2705ebce8ed5055cab6bfa739fcfa202f7 really restored pip as the default)
- We need to install `build[virtualenv]` to at least workaround PyPy venv issue, eventually to allow to pin pip when using build.

Open questions / things to do:
- [ ]  Should dependency constraints be kept ? it feels very hacky https://github.com/pypa/cibuildwheel/pull/521/commits/6f78d73a3cb6a9c71d0390b0c66e11db1d51a82d /  pypa/build#292
- [ ] Extend tests ? all or select some that seem more relevant ? (already added those showing issues that needed a commit)
- [ ] Update docs

I'll most likely be offline for the next 2 weeks so not planning on adding more things here.

joerick:
A thought on options design - rather than `CIBW_PYPA_BUILD`, could we change this to `CIBW_BUILD_FRONTEND`? (I don't love the name 'frontend' because what does that make us...) but that's the [terminology used in PEP 517]. Options would be `pip` or `build`.

More of a future-proofing measure, really. I've noticed some other tools offering themselves as build frontends... e.g. poetry might start building platform wheels, or something else like https://github.com/FFY00/trampolim might crop up... I suppose that in theory they should 'just' be invoking the build backend, but we can't predict the future!

[PEP 517]: https://www.python.org/dev/peps/pep-0517/#terminology-and-goals

henryiii:
I'll change `CIBW_PYPA_BUILD` to `CIBW_BUILD_FRONTEND/tool.cibuildwheel.build-frontend` after #684.

henryiii:
I squashed the previous commits to make it easier to rebase and modify for the new config setup. I then switched to using `build-backend`. Docs still need to be added.

henryiii:
Put some basic docs in, happy to take improvements.

test/conftest.py
```diff
@@ -11,11 +13,18 @@ def pytest_configure(config):
...
     skip_emulation = pytest.mark.skip(reason="need --run-emulation option to run")
     for item in items:
         if "emulation" in item.keywords:
             item.add_marker(skip_emulation)
+
+
+@pytest.fixture(
+    params=[{"CIBW_BUILD_FRONTEND": "pip"}, {"CIBW_BUILD_FRONTEND": "build"}], ids=["pip", "build"]
+)
+def build_mode(request) -> Dict[str, str]:
```
joerick, line 29: Maybe this fixture would be better named as `build_frontend` or `build_frontend_option`?

joerick (commented):
Nice one @henryiii and @mayeut ! Just a few tweaks and this is ready to go, I think.

test/conftest.py
henryiii: I used `build_frontend_env`, since it's an env dict.

cibuildwheel/linux.py
```diff
@@ -180,18 +180,40 @@ def build(options: BuildOptions) -> None:
...
+                                "--wheel",
+                                f"--outdir={built_wheel_dir}",
+                                f"--config-setting={config_setting}",
+                            ],
+                            env=env,
+                        )
+                    else:
+                        raise RuntimeError(
+                            f"build_frontend {options.build_frontend!r} not understood"
+                        )
```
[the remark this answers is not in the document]
henryiii: Loading a literal with a variable is a little odd, but I set up something that works, and it does make the usage nicer.

joerick:
oops, i missed these earlier...

> * [ ]   Should dependency constraints be kept ? it feels very hacky [6f78d73](https://github.com/pypa/cibuildwheel/commit/6f78d73a3cb6a9c71d0390b0c66e11db1d51a82d) /  [pypa/build#292](https://github.com/pypa/build/issues/292)

Yes, this does seem like a bit of a hack.. the use of PIP_CONSTRAINTS seems like exploiting an implementation detail of `build`. That said, it's unlikely to be a problem on the user's end, because we pin the version of `build`. I suppose the other option would be to use `--no-isolation` and ensure the correct versions of the tools we care about are installed in the environment before invoking build. I'd be okay with either, I think.

> * [ ]  Extend tests ? all or select some that seem more relevant ? (already added those showing issues that needed a commit)

Since use of `build` is only an option, currently, this seems okay for now. I'd be curious if the test suite does pass with `CIBW_BUILD_FRONTEND=build`, though, did you try with the default as that?

test/test_before_build.py
```diff
@@ -25,8 +25,13 @@
...
         print('stored_executable', stored_executable)
         print('sys.executable', sys.executable)
+
         # windows/mac are case insensitive
-        assert os.path.realpath(stored_executable).lower() == os.path.realpath(sys.executable).lower()
+        stored_path = os.path.realpath(stored_executable).lower()
+        current_path = os.path.realpath(sys.executable).lower()
+
+        # TODO: This is not valid in an virtual environment
+        # assert stored_path == current_path, '{0} != {1}'.format(stored_path, current_path)
```
joerick, line 34: Whoops, I missed this on the first run through. I guess we can add this test back in, until we're using `build` as the default?

henryiii: We are testing this in build too, so I don't think it can be restored?

joerick: I don't think so? I don't see a `build_frontend_env` fixture in this file...

henryiii:
```suggestion
        # TODO: This is not valid in an virtual environment
        assert stored_path == current_path, '{0} != {1}'.format(stored_path, current_path)
```

henryiii: Let's see, then!

joerick (approved):

henryiii:
> I suppose the other option would be to use --no-isolation

I think this would break PEP 518; I think the current constraints is preferable for now.

> I'd be curious if the test suite does pass with CIBW_BUILD_FRONTEND=build, though, did you try with the default as that?

I think that's how I started, and that's why I had that assert commented out; I was trying to run everything with build. But I could be wrong.

Issue #528, opened 2021-01-10
ARCH and REQUIRES_PYTHON plan


henryiii:
1. We continue with CIBW_ARCHS as we are now; "auto" is the default, and auto (could) include universal2 for macOS, rather like it already includes 32bit for 64bit on the other OSs. I think most users will find this ideal for enabling or playing with the new architectures, since they are using things like `cp36*` now in build and that doesn't have to change _unless they want to split based on it, see the final comments_. macOS 11 considers "universal2" to be auto.
2. We add a `CIBW_MIN_PYTHON` flag, with the usual variants (and maybe command line flag? Don't know what the policy is for that). If it is not set, we read the value from [PEP 621's location](https://www.python.org/dev/peps/pep-0621/#requires-python) if set (eventually), and for now use 2.7 (27?) if not set. For CIBW 2.0, we can change the default to 3.6 (36?), but users could get at least 2.7 back by manually specifying this. This is mostly so we can keep the defaults "nice", while also supporting the full range that we can support. From 2.0 on, we raise the default minimum to always match the official Python EOL. #527 
3. (Optional) We upgrade to using the wcmatch library; this gives `{}` syntax, which I love using on the shell (partially because I'm a fish user and tab completion works correctly when using it, so it's perfect for moves and such). `cp{27,35}*` seems like it should work. Ve can activate the BRACE and NEGATE flags; this is simpler than my example above (basically just BUILD + SKIP), and is already well documented so we don't have to try to teach people how to use it, we can just point at this. Eventually, if this works well, we could push users away from the SKIP flag.
4. (Optional) We could add a little javascript selection in the docs as shown above. I just think that's very handy and will be "flashy" :)

ARCH/MIN_PYTHON are about telling cibuildwheel what you are interested in supporting, and is something we can adjust over time to "ideal" defaults, unless a user overrides them, in which case they specifically want _that_ thing. With SKIP, which is in this category sort of, user overrides blend CPython/PyPy, Python version, and  OS-ARCH, so we can't control each default separately without the new variables. BUILD is more about telling it what "this" job is trying to build out of the total allowed set.

Further; I propose we split the unreleased "special" value for "arch" into three special values:

`"native"`: The native architecture (always one value).
`"auto"`: Produces the native architecture + 32bit on 64bit non-macOS systems, or + universal2 on macOS. Basically "everything you can run tests on". This would be default, just like auto for CIBW_PLATFORM. I think this was first proposed by @mayeut ?
[...]

PR #535, opened 2021-01-14
feat: native / all archs

henryiii:
I rather forget we already had the correct definition of "auto", but this adds the other two as mentioned in #528.

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -16,7 +16,7 @@ import certifi
 import toml
 
 from .environment import ParsedEnvironment
-from .typing import PathOrStr, PlatformName
+from .typing import PathOrStr, PlatformName, assert_never
 
 if sys.version_info < (3, 9):
     from importlib_resources import files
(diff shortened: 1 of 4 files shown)


cibuildwheel/util.py
```diff
@@ -184,6 +188,17 @@ def auto_archs(platform: PlatformName) -> 'Set[Architecture]':
             result.add(Architecture.x86)
         return result
 
+    @staticmethod
+    def all_archs(platform: PlatformName) -> 'Set[Architecture]':
```
YannickJadoul, line 192: Does this overlap with `ALLOWED_ARCHITECTURES`? In values, definitely. But conceptually as well?

henryiii: Yes, they should be the same. Good catch!

cibuildwheel/util.py
```diff
@@ -255,19 +270,13 @@ def detect_ci_provider() -> Optional[CIProvider]:
...
 
-ALLOWED_ARCHITECTURES = {
-    'linux': {Architecture.x86_64, Architecture.i686, Architecture.aarch64, Architecture.ppc64le, Architecture.s390x},
-    'macos': {Architecture.x86_64},
-    'windows': {Architecture.AMD64, Architecture.x86},
-}
-
 
 def allowed_architectures_check(
     name: PlatformName,
```
joerick, line 266: Might have missed this first time around, but generally this variable should be called 'platform'

joerick (commented):
Looks good! I guess we just need these terms adding to the CIBW_ARCHS_LINUX docs?

henryiii:
:eyes:?

joerick (approved):
Great!

PR #536, opened 2021-01-15
feat: requires-python

### LINKED ISSUE #528: ARCH and REQUIRES_PYTHON plan

henryiii:
I think this finishes #528's changes that affect non-Universal builds.

This is using `CIBW_PROJECT_REQUIRES_PYTHON`, though it could be changed to `CIBW_IGNORE_REQUIRES_PYTHON` and just turn off this search instead.

---

Docs example in PR.

----

With a proper pyproject.toml or setup.cfg file, the minimal example for a small package with cibuildwheel could be just 13 lines (using #542)!

```yaml
on: [push, pull_request]

jobs:
  build_wheels:
    strategy:
      matrix:
        os: [ubuntu-20.04, windows-2019, macOS-10.15]
    runs-on: ${{ matrix.os }}

    steps:
      - uses: actions/checkout@v2
        
      - uses: joerick/cibuildwheel@v1.7.4

        with:
          path: ./wheelhouse/*.whl
```

(Most packages should still set up a matrix to take better advantage of the 10+ parallel jobs GHA gives, but still, very nice!)

Fixes #528 
Includes the MyPy update to 0.800, closes #561.


cibuildwheel/__main__.py
```diff
@@ -308,7 +328,7 @@ def print_preamble(platform: str, build_options: BuildOptions) -> None:
 
 
 def print_build_identifiers(
-    platform: str, build_selector: BuildSelector, architectures: Set[Architecture]
+        platform: str, build_selector: BuildSelector, architectures: Set[Architecture]
```
henryiii, line 331: Oops, vi did this, not me! I would never un-Black something! 😳

(Okay, I have been trying to convert `"` -> `'` when I see them to be consistent; most cases of `"` are my fault and because Black has finally retrained my brain to use `"`, I used to be a single-quoter.)

docs/options.md
```diff
@@ -184,6 +184,31 @@ to limit builds for each job; this list could grow in the future.
...
+
+Default: `>=2.7`, all supported versions are built. In the future, this will match
+the EOL of Python, so setting this is recommended if you want to support past-EOL
+Python versions.
+
+Example `pyproject.toml`:
+
+```toml
+[metadata]
+requires-python: ">=3.6"
```
henryiii, line 206:
```suggestion
requires-python = ">=3.6"
```

Thank you, GitHub highlighter! :)

cibuildwheel/__main__.py
henryiii, line 331:
```suggestion
    platform: str, build_selector: BuildSelector, architectures: Set[Architecture]
```

henryiii:
Wait, why can you rerun a single Azure job directly from the GH interface, but not Github Actions?

Czaki:
> Wait, why can you rerun a single Azure job directly from the GH interface, but not Github Actions?

Because Github Actions does not support reboot a single job. Yes, both are provided by a single company.

joerick:
I'm gonna hold off on reviewing this until we have a consensus on the design discussion in #528.

henryiii:
Oops, the name is "project", not "metadata" (though that was listed as the second most popular choice ;) ), I will update everything at once when we decide on #528. Also, from reading the docs, this might be treated as all or nothing by build backends; that is, if someone -only- specifies `project.requires-python`, then setuptools would be allowed or possibly required to require all other non-optional fields be specified, unless they are explicitly listed in `project.dynamic`. So having the workaround might be extra important for the next year or two. As far as I can tell, though, it might only be "name" that's also required. Feel free to check out PEP 621.

henryiii:
I've updated this to my favorite version of the proposal so far:

* CIBW_PROJECT_REQUIRES_PYTHON overrides pyproject.toml, which overrides setup.cfg as specified by PEP 621. The description focuses on setting this for the project, mentining it can be given by an environment variable if required (with warning).
* If unset, no change applied (assuming this will _not_ change in 2.0 in this case). The direction for "nice" builds is to set this via one of the above methods. 

We still need unit tests to verify that cibuildwheel picks up pyproject.toml/setup.cfg expressions correctly.

henryiii:
Added unit tests and fixed a few naming issues (`python_requires` is under `[options]` in `setup.cfg`, not `[metadata]`).

unit_test/build_selector_test.py
```diff
@@ -69,3 +71,20 @@ def test_build_braces():
...
+    assert build_selector('cp37-manylinux1_x86_64')
+    assert not build_selector('cp27-manylinux1_i686')
+    assert build_selector('cp36-manylinux1_i686')
+    assert build_selector('cp37-manylinux1_i686')
+    assert not build_selector('cp27-win32')
+    assert build_selector('cp36-win32')
+    assert build_selector('cp37-win32')
+    assert not build_selector('pp27-win32')
+    assert build_selector('pp36-win32')
+    assert build_selector('pp37-win32')
```
joerick, line 90: Another test with something more exotic like `SpecifierSet(">=2.7.0, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*")` might be valuable. Also something with a patch version, like `SpecifierSet(">=2.7.9")`.

joerick (commented):
I think this API is going to work. The patch version question might throw some spanners into the works, but it shouldn't be insurmountable. Linux is the interesting one, because we don't hardcode the versions, perhaps we can call `python --version` to get it.

cibuildwheel/util.py
```diff
@@ -58,11 +60,25 @@ def read_python_configs(config: PlatformName) -> List[Dict[str, str]]:
...
+            py_ver_str = build_id.split('-')[0]
+            try:
+                major = int(py_ver_str[2])
+                minor = int(py_ver_str[3:])
+            except ValueError:
+                pass
+            else:
+                version = Version(f"{major}.{minor}")
+                if not self.requires_python.contains(version):
+                    return False
```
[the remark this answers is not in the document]
YannickJadoul: > Maybe a package specifies python_requires `>=3.9.1` because they rely on a bug fix?

That would never happen! What [crappy project maintainers](https://github.com/pybind/pybind11/pull/2576) would test the Python release candidates so badly that they [allow this to happen](https://github.com/pybind/pybind11/commit/2a263e087c0aa7a70a6ffc6dcfbfc0d3a5b9a2fa)? _cough cough_

docs/options.md
```diff
@@ -184,6 +184,35 @@ to limit builds for each job; this list could grow in the future.
...
+```toml
+[project]
+requires-python = ">=3.6"
+```
+
+Remember to always set `build-system.requires` and `build-system.build-backend`
+if you add a `pyproject.toml` file.
+
+Platform-specific variants also available:<br/>
+`CIBW_PROJECT_REQUIRES_PYTHON_MACOS` | `CIBW_PROJECT_REQUIRES_PYTHON_WINDOWS` | `CIBW_PROJECT_REQUIRES_PYTHON_LINUX`
```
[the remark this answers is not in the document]
henryiii: That's weird, I was sure I replaced this with your version (clearly remember the header change). Maybe I didn't push it?

unit_test/build_selector_test.py
henryiii: I've never seen a patch version here, but there's no reason it wouldn't be valid. I've been wanting got add minor versions to the macOS identifiers; it would have made the generator script simpler and would have been more consistent (even though we currently never need more than the first two digits. Actually, if we include packaging now, we could keep them Version's and tidy up a bit.

cibuildwheel/__main__.py
```diff
@@ -151,7 +154,30 @@ def main() -> None:
     test_extras = get_option_from_environment('CIBW_TEST_EXTRAS', platform=platform, default='')
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
-    build_selector = BuildSelector(build_config, skip_config)
+    # Passing this in as an environment variable will override pyproject.toml or setup.cfg
+    requires_python_str = get_option_from_environment('CIBW_PROJECT_REQUIRES_PYTHON', platform=platform)
```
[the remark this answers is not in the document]
YannickJadoul: Maybe that's the point where `pyproject.toml` is really the appropriate place to do so?

cibuildwheel/util.py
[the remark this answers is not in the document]
henryiii: `3.9.0 == 3.9` is true, sadly `3.9.1 == 3.9` is false (I thought it was true at one point so I likely have a few bad `version_requires` lying around; you really do need to write `!=3.4.*`).

> test the Python release candidates so badly 

Never! I'm sure it would show up [months before release](https://github.com/pybind/pybind11/issues/2391#issue-678783700)...

unit_test/build_selector_test.py
henryiii: I understand including it for completeness, but a wheel tag does not include the patch version of Python it was built on; a wheel built with 3.9.1 can be loaded on 3.9.0. So I'm not sure how putting a patch version in your Python-Requires could ever be useful. Honestly, it might be better to always pretend to be on the ".0" patch version, so mistaken identifiers like `>=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4, !=3.5` would still work (I'd exactly match pip here; I'd have to see what it does; I'd assume it checks the full version).

henryiii: Python-Requires tells you if the current version of Python is "good enough" to be used; it does not tell you to go download a specific version like an install requirement could. Since Python does not add new features in patch releases (ignore 2.7), you are not likely to "drop" support for a patch version, there's no reason to put a patch version in here. (pip would go back and load an older version of your software that did support a patch version). How about, for now, we add `.99` to any two digit version when checking this? Eventually we could try to compute an actual version, but there really isn't an answer for PyPy, and Linux will depend on the image, and you'd need to load it to detect what Python's where in it.  `>=3.6.3` and `!=3.6.6` would both correctly work if we ask, when picking the identifiers that match, if `3.6.99` matches.

As I've said before, this is about the Python language version, not the CPython version. And the language doesn't have patch versions. So I'd treat placing a patch version in Python-Requires here more as a user error.

joerick: > this is about the Python language version, not the CPython version. And the language doesn't have patch versions. 

This seems like the crux to me. Agreed. I'd be curious what PyPy does for the patch version, do they just put `0`? Anyway, the `.99` hack seems to make the most sense to me.

cibuildwheel/__main__.py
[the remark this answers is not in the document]
henryiii: So does thumbs up agree with leaving the variants in or removing them? 🙄

unit_test/build_selector_test.py
henryiii: I bet these values are the ones used: https://www.pypy.org/compat.html

henryiii: And here's the value for 3.6: https://morepypy.blogspot.com/2020/11/pypy-733-triple-release-python-37-36.html

henryiii:
Okay, so the latest changes:

* I've pulled out config file reading to a class. If we eventually look for anything else in these files, this will extend nicely to that without having to read them multiple times. It's also where the check for existence now happens.
* Checks are made assuming the highest patch version, 3.6 checks as 3.6.99. Eventually we could refactor this to use the actual patch version if there's a need, but there's a couple of things that need to happen first (I'll report a possible idea in #547).
* I've pushed the docs changes based on my slight tweaking on the rewrite in #528, maybe we can iterate from there?
* I needed the new option header from readthedocs to add it to the README, so haven't added it to the readme yet (TODO).

----

By the way, parsing the AST to pull out the value of `setup(python_requires="")` is really easy to do. This is what [setuptools does for `version=attr: ...`](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#specifying-values). Someone please tell me this is a bad idea...

```python
#!/usr/bin/env python3
from __future__ import annotations

import ast
from typing import Optional
import click
from pathlib import Path


class Analyzer(ast.NodeVisitor):
    def __init__(self):
        self.requires_python: Optional[str] = None

    def visit_Call(self, node):
        if getattr(node.func, "id", "") == "setup":
            for kw in node.keywords:
                self.generic_visit(kw)
                if kw.arg == "python_requires" and isinstance(kw.value, ast.Constant):
                    self.requires_python = kw.value.value


@click.command()
@click.argument('filename', type=click.Path(exists=True, file_okay=False, resolve_path=True))
def main(filename):
    with open(Path(filename) / "setup.py") as f:
        tree = ast.parse(f.read())
    analyzer = Analyzer()
    analyzer.visit(tree)
    print(analyzer.requires_python)


if __name__ == "__main__":
    main()
```

A simple assignment should be easy to add, too.

henryiii:
Oops, `CIBW_ARCHS_LINUX` is missing from the README, I tried copying the whole table out this time and noticed that. I'll make a PR that updates the known projects, and will fix that too.

cibuildwheel/__main__.py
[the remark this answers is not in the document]
joerick: Haha, my thumb meant removing them :)

unit_test/build_selector_test.py
YannickJadoul: > As I've said before, this is about the Python language version, not the CPython version. And the language doesn't have patch versions. So I'd treat placing a patch version in Python-Requires here more as a user error.

It shouldn't, no. But in practice, the only real existing standard for Python seems to be the CPython reference implementation (just ask PyPy how much they like this ;-) ). It's very rare/almost non-existant that the language itself changes, though, in patch releases, as I guess that's still not very semantically versioned.

cibuildwheel/projectfile.py
```diff
@@ -0,0 +1,50 @@
...
+
+import toml
+from packaging.specifiers import SpecifierSet
+
+
+class NoProjectFileError(RuntimeError):
+    pass
+
+
+class ProjectFile:
```
YannickJadoul, line 13: To me, this file feels a bit over-engineered. It makes sense in a huge project, but I've always liked that `cibuildwheel` is reasonably straightforward, more or less reads front-to-back what's happening, and doesn't have a dozen files spread around.

I currently don't really see the advantage of this whole class over just keeping the old check for the existence of the file, and then a simple function `utils.get_project_requires_python(package_dir)`. It's not like reading this file is going to cause `cibuildwheel` to become very slow/inefficient?

YannickJadoul: (For the record: I'm just being skeptical/questioning, here; I'm of course open to be convinced (or outvoted ;-) ) of the advantages of this construct, but I'm currently not sure this extra complexity, with custom exceptions etc, is worth it.)

cibuildwheel/__main__.py
```diff
@@ -183,15 +199,11 @@ def main() -> None:
     # This needs to be passed on to the docker container in linux.py
     os.environ['CIBUILDWHEEL'] = '1'
 
-    if not any((package_dir / name).exists()
```
YannickJadoul, line 186: That's funny. Is there a reason we only checked the existence of one of these files so late (after already trying to parse a bunch of options) ?

If so, we should keep this order. If not, well, good to move it up, maybe. But worth thinking what's the logical precedence of failure to the user. I.e. if both `CIBW_ENVIRONMENT` is malformed and there's no `setup.py`, what error is the most severe one, that you expect to be reported? I think I agree with `python_package` not being a Python package is more important to first report :-)

cibuildwheel/projectfile.py
henryiii: Personally, I like having small files with clear names, usually one per main class (helper classes can share files, as can support functions). If I see a class called ProjectFile, I'd like to have a file called projectfile.py. That way, you have a clear, focused set of includes instead of everything jumbled, you don't have items too heavily mixed, etc. `utils.py` is getting a bit crowded. I'd like to see a `architecture.py` with Architecture in it too, I think.

For _logic_, you want it to flow in one place. For classes, you want them in separate files, since they encapsulate logic. As long as the class is unit tested (which I could add) and you can trust it to work, then it simplifies the flow of logic where it is used.

Github even crosslinks if you click the class/function name, so I don't think it's too bad. :)

The design was general so that if we wanted to add more (like a setup.py parser, reading pyproject.toml to get config values, etc), it doesn't cause utils to grow, and all "project" related code could live together - existence checks and reading the files. Note that this always reads the files if present, while the old design was lazy and didn't read if nothing was needed (could use this and make it lazy, too, of course).

Custom exceptions are good! They let you be more specific when catching, and are extremely simple.

YannickJadoul: > If I see a class called ProjectFile, I'd like to have a file called projectfile.py.

That can make sense for a class this size, yes, but I'm doubting the overegineering of making this a class, vs. just the

```
if any_of_these_files_exist:
    exit_with_same_error_as_now()

# ... later, when reading options

if requires_python is None:
    requires_python = utils.get_requires_python(package_dir)
```

Especially the first part is a lot simpler than wrapping `project_file = ProjectFile(package_dir)` in a try-except and catching a new one-time-use exception. It's a simple check, so why not use `if`? I'm currently not convinced the extra complexity is worth it. It makes perfect sense in a large 50+ files project, but not for the current state of `cibuildwheel` (where we've been aiming to keep things transparent and procedural, I believe), to me.

henryiii: I think it makes sense to have a `.exists()` method that checks this, that would remove the exception and allow you to put the check wherever you wanted it to go.

henryiii: Let me try a redesign after the arch stuff goes in, and see if you like it better.

YannickJadoul: > I don't think utils is intended to read top to bottom.

No, utils isn't, ofc.

Maybe @joerick disagrees with me, and says this is more future-roof, though? It seems a bit heavy to me, right now, but ... I don't know what's coming later.

henryiii: Let me rewrite, and see if you like it better, I have some ideas. It was @joerick's original request that this be factored out, I had everything inline, originally.

YannickJadoul: Oh, I missed that. You mean https://github.com/joerick/cibuildwheel/pull/536#discussion_r561357274?

(Which does suggest "as a function", but OK, let's not split hairs too much :-P )

henryiii: "or something", and classes are better than functions :) unless it's just one shot; and this can at least combine existence check and config reading, and if we eventually add reading anything else from the config at all, it should be refactored to a class, so might as well do it here.

joerick: I haven't read all of the arguments above, but I don't think this warrants a class either. The trouble with classes is that they carry around hidden state, so they'd better be a good abstraction, otherwise you need to dig in and understand it and by then it's a net negative because you've got the inner workings of the class to worry about, as well as the task at hand.

I also think that generally, it's better to keep state in the filesystem, rather than in objects - that's a reason we don't have a `Project` object in cibuildwheel - it's just gonna read it from the filesystem anyway, so why try encapsulate that?

This coding style does feel unusual for some people (even sloppy), but we're optimising for lots of first-time readers, and they'll mostly be reading our code to understand why their build is failing. Literal is good.

As for the factoring out, a function seems good to me, because it can have an obvious name, that encapsulates the actual mechanism `read_package_python_requires(path)`.  But admittedly there is a little subjectivity to that!

docs/options.md
```diff
@@ -184,6 +184,61 @@ to limit builds for each job; this list could grow in the future.
 CIBW_ARCHS_LINUX: "auto aarch64"
 ```
 
+
+###  `CIBW_PROJECT_REQUIRES_PYTHON` {: #requires-python}
+> Limit the Python versions to build
```
joerick, line 189: I'm not gonna drop this :) "Limit the Python versions to build" is too close to BUILD/SKIP - that's why I had the 'override compatibility' line in my version.

docs/options.md
```diff
@@ -184,6 +184,61 @@ to limit builds for each job; this list could grow in the future.
...
+    build-backend = "setuptools.build_meta"
+    ```
+
+    If you don't want to use this yet, you can also use the currently
+    recommended location for setuptools in `setup.cfg`. Example `setup.cfg`:
+
+    ```ini
+    [options]
+    python_requires = ">=3.6"
+    ```
```
joerick, line 234: I discovered in 07f972f6d3ed65c5a552223e53d72b1ab738e34a that triple-backtick doesn't work inside the `!!! note` construct, but a 4-space indent does.

henryiii: But then you loose the syntax highlighting! Does `~~~` work?

docs/options.md
henryiii: That's historical, sorry, I didn't replace the title but just edited it, so the old one (or slightly modified old one?) stuck.

Currently trying to see why MyPy in pre-commit is not picking up an error in `__main__` for my rewrite. It's working when I run it manually.

henryiii: Aha, pre-commit adds `--ignore-missing-imports` by default.

docs/options.md
henryiii: No, it doesn't. :'(

henryiii:
Please take a look at the current version. It's still a class, but it has almost no state (just project_dir), and the existence check is optional and can be moved anywhere. I've also taken the one extra step to pull out python_requires if possible, I can back off on that. I've also got unit tests.

Docs need updates, `~~~` doesn't work, and they don't include setup.py (since I might remove the setup.py part).

cibuildwheel/__main__.py
```diff
@@ -156,7 +159,17 @@ def main() -> None:
...
 
-    build_selector = BuildSelector(build_config=build_config, skip_config=skip_config)
+    project_files = ProjectFiles(package_dir)
+
+    if not project_files.exists():
+        print('cibuildwheel: Could not find setup.py, setup.cfg or pyproject.toml at root of package', file=sys.stderr)
+        sys.exit(2)
+
+    # Passing this in as an environment variable will override pyproject.toml, setup.cfg, or setup.py
+    requires_python_str: Optional[str] = os.environ.get('CIBW_PROJECT_REQUIRES_PYTHON') or project_files.get_requires_python_str()
```
joerick, line 169: Since we're into the style discussion, I still think this would be simpler as a function.

While the class is mostly stateless, as a reader reading this code, you can't tell it's stateless without reading the source of the class. So we'd still have to jump to that file to read what 'exists' and 'get_requires_python_str' mean.

Once we get there, we have to understand those property accessors, which reference more accessors, the `dig` function, and then finally the `get_requires_python_str` method,  just to understand what's going on.

Going back to your original solution, but factored as a function-

```python
def get_package_python_requires(package_dir: Path) -> Optional[str]:
    pyproject_toml = package_dir / 'pyproject.toml'
    try:
        info = toml.load(pyproject_toml)
        return info['project']['requires-python']
    except (FileNotFoundError, KeyError):
        pass

    # Read in from setup.cfg:options.python_requires
    setup_cfg = package_dir / 'setup.cfg'
    config = ConfigParser()
    try:
        config.read(setup_cfg)
        return config['options']['python_requires']
    except (FileNotFoundError, KeyError):
        pass
    
    # TODO: read from setup.py
```

It's just... so much clearer!

(There's also a smell in that the error message `print('cibuildwheel: Could not find setup.py, setup.cfg or pyproject.toml at root of package'` understands the mechanism of `project_files.exists()`. I'd consider that a 'leaky abstraction', though it could be solved by a custom exception, it's probably not worth it in this case.)

henryiii:
On the `python_requires` check looking in setup.py, I've scanned all known projects, and here's the result. Green check means found (for setup.py, I've printed the parsed discovery). Grey check means it is set, but not able to be pulled from the AST with the current system (I've manually checked the setup.py's that do this, they all do something weird, like `**kw` or generating the `python_requires` with a `" ".join`). Red x means it is completely missing fro the file, but the file was there. Since it was right 16 times, and missed picking it up 6 times, and never "failed", I think it's a useful addition. Thoughts?

```
matplotlib/matplotlib:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
joerick/pyinstrument_cext:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aaugustin/websockets:
  setup.py: ✅ >=3.6.1
  setup.cfg: ❌
  pyproject.toml: Not found
YannickJadoul/Parselmouth:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
mayeut/pybase64:
  setup.py: ✅ >=3.5
  setup.cfg: ❌
  pyproject.toml: ❌
tommyod/KDEpy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
autopilot-rs/autopy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
pyrogram/tgcrypto:
  setup.py: ✅ ~=3.6
  setup.cfg: Not found
  pyproject.toml: Not found
twisted/twisted-iocpsupport:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
online-ml/river:
  setup.py: ✅ >=3.6.0
  setup.cfg: Not found
  pyproject.toml: Not found
PyAV-Org/PyAV:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aiortc:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aioquic:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pikepdf/pikepdf:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
stfbnc/fathon:
  setup.py: ✅ >=3.5
  setup.cfg: Not found
  pyproject.toml: Not found
etesync/etebase-py:
  setup.py: ✅ >=3
  setup.cfg: Not found
  pyproject.toml: ❌
dimitern/xmlstarlet:
  setup.py: ✅ >=3.6.*
  setup.cfg: ❌
  pyproject.toml: Not found
bxlab/bx-python:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
nedbat/coveragepy:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
scikit-learn/scikit-learn:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-image/scikit-image:
  setup.py: ✅ >=3.7
  setup.cfg: ❌
  pyproject.toml: ❌
google/neuroglancer:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
h5py/h5py:
  setup.py: ✅ >=3.7
  setup.cfg: Not found
  pyproject.toml: ❌
zeromq/pyzmq:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: ❌
python-rapidjson/python-rapidjson:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
mwilliamson/jq.py:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/python_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/cmake_example:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
pybind/scikit_build_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/iminuit:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-hep/pyjet:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
scikit-hep/numpythia:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/boost-histogram:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
Toblerity/rtree:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
giampaolo/psutil:
  setup.py: ✅ >=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
  setup.cfg: Not found
  pyproject.toml: Not found
ets-labs/python-dependency-injector:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pydata/numexpr:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
PyTables/PyTables:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: Not found
DataDog/dd-trace-py:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: ❌
  pyproject.toml: ❌
OpenNMT/Tokenizer:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
matrix-profile-foundation/matrixprofile:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
brentp/cyvcf2:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
OpenNMT/CTranslate2:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
mypyc/mypy_mypyc-wheels:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
czaki/imagecodecs_build:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
deepcharles/ruptures:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
```

joerick:
Nice work! One question- How many of these use the `visit_Assign` part of the AST visitor? That was the only bit that gave me pause, as those assignments could be within an `if`, in which case they might not be executed. Also it's easier to explain if we say the detection only works if you pass python_requires like `setup(python_requires='>=3.7')`

cibuildwheel/__main__.py
YannickJadoul: Yes. I agree with @henryiii that the original class design abstract things better in a large, complex project, but part of the charm of `cibuildwheel` (at least to me) is that it's almost "just a large script" that glues things together, rather than a full OO program.

YannickJadoul: (or at least, that's the current state and style, and it seems like that has worked and made things flexible and intuitive so far)

henryiii:
Python requires is static metadata that gets burned into the wheel (actually it might even be in the SDist metadata too?), so it's very unlikely to be in an if, and we are not currently checking to see if the setup itself is in an if. I didn't quite meant do this, but here's a run where I only look for a keyword called python_requires, ignoring what it is in (so `dict(python_requires=...)` works). This gets all but three.

We could tighten up the check to look for unnested "setup" functions only. It seems alignment is pretty rare; adding it to a dict is a bit more common.

```
matplotlib/matplotlib:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
joerick/pyinstrument_cext:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aaugustin/websockets:
  setup.py: ✅ >=3.6.1
  setup.cfg: ❌
  pyproject.toml: Not found
YannickJadoul/Parselmouth:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
mayeut/pybase64:
  setup.py: ✅ >=3.5
  setup.cfg: ❌
  pyproject.toml: ❌
tommyod/KDEpy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
autopilot-rs/autopy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
pyrogram/tgcrypto:
  setup.py: ✅ ~=3.6
  setup.cfg: Not found
  pyproject.toml: Not found
twisted/twisted-iocpsupport:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
online-ml/river:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
PyAV-Org/PyAV:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aiortc:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aioquic:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pikepdf/pikepdf:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
stfbnc/fathon:
  setup.py: ✅ >=3.5
  setup.cfg: Not found
  pyproject.toml: Not found
etesync/etebase-py:
  setup.py: ✅ >=3
  setup.cfg: Not found
  pyproject.toml: ❌
dimitern/xmlstarlet:
  setup.py: ✅ >=3.6.*
  setup.cfg: ❌
  pyproject.toml: Not found
bxlab/bx-python:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: ❌
  pyproject.toml: Not found
nedbat/coveragepy:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4
  setup.cfg: ❌
  pyproject.toml: Not found
scikit-learn/scikit-learn:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-image/scikit-image:
  setup.py: ✅ >=3.7
  setup.cfg: ❌
  pyproject.toml: ❌
google/neuroglancer:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
h5py/h5py:
  setup.py: ✅ >=3.7
  setup.cfg: Not found
  pyproject.toml: ❌
zeromq/pyzmq:
  setup.py: ✅ >=3.6
  setup.cfg: Not found
  pyproject.toml: ❌
python-rapidjson/python-rapidjson:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
mwilliamson/jq.py:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/python_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/cmake_example:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
pybind/scikit_build_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/iminuit:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-hep/pyjet:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
scikit-hep/numpythia:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/boost-histogram:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
Toblerity/rtree:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
giampaolo/psutil:
  setup.py: ✅ >=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
  setup.cfg: Not found
  pyproject.toml: Not found
ets-labs/python-dependency-injector:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pydata/numexpr:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
PyTables/PyTables:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: Not found
DataDog/dd-trace-py:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: ❌
  pyproject.toml: ❌
OpenNMT/Tokenizer:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
matrix-profile-foundation/matrixprofile:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
brentp/cyvcf2:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
OpenNMT/CTranslate2:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
mypyc/mypy_mypyc-wheels:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
czaki/imagecodecs_build:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
deepcharles/ruptures:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
```

henryiii:
Ahhh, shouldn't have silenced MyPy about conftest, it was right. Ended up wasting a CI run when I could have just listened to MyPy....

cibuildwheel/util.py
```diff
@@ -61,11 +63,25 @@ class IdentifierSelector:
...
     """
-    def __init__(self, *, build_config: str, skip_config: str):
+    def __init__(self, *, build_config: str, skip_config: str, requires_python: Optional[SpecifierSet] = None):
         self.build_patterns = build_config.split()
         self.skip_patterns = skip_config.split()
+        self.requires_python = requires_python
 
     def __call__(self, build_id: str) -> bool:
+        # Filter build selectors by python_requires if set
+        if self.requires_python is not None and '-' in build_id:
```
YannickJadoul, line 73: Shouldn't we assert for `'-' in build_id`?

cibuildwheel/projectfiles.py
```diff
@@ -0,0 +1,82 @@
...
+            if isinstance(node.value, Constant):
+                self.requires_python = get_constant(node.value)
+            elif isinstance(node.value, ast.Name):
+                self.requires_python = self.constants.get(node.value.id)
+
+
+def setup_py_python_requires(content: str) -> Optional[str]:
+    try:
+        tree = ast.parse(content)
+        analyzer = Analyzer()
```
YannickJadoul, line 47: This is very cool, but I'm wondering if it's worth the code. All these projects are probably already setting `CIBW_SKIP`, anyway?

cibuildwheel/projectfiles.py
```diff
@@ -0,0 +1,82 @@
...
+        analyzer.visit(tree)
+        return analyzer.requires_python or None
+    except Exception:
+        return None
+
+
+def get_requires_python_str(package_dir: Path) -> Optional[str]:
+    "Return the python requires string from the most canonical source available, or None"
+
+    setup_py = package_dir / 'setup.py'
```
YannickJadoul, line 57: Nitpicking, but why not just inline these? `package_dir / 'setup.py'` is just as expressive as `setup_py`, no?

cibuildwheel/__main__.py
```diff
@@ -156,7 +159,19 @@ def main() -> None:
     test_extras = get_option_from_environment('CIBW_TEST_EXTRAS', platform=platform, default='')
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
-    build_selector = BuildSelector(build_config=build_config, skip_config=skip_config)
+    setup_py = package_dir / 'setup.py'
+    setup_cfg = package_dir / 'setup.cfg'
+    pyproject_toml = package_dir / 'pyproject.toml'
+
+    if not pyproject_toml.exists() and not setup_cfg.exists() and not setup_py.exists():
```
YannickJadoul, line 166: Not wrong, but the previous

```
if not any((package_dir / name).exists()
           for name in ["setup.py", "setup.cfg", "pyproject.toml"]):
```
was arguably a tiny bit more complex, but also less code duplication. (I have honestly no clue who wrote it.)

cibuildwheel/projectfiles.py
YannickJadoul: For the record, you've done all the work already and I don't think this is something that would be maintained often, so we can just as well keep it? :-)

cibuildwheel/projectfiles.py
henryiii: Just mimicking what was in main. There, `pyproject_toml.exists()` describes exactly what you mean (Does pyproject.toml exist?), while` `(package_dir / "pyproject.toml").exists()` is ugly due to the parenthesis and `package_dir.joinpath("pyproject.toml").exists()` adds an extra term (and is not as common as `/`).

cibuildwheel/__main__.py
henryiii: Could even be:


```python
if not any(p.exists() for p in {setup_py, setup_cfg, pyproject_toml}):
```

Don't have a strong preference, really, it's mostly different than what was there before because I had replaced it then rewrote it.

cibuildwheel/projectfiles.py
henryiii: The goal is to improve the first run experience, that is, #468 - this was just a useful database to run from. You shouldn't have to set up build selectors just to limit your Python versions to the versions you support. A _lot_ of projects use static metadata in setup.py still (mostly due to old examples, old tutorials, and thinking that using "newer" techniques will cause it to break for some users - though it's almost [always the reverse](https://github.com/scikit-hep/uproot4/pull/231), from my experience, and there's a good chance that dependencies use setup.cfg for metadata already, like twine and wheel). It would be nice if, when adding cibuildwheel to a new project, it would just behave nicely, and this improves the chances of that happening.

I actually originally wrote the logic for a script to help with moving from setup.py to setup.cfg, and just adapted it here.

cibuildwheel/util.py
henryiii: We haven't in the past, and there are unit tests that put really weird things in here. That's why this is here. :)

cibuildwheel/projectfiles.py
YannickJadoul: Hmmm, myes, that's the one problem with `pathlib` :-(

Though you're not calling `exists()` here.

cibuildwheel/projectfiles.py
YannickJadoul: Selecting the things to be built has so far always worked, without any complaints. To me, `CIBW_BUILD` and `CIBW_SKIP` are pretty basic options, so I don't fully agree that this is a must-have for the first run experience.
It's a nice-to-have feature, yes, but still not really bullet-proof, either, so I'm not entirely convinced this is worth the code. But not against keeping it, either.

cibuildwheel/__main__.py
YannickJadoul: Then you still have the repetition, though, of `package_dir /` and `setup.py`/`setup_py`.

> it's mostly different than what was there before because I had replaced it then rewrote it.

But I guessed so (and couldn't see a convincing reason for the change), so that's why I wanted to point it out.

cibuildwheel/util.py
YannickJadoul: Aaaaaaaaah, right. Forgot about the unit tests again :-/
Hmmm, that's annoying, changing our code's assumptions to fit unit tests, but... changing unit tests to work around this is almost just as fishy.

henryiii:
Okay, only allowing a top-level function call, no assignments, and still gets us a pretty nice >50%:

```
./bin/inspect_all_known_projects.py
matplotlib/matplotlib:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
joerick/pyinstrument_cext:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aaugustin/websockets:
  setup.py: ✅ >=3.6.1
  setup.cfg: ❌
  pyproject.toml: Not found
YannickJadoul/Parselmouth:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
mayeut/pybase64:
  setup.py: ✅ >=3.5
  setup.cfg: ❌
  pyproject.toml: ❌
tommyod/KDEpy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
autopilot-rs/autopy:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
pyrogram/tgcrypto:
  setup.py: ✅ ~=3.6
  setup.cfg: Not found
  pyproject.toml: Not found
twisted/twisted-iocpsupport:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
online-ml/river:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
PyAV-Org/PyAV:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aiortc:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
aiortc/aioquic:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pikepdf/pikepdf:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: ❌
stfbnc/fathon:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
etesync/etebase-py:
  setup.py: ✅ >=3
  setup.cfg: Not found
  pyproject.toml: ❌
dimitern/xmlstarlet:
  setup.py: ✅ >=3.6.*
  setup.cfg: ❌
  pyproject.toml: Not found
bxlab/bx-python:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
nedbat/coveragepy:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4
  setup.cfg: ❌
  pyproject.toml: Not found
scikit-learn/scikit-learn:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-image/scikit-image:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: ❌
google/neuroglancer:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
h5py/h5py:
  setup.py: ✅ >=3.7
  setup.cfg: Not found
  pyproject.toml: ❌
zeromq/pyzmq:
  setup.py: ✅ >=3.6
  setup.cfg: Not found
  pyproject.toml: ❌
python-rapidjson/python-rapidjson:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
mwilliamson/jq.py:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/python_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
pybind/cmake_example:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: ❌
pybind/scikit_build_example:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/iminuit:
  setup.py: ✅ >=3.6
  setup.cfg: ❌
  pyproject.toml: ❌
scikit-hep/pyjet:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
scikit-hep/numpythia:
  setup.py: ✅ >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
  setup.cfg: Not found
  pyproject.toml: ❌
scikit-hep/boost-histogram:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
Toblerity/rtree:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: ❌
giampaolo/psutil:
  setup.py: ☑️
  setup.cfg: Not found
  pyproject.toml: Not found
ets-labs/python-dependency-injector:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
pydata/numexpr:
  setup.py: ❌
  setup.cfg: Not found
  pyproject.toml: Not found
PyTables/PyTables:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: Not found
DataDog/dd-trace-py:
  setup.py: ☑️
  setup.cfg: ❌
  pyproject.toml: ❌
OpenNMT/Tokenizer:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
matrix-profile-foundation/matrixprofile:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
brentp/cyvcf2:
  setup.py: ❌
  setup.cfg: ❌
  pyproject.toml: Not found
OpenNMT/CTranslate2:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
mypyc/mypy_mypyc-wheels:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
czaki/imagecodecs_build:
  setup.py: Not found
  setup.cfg: Not found
  pyproject.toml: Not found
deepcharles/ruptures:
  setup.py: ❌
  setup.cfg: ✅
  pyproject.toml: ❌
```

YannickJadoul:
Curious intermezzo, slightly off-topic question for @henryiii: is there an advantage to using `pyproject.toml` for this rather than `setup.cfg`?

henryiii:
> is there an advantage to using `pyproject.toml` for this rather than `setup.cfg`?

Well, this will be the canonical location for this information, and will be following a standard (PEP 621), rather than being tool specific like setup.cfg or `[tool.*]` in pyproject.toml. Things like the package name can be specified in one place and picked up regardless of whether you use setuptools, poetry, flit, enscons, or something else.

However, today, there isn't an implementation of a major tool that will pick this up from pyproject.toml yet. Setuptools will take time to add it (Marchish?), and flit/poetry are kind of balking at going first (after they were trying to lead originally...). At least, in poetry's case, they like their complicated requirement system better than the standard one. I personally still see their system as solving a problem that didn't exist, and being more complex than required, so looking forward to being able to use the standard location and spelling. :)

cibuildwheel/projectfiles.py
henryiii: Pretty sure I linked to at least one complaint in my previous comment. ;)

cibuildwheel/__main__.py
henryiii: I went pack to something closer to what we had before.

cibuildwheel/projectfiles.py
YannickJadoul: Hmmm, valid point :-D I thought I had remembered that title to be an internal discussion, but obviously not, if you take the time to actually read.

henryiii: It was renamed, originally was a bug report for Python 2 showing up in 2020. :)

YannickJadoul: And arguably not wrong, as bug report, without a bit of context ;-)

henryiii:
Fiddled with the MyPy configuration quite a bit, part of the 0.800 update and it's extra fixes for namespace packaging causing our workarounds for multiple conftest.py's. I've made them actual, normal packages, which fixes the import check from MyPy as well as lets us do a relative import in the one test that actually imports from a conftest.py. Should be much better now, and we can control the test mypy coverage properly now. The bin directory is now covered in pre-commit under the 3.7 language level; you have to run `mypy --python-version 3.7 bin` manually to test it by hand. Now that I think of it, maybe we should set the language level in the config to 3.7 and then add the 3.6 to pre-commit only...

henryiii:
Enjoy the 3.7 + mypy 0.800 code in the new script. :)

henryiii:
Looks like we were not the first to pick up PEP 621: https://frostming.com/2021/01-22/introducing-pdm/#pdm----a-new-python-package-manager-and-workflow-tool

setup.cfg
```diff
@@ -45,14 +46,15 @@ dev =
...
     pymdown-extensions
-    pytest
+    pytest>=4
     pytest-timeout
     pyyaml
     requests
     typing-extensions
-    packaging>=20.8
     rich>=9.6
+    mypy>=0.800
```
joerick, line 57: Is that mypy in the list twice now? Maybe we need to remove it on line 46?

unit_test/build_selector_test.py
joerick: The underlying question was resolved, but I still think-

> Another test with something more exotic like SpecifierSet(">=2.7.0, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*") might be valuable. Also something with a patch version, like SpecifierSet(">=2.7.9").

cibuildwheel/util.py
joerick: mmmyeah, this does seem a little wrong... also the ValueError catch below. These shouldn't be hit in real use, so I think we should fix the unit tests. It looks like only `test_build_selector` in `main_options_test.py`?

joerick (commented):
Cool, I think this is pretty close.

setup.cfg
henryiii: Oops! 😳

cibuildwheel/projectfiles.py
henryiii: This is even cooler: https://github.com/henryiii/setuptools-modernize/blob/main/src/setuptools_modernize/parse.py :)

unit_test/build_selector_test.py
henryiii: I could have sworn I added this...

joerick (approved):
Looks good!

henryiii:
@YannickJadoul Would you like to do a once-over too?

YannickJadoul:
> @YannickJadoul Would you like to do a once-over too?

I kind of followed this along the way, but let me quickly check!

YannickJadoul (approved):
Slightly surprised this is more code than I had expected, but yeah, I think I had already made my complaints along the way :-)

Good job, as ever, @henryiii! :-)

Issue #529, opened 2021-01-10
CIBW_TEST_SKIP or CIBW_TEST_SELECT


henryiii:
This is a proposal for something like CIBW_TEST_SKIP, using build selectors to either eliminate or pick items to run on. After #527, maybe it should be a "positive" selector, since it's now easy to write both positive and negative selectors in one expression, and "CIBW_SKIP" might fade away. Not sure what to call it, `CIBW_TEST_BUILD` sounds a little odd, but is somewhat obvious that it's the same syntax.

-----

> As it is now, I often have to turn off testing on some jobs due building more wheels than upstream requirements (NumPy, generally). It's not that big of an issue to ship with wheel like NumPy missing further up the chain; many users get NumPy from other locations (package mangers, conda, etc). But it's already irritating that I have to turn off all testing when NumPy does support 1 or 2 in the set (PyPy 3.6, or some specific ARM one). I could split the build even further, so that I can have "testable" vs. "not testable". I almost see the use case for `CIBW_TEST_SKIP=`pp37* ...`. :) But if we do dual-testing on Universal2, that would likely force users to skip all testing if something was not supported, instead of testing half. In fact, if we got ARM runners tomorrow, we might only be able to test on the x86_64 part for some packages! Maybe by the time we get ARM runners, this won't be so common, but some major packages are rather slow to add new wheels. (NumPy. NumPy. NumPy.)

_Originally posted by @henryiii in https://github.com/joerick/cibuildwheel/issues/484#issuecomment-757288860_

joerick:
As noted in #484, this seems like a sensible route to me. Might also be helpful for people building under emulation where the tests are mega slow.

Perhaps worth debating how universal2 tests might fit into this, as maybe we could kill two birds with one stone. Could we have some identifiers like `cp39-macosx_universal2[arm64]` `cp39-macosx_universal2[x86_64]` that identified different archs within a single wheel... that might relieve the need for the previously mentioned `CIBW_TESTS_ARCHS`.

> After #527, maybe it should be a "positive" selector, since it's now easy to write both positive and negative selectors in one expression, and "CIBW_SKIP" might fade away.

I prefer skip, personally, as it's more natural to start with all the builds running tests, then remove them (default: `` 'skip nothing'). (I also don't think CIBW_SKIP is going anywhere, for similar reasons.)

YannickJadoul:
[...]

PR #537, opened 2021-01-15
feat: CIBW_TEST_SKIP

henryiii:
Addressing #529 (no universal yet, but that can be added when #484 is in). Just a SKIP, since that seemed to be the consensus? Without the `!` syntax in BUILD I guess only SKIP makes sense (though it is very easy to add).

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -140,6 +140,7 @@ def main() -> None:
         assert_never(platform)
 
     build_config, skip_config = os.environ.get('CIBW_BUILD', '*'), os.environ.get('CIBW_SKIP', '')
+    test_skip = os.environ.get('CIBW_TEST_SKIP', '')
     environment_config = get_option_from_environment('CIBW_ENVIRONMENT', platform=platform, default='')
     before_all = get_option_from_environment('CIBW_BEFORE_ALL', platform=platform, default='')
     before_build = get_option_from_environment('CIBW_BEFORE_BUILD', platform=platform)
@@ -152,6 +153,7 @@ def main() -> None:
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
     build_selector = BuildSelector(build_config, skip_config)
+    test_selector = BuildSelector("*", test_skip)
 
     try:
         environment = parse_environment(environment_config)
@@ -236,6 +238,7 @@ def main() -> None:
         before_all=before_all,
         build_verbosity=build_verbosity,
         build_selector=build_selector,
+        test_selector=test_selector,
         repair_command=repair_command,
         environment=environment,
         dependency_constraints=dependency_constraints,
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -18,6 +18,7 @@ from cibuildwheel.util import (
     BuildOptions,
     BuildSelector,
     DependencyConstraints,
+    TestSelector,
     Unbuffered,
     detect_ci_provider,
     resources_dir,
@@ -152,8 +153,8 @@ def main() -> None:
     test_extras = get_option_from_environment('CIBW_TEST_EXTRAS', platform=platform, default='')
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
-    build_selector = BuildSelector(build_config, skip_config)
-    test_selector = BuildSelector("*", test_skip)
+    build_selector = BuildSelector(build_config=build_config, skip_config=skip_config)
+    test_selector = TestSelector(skip_config=test_skip)
 
     try:
         environment = parse_environment(environment_config)
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -204,6 +204,7 @@ class BuildOptions(NamedTuple):
     package_dir: Path
     output_dir: Path
     build_selector: BuildSelector
+    test_selector: BuildSelector
     architectures: Set[Architecture]
     environment: ParsedEnvironment
     before_all: str
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -55,7 +55,12 @@ def read_python_configs(config: PlatformName) -> List[Dict[str, str]]:
     return results
 
 
-class IndentifierSelector:
+class IdentifierSelector:
+    """
+    This class holds a set of build/skip patterns. You call an instance with a
+    build identifier, and it returns True if that identifier should be
+    included.
+    """
     def __init__(self, *, build_config: str, skip_config: str):
         self.build_patterns = build_config.split()
         self.skip_patterns = skip_config.split()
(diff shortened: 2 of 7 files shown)


cibuildwheel/util.py
```diff
@@ -204,6 +204,7 @@ class BuildOptions(NamedTuple):
     package_dir: Path
     output_dir: Path
     build_selector: BuildSelector
+    test_selector: BuildSelector
```
YannickJadoul, line 207: Nitpicking, once again, but this should probably move down ... `n` lines, before or after `test_command`?

cibuildwheel/__main__.py
```diff
@@ -152,6 +153,7 @@ def main() -> None:
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
     build_selector = BuildSelector(build_config, skip_config)
+    test_selector = BuildSelector("*", test_skip)
```
YannickJadoul, line 156: If you were removing double quotes, this is one ;-)

cibuildwheel/__main__.py
```diff
@@ -152,6 +153,7 @@ def main() -> None:
     build_verbosity_str = get_option_from_environment('CIBW_BUILD_VERBOSITY', platform=platform, default='')
 
     build_selector = BuildSelector(build_config, skip_config)
```
joerick, line 155: We should probably rename this `BuildSelector` class, now it's being used for build and test. Could be `IdentifierSelector`?

joerick (commented):
Looks good, nice and simple. Just the docs to go, I think :)

cibuildwheel/util.py
henryiii: NamedTuple was designed for adding names to an API that already was using tuples, such as for return values. We probably should be using a dataclass, which does not have an order and the tuple behaviors, which would then not be possible to interact with "by order" so order wouldn't matter.

Anyway, I'll move it.

cibuildwheel/__main__.py
henryiii: I'd rather wait on this, since I'm thinking that if we have add selectors like `*_universal2/arm64`, we might end up splitting this off into a TestSelector, leaving the BuildSelector alone. Depends on the implementation; if if it easy we could leave it one class and then rename it. Or we can make a `IdentifierSelector` with `BuildSelector` and `TestSelector` subclasses to be ready for potential changes like that. :) (But you seem to not like inheritance?)

cibuildwheel/util.py
YannickJadoul: Oh, this isn't about the order of the tuple, btw! Just about having an order in the code that's consistent with the order in the docs, and the docs having a reasonably chronological order (e.g., test options grouped, and things +- in the order in which you need them in the build process).

All for dataclasses, yes! :-) But they need 3.7, right?

henryiii: No, they have a backport for 3.6. Due to the design, that's as far back as you can go. Attrs is also an option, personally I go to that pretty quick because it's so much more powerful and I usually need something it can do. I think we are intending to drop 3.6 soonish? Though I want to see what the default versions of Python 3 on all GitHub runners are first. https://docs.github.com/en/actions/guides/building-and-testing-python#using-the-default-python-version , just in case we can drop the requirement that our action has to have the setup action first.

No changes for now, though, just a potential thought for a refactor. We don't have an external API, so the fact users can access these by order doesn't really hurt us since we are the only users and we know not to do that. :)

YannickJadoul: Ah, didn't know that. But yes, agreed it's better. One of the satisfying things of dropping older Python versions ;-)

YannickJadoul: > We don't have an external API, so the fact users can access these by order doesn't really hurt us since we are the only users and we know not to do that. :)

Exactly, purely nitpicking to make our future live easier when looking for that one option, switching back and forth between code and docs. Well, and consistency sometimes helps to spot bugs/weird things ;-)

henryiii: Not against nitpicking, or strict requirements. :)

YannickJadoul: Alright. If there's nothing else to complain about, then... well, yeah, I need to start nitpicking ;-)

(Open for other things, btw; I just really like keeping an intuitive/consistent order, here, for some reason :-) )

cibuildwheel/__main__.py
henryiii: I've updated this, see if you like it better.  I have an `IdentifierSelector` with `BuildSelector` and `TestSelector`; after Universal2 does in, TestSelector can have extra logic in it if needed for Universal2.

cibuildwheel/util.py
henryiii: Moved.

cibuildwheel/util.py
```diff
@@ -55,8 +55,8 @@ def read_python_configs(config: PlatformName) -> List[Dict[str, str]]:
     return results
 
 
-class BuildSelector:
-    def __init__(self, build_config: str, skip_config: str):
+class IndentifierSelector:
```
joerick, line 58: Typo- `IdentifierSelector`. But otherwise this design works for me :)

cibuildwheel/__main__.py
joerick: I doubt we'd need extra logic, but the different constructor makes some sense.

cibuildwheel/util.py
henryiii: I had to stare at this for a minute to figure out what was different. 😳

henryiii:
Docs are ready for review too. ;)

docs/options.md
```diff
@@ -518,6 +518,18 @@ Platform-specific variants also available:<br/>
 CIBW_TEST_EXTRAS: test,qt
 ```
 
+### `CIBW_TEST_SKIP` {: #test-skip}
+> Match selectors that should be skipped
```
joerick, line 522: Maybe this is more descriptive of what it's for?
```suggestion
> Skip running tests on some builds
```

joerick (approved):
Thanks @henryiii , this looks good! 

Just one thing before merge, could you update the options table in the README with this new option?

(A tip - if you open the JS console on the options page e.g. https://cibuildwheel--537.org.readthedocs.build/en/537/options/ , generated README markdown can be copied from there! Kinda weird I know but that was how it was created the first time :) )

YannickJadoul:
> https://cibuildwheel--537.org.readthedocs.build/en/537/options/

Huh? There's a dedicated docs page of each PR? I never knew about this! Did you perform any dark magic to get this? :-O

henryiii:
Click on the "readthedocs" "check" and it opens the docs for the PR itself, wow. :)

YannickJadoul:
Oh, right. Seems I need to read up on this: https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html :-)

YannickJadoul:
Of course. I should read my emails/notifications not in random order ;-) https://github.com/joerick/cibuildwheel/pull/550#issuecomment-764189264

PR #539, opened 2021-01-15
fix: improve correctness of exits

henryiii:
Changes some raw uses of `exit()` to either `sys.exit` or `raise Exit` from click. Just a correctness fix (unless site.py is not loaded).

### DIFF
--- a/bin/bump_version.py
+++ b/bin/bump_version.py
@@ -4,11 +4,11 @@
 import glob
 import os
 import subprocess
+import sys
 import urllib.parse
 from pathlib import Path
 
 import click
-from click.exceptions import Exit
 from packaging.version import InvalidVersion, Version
 
 import cibuildwheel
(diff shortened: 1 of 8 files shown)


joerick:
Could we use `sys.exit` rather than `raise Exit`? Seems like the former is more universal.

henryiii:
Sure. `sys.exit` really just raises `SystemExit` anyway, so they are about the same.

joerick:
Thanks. I like to keep things stdlib where possible, keeps the codebase more approachable to newcomers :)

henryiii:
If this was The Way™ to do it in Click, I'd say we should use it, because we are using Click, and we should look like other Click programs. But AFAICT, it is not more common or even that advertised. I _think_ the most "normal" way to do it is to call `sys.exit(ClickApp())` and return an int from the app. Click is kind of specialized for chaining calls.

henryiii:
By the way, do you like `sys.exit(1)`, or return 1 and then wrapping the call in `sys.exit`? I'll do the former tonight unless you like the latter.

joerick (approved):
The `sys.exit` version seems good to me

PR #541, opened 2021-01-16
ci: experiments

henryiii:
This might have an issue, noticed when working on the action.﻿

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-20.04, windows-latest, macos-latest]
         python_version: ['3.7']
     timeout-minutes: 180
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,12 +43,8 @@ jobs:
       run: |
         choco install vcpython27 -f -y
 
-    - name: Get PR labels
-      id: pr-labels
-      uses: joerick/pr-labels-action@v1.0.6
-
     - name: Sample build
-      if: contains(steps.pr-labels.outputs.labels, ' ci-sample-build ')
+      if: "contains(github.event.pull_request.labels.*.name, 'CI: Sample build')"
       run: |
         python bin/sample_build.py
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,7 @@ jobs:
         python bin/sample_build.py
 
     - name: Test cibuildwheel
+      shell: pwsh
       run: |
         python ./bin/run_tests.py
 


henryiii:
```
Run joerick/pr-labels-action@v1.0.6
No labels found
```

?

henryiii:
Curious, this works, will continue to investigate why my run failed on 20.04. Actually, I am using a custom manylinux1 image that might have issues with 20.04 somehow?

I dropped the custom action and am just using native syntax for accessing labels, and fixed the label name. Since 20.04 is about to become the default, maybe merging this PR is a good idea? Should we update the examples?

.github/workflows/test.yml
```diff
@@ -43,12 +43,8 @@ jobs:
       run: |
         choco install vcpython27 -f -y
 
-    - name: Get PR labels
-      id: pr-labels
-      uses: joerick/pr-labels-action@v1.0.6
```
YannickJadoul, line 48: @joerick, this seems like an interesting/important consequence towards your custom action/repository?

YannickJadoul:
Shouldn't/couldn't we just grow along with `-latest`? Of course, Ubuntu 22.04 is still quite far away, so it won't really make that much of a difference, afterwards.

henryiii:
Since `-latest` is about to change, I at least want(ed) to test with it.

.github/workflows/test.yml
henryiii: Maybe this pre-dates the syntax for accessing labels?

.github/workflows/test.yml
```diff
@@ -43,12 +43,8 @@ jobs:
...
-
     - name: Sample build
-      if: contains(steps.pr-labels.outputs.labels, ' ci-sample-build ')
+      if: "contains(github.event.pull_request.labels.*.name, 'CI: Sample build')"
```
henryiii, line 47: We should also decide what we want to do for non pull_request builds. Should we only build samples if requested on pull_requests? It should also be possible via `workflow_dispatch`, what about on master?

YannickJadoul: What if we set up some kind of weekly cron trigger? I had already forgotten there was this label to trigger things :-/ Once a week (or so) doesn't seem to crazy when it comes to using resources?

henryiii:
Trying to see if I can get `cibuildwheel` to run directly in powershell. Nope, that didn't trigger what I've been seeing.

.github/workflows/test.yml
joerick: No idea why it wasn't working, it [still works on my repo](https://github.com/joerick/pr-labels-action/pull/5/checks?check_run_id=1717049797). Labels can be missed depending on when they're added - they're snapshotted when the build was triggered - perhaps the label was added after the build was triggered?

henryiii: I restarted the build after adding the label, it should have been picked up? Though I didn't try closing and reopening.

Is there a benefit to the action over just using an if?

henryiii:
Closing so as to not take focus off of more important PRs. Though still wonder if this can be simplified with a native if instead of a custom action.

PR #543, opened 2021-01-17
fix: empty image check

henryiii:
Would have helped me debug an issue testing out #542, where I changed a matrix and ended up with empty image names.﻿

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -30,6 +30,9 @@ class DockerContainer:
     bash_stdout: IO[bytes]
 
     def __init__(self, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None):
+        if not docker_image:
+            raise RuntimeError("Must have a non-empty docker image to run.")
+
         self.docker_image = docker_image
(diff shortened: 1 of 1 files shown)


cibuildwheel/docker_container.py
```diff
@@ -30,6 +30,9 @@ class DockerContainer:
     bash_stdout: IO[bytes]
 
     def __init__(self, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None):
+        if not docker_image:
```
YannickJadoul, line 33: Oh, I was expecting it in `__main__`, where we're reading the docker image. But maybe this makes even more sense.

cibuildwheel/docker_container.py
```diff
@@ -30,6 +30,9 @@ class DockerContainer:
...
+            raise RuntimeError("Must have a non-empty docker image to run.")
```
joerick, line 34: To nit-pick the most minor of minor points... 
```suggestion
            raise ValueError("Must have a non-empty docker image to run.")
```

joerick (approved):
Good idea, thanks for this

PR #545, opened 2021-01-18
feat: error if empty build selection, Travis 2.7 Win w/ workaround

henryiii:
This adds an error message and exit code if no build selectors match, catching runs that don't actually build anything at all. I think this came up once before? Can't find it now. Hopefully the conclusion was it was a good idea?

#536 could build on this and print a smart error message if it's the requires-python part causing it to be empty.

This uncovered a few issues in testing, most importantly that `test_cpp17_py27_modern_msvc_workaround` passes even if not supported, since it's comparing an empty set with an empty set. I've allowed Travis to participate in the MSVC workaround to fix it.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -195,9 +195,6 @@ def main() -> None:
     archs = Architecture.parse_config(archs_config_str, platform=platform)
 
     identifiers = get_build_identifiers(platform, build_selector, archs)
-    if not identifiers:
-        print("ERROR: No build identifiers selected!")
-        sys.exit(3)
 
     if args.print_build_identifiers:
         for identifier in identifiers:
@@ -268,6 +265,10 @@ def main() -> None:
         cibuildwheel.macos.build(build_options)
     else:
         assert_never(platform)
+
+    if not identifiers:
+        print("ERROR: No build identifiers selected!")
+        sys.exit(3)
 
 
 def detect_obsolete_options() -> None:
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -20,6 +20,7 @@ from cibuildwheel.util import (
     DependencyConstraints,
     TestSelector,
     Unbuffered,
+    allowed_architectures_check,
     detect_ci_provider,
     resources_dir,
 )
@@ -256,6 +257,12 @@ def main() -> None:
     sys.stdout = Unbuffered(sys.stdout)  # type: ignore
 
     print_preamble(platform, build_options)
+
+    try:
+        allowed_architectures_check(platform, build_options)
+    except ValueError as err:
+        print("cibuildwheel:", *err.args, file=sys.stderr)
+        sys.exit(4)
 
     if not identifiers:
         print(f'cibuildwheel: No build identifiers selected: {build_selector}', file=sys.stderr)
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -95,6 +95,9 @@ def main() -> None:
     parser.add_argument('--print-build-identifiers',
                         action='store_true',
                         help='Print the build identifiers matched by the current invocation and exit.')
+    parser.add_argument('--allow-empty',
+                        action='store_true',
+                        help='Do not report an error code if the build does not match any wheels.')
 
     args = parser.parse_args()
 
@@ -267,8 +270,9 @@ def main() -> None:
         assert_never(platform)
 
     if not identifiers:
-        print("ERROR: No build identifiers selected!")
-        sys.exit(3)
+        print(f'cibuildwheel: No build identifiers selected: {build_selector}', file=sys.stderr)
+        if not args.allow_empty:
+            sys.exit(3)
 
 
 def detect_obsolete_options() -> None:
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -257,6 +257,11 @@ def main() -> None:
 
     print_preamble(platform, build_options)
 
+    if not identifiers:
+        print(f'cibuildwheel: No build identifiers selected: {build_selector}', file=sys.stderr)
+        if not args.allow_empty:
+            sys.exit(3)
+
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
@@ -268,11 +273,6 @@ def main() -> None:
         cibuildwheel.macos.build(build_options)
     else:
         assert_never(platform)
-
-    if not identifiers:
-        print(f'cibuildwheel: No build identifiers selected: {build_selector}', file=sys.stderr)
-        if not args.allow_empty:
-            sys.exit(3)
 
 
 def detect_obsolete_options() -> None:
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -11,16 +11,15 @@ import cibuildwheel
 import cibuildwheel.linux
 import cibuildwheel.macos
 import cibuildwheel.windows
+from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.environment import EnvironmentParseError, parse_environment
 from cibuildwheel.typing import PLATFORMS, PlatformName, assert_never
 from cibuildwheel.util import (
-    Architecture,
     BuildOptions,
     BuildSelector,
     DependencyConstraints,
     TestSelector,
     Unbuffered,
-    allowed_architectures_check,
     detect_ci_provider,
     resources_dir,
 )
@@ -259,7 +258,7 @@ def main() -> None:
     print_preamble(platform, build_options)
 
     try:
-        allowed_architectures_check(platform, build_options)
+        allowed_architectures_check(platform, build_options.architectures)
     except ValueError as err:
         print("cibuildwheel:", *err.args, file=sys.stderr)
         sys.exit(4)
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -5,7 +5,7 @@ import textwrap
 import traceback
 from configparser import ConfigParser
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, overload
+from typing import Dict, List, Optional, Set, Union, overload
 
 import cibuildwheel
 import cibuildwheel.linux
@@ -194,8 +194,14 @@ def main() -> None:
         archs_config_str = get_option_from_environment('CIBW_ARCHS', platform=platform, default='auto')
     archs = Architecture.parse_config(archs_config_str, platform=platform)
 
+    identifiers = get_build_identifiers(platform, build_selector, archs)
+    if not identifiers:
+        print("ERROR: No build identifiers selected!")
+        sys.exit(3)
+
     if args.print_build_identifiers:
-        print_build_identifiers(platform, build_selector, archs)
+        for identifier in identifiers:
+            print(identifier)
         sys.exit(0)
 
     manylinux_images: Optional[Dict[str, str]] = None
@@ -311,20 +317,22 @@ def print_preamble(platform: str, build_options: BuildOptions) -> None:
     print('\nHere we go!\n')
 
 
-def print_build_identifiers(
-    platform: str, build_selector: BuildSelector, architectures: Set[Architecture]
-) -> None:
-
-    python_configurations: List[Any] = []
+def get_build_identifiers(
+    platform: PlatformName, build_selector: BuildSelector, architectures: Set[Architecture]
+) -> List[str]:
+    python_configurations: Union[List[cibuildwheel.linux.PythonConfiguration],
+                                 List[cibuildwheel.windows.PythonConfiguration],
+                                 List[cibuildwheel.macos.PythonConfiguration]]
     if platform == 'linux':
         python_configurations = cibuildwheel.linux.get_python_configurations(build_selector, architectures)
     elif platform == 'windows':
         python_configurations = cibuildwheel.windows.get_python_configurations(build_selector, architectures)
     elif platform == 'macos':
         python_configurations = cibuildwheel.macos.get_python_configurations(build_selector)
+    else:
+        assert_never(platform)
 
-    for config in python_configurations:
-        print(config.identifier)
+    return [config.identifier for config in python_configurations]
 
 
 def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,9 +5,3 @@ requires = [
 ]
 
 build-backend = "setuptools.build_meta"
-
-
-[tool.pytest.ini_options]
-markers = [
-    "allow_empty(platform): Allow main tests on this platform(s) to be empty",
-]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,9 @@ requires = [
 ]
 
 build-backend = "setuptools.build_meta"
+
+
+[tool.pytest.ini_options]
+markers = [
+    "allow_empty(platform): Allow main tests on this platform(s) to be empty",
+]
--- a/test/utils.py
+++ b/test/utils.py
@@ -77,7 +77,7 @@ def cibuildwheel_run(project_path, package_dir='.', env=None, add_env=None, outp
 
 def expected_wheels(package_name, package_version, manylinux_versions=None,
                     macosx_deployment_target='10.9', machine_arch=None, *,
-                    filter_27=IS_WINDOWS_RUNNING_ON_TRAVIS):
+                    exclude_27=IS_WINDOWS_RUNNING_ON_TRAVIS):
     '''
     Returns a list of expected wheels from a run of cibuildwheel.
     '''
@@ -135,8 +135,9 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
         for platform_tag in platform_tags:
             wheels.append(f'{package_name}-{package_version}-{python_abi_tag}-{platform_tag}.whl')
 
-    if filter_27:
-        # Python 2.7 isn't supported on Travis.
+    # Travis on Windows does not support using the default Python 2.7 compiler,
+    # so we support skipping here.
+    if exclude_27:
         wheels = [w for w in wheels if '-cp27-' not in w and '-pp2' not in w]
 
     return wheels
--- a/test/utils.py
+++ b/test/utils.py
@@ -76,7 +76,8 @@ def cibuildwheel_run(project_path, package_dir='.', env=None, add_env=None, outp
 
 
 def expected_wheels(package_name, package_version, manylinux_versions=None,
-                    macosx_deployment_target='10.9', machine_arch=None):
+                    macosx_deployment_target='10.9', machine_arch=None, *,
+                    filter_27=IS_WINDOWS_RUNNING_ON_TRAVIS):
     '''
     Returns a list of expected wheels from a run of cibuildwheel.
     '''
@@ -134,7 +135,7 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
         for platform_tag in platform_tags:
             wheels.append(f'{package_name}-{package_version}-{python_abi_tag}-{platform_tag}.whl')
 
-    if IS_WINDOWS_RUNNING_ON_TRAVIS:
+    if filter_27:
         # Python 2.7 isn't supported on Travis.
         wheels = [w for w in wheels if '-cp27-' not in w and '-pp2' not in w]
 
--- a/unit_test/main_tests/main_options_test.py
+++ b/unit_test/main_tests/main_options_test.py
@@ -11,6 +11,7 @@ from cibuildwheel.util import BuildSelector
 # CIBW_PLATFORM is tested in main_platform_test.py
 
 
+@pytest.mark.allow_empty('windows')
 def test_output_dir(platform, intercepted_build_args, monkeypatch):
     OUTPUT_DIR = Path('some_output_dir')
 
@@ -21,12 +22,14 @@ def test_output_dir(platform, intercepted_build_args, monkeypatch):
     assert intercepted_build_args.args[0].output_dir == OUTPUT_DIR
 
 
+@pytest.mark.allow_empty('windows')
 def test_output_dir_default(platform, intercepted_build_args, monkeypatch):
     main()
 
     assert intercepted_build_args.args[0].output_dir == Path('wheelhouse')
 
 
+@pytest.mark.allow_empty('windows')
 @pytest.mark.parametrize('also_set_environment', [False, True])
 def test_output_dir_argument(also_set_environment, platform, intercepted_build_args, monkeypatch):
     OUTPUT_DIR = Path('some_output_dir')
@@ -40,6 +43,7 @@ def test_output_dir_argument(also_set_environment, platform, intercepted_build_a
     assert intercepted_build_args.args[0].output_dir == OUTPUT_DIR
 
 
+@pytest.mark.allow_empty
 def test_build_selector(platform, intercepted_build_args, monkeypatch):
     BUILD = 'some build* *-selector'
     SKIP = 'some skip* *-selector'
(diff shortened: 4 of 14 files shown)


YannickJadoul:
No, the conclusion back then was that it wasn't very desirable, I believe: #179

henryiii:
To me (on quickly looking over the issue, such as this comment https://github.com/joerick/cibuildwheel/pull/179#issuecomment-552614268) seems to be not that it was a bad idea, just that the implementation was not ideal.

If you have a matrix and the macOS job doesn't match anything, most of the time, that's a user error, and when it's not, the macOS job should be removed or the step should be if'd out, since it's not doing anything and confusing readers. Or a user could do `cibuildwheel || echo "Nothing built"` to explicitly "not care" that nothing was built.

For example, currently, someone seeing the new arch support might write this:

```yaml
      - name: Build wheels
        run: pipx run cibuildwheel==1.8.0
        env:
          CIBW_BUILD: "*_aarch64"
```

But since they didn't turn on the needed arch, currently this just silently "succeeds". They might not know they don't have aarch64 wheels until after a release! I think it's critical that CI jobs "fail" if something is wrong, so they can be fixed; no one reads CI logs unless there's a failure.

Another example is if we drop Python 3.5, this should immediately fail loudly:

```yaml
      - uses: joerick/cibuildwheel@v2.0.0
        env:
          CIBW_BUILD: "cp35-*"
```

So the user who just updated their CI configs can remote this step instead of thinking (and indicating to readers) that they still are making 3.5 wheels. Or if they want 3.5 wheels, they will know not to merge the PR with the updated cibuildwheel version.

YannickJadoul:
No I think it's more fundamental: https://github.com/joerick/cibuildwheel/pull/179#issuecomment-549528543

> If you have a matrix and the macOS job doesn't match anything, most of the time, that's a user error, and when it's not, the macOS job should be removed or the step should be if'd out, since it's not doing anything and confusing readers. Or a user could do `cibuildwheel || echo "Nothing built"` to explicitly "not care" that nothing was built.

That's the design/workaround @Czaki suggested (https://github.com/joerick/cibuildwheel/pull/179#issuecomment-549568383) and what @joerick didn't like.

> Running cibuildwheel and getting nothing is clearly wrong?

I don't know. I felt reasonably neutral in the discussion, and a quick scan tells me there were some good arguments for both.

> ```yaml
>       - name: Build wheels
>         run: pipx run cibuildwheel==1.8.0
>         env:
>           CIBW_BUILD: "*_aarch64"
> ```

This could still silently fail if there's `CIBW_BUILD: "*_x86_64 *_aarch64"`, btw, while there's still a mistake. So this might give a false sense of security?

Also, on an implementation note: the fact that you need to hack around the tests with `pytest.mark.allow_empty` might suggest that this wasn't really meant to error? Again, I think I could go either way, but this was definitely discussed before.

henryiii:
The allow empty hack is there because these are unit tests. We are setting unrealistic things and trying them out.

It also exposed a bug in our tests. We are faking the platform, but missed faking the arch, so a bunch of them were empty that shouldn't have been. So I'd say that's another argument in favor.

We can't fix all build selectors. But we can error on cases where the configuration is so bad nothing builds.

Unlike the earlier discussion and PR, this is not about build/skip at all. It is about running with no wheels to build. It could be due to build/skip, or arch, or anything in the future, like a Requires.Python. CI build wheel should not report "success" if it does nothing at all.

henryiii:
We could also check that every build / skip selector term selects at least only thing, but that's a different, orthogonal check and is separate from this. I think if you write cibuildwheel in a config and it builds nothing, that's confusing to readers, you should make your intention clear with the `||` trick, for example. You wouldn't expect `pip install x` to silently do nothing if x can't be installed on this platform? Yes, it might save some typing, but if you misspell x or thought it was supported, that's more important.

henryiii:
If there's a use case for running an empty build, how about an `--allow-empty` flag? Then it's also explicit and cross platform. I don't see a sting need for it, but if someone else does?

YannickJadoul:
> Unlike the earlier discussion and PR, this is not about build/skip at all. It is about running with no wheels to build.

I kind of disagree on this being different, since @Czaki's original PR was doing the same: checking if the configuration that's running is not going to build anything.

But I do agree that the tests that currently weren't/aren't running are quite worrying and an argument. And recent changes have made the situation slightly more complex, making it possible that the situation has changed, so revisiting the idea might be a good plan.
But I'm mainly saying that @joerick was against the previous instantiation of this idea, and you'll have to convince him ;-) To me, it would probably make sense (I had already approved #179, long long time ago ;-) ), but I never felt the strong need to urgently have this, either :-)

> If there's a use case for running an empty build, how about an `--allow-empty` flag? Then it's also explicit and cross platform. I don't see a strong need for it, but if someone else does? (Though I would steal it for the tests ;) )

Not a big fan of this, for the record. Then `||` should be better, no? (though not cross-platform, but I don't really think it's worth adding an extra flag over; unless someone brings up a good use case, indeed)

YannickJadoul:
Actually, the more I think about it, the more it does indeed make sense to me (the idea, that is; this is not a review of the code; I'll do that when I know something about the chances of @joerick being convinced ;-) ).

henryiii:
Haha, I've been kicking travis since I thought it was broken as usual. Then I realized, this is actually because:

```
test_cpp17_py27_modern_msvc_workaround
```

We only have an explicit skip for Appveyor; we don't support the workaround on travis (Note that building Windows Python 2.7 wheels on Travis is unsupported.), but this test used to "pass" anyway because it's making an empty set and expecting and empty set. 🤦 Another +1 for adding this check!!!

We probably should relax this to allow this workaround on Travis, honestly, since it should support it, it just doesn't support non-workaround Windows 2.7 wheels.

joerick:
Ah! It comes up again! Maybe my instinct is just wrong about this...

I struggle to disagree with what I wrote last time, from #179:

> Also, there might be reasons a user would like no action. E.g., a user that wants to temporarily skip wheels on Windows, for example, might set `CIBW_SKIP=*-win*` in their travis config. Or maybe somebody is having a tough day fighting the CI, they might set `CIBW_SKIP=*`. In both cases, the option is fairly clear that the user would like no builds, and cibuildwheel can happily succeed at doing nothing.

Maybe, I'm thinking about this wrong? Here I'm thinking about cibuildwheel as an API, and in an API, you ask for something to be done 0 times, and it will not raise an error. Because it can do what you ask. In an API, an error should happen when you ask the computer to do something and it cannot do it. But here, it can. It can do nothing.

Buuut I suppose the other way to look at cibuildwheel is as a user interface. You can't open 0 items. You can't add 0 songs to a playlist. Most UIs don't _let_ you do these things. The button is disabled, the menu item greyed out.

I guess command line apps fall in the middle of these things. We're both UIs and APIs. So both arguments work.

```shell
$ cp *.txt /tmp
zsh: no matches found: *.txt
```

> I think if you write cibuildwheel in a config and it builds nothing, that's confusing to readers, you should make your intention clear with the || trick, for example. You wouldn't expect pip install x to silently do nothing if x can't be installed on this platform?

Yeah. I can't really argue with that either... and added to the archs/requires-python thing, I think I'm coming round to the idea. I don't have time to review now, but I think in principle I'm okay with it :)

YannickJadoul:
OK, I'll give a quick review tomorrow (taking into account and comparing to @Czaki's #179).

One way to go would be to include this in a release, and wait for complaints to come in? If they suddenly do, a new patch version is released quickly enough? At the very least, this would be safe: users would only risk getting unwanted errors, but no silently ignored unwanted results :-)

henryiii:
Also keep in mind **ci**buildwheel is designed for CI. In CI, no one is monitoring the jobs most of the time, they are just looking for green checks. Having cibuildwheel happily claim it did its job when there's a mistake in BUILD, a bad SKIP config, etc. We just had a bug in the tests because running on travis windows silently skips Python 2.7. Someone making a special Windows 2.7 workaround build with `CIBW_BUILD=?p27-*` on Travis would just get a happy cibuildwheel run, no error (or even warning, currently), and no wheel. When we drop cp27, the jobs that run `cp27*` are still going to "succeed", just no wheels will be produced.

Azure made a huge mistake in not activating `set -e` on multiline scripts; they were unable to fix it after shipping, so they fixed it in GitHub Actions when it came out. Passing errors in CI is very problematic! If you do like this idea, there could be `--allow-empty` flag, I think it's better to comment out cibuildwheel or similar, but it would be better than silently succeeding on bad configurations by default.

> Also, there might be reasons a user would like no action

Then skip the job (or comment out the cibuildwheel line if you really have to run everything else). Starting up the CI, running the job, getting a green check, but _not_ building anything is very likely to cause big problems later. You have to remember to un-skip it. And you might have the parts of the pipeline that depend on the wheels produced, or have expensive setup; it's better to turn off the job, I think.

> One way to go would be...

I rather expect this might catch some mistakes. :) I'm now suddenly apprehensive for my own projects now. I actually always assumed it wouldn't let me build nothing and still succeed, so if it passes, I don't check to see if nothing was produced.

cibuildwheel/__main__.py
```diff
@@ -191,8 +191,11 @@ def main() -> None:
         archs_config_str = get_option_from_environment('CIBW_ARCHS', platform=platform, default='auto')
     archs = Architecture.parse_config(archs_config_str, platform=platform)
 
+    identifiers = get_build_identifiers(platform, build_selector, archs)
+
     if args.print_build_identifiers:
-        print_build_identifiers(platform, build_selector, archs)
+        for identifier in identifiers:
+            print(identifier)
         sys.exit(0)
```
henryiii, line 199: I chose to always return 0 if you print build identifiers. Technically, you could return 3 here too and use it in an api to ask if something works - but I don't think anyone is asking for that, and normal usage for this is by hand AFAIK.

test/utils.py
```diff
@@ -134,7 +134,8 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
         for platform_tag in platform_tags:
             wheels.append(f'{package_name}-{package_version}-{python_abi_tag}-{platform_tag}.whl')
 
-    if IS_WINDOWS_RUNNING_ON_TRAVIS:
+    custom_compiler = os.environ.get('DISTUTILS_USE_SDK') and os.environ.get('MSSdk')
+    if IS_WINDOWS_RUNNING_ON_TRAVIS and not custom_compiler:
```
henryiii, line 138: The Windows 2.7 newer compiler workaround now works on Travis. There's an issue somewhere where it was suggested that we allow this check to be loosened.

cibuildwheel/__main__.py
```diff
@@ -259,6 +262,10 @@ def main() -> None:
     else:
         assert_never(platform)
 
+    if not identifiers:
+        print("ERROR: No build identifiers selected!")
```
YannickJadoul, line 266: The current kind of errors are something like:

```py
print(f'cibuildwheel: Unsupported platform: {platform}', file=sys.stderr)
```

So I suggest following this convention, and going with:

```py
print(f'cibuildwheel: No build identifiers selected: {build_selector}', file=sys.stderr)
```

pyproject.toml
```diff
@@ -5,3 +5,9 @@ requires = [
 ]
 
 build-backend = "setuptools.build_meta"
+
+
+[tool.pytest.ini_options]
+markers = [
+    "allow_empty(platform): Allow main tests on this platform(s) to be empty",
```
YannickJadoul, line 12: Still not a fan of the marker, actually. Especially because we don't want to document it here. It's an internal thing, and not part of the interface, so nothing that should need to be added here.

unit_test/main_tests/main_platform_test.py
```diff
@@ -112,6 +114,7 @@ def test_archs_platform_specific(platform, intercepted_build_args, monkeypatch):
         assert build_options.architectures == {Architecture.x86_64}
 
 
+@pytest.mark.allow_empty('windows')
```
YannickJadoul, line 117: Why don't we just skip Windows, then? Or fix the `machine`?

cibuildwheel/__main__.py
```diff
@@ -259,6 +262,10 @@ def main() -> None:
     else:
         assert_never(platform)
 
+    if not identifiers:
```
YannickJadoul, line 265: 2 things:

1. So, the main difference with #179 is that @Czaki put the detection in the `build` function, and raised an exception there. I think I prefer this approach (checking in `__main__`), since we don't really care about doing this twice?
   -  @Czaki's approach does fix the whole "marker for empty build" issue, since `build` gets intercepted.
2. Why do this after you print the preamble and after calling `build`? If we're catching errors in `__main__.py`, they should probable be thrown before we give the impression that `cibuildwheel` started running successfully?

henryiii: > Why do this after you print the preamble

Because that information could potentially be useful in helping debug what went wrong. The preamble is also not build-specific, it doesn't get repeated N times for N builds.

> since we don't really care about doing this twice

You mean three times? Once per OS file?

unit_test/main_tests/main_platform_test.py
henryiii: I originally had a fixture for this. Then I realized there were a lot of these, so I added the marker. But then I fixed most of them by making the arch realistic for the testing, so it probably could be a fixture again.

henryiii:
I'm not against the idea of a `--allow-empty` flag. The problem with `cibuildwheel || true` is that it also masks build errors, while `--allow-empty` would just be a safe way to opt back into the old behavior. And it's handy for a few of our unit tests, too. ;)

cibuildwheel/__main__.py
henryiii: Though it's not as important now that we have the build selector. However, basic info like what it's running on, etc. still seem quite useful.

unit_test/main_tests/main_platform_test.py
henryiii: I've reworked it into a fixture, and fixed several of the tests to no longer need it (actually tests more thoroughly now :) ), and I've added an `--allow-empty` option that lets this pass in a shell-independent way. I think it will help the transition, especially if it turns out someone does like using `SKIP=*` or similar as an API. If we had never let empty builds pass, I don't think it would be needed.

henryiii: I would be okay to leave it an undocumented option and just see if users end up needed it. If enough do, we could document it.

cibuildwheel/__main__.py
YannickJadoul: Hmmm, myeah, maybe. We don't do that for the other errors, though; but you could argue that the other errors are a bit "simpler".

> > since we don't really care about doing this twice
> 
> You mean three times? Once per OS file?

That wasn't clear. But I meant determining the selected build identifiers twice. It's now done to check they're not empty, then again in `{linux,macos,windows}.build()` to actually loop over the different builds.

So I meant we don't really care about the performance of doing this twice, in this case (not like it's going to be measurable), and there's hardly any code duplication.

unit_test/main_tests/main_options_test.py
```diff
@@ -220,7 +229,7 @@ def test_build_verbosity(build_verbosity, platform_specific, platform, intercept
     ('*-macosx_10_9_x86_64', ['*-macosx_x86_64']),
     ('cp37-macosx_10_9_x86_64', ['cp37-macosx_x86_64']),
 ])
-def test_build_selector_migrations(intercepted_build_args, monkeypatch, option_name, option_value, build_selector_patterns):
+def test_build_selector_migrations(intercepted_build_args, monkeypatch, option_name, option_value, build_selector_patterns, allow_empty):
```
YannickJadoul, line 232: Thanks, @henryiii! Not entirely sure why, but I somehow like this better :-)

cibuildwheel/__main__.py
YannickJadoul: Agreed. I think this wouldn't be an error; you're just asking for a list of what would be built, right?

YannickJadoul:
> I'm not against the idea of a `--allow-empty` flag. The problem with `cibuildwheel || true` is that it also masks build errors, while `--allow-empty` would just be a safe way to opt back into the old behavior. And it's handy for a few of our unit tests, too. ;)

Oh, that's true as well, ofc. I don't really mind either way. I don't really think it would be used that often, but it can't really hurt either? (It's a pretty "shallow" flag, anyway, as it's already handled in `__main__.py`.)

YannickJadoul (approved):
As far as I'm concerned, this makes sense. And the amount of code that was necessary is also reasonably little, which provides evidence it's not a crazy idea/design :-)

cibuildwheel/__main__.py
henryiii: I'm happy to print it out before if that's better; including the build specifier helps quite a bit.

cibuildwheel/__main__.py
joerick: Yes `0` is right, because it successfully printed.

cibuildwheel/__main__.py
joerick: I think it's good to print this after the preamble, printing the options (and configuration warnings) might be useful information. But after calling `platform-specific.build` is strange to me... as this check is designed to flag a misconfiguration. Also, just a few lines up we create the output_dir ready to receive wheels, we should be doing this check and failing before that I think.

test/utils.py
```diff
@@ -134,7 +135,7 @@ def expected_wheels(package_name, package_version, manylinux_versions=None,
         for platform_tag in platform_tags:
             wheels.append(f'{package_name}-{package_version}-{python_abi_tag}-{platform_tag}.whl')
 
-    if IS_WINDOWS_RUNNING_ON_TRAVIS:
+    if filter_27:
         # Python 2.7 isn't supported on Travis.
```
joerick, line 139: This comment is obsolete, or at least in the wrong place.
```suggestion
```

unit_test/main_tests/main_platform_test.py
```diff
@@ -68,19 +68,24 @@ def test_platform_environment(platform, intercepted_build_args, monkeypatch):
...
-    monkeypatch.setattr(platform_module, 'machine', lambda: 'x86_64')
+    monkeypatch.setattr(platform_module, 'machine', lambda: 'AMD64' if platform == 'windows' else 'x86_64')
 
     main()
     build_options = intercepted_build_args.args[0]
 
     if platform == 'linux':
         assert build_options.architectures == {Architecture.x86_64, Architecture.i686}
+    elif platform == 'windows':
+        assert build_options.architectures == {Architecture.AMD64, Architecture.x86}
```
joerick, line 79: Ooh, that's a good catch :)

unit_test/main_tests/main_platform_test.py
```diff
@@ -68,19 +68,24 @@ def test_platform_environment(platform, intercepted_build_args, monkeypatch):
...
     else:
         assert build_options.architectures == {Architecture.x86_64}
 
 
 @pytest.mark.parametrize('use_env_var', [False, True])
 def test_archs_argument(platform, intercepted_build_args, monkeypatch, use_env_var):
+    if platform == 'windows':
+        pytest.skip('Will have empty build selectors on Windows')
```
joerick, line 87: I'm confused. I would have expected this to be `if platform != 'linux':` Why does this test pass on macOS? There shouldn't be any selectors active when `archs` is `ppc64le` on macOS?

unit_test/main_tests/main_platform_test.py
```diff
@@ -113,7 +118,7 @@ def test_archs_platform_specific(platform, intercepted_build_args, monkeypatch):
 
 
 def test_archs_platform_native(platform, intercepted_build_args, monkeypatch):
-    monkeypatch.setattr(platform_module, 'machine', lambda: 'x86_64')
+    monkeypatch.setattr(platform_module, 'machine', lambda: 'AMD64' if platform == 'windows' else 'x86_64')
```
joerick, line 121: I think this monkeypatch is already done in the `platform` fixture?

unit_test/main_tests/main_platform_test.py
```diff
@@ -68,19 +68,24 @@ def test_platform_environment(platform, intercepted_build_args, monkeypatch):
 
 
 def test_archs_default(platform, intercepted_build_args, monkeypatch):
-    monkeypatch.setattr(platform_module, 'machine', lambda: 'x86_64')
+    monkeypatch.setattr(platform_module, 'machine', lambda: 'AMD64' if platform == 'windows' else 'x86_64')
```
joerick, line 71: I think this is already done in the monkeypatch fixture?

test/utils.py
```diff
@@ -76,7 +76,8 @@ def cibuildwheel_run(project_path, package_dir='.', env=None, add_env=None, outp
 
 
 def expected_wheels(package_name, package_version, manylinux_versions=None,
-                    macosx_deployment_target='10.9', machine_arch=None):
+                    macosx_deployment_target='10.9', machine_arch=None, *,
+                    filter_27=IS_WINDOWS_RUNNING_ON_TRAVIS):
```
joerick, line 80: I don't love this signature growing but I can't see a way around it either! Nitpick, but `filter_27` might be a little ambiguous, maybe `exclude_27` would be clearer?

joerick (commented):
Looks good generally, a few minor things, and I think we should move the location of the check in the `main` function.

Once those are fixed, I'm happy to merge this and get it into 1.8 - it makes sense to land it at the same time as ARCHS.

unit_test/main_tests/main_platform_test.py
henryiii: macOS does not take the arch into account at all, it's not even passed. So it currently completly ignores the arch when making build selectors (until the Universal2 patch, anyway).

See:
```python
  elif platform == 'macos':
      python_configurations = cibuildwheel.macos.get_python_configurations(build_selector)
```

henryiii: Should the arch check protect against any unsupported arch, though? Why is this passing that?

unit_test/main_tests/main_platform_test.py
henryiii: Not when I first wrote it. 😳

unit_test/main_tests/main_platform_test.py
henryiii: Ah, that happens in the build function. So if we move this before the build attempt, then the empty selector check happens before the architecture check, and you won't get the nicer message for that check. I think, since they are all the same, we should pull the arch check out and place it with the selector check (it gets repeated 3x anyway).

henryiii:
Okay, I've moved architecture out into a separate file, and the check now happens inline rather than in each build function, so it can precede the empty selector check. It's also nice to have all possible exits in the same place and have a clearer idea of what's being checked directly in the main function, reading top to bottom. :) (CC @YannickJadoul).

Also test for the SystemExit now, rather than just skipping, even on Windows.

joerick (approved):
Looks good to me!

test/utils.py
YannickJadoul: Is there a reason to go for kwargs-only arguments? This might be the only place in the whole of cibuildwheel, currently?

henryiii: No, there are some in other places (build selectors, for example). They force you to remain readable. I'd like to move the star farther to the left; though `"10.9" or `"x86_64"` might be guessable on it's own, it's much clearer with the keyword. And a random True at the end would be a disaster!

henryiii: Ehh, maybe it's 50/50. I didn't write the ones in logger or c; the main reason not to have keyword only arguments is for compatibility with Python 2. They should always be used if it's not obvious from the context or type of argument.

henryiii:
The emulated wheel check is slow...........

test/utils.py
YannickJadoul: Oh, there is more. Eh, not worth changing, then. :-)

YannickJadoul: (It's just ... internal API; I don't care too much about that; it's mostly the calling code that should be careful, then. But it's not _wrong_ either, ofc)

henryiii:
Why did this suddenly take so much longer, even timing out on the emulate build?

henryiii:
Must be an issue with GHA/Azure, the Travis runs were exactly the same length as before. All the GHA/Azure runs took much longer than normal (compared to a recent PR).

joerick:
Well, the emulation test passed. No idea what happened there. I can't see the old log and the results of the previous check because of the force-push (you guys with your force pushing! but i digress). I think the other tests had passed last time around?

henryiii:
Much better, 16 mins instead of 3 hours. :)

joerick:
Pretty sure the previous failure was just a glitch, and want to get on with this release!

henryiii:
The force push put it on latest master to pick up the readme update, and changed the name from `err` to `exit` in one spot for consistency. Nothing else changed. All the old logs are in the "actions" tab on the top. But I'm pretty sure this was a weird glitch in scheduling, several jobs were longer than they should have been.

PR #550, opened 2021-01-21
ci: add bot to remind users about docs builds if docs have changed

henryiii:
@joerick, you might be interested. Have no idea if it actually works, based on the example here: https://github.com/actions/github-script#comment-on-an-issue .

### DIFF
--- a/.github/workflows/prbot.yml
+++ b/.github/workflows/prbot.yml
@@ -1,0 +1,22 @@
+name: Reminder bot
+on:
+  pull_request:
+    types: [opened, edited]
+    paths:
+    - 'docs/*.md'
+    - 'README.md'
+
+jobs:
+  docs-reminder:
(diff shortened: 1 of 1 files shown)


henryiii:
Honestly, just noticed the readthedocs line in the checks, maybe that's enough if there's a way to get to it from there.

henryiii:
Yes. clicking details jumps to the build. Fine to just do that, I just didn't notice it in the list of checks. ~~How did you set that up?~~ Looks like it's here: https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

Hmmm, have to try that other places. :)

henryiii:
Since there's a spot in the checks for it, and I've added it to my repos so I'm getting used to it, I think that's fine, and this is not needed.

PR #553, opened 2021-01-21
feat: auto32/auto64

### LINKED ISSUE #551: Idea: Arch 32/64 shortcut
When looking at the build matrix here:

https://github.com/deepcharles/ruptures/blob/404957563c0391f073200899b9b041caa9277cc6/.github/workflows/upload-to-pypi.yml#L40-L80

It seems we could have shortcut values for `ARCHS`, "32" and "64", that would select 32 bit and 64 bit native architectures. So on 64 bit Windows or Linux, you could select 32 or 64. On macOS, you could only select 64, 32 would be empty (so an error after #545 unless `--allow-empty` was passed) or we could just make it an error. Same thing for 64 on 32bit OS's. On alternate archs, these remain the "native" on that arch 32/64 values.

I believe @mayeut wanted something like this too? Don't have the issue reference handy for it.

This would take a lot of pressure off the build selector. ARCHS, like BUILD/SKIP, already helps define the build process.

<details><summary>Old version</summary>

```yaml
matrix:
  os: [ubuntu-latest, windows-latest, macos-latest]
  bitness: [32, 64]
  python: [36, 37, 38, 39]
  include:
    # Run 32 and 64 bits version in parallel for Linux and Windows
    - os: windows-latest
      bitness: 64
      platform_id: win_amd64
    - os: windows-latest
      bitness: 32
      platform_id: win32
    - os: ubuntu-latest
      bitness: 64
      platform_id: manylinux_x86_64
    - os: ubuntu-latest
      bitness: 32
      platform_id: manylinux_i686
    - os: macos-latest
      bitness: 64
      platform_id: macosx_x86_64
  exclude:
    - os: macos-latest
      bitness: 32
    - os: windows-latest
      python: 39
steps:
- uses: actions/checkout@v2
  with:
    fetch-depth: 0
  name: Install Python
  with:
    python-version: '3.8'
- name: Install cibuildwheel
  run: |
    python -m pip install --upgrade pip
    python -m pip install cibuildwheel
- name: Build wheels
  env:
    CIBW_BUILD: cp${{ matrix.python }}-${{ matrix.platform_id }}
```

</details>

<details><summary>New version</summary>

```yaml
matrix:
  os: [ubuntu-latest, windows-latest, macos-latest]
  bitness: [32, 64]
  python: [36, 37, 38, 39]
  exclude:
    - os: macos-latest
      bitness: 32
    - os: windows-latest
      python: 39
steps:
- uses: actions/checkout@v2
  with:
    fetch-depth: 0
- name: Build wheels
  env:
    CIBW_BUILD: cp${{ matrix.python }}-*
  run: pipx run cibuildwheel==1.80 --output-dir wheelhouse --archs ${{ matrix.bitness }}
```

</details>

Thoughts? Hopefully @deepcharles doesn't mind me using his yaml as an example. :) Also used pipx for the "new" version, because why not, it's preinstalled. :)

henryiii:
Add `auto32`/`auto64`.

Closes #551.

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -4,7 +4,7 @@ import re
 from enum import Enum
 from typing import Set
 
-from .typing import PlatformName, assert_never
+from .typing import Literal, PlatformName, assert_never
 
 PRETTY_NAMES = {'linux': 'Linux', 'macos': 'macOS', 'windows': 'Windows'}
 
@@ -44,6 +44,10 @@ class Architecture(Enum):
                 result.add(Architecture(platform_module.machine()))
             elif arch_str == 'all':
                 result |= Architecture.all_archs(platform=platform)
+            elif arch_str == 'auto64':
+                result |= Architecture.bitiness_archs(platform=platform, bitiness="64")
+            elif arch_str == 'auto32':
+                result |= Architecture.bitiness_archs(platform=platform, bitiness="32")
             else:
                 result.add(Architecture(arch_str))
         return result
@@ -76,6 +80,26 @@ class Architecture(Enum):
             return {Architecture.x86, Architecture.AMD64}
         else:
             assert_never(platform)
+
+    @staticmethod
+    def bitiness_archs(platform: PlatformName, bitiness: Literal["64", "32"]) -> 'Set[Architecture]':
+        native_architecture = Architecture(platform_module.machine())
+
+        if native_architecture in {Architecture.x86_64, Architecture.aarch64, Architecture.ppc64le, Architecture.s390x, Architecture.AMD64}:
+            if bitiness == "64":
+                return {native_architecture}
+            else:
+                if native_architecture == Architecture.x86_64 and platform != "macos":
+                    return {Architecture.i686}
+                elif native_architecture == Architecture.AMD64:
+                    return {Architecture.x86}
+                else:
+                    return set()
+        elif native_architecture in {Architecture.i686, Architecture.x86}:
+            return {native_architecture} if bitiness == "32" else set()
+        else:
+            # assert_never doesn't work here, oddly, maybe due to set checking above
+            raise RuntimeError("Cannot be reached")
 
 
 def allowed_architectures_check(
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -82,24 +82,16 @@ class Architecture(Enum):
             assert_never(platform)
 
     @staticmethod
-    def bitiness_archs(platform: PlatformName, bitiness: Literal["64", "32"]) -> 'Set[Architecture]':
-        native_architecture = Architecture(platform_module.machine())
+    def bitiness_archs(platform: PlatformName, bitiness: Literal['64', '32']) -> 'Set[Architecture]':
+        archs_32 = {Architecture.i686, Architecture.x86}
+        auto_archs = Architecture.auto_archs(platform)
 
-        if native_architecture in {Architecture.x86_64, Architecture.aarch64, Architecture.ppc64le, Architecture.s390x, Architecture.AMD64}:
-            if bitiness == "64":
-                return {native_architecture}
-            else:
-                if native_architecture == Architecture.x86_64 and platform != "macos":
-                    return {Architecture.i686}
-                elif native_architecture == Architecture.AMD64:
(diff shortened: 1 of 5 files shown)


henryiii:
I've rebased on top of #545 since it looks like that one will go in. When looking at this PR, just look at the last commit. Looks like #545 is based on a slightly older master than I intended, so the `CIBW_ARCHS` README change isn't picked up, so I'll re-rebase when #545 is in.  @joerick I've updated the docs quite a bit on `CIBW_ARCHS`, to be more consistent with the rest of the options, especially now that we have some useful on-all-platforms arch settings. It would be great if we could get this in as the last PR before 1.8.0.

I've also normalized a names in the tests for better consistency; could have gone into #545 but didn't realize it at the time I pushed that one.

henryiii:
@joerick and/or @YannickJadoul Feel free to look at 7ccca9e by itself to speed up the process, as soon as #545 goes in that's what I'll rebase on master.

joerick:
#545 is the last PR for 1.8.0, otherwise we'll never get it out! I think the docs currently on master are fine for documenting the emulation use-case, so we don't need to rush to get this in.

cibuildwheel/architecture.py
```diff
@@ -77,6 +81,26 @@ def all_archs(platform: PlatformName) -> 'Set[Architecture]':
...
+                    return {Architecture.i686}
+                elif native_architecture == Architecture.AMD64:
+                    return {Architecture.x86}
+                else:
+                    return set()
+        elif native_architecture in {Architecture.i686, Architecture.x86}:
+            return {native_architecture} if bitiness == "32" else set()
+        else:
+            # assert_never doesn't work here, oddly, maybe due to set checking above
+            raise RuntimeError("Cannot be reached")
```
joerick, line 102: This looks a little overcomplicated to me... maybe this would work?

```python
    archs_32 = {Architecture.i686, Architecture.x86}
    auto_archs = Architecture.auto_archs(platform)

    if bitness == '64':
        return auto_archs - archs_32
    elif bitness == '32':
        return auto_archs.intersection(archs_32)
    else:
        assert_never(bitness)
```

docs/options.md
```diff
@@ -191,6 +193,9 @@ If not listed above, `auto` is the same as `native`.
 [setup-qemu-action]: https://github.com/docker/setup-qemu-action
 [binfmt]: https://hub.docker.com/r/tonistiigi/binfmt
 
+Platform-specific variants also available:<br/>
+ `CIBW_ARCHS_MACOS` | `CIBW_ARCHS_WINDOWS` | `CIBW_ARCHS_LINUX`
+
```
joerick, line 198: Good catch, I forgot that 👍

docs/options.md
```diff
@@ -172,6 +172,8 @@ Options:
 - macOS: `x86_64` `arm64` `universal2`
 - Windows: `AMD64` `x86`
 - `auto`: The default archs for your machine - see the table below.
+    - `auto64`: Just the 64-bit auto archs (will result in an empty set on a 32-bit runner)
+    - `auto32`: Just the 32-bit auto archs (will result in an empty set on macOS)
```
joerick, line 176: Maybe we could remove the parens here, but add auto64 and auto32 columns to the table below?

cibuildwheel/architecture.py
henryiii: Didn't think of that, very clean. :)

docs/options.md
henryiii: Actually, this was the result of a bad rebase - there are two of these now! 😳

henryiii: I think the "normal" place is here, right before examples, rather than after. I've kept this one.

cibuildwheel/architecture.py
```diff
@@ -77,6 +81,18 @@ def all_archs(platform: PlatformName) -> 'Set[Architecture]':
         else:
             assert_never(platform)
 
+    @staticmethod
+    def bitiness_archs(platform: PlatformName, bitiness: Literal['64', '32']) -> 'Set[Architecture]':
```
joerick, line 85: Oops I thought I put a comment about this earlier, must have deleted it accidentally. I think the word is 'bitness', 'bitiness' :)

joerick (approved):
Looks good!

PR #555, opened 2021-01-23
ci: bump pre-commit with other deps

henryiii:
Let's bump pre-commit repos too when we bump other deps.

Should be testable tomorrow with mypy 0.800.

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -16,6 +16,8 @@ jobs:
       with:
         python-version: 3.9
         architecture: x64
+    - name: Pre-commit update
+      run: pipx run pre-commit autoupdate
     - name: Install dependencies
       run: python -m pip install ".[dev]"
     - name: "Run update: dependencies"


joerick:
Sure, why not. I can see an argument that this should be a separate workflow, to separate runtime and dev dependencies, but I have no idea how often bumping pre-commits will cause errors. I suppose MyPy updates are the most likely to find things to complain about? But I'm happy to try this too.

henryiii:
There is an alternative: you could set this up: https://pre-commit.ci if you'd like. It updates every Monday via PR, and also is a bit faster than the pre-commit GHA check, and will auto-fix if a modifying check (endines, black, etc) make a change. Only downside I've faced so far with it is that doesn't support docker steps (it's already in a container), but we don't have any of those.

joerick:
I've heard good things about https://pre-commit.ci actually. Maybe we should give it a shot? Do you recommend it @henryiii ?

henryiii:
I've liked it so far. I can't use it with clang format due the docker issue, but on pure Python repos I've seen it help out (usually I run pre-commit a lot or have it installed as a hook, so don't have it commit on my PRs much). And I like the auto update PR.

joerick:
Cool, I enabled it. It's also doing the same job as the Github Actions runner, so I suppose we can remove that one?

henryiii:
Yes, we can. Frees up a parallel run. :)

joerick:
I guess we can close this, now that pre-commit.ci is enabled?

henryiii:
Yep

PR #557, opened 2021-01-24
ci: bump get-pip and use python -m pip

henryiii:
Just testing.

TODO:
- keep isort from changing get-pip.

Bumping get-pip is probably not necessary. Didn't realize it was a 20K line change.


henryiii:
Looks like the problem wasn't pip, the separates output confused me; the real problem was a permissions error again.

PR #558, opened 2021-01-25
fix: try to avoid error on CircleCI by avoiding explicit an use of /tmp directly

henryiii:
Trying to see if this helps CircleCI.

Ideally we should refactor temporary directory handling, using the high level interface (TemporaryDirectory) and avoiding direclty using `/tmp` too much.﻿

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -82,8 +82,10 @@ def install_cpython(version: str, url: str) -> Path:
             tmp_pkg = Path(tmp_name) / 'Python.pkg'
             # download the pkg
             download(url, tmp_pkg)
+            call(['ls', '-l', tmp_pkg])
             # install
             call(['sudo', 'installer', '-pkg', tmp_pkg, '-target', '/'])
+            raise RuntimeError("quiting")
             # patch open ssl
             if version == '3.5':
                 tmp_patch = Path(tmp_name) / 'python-patch.tar.gz'
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -78,15 +78,18 @@ def install_cpython(version: str, url: str) -> Path:
     installation_bin_path = Path(f'/Library/Frameworks/Python.framework/Versions/{version}/bin')
 
     if python_package_identifier not in installed_system_packages:
-        # download the pkg
-        download(url, Path('/tmp/Python.pkg'))
-        # install
-        call(['sudo', 'installer', '-pkg', '/tmp/Python.pkg', '-target', '/'])
-        # patch open ssl
-        if version == '3.5':
-            open_ssl_patch_url = f'https://github.com/mayeut/patch-macos-python-openssl/releases/download/v1.1.1h/patch-macos-python-{version}-openssl-v1.1.1h.tar.gz'
-            download(open_ssl_patch_url, Path('/tmp/python-patch.tar.gz'))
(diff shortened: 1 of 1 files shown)


henryiii:
Hard to tell, but I think this might fix it. I'd like to refactor all the tmp usage to use TemporaryDirectory and a context manager, I think that might fully fix our CI flakiness about `/tmp`. It would also nicely cleanup after runs and help with thread safety (@Czaki's attempt to thread likely hit the fact that all jobs download to exactly the same `/tmp/python.pkg`?)

Czaki:
>  @Czaki's attempt to thread likely hit the fact that all jobs download to exactly the same `/tmp/python.pkg`

No. My attempts fail because `pip` fail when try to perform two installation at same time. And in contradiction to windows. Current macos python installation method does not allow to install python twice on machine.

henryiii:
Are you sure it wasn't because it was actually because it was the same underlying Python, since python.pkg was the same between two runs?

Czaki:
I first run basic_0 stem, and then, when Python is installed I try to run in parallel other tests.

joerick:
I restarted it once this morning, so that's two runs without issues. The `/tmp/Python.pkg` thing is a relic from the very early days (I was trying to make the paths descriptive (like `/tmp/repaired_wheels`), so the logs would make more sense...)

If we wanted to run builds in parallel, we'd need to create virtualenvs for each build, to properly isolate them. That would add to the run-time, though.

joerick (approved):

joerick:
Just restarted it again. If this one's green, that's a decent chance we've fixed it 🤞

henryiii:
```
+ Download https://www.python.org/ftp/python/3.7.9/python-3.7.9-macosx10.9.pkg to /var/folders/gk/lkr9pm5x039fx6d3j9r52rv80000gn/T/tmptu6wfpr5/Python.pkg
+ sudo installer -pkg /var/folders/gk/lkr9pm5x039fx6d3j9r52rv80000gn/T/tmptu6wfpr5/Python.pkg -target /
installer: Error the package path specified was invalid: '/var/folders/gk/lkr9pm5x039fx6d3j9r52rv80000gn/T/tmptu6wfpr5/Python.pkg'.
```

Wow, that's really weird. It worked the first time, unlike before, but not the second. I still think #560 is a very good idea.

henryiii:
For debugging, maybe we can add a `ls -l` to this path to see what it's reporting as when it fails, and if the size is wrong.

henryiii:
I believe we are silently passing exceptions in `download`. It's trying three times, failing, and then instead of raising an error, it just returns as if nothing went wrong (as far as I read it). `download` should report and raise an error if it fails.

henryiii:
Failed:
```
-rw-r--r--  1 distiller  staff  27073838 Jan 25 06:49 /var/folders/gk/lkr9pm5x039fx6d3j9r52rv80000gn/T/tmptilibqdo/Python.pkg
```
This time it gave this error:
```
http.client.IncompleteRead: IncompleteRead(219264 bytes read, 24670021 more expected)
```

Working, on Azure (Python 3.5):
```
-rw-r--r--  1 runner  staff  24874014 Jan 25 14:53 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpgplaqpu4/Python.pkg
+ sudo installer -pkg /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpgplaqpu4/Python.pkg -target /
```

henryiii:
I'd like to refactor a bit eventually, but not needed here yet.

Issue #229, opened 2019-11-26
Consider including a link to Azure templates


astrofrog:
Building on top of cibuildwheel, @Cadair and I have developed an Azure template that makes it very easy for developers to add wheel and sdist building to their repositories:

https://github.com/OpenAstronomy/azure-pipelines-templates/

With this template, the Azure configuration can look as simple as:

```
jobs:
- template: publish.yml@OpenAstronomy
  parameters:
    pypi_connection_name: 'pypi_endpoint'
    targets:
    - sdist
    - wheels_linux
    - wheels_macos
    - wheels_windows
```

But still allows the ``CIBW`` env variables to be set if needed. I thought it might be worth linking to this from the README here? (but feel free to close if not)

henryiii:
Wait, can you load a remote template in Azure? With parameters? [https://github.com/scikit-hep/azure-wheel-helpers](azure-wheel-helpers) would likely have lasted longer if I knew about this (but probably best I didn't, as the transition to cibuildwheel has been fantastic). (or maybe it's a newer feature)

Also, there are lots of thumbs up, so maybe you could propose the wording to use as a note in the Azure section? Also, what does "opinionated" mean? That should probably be noted a bit better in the README or in the note. Do you have to use tox, PEP 517, etc.

Just posting it here is fine, I can copy it to the docs. I'm working on something that might help here, too.

joerick:
I _nearly_ considered using a template as the 'minimal config' in cibuildwheel, but I decided in the end that it's better to spell it out natively, so there isn't a 'config eject' moment, where a small tweak requires a large change away from the template to a native approach.

But, I expect for many people a template would do just fine. However, I do hesitate to include this in our docs as a first-party example because these templates have their own 'API' and docs - https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html . Perhaps, a link out to [their docs site](https://openastronomy-azure-pipelines.readthedocs.io/en/latest/index.html) from the Azure setup section would be more appropriate.

Cadair:
I also think a link to the docs would be the best option.

PR #567, opened 2021-01-29
docs: cleanup

henryiii:
Adding three things (one commit each):

* Support for `[docs]` and `[test]` extras, allowing these to be installed individually. Dev no longer includes the docs requirements (but the requirements-dev file does both for you, just like before). Simplifies RtD build a little bit. Bumped the pinned versions of docs requirements. Edit: `[test]` not `[tests]`, more common (checked Pandas).
* Bump to showing the 20.04 images, since the default should be changing soon. (GHA only, not sure about Azure or others)
* ~~Support nested markdown in other constructs. @joerick I think there might be one other place where this was needed?~~ Removed for now, might be ruining styling

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,1 +1,1 @@
--e .[dev]
+-e .[dev,docs]
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,21 +39,29 @@ install_requires =
 cibuildwheel = resources/*
 
 [options.extras_require]
-dev =
-    click
+docs =
     mkdocs-include-markdown-plugin==2.1.1
     mkdocs==1.0.4
+    pymdown-extensions
+test =
+    jinja2
+    pytest
+    pytest-timeout
+dev =
+    click
+    ghapi
+    jinja2
+    mypy>=0.800
+    packaging>=20.8
     pip-tools
     pygithub
-    ghapi
-    pymdown-extensions
-    pytest>=4
     pytest-timeout
+    pytest>=4
     pyyaml
     requests
-    typing-extensions
+    requests
     rich>=9.6
-    mypy>=0.800
+    typing-extensions
 
 [options.packages.find]
 include =
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,31 +38,6 @@ install_requires =
 [options.package_data]
 cibuildwheel = resources/*
 
-[options.extras_require]
-docs =
-    mkdocs-include-markdown-plugin==2.1.1
-    mkdocs==1.0.4
-    pymdown-extensions
-test =
-    jinja2
-    pytest
-    pytest-timeout
-dev =
-    click
-    ghapi
-    jinja2
-    mypy>=0.800
-    packaging>=20.8
-    pip-tools
-    pygithub
-    pytest-timeout
-    pytest>=4
-    pyyaml
-    requests
-    requests
-    rich>=9.6
-    typing-extensions
-
 [options.packages.find]
 include =
     cibuildwheel
--- a/.gitignore
+++ b/.gitignore
@@ -106,4 +106,11 @@ env3?/
 # MyPy cache
 .mypy_cache/
 
+# Output cache for setup.py checker
 all_known_setup.yaml
+
+# mkdocs
+site/
+
+# Virtual environments
+venv*
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
   rev: 20.8b1
   hooks:
   - id: black
-    files: ^bin/update_pythons.py$
+    files: ^bin/update_pythons.py|setup.py$
 
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.800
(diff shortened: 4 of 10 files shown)


requirements-dev.txt
```diff
@@ -1 +1 @@
--e .[dev]
+-e .[dev,docs]
```
joerick, line 1: This seems a little odd - shouldn't requirements-dev match `[dev]`?  and `[dev]` include the requirements for building docs? Could the `[dev]` extra depend on other extras?

joerick (changes requested):
Something seems to have gone wrong with our styling, here...

| Before | After |
|---|---|
| home-before | home-after
| setup-before | setup-after
| option-before | option-after

requirements-dev.txt
henryiii: Extras can't depend on other extras (my biggest pet peeve - and one that Poetry seems to not have solved when they built a more complicated requirements system), so they'd have to be listed twice. However, this means there might something missing from docs. I thought it was a little nicer to split "docs" from "dev", as often developers are not rebuilding the docs, so it gives more freedom there (since it's easy to ask for multiple extras. You can't as easily make multiple requirements files (it's ugly), and for historical purposes, the requirements-dev should stay the same. But I can also list them twice if better.

henryiii:
That is really weird, I don't see what would cause such a change in the styling... Wait, I wonder if it's super fences? It's now able to "nest" syntax, so... Still don't see how that would change un-fenced regular styling, like the table.

joerick:
it might be the mkdocs update, maybe they changed the theme?

henryiii:
Maybe it could be the newer versions? If we have customizations depending on the version of these dependencies, perhaps? Later I'll try to bisect it and see which change causes the failure.

requirements-dev.txt
henryiii: Note: you _can_ move extras back to setup.py, then you can make them depend on either other via code, I do this for extras sometimes: https://github.com/scikit-hep/boost-histogram/blob/918caebdf922097fde92df5a5de5442d86aa04c7/setup.py#L51-L64

I often have an "all" if I do this this way.

henryiii:
I think it was the super fences causing different HTML produced for nested constructs.

My real goal was to try to set up a way to add tabs (which I didn't manage yet / here, this was just some cleanup while playing with ideas). I think having tabs could really help in a few places:

* We could add `action | classic | pipx` setup for GitHub Actions
* We could add `classic | template` for Azure (with #229)
* We could make the example for setting an environment variable tabbed https://cibuildwheel.readthedocs.io/en/stable/options/#setting-options.

The tab support for mkdocs seems to be for other themes, so we'd probably have to roll our own.

henryiii:
Should we change https://cibuildwheel.readthedocs.io/en/stable/options/#examples_4 to use https://github.com/scipy/oldest-supported-numpy ?

YannickJadoul:
> Should we change https://cibuildwheel.readthedocs.io/en/stable/options/#examples_4 to use https://github.com/scipy/oldest-supported-numpy ?

I guess so, if it's good practice, even though that would show off less the possibilities of `pyproject.toml`

henryiii:
I would use:

```toml
    "numpy==1.13.3; python_version<'3.5'",
    "oldest-supported-numpy; python_version>='3.5'",
```

This would still show off the abilities of toml, but it would not have to be updated on every Python release, and it includes things like PyPy too!

YannickJadoul:
That would be perfect, indeed, I'd say :-)

henryiii:
Okay, pushed _just_ the cleanups; the version updates can go in later, not mixed into this PR, since they are affecting the style and will need more work. Superfences too, maybe. I'd like to work on adding tabs to the above mentioned places as well eventually.

setup.cfg
```diff
@@ -39,21 +39,29 @@ install_requires =
...
+test =
+    jinja2
+    pytest
+    pytest-timeout
+dev =
+    click
+    ghapi
+    jinja2
+    mypy>=0.800
+    packaging>=20.8
```
joerick, line 55: Looks like we lost pytest from `dev`. Should we add the `test` extra to the requirements-dev? or add pytest in twice?

Maybe the best solution would be to define these in setup.py, after all. then `requirements-dev` can match `dev` and `dev can contain everything.

joerick (approved):
Just the one note on pytest dev requirement below, then I'm happy

setup.cfg
henryiii: Okay, moved them over, also added an "all" which installs everything. I'd rather keep the docs requirements out of dev, since I don't use the same venv for dev and docs; it's easy to specify two extras (or `all` now).

henryiii:
If you'd rather, I could rename all -> dev, so that requirements-dev.txt would match `[dev]`, then the current "dev" could be "scripts" or "bin" or something like that to indicate it's the requirements for the bin directory. I think this naming is more common, where you have a dev for developer requirements, a docs for building docs, and test which is just the part of dev that builds tests. Developers who want to build docs can do `[dev,docs]`.

henryiii:
We can tweak or iterate later, it's approved and I fixed the issue mentioned (though, technically, pytest is still there, it just is mismatched (one has a >4, one doesn't), so I plan to merge soonish. I've now started alphabetically sorting the includes, just like in requirements.txt, so that these don't get added multiple times or in different places.

PR #570, opened 2021-02-01
feat: print new files after building

### LINKED ISSUE #566: Tiny feat: Add printout of produced wheels at end of logs
One thing I have in common for all my uses of cibuildwheel is:

```yaml
    - uses: joerick/cibuildwheel@v1.8.0

    - name: Show files
      run: ls -lh wheelhouse
      shell: bash
```

It might be a nice to list the built wheel names at the end of the logs, so it's easy to tell what's been produced. (If actions/upload-artifact printed the name of the files it was uploading, this wouldn't be so important, but it doesn't)

henryiii:
This adds a printout after cibuildwheel runs to list exactly what changes. Serves as a good way to check to see exactly what built, and should show if any (un)expected extra files show up.

Closes #566

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -281,6 +281,8 @@ def main() -> None:
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
+    existing_contents = set(output_dir.iterdir())
+
     if platform == 'linux':
         cibuildwheel.linux.build(build_options)
     elif platform == 'windows':
@@ -289,6 +291,10 @@ def main() -> None:
         cibuildwheel.macos.build(build_options)
     else:
         assert_never(platform)
+
+    final_contents = set(output_dir.iterdir())
+    new_contents = final_contents - existing_contents
+    print("Produced wheels:", *sorted(f.name for f in new_contents))
 
 
 def detect_obsolete_options() -> None:
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -282,7 +282,7 @@ def main() -> None:
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
-    with cibuildwheel.util.print_new_wheels("{n} wheels produced:", output_dir):
+    with cibuildwheel.util.print_new_wheels("{n} wheels produced in {m:.0f} minutes:", output_dir):
         if platform == 'linux':
             cibuildwheel.linux.build(build_options)
         elif platform == 'windows':
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -1,18 +1,18 @@
 import argparse
-import contextlib
 import os
 import sys
 import textwrap
 import traceback
 from configparser import ConfigParser
 from pathlib import Path
-from typing import Dict, Iterator, List, Optional, Set, Union, overload
+from typing import Dict, List, Optional, Set, Union, overload
 
 from packaging.specifiers import SpecifierSet
 
 import cibuildwheel
 import cibuildwheel.linux
 import cibuildwheel.macos
+import cibuildwheel.util
 import cibuildwheel.windows
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.environment import EnvironmentParseError, parse_environment
@@ -282,7 +282,7 @@ def main() -> None:
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
-    with print_new_wheels("{n} wheels produced:", output_dir):
+    with cibuildwheel.util.print_new_wheels("{n} wheels produced:", output_dir):
         if platform == 'linux':
             cibuildwheel.linux.build(build_options)
         elif platform == 'windows':
@@ -314,18 +314,6 @@ def detect_obsolete_options() -> None:
             if option in os.environ and deprecated in os.environ[option]:
                 print(f"Build identifiers with '{deprecated}' have been deprecated. Replacing all occurences of '{deprecated}' with '{alternative}' in the option '{option}'")
                 os.environ[option] = os.environ[option].replace(deprecated, alternative)
-
-
-@contextlib.contextmanager
-def print_new_wheels(msg: str, output_dir: Path) -> Iterator[None]:
-    existing_contents = set(output_dir.iterdir())
-    try:
-        yield
-    finally:
-        final_contents = set(output_dir.iterdir())
-        new_contents = final_contents - existing_contents
-        n = len(new_contents)
-        print(msg.format(n=n), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
 
 
 def print_preamble(platform: str, build_options: BuildOptions) -> None:
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -1,11 +1,12 @@
 import argparse
+import contextlib
 import os
 import sys
 import textwrap
 import traceback
 from configparser import ConfigParser
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Union, overload
+from typing import Dict, Iterator, List, Optional, Set, Union, overload
 
 from packaging.specifiers import SpecifierSet
 
@@ -281,20 +282,15 @@ def main() -> None:
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
-    existing_contents = set(output_dir.iterdir())
-
-    if platform == 'linux':
-        cibuildwheel.linux.build(build_options)
-    elif platform == 'windows':
-        cibuildwheel.windows.build(build_options)
-    elif platform == 'macos':
-        cibuildwheel.macos.build(build_options)
-    else:
-        assert_never(platform)
-
-    final_contents = set(output_dir.iterdir())
-    new_contents = final_contents - existing_contents
-    print("Produced wheels:", *sorted(f.name for f in new_contents))
+    with print_new_wheels("{n} wheels produced:", output_dir):
+        if platform == 'linux':
+            cibuildwheel.linux.build(build_options)
+        elif platform == 'windows':
+            cibuildwheel.windows.build(build_options)
+        elif platform == 'macos':
+            cibuildwheel.macos.build(build_options)
+        else:
+            assert_never(platform)
 
 
 def detect_obsolete_options() -> None:
@@ -318,6 +314,18 @@ def detect_obsolete_options() -> None:
             if option in os.environ and deprecated in os.environ[option]:
                 print(f"Build identifiers with '{deprecated}' have been deprecated. Replacing all occurences of '{deprecated}' with '{alternative}' in the option '{option}'")
                 os.environ[option] = os.environ[option].replace(deprecated, alternative)
+
+
+@contextlib.contextmanager
+def print_new_wheels(msg: str, output_dir: Path) -> Iterator[None]:
+    existing_contents = set(output_dir.iterdir())
+    try:
+        yield
+    finally:
+        final_contents = set(output_dir.iterdir())
+        new_contents = final_contents - existing_contents
+        n = len(new_contents)
+        print(msg.format(n=n), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
 
 
 def print_preamble(platform: str, build_options: BuildOptions) -> None:
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -1,3 +1,4 @@
+import contextlib
 import fnmatch
 import itertools
 import os
@@ -8,7 +9,7 @@ import urllib.request
 from enum import Enum
 from pathlib import Path
 from time import sleep
-from typing import Dict, List, NamedTuple, Optional, Set
+from typing import Dict, Iterator, List, NamedTuple, Optional, Set
 
 import bracex
 import certifi
@@ -251,3 +252,15 @@ def unwrap(text: str) -> str:
     text = text.strip()
     # remove consecutive whitespace
     return re.sub(r'\s+', ' ', text)
+
+
+@contextlib.contextmanager
+def print_new_wheels(msg: str, output_dir: Path) -> Iterator[None]:
+    existing_contents = set(output_dir.iterdir())
+    try:
+        yield
+    finally:
+        final_contents = set(output_dir.iterdir())
+        new_contents = final_contents - existing_contents
+        n = len(new_contents)
+        print(msg.format(n=n), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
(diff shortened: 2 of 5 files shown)


henryiii:
I could make this a context manager if you prefer, something like this:

```path
@contextlib.contextmanager
def print_change(msg: str, output_dir: Path):
    existing_contents = set(output_dir.iterdir())
    yield
    final_contents = set(output_dir.iterdir())
    new_contents = final_contents - existing_contents
    print(msg, *sorted(f.name for f in new_contents))
```

cibuildwheel/__main__.py
```diff
@@ -290,6 +292,10 @@ def main() -> None:
     else:
         assert_never(platform)
 
+    final_contents = set(output_dir.iterdir())
+    new_contents = final_contents - existing_contents
+    print("Produced wheels:", *sorted(f.name for f in new_contents))
```
YannickJadoul, line 297: 2 suggestions:
- Can we make this foldable, the folded thing just showing "Produced N wheels"?
- Maybe one wheel per line, `ls` style? Especially if this is foldable, this will be more clear to look up which wheel is missing, I'd expect.

cibuildwheel/__main__.py
```diff
@@ -281,6 +281,8 @@ def main() -> None:
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
 
+    existing_contents = set(output_dir.iterdir())
```
YannickJadoul, line 284: Good catch; I hadn't thought of there already being wheels present. What about `output_dir.glob("*.whl")`, since you're saying "produced wheels"?

YannickJadoul (commented):
Not sure how I feel about the context manager. I don't think I would, as it would just increase nesting and we don't want to per se be "exception-safe" and print the wheels afterwards? If there's an exception, do you really still want to print "This is how many wheels you have"?

cibuildwheel/__main__.py
henryiii: Well, we shouldn't be producing non-wheels, and if we do, we don't want to hide it, that would make debugging harder?

henryiii:
I'd think it might help for debugging to see what was produced if it failed part way through. But no strong option on whether it should print if exceptions are thrown. I did not write the above example to handle exceptions, there's no try/finally.

cibuildwheel/__main__.py
henryiii: We could, I considered that. We don't' collapse the header, though, so was thinking this might be best uncollapsed too. But I was on the fence.

YannickJadoul: But this is something you don't always want to see? I my case, I know e.g. that I'm expecting (_looks at my releases_) 32 wheels. So if that's fine, without an error, I feel like that's more clear than always seeing everything.
At the very least, everything on one line loses all oversight, for me, when I checked the example build.

cibuildwheel/__main__.py
YannickJadoul: There's something to be said about that. So if that's the case, we might even want to check that all produced files end in `.whl`?

cibuildwheel/__main__.py
henryiii: Yes, I'll collapse and do one per line. Though, if you scroll all the way down the log to the end anyway, not sure you should then have to uncollapse it - there's nothing past it, after all, in that step, usually. :)

cibuildwheel/__main__.py
YannickJadoul: But yeah, nothing urgent either.

henryiii: It shouldn't be common (or possible?), but I'd currently just like not to hide it if it happens, rather than erroring out quite yet on it happening.

The thing that would help me is if you could detect if the git dir was dirty. That causes badly named wheels when using setuptools_scm, which can't upload to PyPI.

YannickJadoul:
> I did not write the above example to handle exceptions, there's no try/finally.

Hmmm, true. I thought it would handle exceptions, but not really. At least, that's most often my reason to use context managers, to have some kind of C++ RAII-like thing. So I'm not sure if this is very useful, here.

> I'd think it might help for debugging to see what was produced if it failed part way through.

Maybe. But wouldn't you check anyway? At the point where things fail, you'll go have a look at where it did fail, I'd say.

cibuildwheel/__main__.py
YannickJadoul: OK! Hmmm, indeed, I thought GHA would scroll down for me, but not really.

But I switched around my reasoning, it seems. I first thought of 2., to see the different wheel names nicely lined up underneath each other (also catches outliers better, if there's files that shouldn't be there!), and only then thought about 1. since it would be too long and would need to get collaped :-)

henryiii:
But checking to see what was produced either causes you to scroll all the way through the logs (and the logs can get a little scrambled if something fails), or to add a "ls" line and run again.

I could have it handle exceptions, I just didn't. If it is useful to still print in case of exceptions, it would be better as a context manager. We aren't too deeply nested here (and if we were, that means something should be refactored).

YannickJadoul:
> But checking to see what was produced either causes you to scroll all the way through the logs (and the logs can get a little scrambled if something fails), or to add a "ls" line and run again.

No, but the failed build will be the last one, anyway, no?  Somehow "Produced wheels" feels like a report of a probably successful thing to me.

> I could have it handle exceptions, I just didn't. If it is useful to still print in case of exceptions, it would be better as a context manager. We aren't too deeply nested here (and if we were, that means something should be refactored).

I personally don't see the need, but I'll leave this up to @joerick. Reasonably neutral.

henryiii:
Pulling folding out of logger.py without printing times and such might take a little work to make it clean to do. So I'll wait and see what directions @joerick likes here.

YannickJadoul:
> Pulling folding out of logger.py without printing times and such might take a little work to make it clean to do. So I'll wait and see what directions @joerick likes here.

Oh, I was hoping you could reuse `logger` :-( If not, that might not be worth it, then, indeed :-/

henryiii:
I think I can, but I'll need to refactor the folding part to be able to be used standalone. It's not a huge change, but just need to see what direction we want to go first. If it was a contextmanager, those can also be decorators and I would simply put @fold(...) on my function. :)

henryiii:
@joerick Questions for you: inline or contextmanager, do you want it printing on success or always (the later strongly pushes toward context manager method), and do you want it folding with one file per line, or not folded and just chained together with commas.

cibuildwheel/__main__.py
joerick: One-per-line seems like a good idea. Maybe indent it too so they're under the header? like

```
16 wheels produced:
  spam-0.1.0-cp27-manylinux1_x86_64.whl
  spam-0.1.0-cp35-manylinux2010_x86_64.whl
  spam-0.1.0-cp35-manylinux1_x86_64.whl
  spam-0.1.0-cp35-manylinux2010_x86_64.whl
  spam-0.1.0-cp36-manylinux1_x86_64.whl
  spam-0.1.0-cp36-manylinux2010_x86_64.whl
  spam-0.1.0-cp37-manylinux1_x86_64.whl
  spam-0.1.0-cp37-manylinux2010_x86_64.whl
  spam-0.1.0-cp38-manylinux1_x86_64.whl
  spam-0.1.0-cp38-manylinux2010_x86_64.whl
  spam-0.1.0-cp39-manylinux1_x86_64.whl
  spam-0.1.0-cp39-manylinux2010_x86_64.whl
  ...
```

Unsure about collapsing, I see the argument, but I feel like in a success case the actual output of all the logs above seems important enough to be unhidden by default.

henryiii: Yes, and the overall step is always collapsed in all CI's, so collapsing inside the step wouldn't help anything.

joerick:
ah, apologies! I missed this and replied to the older source-line comment!

> @joerick Questions for you: inline or contextmanager, 

inline seems good. contextmanager is overkill imo

> do you want it printing on success or always (the later strongly pushes toward context manager method),

Printing should be only on success - we shouldn't talk about any successes after an error was encountered, the error should be the last thing in the log.

> and do you want it folding with one file per line, or not folded and just chained together with commas.

no folding, but one file per line :) like-

```
16 wheels produced:
  spam-0.1.0-cp27-manylinux1_x86_64.whl
  spam-0.1.0-cp35-manylinux2010_x86_64.whl
  spam-0.1.0-cp35-manylinux1_x86_64.whl
  spam-0.1.0-cp35-manylinux2010_x86_64.whl
  spam-0.1.0-cp36-manylinux1_x86_64.whl
  spam-0.1.0-cp36-manylinux2010_x86_64.whl
  spam-0.1.0-cp37-manylinux1_x86_64.whl
  spam-0.1.0-cp37-manylinux2010_x86_64.whl
  spam-0.1.0-cp38-manylinux1_x86_64.whl
  spam-0.1.0-cp38-manylinux2010_x86_64.whl
  spam-0.1.0-cp39-manylinux1_x86_64.whl
  spam-0.1.0-cp39-manylinux2010_x86_64.whl
  ...
```

henryiii:
Making it a context manager makes it much easier to avoid in the unit_test's; before, I had to hack around the iterdir; now I can just monkeypatch in an do-nothing context manager.

I can avoid the wheel printout if there's an error (currently it prints no matter what).

YannickJadoul:
FWIW, I wasn't sure about the context manager before, but seeing things now, it looks pretty nice and things that belong together are grouped :-)

joerick:
Looks a little too clever, for my taste!

henryiii:
So do you like it this way or before? Remember to _look at the tests_ to see how we had to hack around this before!

Maybe we could just link to [level-up-your-python: Context Managers](https://henryiii.github.io/level-up-your-python/notebooks/2.4%20Context%20Managers.html) for anyone who finds them confusing. :)

joerick:
I prefer before, the inline style, not the context manager. We're talking about [this commit](https://github.com/joerick/cibuildwheel/pull/570/commits/635156ad7bd1b4688906fa1a581c4cdf8d803203) right? Looks fine to me - we're already patching `Path.exists`, and regardless, I'd expect unit tests to involve a bit of mocking.

henryiii:
Yes, that commit. Mocking `Path.iterdir` could affect/break other things not related to this one step; what if we later iterate somewhere else, and have no idea why it's coming up empty in the tests? Mocking just this one context manager function is nicely localized. We could also add a unit test that verifies that this function works; while if it's inline, all the code is mangled with the build call and it's much harder to test. And the collect before and after logic is nicely organized in once place this way. But I'll restore the inline version if you really do want it.

I can at least keep this broken out into a separate monkeypatch so I don't have to do it twice like before if you are sure you like the inline.

henryiii:
Grouping related code in separate functions, context managers, and classes is not bad design - piling everything in "linear" fashion is unmanageable, untestable,  and unrefactorable. I'm currently having to [refactor this monstrosity](https://github.com/scikit-hep/mplhep/blob/1bc66451378004a8c2b31b6c74804f85f1eab43f/src/mplhep/plot.py#L106-L427) to handle something new, and it's completely impossible to change slightly without random things breaking everywhere. (note: this is the contents of a _single function_!)

YannickJadoul:
Wow. We're luckily not at _that_ level of complexity in `cibuildwheel` :-D

henryiii:
> Looks a little too clever, for my taste!

Didn't you write [this](https://github.com/joerick/cibuildwheel/blob/0fb2ead6fb5757076862910d05dfd882c9ba0272/cibuildwheel/logger.py#L194-L217) at one point? A context manager, which is built into the language, is "too clever" but you are happy to write a class in a class is ["fun"](https://github.com/joerick/cibuildwheel/pull/509#pullrequestreview-560651446)? BTW, the oddities of Python scoping for classes is [something that even Guido finds "broken"](https://twitter.com/gvanrossum/status/1354305179244392453?s=20).

I am totally okay with reverting to no context manager. But had to point that out. :)

joerick:
My sympathies! I don't really think the two are comparable, though :) your linked example has way too much unencapsulated state to my eyes. 

Here, in the inline case, it's a simple log statement, with a 'before' variable, within the main function of the program. It's not hard to read, and easy to ignore. As a context manager, it looks to me more complicated than it really is - I naturally wonder what it could mean, and there's more indirection and the control flow is more convoluted.

joerick:
But maybe there's something I'm not seeing, if you really feel this is the best route, we can do that too :)

joerick:
> Didn't you write this at one point? A context manager, which is built into the language, is "too clever" but you are happy to write a class in a class and call it "fun"? BTW, the oddities of Python scoping for classes is something that even Guido finds "broken".

Hahaha, fair enough. No accounting for taste ;) I thought it read quite well. Admittedly a total abuse of the 'class' though :)

henryiii:
Added a total time printout too and some unit tests.

If you want to avoid the context manager, since we are not "printing no matter what" on exceptions, then this could work:

```python
class WheelPrinter(NamedTuple):
    '''
    Prints the new items in a directory when print_new_wheels is called.

    Use like this:

    wheel_printer = WheelPrinter(time.time(), output_dir.iterdir())
    ... # Make wheels
    wheel_printer.print_new("Produced {n} new wheels in {m:.0} minutes", output_dir.iterdir())
    '''
    start_time: float
    existing_contents: Iterator[Path]

    def print_new(msg: str, final_contents: Iterator[Path]):
        '''
        The message to display
        can include {n} for number of wheels, {s} for total number of seconds,
        and/or {m} for total number of minutes. Does not print anything if this
        exits via exception.
        '''
        new_contents = set(final_contents) - set(self.existing_contents)
        n = len(new_contents)
        s = time.time() - self.start_time
        m = s / 60
        print(msg.format(n=n, s=s, m=m), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
```

It would look better as a dataclass, I think, but we don't require the backport yet, IIRC.

henryiii:
Here's the dataclass (would need the backport, but only for 3.6):

```python
@dataclass(frozen=True, init=False)
class WheelPrinter:
    '''
    Prints the new items in a directory when print_new_wheels is called.

    Use like this:

        wheel_printer = WheelPrinter(output_dir)
        ... # Make wheels
        wheel_printer.print_new("Produced {n} new wheels in {m:.0} minutes")
    '''
    output_dir: Path
    start_time: float
    existing_contents: Set[Path]

    def __init__(self, output_dir: Path) -> None:
        self.output_dir = output_dir
        self.start_time = time.time()
        self.existing_contents = set(output_dir.iterdir())

    def print_new(msg: str) -> None:
        '''
        The message to display
        can include {n} for number of wheels, {s} for total number of seconds,
        and/or {m} for total number of minutes. Does not print anything if this
        exits via exception.
        '''
        final_contents = set(self.output_dir.iterdir())
        new_contents = final_contents - self.existing_contents
        n = len(new_contents)
        s = time.time() - self.start_time
        m = s / 60
        print(msg.format(n=n, s=s, m=m), *sorted(f"  {f.name}" for f in new_contents), sep="\n")
```

joerick (approved):
We can do the context manager, that seems okay. Let's get this in, I'm feeling a release brewing!

PR #573, opened 2021-02-01
fix: updates should include universal2 Pythons

henryiii:
Fixes issues seen in #572

Small cleanup tweaks to black and MyPy, to make running by hand a little easier.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,6 @@ repos:
   hooks:
   - id: black
     files: ^bin/update_pythons.py$
-    args: ["--line-length=120"]
 
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.800
@@ -29,6 +28,7 @@ repos:
   - id: mypy
     exclude: ^(bin/|cibuildwheel/resources/).*py$
     additional_dependencies: [packaging]
+    args: ["--python-version=3.6", "--ignore-missing-imports", "--scripts-are-modules"]
   - id: mypy
     name: mypy 3.7+ on bin/
     files: ^bin/.*py$
--- a/bin/update_pythons.py
+++ b/bin/update_pythons.py
@@ -213,7 +213,7 @@ class AllVersions:
         self.windows_64 = WindowsVersions("64")
         self.windows_pypy = PyPyVersions("32")
 
-        self.macos_x86_64 = CPythonVersions()
+        self.macos_cpython = CPythonVersions()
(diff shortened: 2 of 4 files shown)


bin/update_pythons.py
```diff
@@ -229,11 +225,12 @@ def update_config(self, config: Dict[str, str]) -> None:
...
 
         # We need to use ** in update due to MyPy (probably a bug)
-        if "macosx_x86_64" in identifier:
+        if "macos" in identifier:
             if identifier.startswith("pp"):
                 config_update = self.macos_pypy.update_version_macos(spec)
             else:
-                config_update = self.macos_9.update_version_macos(spec) or self.macos_6.update_version_macos(spec)
+                config_update = self.macos_x86_64.update_version_macos(identifier, spec)
+
```
henryiii, line 233: This is much faster, especially when adding a check for 11 first.

.pre-commit-config.yaml
```diff
@@ -21,14 +21,14 @@ repos:
   hooks:
   - id: black
     files: ^bin/update_pythons.py$
-    args: ["--line-length=120"]
```
YannickJadoul, line 24: Why is this being removed? Feels unrelated.

YannickJadoul: Oh, nevermind. Moved down, I see now!

henryiii: It had to be manually given if running black directly. Now, it isn't required if running black directly.

bin/update_pythons.py
```diff
@@ -215,9 +213,7 @@ def __init__(self) -> None:
         self.windows_64 = WindowsVersions("64")
         self.windows_pypy = PyPyVersions("32")
 
-        self.macos_6 = CPythonVersions(plat_arch="macosx_x86_64", file_ident="macosx10.6.pkg")
-        self.macos_9 = CPythonVersions(plat_arch="macosx_x86_64", file_ident="macosx10.9.pkg")
-        self.macos_u2 = CPythonVersions(plat_arch="macosx_universal2", file_ident="macos11.0.pkg")
+        self.macos_x86_64 = CPythonVersions()
```
joerick, line 216: i guess this variable should be just `self.macos_cpython`? I don't see x86_64 parts...?

joerick (approved):
Thanks @henryiii - proof is in the runnin' i suppose but this looks okay to me 👍

bin/update_pythons.py
henryiii: Sure.

PR #576, opened 2021-02-04
docs: support tabs

henryiii:
This is a rough draft example of what tabs would look like. I've applied it to one place (GHA example), and the implementation is highly manual and not reusable, so will need to be improved if we want to go this way (@joerick might be able to be a lot more elegant at this than me :), or maybe there's a built-in tool somewhere (though I think we would need to support super fences, which we don't yet). Previously mentioned here: https://github.com/joerick/cibuildwheel/pull/567#issuecomment-770572043

Thoughts? [Visit the RtD PR check to view the docs, setup page](https://cibuildwheel--576.org.readthedocs.build/en/576/setup/). I've recently added tabs here: https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata (though that was in sphinx).

### DIFF
--- a/docs/extra.js
+++ b/docs/extra.js
@@ -1,4 +1,49 @@
-
+// add classes for code-block-filename styling
 $('.rst-content pre')
     .prev('blockquote')
     .addClass('code-block-filename');
+
+var i=0
+
+// convert tab admonition to tabs
+while (true) {
+  const firstTab = $('.rst-content .admonition.tab').first()
+  if (firstTab.length == 0) break;
+
+  const otherTabs = firstTab.nextUntil(':not(.admonition.tab)');
+  const allTabs = $($.merge($.merge([], firstTab), otherTabs));
+
+  const tabContainer = $('<div>').addClass('tabs');
+  const headerContainer = $('<div>').addClass('tabs-header');
+  const contentContainer = $('<div>').addClass('tabs-content');
+  console.log(allTabs)
+
+  tabContainer.insertBefore(firstTab);
+  tabContainer.append(headerContainer, contentContainer)
+
+  const selectTab = function (index) {
+    headerContainer.children().removeClass('active')
+    headerContainer.children().eq(index).addClass('active')
+    contentContainer.children().hide()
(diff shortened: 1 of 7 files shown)


joerick:
Hey @henryiii , I hope you don't mind I had a tinker with this, thought I'd try using the markdown 'admonition' syntax, which is like

```markdown

!!! tab "tab name"
    tab content

!!! tab "tab name"
    tab content
```

Then it's transformed into tabs with jQuery. What do you think? I also had a go at some styling, not 100% sure about that yet but there's something there.

image

To get those code snippets working inside the tabs, I also managed to get the styling of the superfences extension working (the problem before was that it was doing pygments highlighting, which wasn't supported by the theme we're using, which uses highlightjs)

henryiii:
Looks good! I can rebase and tweak a bit, then look at what else can be tabbified. Not sure if we need it, but can this do "grouped" tabs, where all matching tabs change if one changes? That could be possibly useful for the options examples, perhaps; you could select "travis" and all the examples of options change to the travis tab.

henryiii:
I've tried to unify on the same order for the CI listing, I've made the list of examples at the top of the options page tabbed (and added Gitlab CI, which was missing), and reordered/updated the original tabs to better reflect the current examples (though, I just realized, we've moved the main example to actions, so now we probably have two actions examples. Will fix later.

henryiii:
I really like the tabbed options examples. https://cibuildwheel--576.org.readthedocs.build/en/576/options/#setting-options :)

docs/extra.js
```diff
@@ -1,4 +1,49 @@
-
+// add classes for code-block-filename styling
 $('.rst-content pre')
     .prev('blockquote')
     .addClass('code-block-filename');
+
+var i=0
```
joerick, line 6:
```suggestion
var tabConversionIterations = 0
```

joerick (approved):

henryiii:
@joerick Extra important to `git grep` for the old version when bumping to a new version next time; it's possible a doc change (like this one) could miss a form for the version number.

PR #581, opened 2021-02-07
docs: fix bullets in 1.6.0 release notes

henryiii:
These were not rendering correctly:

Screen Shot 2021-02-06 at 8 49 50 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ _9 September 2020_
   [still on Python 3.9b3](https://github.com/pypa/manylinux/issues/758), not a
   release candidate. We'd advise holding off on distributing 3.9 ppc64le wheels
   until a subsequent version of cibuildwheel.
-  
+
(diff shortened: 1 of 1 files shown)


henryiii:
Actually, @joerick, the last part of 1.9.0 in the docs renders poorly as well:

Screen Shot 2021-02-06 at 11 16 06 PM


Maybe you could try to fix that? I might be able to get to it at some point, but writing a lot currently.

joerick:
Ohh, are these problems in the documentation version? Because it renders fine in the README in Github-

image

joerick:
the second problem referenced will probably be fixed by the new superfences codeblock thingy from #576

henryiii:
> are these problems in the documentation version?

Yep.

joerick:
btw, that pre-commit auto fix is really cool!

henryiii:
Yes it is. Enables me to be lazy and use the web editor sometimes. :)

PR #583, opened 2021-02-08
docs: add missing AS indentifiers

henryiii:
https://cibuildwheel--583.org.readthedocs.build/en/583/options/#build-skip

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -79,17 +79,17 @@ When setting the options, you can use shell-style globbing syntax, as per [`fnma
 
 <div class="build-id-table-marker"></div>
 
-|                | macOS                                                               | Windows                       | Manylinux Intel                               | Manylinux Other                                                            |
-|----------------|---------------------------------------------------------------------|-------------------------------|-----------------------------------------------|----------------------------------------------------------------------------|
-| Python 2.7     | cp27-macosx_x86_64                                                  | cp27-win_amd64<br/>cp27-win32 | cp27-manylinux_x86_64<br/>cp27-manylinux_i686 |                                                                            |
-| Python 3.5     | cp35-macosx_x86_64                                                  | cp35-win_amd64<br/>cp35-win32 | cp35-manylinux_x86_64<br/>cp35-manylinux_i686 | cp35-manylinux_aarch64<br/>cp35-manylinux_ppc64le<br/>cp35-manylinux_s390x |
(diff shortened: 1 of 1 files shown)


YannickJadoul:
Why aren't we adding more columns, here? There e.g. also "Manylinux Armv8 64bit"? (This also works, though; hadn't thought about this option yet!)

henryiii:
Because it makes the table really, really wide, and one item column look bad, I think. But could be either way, was just trying this.

joerick:
I wonder if we could do more line-breaking actually. If they were grouped under the columns: `macOS`, `Windows`, `Manylinux Intel`, `Manylinux Other` would that make the table more manageable? Sorry I don't have time to try this right now! this also LGTM.

henryiii:
I almost tried this, actually! I can it this out soonish.

henryiii:
With a bit of fighting, I think I have it pretty much into what fits on the screen (with about 10 more pixels, it would fit without scrolling for me, but I think this is close enough). I've removed the beta label from PyPy3.7, which took up a lot of space, and is missing now from their docs. Given that NumPy 1.20 only ships with PyPy3.7, it probably is not helpful to keep the beta label, anyway. I also removed the space between `PyPy` and the number, since that's how they officially list it, and it saved another sliver of space.

Screen Shot 2021-02-08 at 10 04 21 AM

I quite like it, I think. Much denser and easier to see general structure.

henryiii:
If we wanted to fully get it into one page, we could put the pypy versions on the second line, or drop the patch digit from the PyPy version (after all, the final patch digit is not relevant for a wheel's compatibility - probably should be removed anyway)

henryiii:
Yes, that does it:

Screen Shot 2021-02-08 at 10 12 32 AM

(at least for my font and such)

YannickJadoul:
I still have a tiny scroll bar :-p

image

But this looks fine to me, yes! (If it were up to me, we could even still group all `manylinux` IDs, but ... eh, this is good :-) )

henryiii:
macOS has slightly tighter fonts than Linux, I've noticed that before. Was really a pain when developing the flowmeter software, I would nail the design locally, then it would be 5% of the screen on the Raspberry π powered nurse station (the final design ended up shipping with Windows NUC's, not sure if they were closer to macOS or Linux, but it wasn't wider than Linux, anyway). But that's where I wanted to be - you can see all the content (even if not the visual table edge) without scrolling, still.

I like the fact that if we add more manylinuxes, this will grow vertically, not horizontally. :)

joerick:
Nice! Looks good :)

joerick:
(the last commit is unrelated, just something I noticed when testing)

Issue #587, opened 2021-02-10
Output mangled


henryiii:
I made a mistake, and cibuildwheel was very helpful in erroring out (otherwise I would have missed the mistake), but the error message was not at all where I was expecting it:

```
2021-02-10T05:15:42.7037690Z ##[group]Run joerick/cibuildwheel@v1.9.0
2021-02-10T05:15:42.7038170Z with:
2021-02-10T05:15:42.7038500Z   package-dir: .
2021-02-10T05:15:42.7038900Z   output-dir: wheelhouse
2021-02-10T05:15:42.7039250Z env:
2021-02-10T05:15:42.7039620Z   SETUPTOOLS_SCM_PRETEND_VERSION: 
2021-02-10T05:15:42.7040240Z   CIBW_ENVIRONMENT: PIP_ONLY_BINARY=numpy SETUPTOOLS_SCM_PRETEND_VERSION=
2021-02-10T05:15:42.7040790Z   CIBW_TEST_EXTRAS: test
2021-02-10T05:15:42.7041250Z   CIBW_TEST_COMMAND: pytest {project}/tests
2021-02-10T05:15:42.7041760Z   CIBW_TEST_SKIP: pp*macos* *universal2:arm64
2021-02-10T05:15:42.7042170Z   CIBW_BUILD: 
2021-02-10T05:15:42.7042530Z   CIBW_SKIP: cp27-win* pp27-win*
2021-02-10T05:15:42.7042940Z   CIBW_MANYLINUX_I686_IMAGE: 
2021-02-10T05:15:42.7043340Z   CIBW_MANYLINUX_X86_64_IMAGE: 
2021-02-10T05:15:42.7043730Z   CIBW_ARCHS: universal2
2021-02-10T05:15:42.7044100Z ##[endgroup]
2021-02-10T05:16:02.6041070Z 
2021-02-10T05:16:02.6042070Z      _ _       _ _   _       _           _
2021-02-10T05:16:02.6042490Z  ___|_| |_ _ _|_| |_| |_ _ _| |_ ___ ___| |
2021-02-10T05:16:02.6044500Z [?25l[?25hcibuildwheel: No build identifiers selected: BuildSelector('' - 'cp27-win* pp27-win*')
2021-02-10T05:16:02.6045630Z |  _| | . | | | | | . | | | |   | -_| -_| |
2021-02-10T05:16:02.6046050Z |___|_|___|___|_|_|___|_____|_|_|___|___|_|
[...]

PR #588, opened 2021-02-11
fix: empty CIBW_BUILD should still count as "*"

henryiii:
Mentioned in #587 - currently, there's an asymmetry between not setting `CIBW_BUILD`, and setting it to "", that can be surprising, especially in a matrix. This treats an empty string the same way as being undefined.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -147,7 +147,8 @@ def main() -> None:
     else:
         assert_never(platform)
 
-    build_config, skip_config = os.environ.get('CIBW_BUILD', '*'), os.environ.get('CIBW_SKIP', '')
+    build_config = os.environ.get('CIBW_BUILD') or '*'
+    skip_config = os.environ.get('CIBW_SKIP', '')
     test_skip = os.environ.get('CIBW_TEST_SKIP', '')
     environment_config = get_option_from_environment('CIBW_ENVIRONMENT', platform=platform, default='')
     before_all = get_option_from_environment('CIBW_BEFORE_ALL', platform=platform, default='')


henryiii:
Ouch, tests are pulling from latest? Manylinux2010 tests will need to stop trying to do that for Py2 for latest.

cibuildwheel/__main__.py
```diff
@@ -147,7 +147,8 @@ def main() -> None:
     else:
         assert_never(platform)
 
-    build_config, skip_config = os.environ.get('CIBW_BUILD', '*'), os.environ.get('CIBW_SKIP', '')
+    build_config = os.environ.get('CIBW_BUILD', '*') or '*'
```
YannickJadoul, line 150: Just `os.environ.get('CIBW_BUILD') or '*'`, then?

YannickJadoul (approved):

YannickJadoul:
Huh, wait. How are things broken?

henryiii:
Tests are dead until #591 goes in.

YannickJadoul:
> Tests are dead until #591 goes in.

Sorry, yeah. I'd just noticed :-)

cibuildwheel/__main__.py
henryiii: Sure.

joerick (approved):

henryiii:
PS: I also am somewhere between disliking and vehemently hating the `a, b = c, d` syntax. You'll notice the long block of code in mplhep I was refactoring had tons of it. I don't find it any better than writing `a = c; b = d;`, which is a fake way to compress lines, losing readability. And if you are trying to scan the code base for `b =`, it hides this by being the only construct in Python (except for the walrus operator, in 3.8+) that does have the assignment as the first item in a line. For `a, b = c` or `a, b = b, a`, you are doing something explicit (tuple unpacking or variable swapping), so losing first item assignment is a valid price to pay. But not for assignment grouping - a blank line does that much better.

(I'm planning to add do/do not's to [Level Up Your Python](https://henryiii.github.io/level-up-your-python), and this is one I'm planning to add. Mentioning it here gives the option for rebuttals before I add it. :) )

PR #590, opened 2021-02-11
fix: set MACOSX_DEPLOYMENT_TARGET for BEFORE_ALL

### LINKED ISSUE #563: set MACOSX_DEPLOYMENT_TARGET for all stages?
I recently adopted cibuildwheel (it is wonderful, thank you!), and made a broken release on older macs. The mistake I made was failing to set MACOSX_DEPLOYMENT_TARGET for the CIBW_BEFORE_ALL stage. I was surprised to see that CIBW has the feature to set MACOSX_DEPLOYMENT_TARGET, but doesn't do it for all stages.

#306 sets MACOSX_DEPLOYMENT_TARGET for python commands, but I was building a library dependency in CIBW_BEFORE_ALL.

My wheels are roughly:

- CIBW_BEFORE_ALL compiles and installs a library
- build my wheel that links the library
- repair (auditwheel/delocate/delvewheel) bundles the library in the wheel

So my question is: should I have built the library in a different stage, or does it make sense to set the default MACOSX_DEPLOYMENT_TARGET for all commands, not just the python environment?

Fixing it was easy enough (set MACOSX_DEPLOYMENT_TARGET in global env)

henryiii:
Looking at it, I think just a simple fix is likely okay; there's not a way to combine these easily, and it's a bit specific to this one variable. Fixes #563.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -288,6 +288,7 @@ def build(options: BuildOptions) -> None:
         if options.before_all:
             log.step('Running before_all...')
             env = options.environment.as_dictionary(prev_environment=os.environ)
+            env.setdefault('MACOSX_DEPLOYMENT_TARGET', '10.9')
             before_all_prepared = prepare_command(options.before_all, project='.', package=options.package_dir)
             call([before_all_prepared], shell=True, env=env)
 


YannickJadoul:
Rebasing this as well, since CircleCI is still broken.

YannickJadoul (approved):
This is the same order as for the other times we add `MACOSX_DEPLOYMENT_TARGET` (in `CIBW_BEFORE_BUILD` and for the main build): after evaluating `CIBW_ENVIRONMENT` (so `CIBW_ENVIRONMENT` can't access the variable).

So, looks good to me!

YannickJadoul:
@henryiii, Azure is a bit confused, but depending on how much you are in a hurry, this could get merged :-)

joerick (approved):
thanks @henryiii 👍

PR #591, opened 2021-02-11
tests: don't expect CPython 2.7 in manylinux2010 images

henryiii:
Fixing our tests, stage 1. Stage 2 will be when we update our docker images. We might want to pin to the last manylinux2010 with Python2 for the cibuildwheel 1.9.x series. The next release could drop Python 2 when building with manylinux2010, just like 2014 does now, I believe.


YannickJadoul:
Let's merge this. There's no discussion to be had about pinning these dockcross images, anyway.

PR #592, opened 2021-02-12
refactor: subprocess.run

henryiii:
In Python 3.5+, `run` is [recommended](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) over `check_*`. This converts to using the newer syntax, as well as catches two issues: `check_call` either returns 0 or throws an error, so the return int functions were misleading. And there was a case of `check_output` that was not using the output. (If it was meant to hide the output, I can tell it to do that via run).

- change check_call to run
- refactor: change check_output to run

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -49,7 +49,8 @@ def get_python_configurations(
 
 def build(options: BuildOptions) -> None:
     try:
-        subprocess.run(['docker', '--version'], check=True)
+        # check docker is installed
+        subprocess.run(['docker', '--version'], check=True, stdout=subprocess.DEVNULL)
     except Exception:
         print('cibuildwheel: Docker not found. Docker is required to run Linux builds. '
               'If you\'re building on Travis CI, add `services: [docker]` to your .travis.yml.'
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -49,7 +49,7 @@ def get_python_configurations(
 
 def build(options: BuildOptions) -> None:
     try:
-        subprocess.check_output(['docker', '--version'])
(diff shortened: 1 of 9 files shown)


cibuildwheel/linux.py
```diff
@@ -49,7 +49,7 @@ def get_python_configurations(
 
 def build(options: BuildOptions) -> None:
     try:
-        subprocess.check_output(['docker', '--version'])
+        subprocess.run(['docker', '--version'], check=True)
```
henryiii, line 52:
```suggestion
        subprocess.run(['docker', '--version'], check=True, stdout=subprocess.DEVNULL)
```

This probably was to hide the output?

henryiii:
Not sure why that one job timed out, but we can add `timeout-minutes:60` or something like that so we don't have multi-hour waits if something hangs.

cibuildwheel/linux.py
joerick: Yes, that was the intention - we didn't care about the version, really it was just to check that docker was installed.

joerick: perhaps adding a comment to that effect would be good, too
```suggestion
        # check docker is installed
        subprocess.run(['docker', '--version'], check=True, stdout=subprocess.DEVNULL)
```

joerick (approved):
Looks good to me!  I tried this the other day but couldn't figure it out. `stdout=subprocess.PIPE` is the incantation I was missing - I had been trying `capture_output=True` and mypy was giving some craaazy error messages. In hindsight it was probably because Python3.6 doesn't have capture_output.

joerick:
Travis also being flakey.... but that's a separate discussion

PR #593, opened 2021-02-12
refactor: string and setup.cfg normalization

henryiii:
I found (from pypa/build) a way to enforce blackwards quotes (single quotes instead of double quotes). Black's argument (in English, single quotes are more common, so using double quotes avoids escaping as often) is reasonable, though I really just like a standard regardless of which one, and black is reasonable. But if we are going with single quotes (for now?), let's be consistent. (And it's mostly my fault that we are not, I think, because I revert to blackish).

Also adding a few extra pre-commit checks, and _almost_ applying consistent styling in pre-commit to setup.cfg - the parser loses comments, so can't activate it as as default check.

Look at the changes by commit for easier review.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,8 @@ repos:
   - id: check-case-conflict
   - id: check-merge-conflict
   - id: check-yaml
+  - id: check-toml
+  - id: debug-statements
   - id: end-of-file-fixer
     exclude: (cibuildwheel/resources/pinned_docker_images.cfg)|(.svg$)
   - id: mixed-line-ending
@@ -21,6 +23,12 @@ repos:
   hooks:
   - id: black
     files: ^bin/update_pythons.py|setup.py$
+
+- repo: https://github.com/asottile/setup-cfg-fmt
+  rev: v1.16.0
+  hooks:
+  - id: setup-cfg-fmt
+    stages: [manual]
 
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v0.812
@@ -54,8 +62,6 @@ repos:
     exclude: cibuildwheel/resources/
     additional_dependencies: [flake8-bugbear]
 
-
-# The original pybind11 checks for a few C++ style items
 - repo: local
   hooks:
   - id: disallow-caps
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,4 +8,5 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 120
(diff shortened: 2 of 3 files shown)


joerick:
on the single quote: yeah, I guess? The inconsistency doesn't really bother me, I somehow trained my brain not to see any difference, but if the precommit hook can just auto-fix it, I suppose it's no extra work for anyone, so why not.

> and almost applying consistent styling in pre-commit to setup.cfg - the parser loses comments, so can't activate it as as default check.

How does this work then? You can manually invoke the fixer, then manually add back the comments?

henryiii:
>  The inconsistency doesn't really bother me

Would you be fine with black-like double quotes everywhere, then? :) Not sure I can quite as easily add them without full Black, though, since I suspect this fixer was added entirely to be the opposite of black.

> How does this work then? You can manually invoke the fixer, then manually add back the comments?

Yes (manually add partial file changes via `git add -p` or `git gui`). Assuming this doesn't change much, so not something to normally worry about.

YannickJadoul:
Not a suggestion, but just to (probably) annoy @henryiii: I've even tried to take up the habit of having `'` and `"` mean something differently, in my code. `"` is for text and things that get printed, while `'` would be for constants or column names of a data frame or so. I'm not sure it gives a lot of extra expressiveness, but if Python allows you some freedom, you'd better take it? ;-)

henryiii:
Python gives you a lot of freedom you should not take. The focus should be on the code you write, not on patterns that do not affect Python. Consistency helps readability. If you can enforce the single/double quote convention in pre-commit, then go ahead, that is useful; but otherwise it will never be effective - readers are not going to be able to depend on it because it might be wrong. It also generates merge conflicts; two identical changes should not conflict due to ambiguity in quotes. It's still okay to use the opposite quote character to contain a string with the other one.

In Ruby, double quotes are the equivalent of f-strings, so the rule Rubocop imposes is double quotes only when there is an interpolated value (like the similar f-string rule in flake8). However, brew, for example, enforces double quotes everywhere regardless, I think to keep it simpler for users not familiar with Ruby.

There are two things going for double quotes; they are less common in English (this post has several single quotes, but doesn't have any double), and the C family requires double quotes.

joerick:
Perhaps we could merge this without the single/double quote fixer?

henryiii:
Yep. I like double quotes better, partially due to black, and partially because there really are a lot more instances of `'` than `"` in strings. So I'll drop this in hope of getting black eventually, but will leave the setup.cfg part.

joerick (approved):

PR #594, opened 2021-02-14
Pin `manylinux2010` images (CPython 2.7 support)

mayeut:
`manylinux2010` dropped CPython 2.7
Use the tag `2021-02-06-3d322a5` which is the last supporting CPython 2.7

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -23,7 +23,7 @@ jobs:
     - name: "Run update: python configs"
       run: python ./bin/update_pythons.py --force
     - name: Create Pull Request
-      if: github.ref == 'refs/heads/master'
+      if: github.ref == 'refs/heads/master' && github.repository == 'joerick/cibuildwheel'
       uses: peter-evans/create-pull-request@v3
       with:
         commit-message: Update dependencies
--- a/bin/update_dependencies.py
+++ b/bin/update_dependencies.py
@@ -53,28 +53,34 @@ Image = namedtuple('Image', [
     'manylinux_version',
     'platform',
     'image_name',
+    'tag',
 ])
 
 images = [
-    Image('manylinux1', 'x86_64', 'quay.io/pypa/manylinux1_x86_64'),
-    Image('manylinux1', 'i686', 'quay.io/pypa/manylinux1_i686'),
+    Image('manylinux1', 'x86_64', 'quay.io/pypa/manylinux1_x86_64', None),
+    Image('manylinux1', 'i686', 'quay.io/pypa/manylinux1_i686', None),
 
-    Image('manylinux2010', 'x86_64', 'quay.io/pypa/manylinux2010_x86_64'),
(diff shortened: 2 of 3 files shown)


mayeut:
This could be good for a `cibuildwheel 1.9.x LTS` as mentioned in https://github.com/joerick/cibuildwheel/discussions/580#discussioncomment-348485

I'll also open a PR for dropping Python 2.7 altogether.

YannickJadoul (approved):
Looks good! The 1.9 and backporting might need some further discussion and practicalities sorted out (about how we're going to achieve it, and what's the policy for backporting, etc), but this PR is independent of that. We'll need to pin this anyway :-)

YannickJadoul:
Although, hmmm, wait. Now anyone building only Python 3.6 wheels would also be stuck on old manylinux2010 images?

We could also introduce `CIBW_MANYLINUX_X86_64_IMAGE=python27`, and try detecting whether Python 2.7 is present?

henryiii:
> would also be stuck on old manylinux2010 

Yes. That's the idea behind making this an LTS then moving forward. I was more interested in making 2010 Py3 only, but still supporting 2.7 until it is dropped (or manylinux1 drops 2.7 support ;) ). I really wanted cibuildwheel to match manylinux. I had a discussion opened about it but @mayeut seems to have gone ahead anyway in the next PR.

What do we do now for manylinux 2014?

YannickJadoul:
> What do we do now for manylinux 2014?

I suppose builds with `manylinux2014` fail if Python 2.7 isn't deselected?

mayeut:
> I suppose builds with manylinux2014 fail if Python 2.7 isn't deselected?

Yes, as mentioned in the docs (`Note that manylinux2014 doesn't support builds with Python 2.7 - when building with manylinux2014, skip Python 2.7 using CIBW_SKIP (see example below).`)

> Although, hmmm, wait. Now anyone building only Python 3.6 wheels would also be stuck on old manylinux2010 images?

> We could also introduce CIBW_MANYLINUX_X86_64_IMAGE=python27, and try detecting whether Python 2.7 is present?

@YannickJadoul,

This specific PR is in fact only here to allow things moving forward for everything except `manylinux2010`. If you look at the current situation, everything is stuck right now (unless more manual actions are taken) because the update pipeline is failing.
This only fixes this situation to allow everything else to update correctly. The points you are raising are valid but might be addressed later on.


> I was more interested in making 2010 Py3 only, but still supporting 2.7 until it is dropped (or manylinux1 drops 2.7 support ;) ). I really wanted cibuildwheel to match manylinux. I thought there was open discussion about it but @mayeut seems to have gone ahead anyway in #596.

@henryiii, I'm happy to discuss again the different options in #596 and I'll expand on that in that issue.

henryiii (approved):
This is a minimal required step to keep moving forward for 1.9.x. I think it's a no-brainer. :)

mayeut:
Let's merge then.

PR #595, opened 2021-02-14
Add manylinux_2_24 support

mayeut:
Add manylinux_2_24 support.

Builds on top of #594


henryiii:
What is the status of the aliasing of older manylinuxs? Would it make sense to allow the new names to used for the old images, too? Or is it pretty much a given that most projects will keep the old names and the new ones are mostly useful going forward (probably due to pip support)?

mayeut:
> What is the status of the aliasing of older manylinuxs? Would it make sense to allow the new names to used for the old images, too? Or is it pretty much a given that most projects will keep the old names and the new ones are mostly useful going forward (probably due to pip support)?

I think that, at the moment, I only know of `pip 20.3+`/`packaging 20.5+` supporting those aliases. For now, there are no plan to alias the `manylinux` images and there are no plans in `auditwheel` to support aliasing.

IMHO, most project will keep the old names (pip support being one reason). The new ones will probably only be used for PEP600 only wheels.

YannickJadoul (approved):
Looks good, at a quick glance! :-)

joerick:
Thanks all! Apologies for missing this, I have been crazy busy of late. it's probably worth a minor release, I reckon.

joerick:
Released as v1.10.0!

PR #604, opened 2021-02-19
Add time-machine as working example

adamchainz:


YannickJadoul:
@adamchainz Mind just adding this to the list at #485? We're making batches of these, these days, as it's more complicated than this PR (i.e., the docs weren't regenerated from this change in the yaml file).

henryiii:
I don't mind PRs either, we can just rerun the generation and then merge. Possibly could even add to the CI at some point.

YannickJadoul:
Oh... maybe the README is missing something, then, though?

> or send a PR, adding your information to `docs/data/projects.yml`.

Sorry, didn't know that was stated there.

YannickJadoul:
> I don't mind PRs either, we can just rerun the generation and then merge. Possibly could even add to the CI at some point.

Right, yes, that would make sense; it's not really indicated in the README that this needs to happen, so the more automatic it would be the better?

henryiii:
It's only a few lines for me to do it:

```bash
gh pr checkout 604
./bin/projects.py docs/data/projects.yml --readme README.md --auth $HOMEBREW_GITHUB_API_TOKEN
git add -u
git commit -m "docs: rerun projects.py"
git push
git switch -
```

(Unless someone leaves "master" as the branch name, then it's a headache...)

YannickJadoul:
> It's only a few lines for me to do it:

That's what I thought, yes. That's why I had suggested the #485 for a full batch. But this is easy enough as well :-)

YannickJadoul:
> (Unless someone leaves "master" as the branch name, then it's a headache...)

Even without GitHub CLI, it's reasonably easy to check out PRs, though: https://gist.github.com/piscisaureus/3342247

henryiii:
I prefer the one simple line, though. :) And it is more of a headache with a matching branch name by hand, too.

PS: you don't need to edit a file to do it by hand.

henryiii:
Oh, I see. That's a pretty nice trick, and Amos think affected by branch name! Filed away.

YannickJadoul:
The disadvantage is that you end up with a detached HEAD and that you don't have the upstream branch to push set (like GH CLI does), but still no need for a one-time remote :-)

adamchainz:
Thanks!

PR #606, opened 2021-02-21
docs: add PyPI field and a few more projects

henryiii:
Adding a few more from #485, also a new `pypi:` field to make #605 easier.

### DIFF
--- a/docs/data/projects.yml
+++ b/docs/data/projects.yml
@@ -2,6 +2,7 @@
 # name: required, printed name
 # gh: GitHub repo (eventually a url: could be added for non-github projects)
 # stars: GitHub repo (optional, if different from package, such as for Twisted)
+# pypi: The pypi name, if different from the GitHub package name
 # os: Operating system list, [windows, apple, linux] (optional)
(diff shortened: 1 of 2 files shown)


docs/data/projects.yml
```diff
@@ -2,6 +2,7 @@
 # name: required, printed name
 # gh: GitHub repo (eventually a url: could be added for non-github projects)
 # stars: GitHub repo (optional, if different from package, such as for Twisted)
+# pypi: The pypi name, if different from the github package name
```
henryiii, line 5:
```suggestion
# pypi: The pypi name, if different from the GitHub package name
```

henryiii:
We could in instead always use the pypi name as the "name".

YannickJadoul:
You don't want PyPI links for all? Because e.g. for my Parselmouth, there's also a difference between "Parselmouth" and `praat-parselmouth`.

joerick:
Seems like a good idea to me! I quite like keeping the separate name field, because some project pypi names are a little misleading, but I also wouldn't mind changing that if you prefer.

PR #609, opened 2021-02-26
Minor docs updates

joerick:
Fixing a few out-of-date things on the contributing page of the docs


henryiii:
(Failure of Gitlab CI has nothing to do with the PR, though maybe should be checked?)

PR #614, opened 2021-03-03
fix: macos11 universal naming corrected

henryiii:
Should fix the update script (as seen in #610)


henryiii:
Should be safe to merge so I can rerun the update trigger to see if it works.

joerick:
Thanks @henryiii ! Yes this does make sense, I was surprised by the name change also

PR #616, opened 2021-03-04
add search plugin

mattip:
Enable the search box in docs

xref mondeja/mkdocs-include-markdown-plugin#24

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -39,3 +39,4 @@ markdown_extensions:
 
 plugins:
   - include-markdown
+  - search


YannickJadoul (approved):
Nice, thanks for noticing and catching!!

Do you (by chance) know if we're missing out on other plugins that would be enabled by default?

joerick (approved):
Looks good! Thank you and well spotted, @mattip.

It looks like on https://www.mkdocs.org/user-guide/configuration/#plugins that `search` is the only one that's built-in to mkdocs.

YannickJadoul:
> It looks like on https://www.mkdocs.org/user-guide/configuration/#plugins that `search` is the only one that's built-in to mkdocs.

I did look, but couldn't immediately find a list and wanted to be sure I didn't overlook anything :-)

PR #623, opened 2021-03-18
mention apt-get for manylinux2_24

mattip:
add mentions of `apt-get` for manylinux2_24, also reflow the faq.md document for 80 columns


docs/options.md
```diff
@@ -354,14 +368,29 @@ CIBW_BEFORE_ALL: make -C third_party_lib
 CIBW_BEFORE_ALL_LINUX: yum install -y libffi-dev
 ```
 
+Note that manylinux2_24 builds occur inside a Debian9 docker, where
+manylinux2010 and manylinux2014 builds occur inside a CentOS one. So for
+`manylinux2_24` the `CIBW_BEFORE_ALL_LINUX` command must use `apt-get -y`
+instead.
+
```
mattip, line 375: This is the added blurb, it kind of got lost in the reflow.

henryiii (approved):

docs/options.md
henryiii: Yeah, GH's ignore whitespace doesn't work on reflowing things.

docs/options.md
```diff
@@ -446,16 +476,51 @@ CIBW_REPAIR_WHEEL_COMMAND_LINUX: "auditwheel repair --lib-sdir . -w {dest_dir} {
...
+Note that `auditwheel` detects the version of the `manylinux` standard in the
+Docker image through the `AUDITWHEEL_PLAT` environment variable, as
+`cibuildwheel` has no way of detecting the correct `--plat` command line
+argument to pass to `auditwheel` for a custom image. If a Docker image does not
+correctly set this `AUDITWHEEL_PLAT` environment variable, the
+`CIBW_ENVIRONMENT` option can be used to do so (e.g.,
+`CIBW_ENVIRONMENT='AUDITWHEEL_PLAT="manylinux2010_$(uname -m)"'`).
+
+Note that `manylinux2014`/`manylinux_2_24` don't support builds with Python 2.7
+- when building with `manylinux2014`/`manylinux_2_24`, skip Python 2.7 using
```
henryiii, line 522: Is this okay?

henryiii: Yep, looks fine on RtD.

docs/options.md
YannickJadoul: What about the whitespace? If we want to do this, we should probably be consistent (and do it in a different PR) ?

I can push a commit to minimize the diff, if you want?

YannickJadoul: Alright, I did so (together with a minor typo). Feel free to kick out the last commit (67903ae0a2e7e5d34f864ded03bd66ea8cd964ea) again, if you disagree :-)

YannickJadoul (approved):
Thanks, @mattip!

PR #624, opened 2021-03-20
Make GHA pre-commit check a precursor to main test

joerick:
This check was the same as the check performed by pre-commit.ci, except that the pre-commit.ci is a bit faster and can fix some issues itself, so I'm removing the GHA one in favour of the pre-commit.ci one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,8 +11,17 @@ on:
     # allow manual runs on branches without a PR
 
 jobs:
+  pre-commit:
+    name: Pre-commit checks (mypy, flake8, etc.)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


YannickJadoul:
GitLab just timed out. I don't know if reopening helps? Let's try.

henryiii:
Looks fine. I thought I'd added a manual stage check (check-manifest), but apparently I haven't so we could remove for now. Or I could add that check then modify this to only run the manual stage check.

Czaki:
In one of my project, I keep this step but use it as a pre requirements for other steps. In `cibuildwheel` there is a problem with the time of test execution so maybe instead remove it make it required for the next jobs.

henryiii:
That's not a bad idea. :)

henryiii:
I like this, and if/when I add a check to make sure the SDist is not missing files, it would likely only run in the GHA check.

henryiii (approved):

PR #626, opened 2021-03-22
style: apply some cleanups from Sourcery.ai

henryiii:
Applying some of the fixes (slightly cleaned up) as suggested  by running Sourcery.ai. See the original PR with reasoning on each change here: https://github.com/henryiii/cibuildwheel/pull/1. It _mostly_ did a pretty good job, though it liked lists over sets, and didn't handle the static type checks over sets very well. But more than 50% were nice cleanups.

Note that it produces "Black" style output for a change. PyUpgrade does too, IIRC.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -367,9 +367,10 @@ def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
     for option_name in ['test_command', 'before_build']:
         option_value = getattr(build_options, option_name)
 
-        if option_value:
-            if '{python}' in option_value or '{pip}' in option_value:
-                warnings.append(option_name + ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
+        if option_value and (
+            '{python}' in option_value or '{pip}' in option_value
+        ):
+            warnings.append(option_name + ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
 
     return warnings
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -367,9 +367,7 @@ def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
     for option_name in ['test_command', 'before_build']:
         option_value = getattr(build_options, option_name)
 
-        if option_value and (
-            '{python}' in option_value or '{pip}' in option_value
-        ):
+        if option_value and ('{python}' in option_value or '{pip}' in option_value):
             warnings.append(option_name + ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
 
     return warnings
--- a/cibuildwheel/bashlex_eval.py
+++ b/cibuildwheel/bashlex_eval.py
@@ -85,9 +85,7 @@ def evaluate_nodes_as_compound_command(nodes: Sequence[bashlex.ast.node], contex
         if node.kind == 'command':
             result += evaluate_command_node(node, context=context)
         elif node.kind == 'operator':
-            if node.op == ';':
-                pass
-            else:
+            if node.op not in {';'}:
                 raise ValueError(f'Unsupported bash operator: "{node.op}"')
         else:
             raise ValueError(f'Unsupported bash node in compound command: "{node.kind}"')
--- a/cibuildwheel/bashlex_eval.py
+++ b/cibuildwheel/bashlex_eval.py
@@ -85,7 +85,7 @@ def evaluate_nodes_as_compound_command(nodes: Sequence[bashlex.ast.node], contex
         if node.kind == 'command':
             result += evaluate_command_node(node, context=context)
         elif node.kind == 'operator':
-            if node.op not in {';'}:
+            if node.op != ';':
                 raise ValueError(f'Unsupported bash operator: "{node.op}"')
         else:
             raise ValueError(f'Unsupported bash node in compound command: "{node.kind}"')
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -257,23 +257,29 @@ def setup_python(python_configuration: PythonConfiguration,
                 # set ARCHFLAGS otherwise CPython sets it to `-arch x86_64`
                 env.setdefault('ARCHFLAGS', '-arch arm64 -arch x86_64')
 
-    if python_configuration.identifier.endswith('arm64') or python_configuration.identifier.endswith('universal2'):
-        if get_macos_version() < (10, 16) and 'SDKROOT' not in env:
-            # xcode 12.2 or higher can build arm64 on macos 10.15 or below, but
-            # needs the correct SDK selected.
-            sdks = get_macos_sdks()
+    if (
+        (
+            python_configuration.identifier.endswith('arm64')
+            or python_configuration.identifier.endswith('universal2')
+        )
+        and get_macos_version() < (10, 16)
+        and 'SDKROOT' not in env
+    ):
+        # xcode 12.2 or higher can build arm64 on macos 10.15 or below, but
+        # needs the correct SDK selected.
+        sdks = get_macos_sdks()
 
-            # Different versions of Xcode contain different SDK versions...
-            # we're happy with anything newer than macOS 11.0
-            arm64_compatible_sdks = [s for s in sdks if not s.startswith('macosx10.')]
+        # Different versions of Xcode contain different SDK versions...
+        # we're happy with anything newer than macOS 11.0
+        arm64_compatible_sdks = [s for s in sdks if not s.startswith('macosx10.')]
 
-            if not arm64_compatible_sdks:
-                log.warning(unwrap('''
+        if not arm64_compatible_sdks:
+            log.warning(unwrap('''
                     SDK for building arm64-compatible wheels not found. You need Xcode 12.2 or later
                     to build universal2 or arm64 wheels.
                 '''))
-            else:
-                env.setdefault('SDKROOT', arm64_compatible_sdks[0])
+        else:
+            env.setdefault('SDKROOT', arm64_compatible_sdks[0])
 
     log.step('Installing build tools...')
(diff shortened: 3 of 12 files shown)


test/test_dependency_versions.py
```diff
@@ -40,11 +40,10 @@
...
 
-    versions = {}
-    for package, version in re.findall(VERSION_REGEX, constraint_file_text):
-        versions[package] = version
-
-    return versions
+    return {
+        package: version
+        for package, version in re.findall(VERSION_REGEX, constraint_file_text)
+    }
```
henryiii, line 46: This was a pretty clever rewrite.

cibuildwheel/__main__.py
```diff
@@ -367,9 +367,10 @@ def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
     for option_name in ['test_command', 'before_build']:
         option_value = getattr(build_options, option_name)
 
-        if option_value:
-            if '{python}' in option_value or '{pip}' in option_value:
-                warnings.append(option_name + ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
+        if option_value and (
+            '{python}' in option_value or '{pip}' in option_value
+        ):
```
YannickJadoul, line 372: Given the length of the next line, this could just as well be on one line?

cibuildwheel/bashlex_eval.py
```diff
@@ -85,9 +85,7 @@ def evaluate_nodes_as_compound_command(nodes: Sequence[bashlex.ast.node], contex
         if node.kind == 'command':
             result += evaluate_command_node(node, context=context)
         elif node.kind == 'operator':
-            if node.op == ';':
-                pass
-            else:
+            if node.op not in {';'}:
```
YannickJadoul, line 88: Or `node.op != ';'`, for now? I don't know if/how this would expand

cibuildwheel/macos.py
```diff
@@ -257,23 +257,29 @@ def setup_python(python_configuration: PythonConfiguration,
...
-            if not arm64_compatible_sdks:
-                log.warning(unwrap('''
+    if (
+        (
+            python_configuration.identifier.endswith('arm64')
+            or python_configuration.identifier.endswith('universal2')
+        )
+        and get_macos_version() < (10, 16)
+        and 'SDKROOT' not in env
+    ):
```
YannickJadoul, line 267: This might stick slightly closer to the previous meaning of checking this in two steps/levels?

```suggestion
    building_arm64 = python_configuration.identifier.endswith('arm64') or python_configuration.identifier.endswith('universal2')
    if building_arm64 and get_macos_version() < (10, 16) and 'SDKROOT' not in env:
```

unit_test/main_tests/main_platform_test.py
```diff
@@ -123,12 +123,10 @@ def test_archs_platform_native(platform, intercepted_build_args, monkeypatch):
     main()
     build_options = intercepted_build_args.args[0]
 
-    if platform == 'linux':
+    if platform in {'linux', 'macos'}:
```
YannickJadoul, line 126: We might want to think about this, since we're used seeing (and checking for!) the 3-part if of "platform is Linux/mac/Windows"?

Then again, now it's Unix vs. Windows, so that also makes kind of sense?

YannickJadoul (approved):
Wow, I'm quite impressed by some of these, indeed!

A few thoughts regarding how some constructs map to the more high-level/human reasoning, but overall, this looks good! :-)

YannickJadoul:
> Note that it produces "Black" style output for a change.

Since we haven't completely fixed a style, Black is also a valid style? ;-)

YannickJadoul:
> though it liked lists over sets

I guess `if platform in ['linux', 'macos']:` is in general slightly faster since a list is a simpler structure than a set? Though, eh, it shouldn't matter anything, so let's go for the one that makes most intuitive sense.

henryiii:
> > though it liked lists over sets
> 
> I guess `if platform in ['linux', 'macos']:` is in general slightly faster since a list is a simpler structure than a set? Though, eh, it shouldn't matter anything, so let's go for the one that makes most intuitive sense.

See https://github.com/sourcery-ai/sourcery/issues/123. Even for small numbers of items, sets are faster because they are designed for fast containership testing. There are two requirements to that - it has to be an in-place literal (if you assign to a variable, then the set construction time is at runtime on each pass, and it's actually a set rather than a frozenset), and it has to have hashable entries (standard set requirement, that's how the search is fast). Also, the first item in a list can beat the set testing, but the second item or more, or a missing item, is slower, and the timing is constant for the set. Sets were designed for containership testing; lists are just linear structures that happen to support `in` as a shortcut.

Mostly, though, regardless of the performance, the set is the correct language for what we are doing. Is the platform in the set 'linux', 'macos'? The order is not important, it would be invalid to list the same item twice, etc. It's logically a set, not a list.

unit_test/main_tests/main_platform_test.py
henryiii: In the regular code, I use completeness checking to ensure we have all three, and (I am pretty sure) MyPy can't do completeness checking if you use a set (at least, currently). But here, we aren't checking as carefully with MyPy, and this combines the similar platforms, keeping the tests a little simpler.

cibuildwheel/bashlex_eval.py
henryiii:
```suggestion
            if node.op != ';':
```

I was thinking that if another op came up, it would be passed too, but maybe only `;` should ever pass.

cibuildwheel/__main__.py
henryiii: This doesn't know that we've increased black's line width to 120. This would be the normal way to do it if running with 80/88 line width. It makes it much harder to miss the `()`, too.

```suggestion
        if option_value and ('{python}' in option_value or '{pip}' in option_value):
```

cibuildwheel/macos.py
henryiii: I think that's a nice improvement (and the correct way to clean up black, too, by the way :) )

cibuildwheel/bashlex_eval.py
YannickJadoul: Yes, you're right, but I think this has been reasonably stable for a long time, so I reckoned this was slightly premature (and not just a rewrite?). I'm fine with keeping this as well, btw.

cibuildwheel/macos.py
YannickJadoul: I thought so. Learned from the best ;-)

unit_test/main_tests/main_platform_test.py
YannickJadoul: Sure, didn't notice this was in the tests, actually. So then it's definitely OK :-)

YannickJadoul:
> See [sourcery-ai/sourcery#123](https://github.com/sourcery-ai/sourcery/issues/123). Even for small numbers of items, sets are faster because they are designed for fast containership testing.

Ooooh, that's interesting! I had overlooked the possibility that these literals are just stored at bytecode generation time (I'd expected that creating a set at runtime is slower than a list at runtime and would undo the gain of the fast access).

> Mostly, though, regardless of the performance, the set is the correct language for what we are doing. Is the platform in the set 'linux', 'macos'? The order is not important, it would be invalid to list the same item twice, etc. It's logically a set, not a list.

More or less agreed there. That's what I meant by

> Though, eh, it shouldn't matter anything, so let's go for the one that makes most intuitive sense.

I was just trying to figure out if there would be a good reason for Sourcery.ai, but no, then.

YannickJadoul:
> (I'd expected that creating a set at runtime is slower than a list at runtime and would undo the gain of the fast access).

For the record, at least that intuition works, without Python's optimization of the stored literal. Here's your code from sourcery-ai/sourcery#123, slightly extended ;-)

```pycon
>>> from timeit import Timer
>>> def fn(value):
...     return value in ('A', 'B', 'C', 'D', 'E')
... 
>>> def fnS(value):
...     return value in {'A', 'B', 'C', 'D', 'E'}
... 
>>> def fnV(value):
...     x = ('A', 'B', 'C', 'D', 'E')
...     return value in x
... 
>>> def fnSV(value):
...     x = {'A', 'B', 'C', 'D', 'E'}
...     return value in x
... 
>>> Timer("fn('E')", setup="from __main__ import fn as fn").autorange()
(10000000, 1.0219394589985313)
>>> Timer("fn('E')", setup="from __main__ import fnS as fn").autorange()
(10000000, 0.62776606099942)
>>> Timer("fn('E')", setup="from __main__ import fnV as fn").autorange()
(10000000, 1.065856844001246)
>>> Timer("fn('E')", setup="from __main__ import fnSV as fn").autorange()
(1000000, 0.20328924400018877)
```

cibuildwheel/bashlex_eval.py
joerick: henry's suggestion seems good to me. i guess maybe i was thinking that other operations would have other behaviours, but I guess that never happened :)

unit_test/docker_container_test.py
```diff
@@ -94,12 +94,12 @@ def test_binary_output():
...
         data = bytes(output, encoding='utf8', errors='surrogateescape')
 
-        for i in range(0, 512):
+        for i in range(512):
             assert data[i] == i % 256
 
         # check that environment variables can carry binary data, except null characters
         # (https://www.gnu.org/software/libc/manual/html_node/Environment-Variables.html)
-        binary_data = bytes(n for n in range(1, 256))
+        binary_data = bytes(range(1, 256))
```
joerick, line 102: I'm less sure about this change, myself. Clearly they're equivalent, but somehow `binary_data = bytes(n for n in range(1, 256))` is more obviously a buffer containing every byte value from 1 to 256 than `bytes(range(1, 256))`, which looks like it was written by a computer :P
```suggestion
        binary_data = bytes(n for n in range(1, 256))
```

joerick (commented):
This is kinda neat! I think there are a couple of corrections to go in from the comments, but then this LGTM.

unit_test/docker_container_test.py
henryiii: I like the simpler one that does not make an extra generator, but okay. Given the comment above especially, I don't `bytes(range(` could be misunderstood.

joerick (approved):

PR #629, opened 2021-03-25
Try out PyPy 7.3.4rc2

YannickJadoul:
Still in draft, but preparing for when a release is made. The Python 3.6-compatible version of PyPy was dropped﻿, as the Python 3.7-compatibility was deemed completed (cfr. https://doc.pypy.org/en/latest/release-v7.3.4.html).

I still need to get an updated manylinux docker image.


YannickJadoul:
> The Python 3.6-compatible version of PyPy was dropped﻿, as the Python 3.7-compatibility was deemed completed (cfr. https://doc.pypy.org/en/latest/release-v7.3.4.html).

Something to think about, btw: since this is just a patch version increase of PyPy, do we actually want to keep building `pp36-pypy36_pp73` wheels with 7.3.3? I'll ask the same question when submitting a PR for PyPy's manylinux images.

henryiii:
The targets have changed in a patch version before (3.7 got added). I think it's fine to change on PyPy - 3.6 (due to NumPy dropping it) is rapidly getting harder to use.

Wait, are there no 32 bit binaries for Windows? They should have both, not drop 32 for 64? I know 32-bit CPython was rather popular (maybe heavily due to the fact it was the default download choice until very recently more than due to OS requirements).

YannickJadoul:
> The targets have changed in a patch version before (3.7 got added). I think it's fine to change on PyPy - 3.6 (due to NumPy dropping it) is rapidly getting harder to use.

Right, but adding something is more backwards compatible, ofc. But yeah, fair enough; I guess it's PyPy's decision to release it as 7.3.4 (and ofc not worth breaking the ABI version for).

> Wait, are there no 32 bit binaries for Windows? They should have both, not drop 32 for 64? I know 32-bit CPython was rather popular (maybe heavily due to the fact it was the default download choice until very recently more than due to OS requirements).

I had already popped onto IRC to ask, yes. Nothing that seems to secret, so I'll post it here:

> ```
> <YannickJadoul> Hi all. I'm trying out PyPy 7.3.4rc1 in cibuildwheel, for the interested: https://github.com/joerick/cibuildwheel/pull/629
> <YannickJadoul> Quick question: there's no win32 anymore? PyPy won't provide both 32-bit as well as 64, just like CPython still seems to do?
> <mattip> hmm. Just because we can, should we? It is a small maintenance burden
> <YannickJadoul> Ah, no, not saying that. Just curious, if the switch will be total
> <YannickJadoul> Important to update our docs and think about releases and version numbers, etc
> <mattip> unless there is a user who says otherwise: yes, the switch will be total
> <YannickJadoul> OK, thanks!
> ```

I'll just ping @mattip, in case you'd like to convince him about the popularity of 32-bit :-)

mattip:
I am willing to be convinced to release 32-bit windows, but it would have to be by a user who can explain how it is worth the (admittedly small) maintenance burden).

YannickJadoul:
@Czaki, I don't remember the details, but I'm seeing "PyPy patch not applied". Do you know by heart whether we can remove that patch and workaround, or should I check the PR that introduced it again?

Czaki:
Where did you see this? In log?

Czaki:
Patch are here: https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/resources/pypy_venv_27.patch

If I good remember there was a problem that in virtualenv compiler cannot find "Python.h"

YannickJadoul:
> maybe current version fix it already.

Yeah, didn't you submit a PR/issue to PyPy? Any idea whether that got merged/resolved?

YannickJadoul:
OK, I stopped being lazy for a second and looked it up myself: https://github.com/joerick/cibuildwheel/issues/501#issuecomment-751369614

YannickJadoul:
So, something's wrong with the headers of [pypy2.7-v7.3.4rc1-win64.zip](https://downloads.python.org/pypy/pypy2.7-v7.3.4rc1-win64.zip); reported on IRC already. Apart from that, things seem to be working :-)

mattip:
The headers are produced as part of an optional module in PyPy called cpyext that provides the C-API compatibility layer. We do not compile pypy2.7 on win64 with cpyext. Would it be hard to do `if pypy and windows and python2.7: <don't check compiling a library>` ? Let's see if there are any people who this is critical for them, then we can discuss whether to ship a 32-bit windows version (which does have cpyext) or to get cpyext going on win64-python2.7

<details>

I started work on this in the win64-cpyext-default branch, but then someone pointed out that tp_hash and hashes in general on python2.7 are `long`, which are 4-bytes even on win64. This messes with some of the RPython assumptions about `sizeof(long)` so it is a bit of a mess: we need to be careful not to mix `long` and pointers

Czaki:
> The headers are produced as part of an optional module in PyPy called cpyext that provides the C-API compatibility layer. We do not compile pypy2.7 on win64 with cpyext. Would it be hard to do `if pypy and windows and python2.7: <don't check compiling a library>` ? Let's see if there are any people who this is critical for them, then we can discuss whether to ship a 32-bit windows version (which does have cpyext) or to get cpyext going on win64-python2.7

But the main task of `cibuildwheel` is compiling libraries. And most of the extensions need `Python.h` header.

YannickJadoul:
> Would it be hard to do `if pypy and windows and python2.7: <don't check compiling a library>` ?

I guess we could plainly not offer PyPy Python 2.7 on Windows, then? I think `cibuildwheel` is only useful in case of binary extensions, right?

Oh, no, wait. There's still `cffi`, of course. But `cibuildwheel` doesn't really have a way to test what kind of extension is getting compiled :-/

> This messes with some of the RPython assumptions about `sizeof(long)` so it is a bit of a mess: we need to be careful not to mix `long` and pointers

(sorry to hear about that btw; sounds like a horrible issue :-/ )

henryiii:
Biased for pybind11, but not having Python.h kills pybind11 extensions.

Not the best solution quite yet, but I guess needs to be mentioned: dropping Python 2 support would fix this issue...

> long, which are 4-bytes even on win64

ROOT has been fighting with this for a while, still don't have a 64-bit windows build due to it. It is clearly written in the standard that long does not have to be 64-bits, but that hasn't stopped assumptions based on Linux/macOS or 32-bit Windows.

mattip:
> Biased for pybind11, but not having Python.h kills pybind11 extensions.

Right. How many projects have a user base that needs testing/releasing with PyPy on windows for python2.7? Before  I spend a few days solving this, I would like to know it is going to be used.

YannickJadoul:
Yeah, dropping 2.7 would of course also work. Or dropping PyPy 2.7 on Windows, for now? (or probably just everywhere to be consistent)
Apart from not being able to test it, I'm mostly concerned by offering PyPy 2.7 on Windows but having users' builds fail because they expect the C API to be there. I wouldn't be surprised if that resulted in a bunch of issues blaming `cibuildwheel` for offering `pp27-win_amd64`.

henryiii:
Honestly, I don't know how much of a user base PyPy2 has. With NumPy requiring 3.7 now, it's likely a bit limited. And 2.7 on Windows is traditionally small due to other compiler reasons (and the fact it's easier to upgrade on Windows than having a build-in Python 2 on Linux/macOS). I do think we are going to be dropping PyPy2 when dropping CPython 2.7 in 2.0. But that's not quite here yet.

YannickJadoul:
So, we're still failing on PyPy 2.7's missing headers on Windows. @joerick, where would you like to go with this? Still build 2.7 32-bit for now, using 7.3.3? Or just drop PyPy 2.7 on Windows, or ... ?

mattip:
I would vote for dropping it until you hear from users.

Reasoning: If you drop PyPy2.7 for windows, you would quickly hear from any users who need it, and then PyPy could invest the effort to support cpyext for that verison. Any other work-around will encourage users to stay with an older version.

README.md
```diff
@@ -30,9 +30,8 @@ What does it do?
 | CPython 3.7     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.8     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.9     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| PyPy 2.7 v7.3.3 | ✅ |    |    | ✅  | ✅ |    |    |    |   |
-| PyPy 3.6 v7.3.3 | ✅ |    |    | ✅  | ✅ |    |    |    |   |
-| PyPy 3.7 (beta) v7.3.3 | ✅ | |  | ✅  | ✅ |    |    |   |    |
+| PyPy 2.7 v7.3.4 | ✅ |    |    |    |    | ✅ |    |    |   |
+| PyPy 3.7 v7.3.4 | ✅ |    |    | ✅  |    | ✅ |    |    |   |
```
nulano, line 34:
```suggestion
| PyPy 3.7 v7.3.4 | ✅ |    | ✅ |    |    | ✅ |    |    |   |
```
I think this is in the wrong column?

cibuildwheel/resources/build-platforms.toml
```diff
@@ -63,7 +61,5 @@ python_configurations = [
   { identifier = "cp38-win_amd64", version = "3.8.9", arch = "64" },
   { identifier = "cp39-win32", version = "3.9.4", arch = "32" },
   { identifier = "cp39-win_amd64", version = "3.9.4", arch = "64" },
-  { identifier = "pp27-win32", version = "2.7", arch = "32", url = "https://downloads.python.org/pypy/pypy2.7-v7.3.3-win32.zip" },
-  { identifier = "pp36-win32", version = "3.6", arch = "32", url = "https://downloads.python.org/pypy/pypy3.6-v7.3.3-win32.zip" },
-  { identifier = "pp37-win32", version = "3.7", arch = "32", url = "https://downloads.python.org/pypy/pypy3.7-v7.3.3-win32.zip" },
+  { identifier = "pp37-win_amd64", version = "3.7", arch = "32", url = "https://downloads.python.org/pypy/pypy3.7-v7.3.4rc2-win64.zip" },
```
nulano, line 64:
```suggestion
  { identifier = "pp37-win_amd64", version = "3.7", arch = "64", url = "https://downloads.python.org/pypy/pypy3.7-v7.3.4rc2-win64.zip" },
```
Update arch?

YannickJadoul:
@mattip, sure. This seems to be working. The downside is that users also won't be able to build cffi wheels for PyPy 2.7, ofc. After all, this is just a decision to be made by PyPy, I'd say. (Unless you want us to support PyPy 2.7 (for cffi wheels), without C API headers; then it'd be up to @joerick to decide whether that's risking a lot of complaints about failing builds.)

mattip:
@YannickJadoul let's give it a shot. If it turns out lots of people want to package for pypy2.7 on windows, we can think harder about it.

README.md
YannickJadoul: How do you mean? That's changing from 32-bit Windows (4th column) to 64-bit Windows (3rd column) for PyPy, right? Or am I miscounting?

cibuildwheel/resources/build-platforms.toml
YannickJadoul: Oh, great point! So we don't use this? Or we don't check it?
Let me double check.

YannickJadoul: Done (and added a minor extra check), thanks!

README.md
nulano: If I'm counting right, the PR puts the checkmark in the 32-bit column, so my suggestion above is to put it in the 64-bit column (since 7.3.4 is win64 only).

nulano: Actually, the manylinux column might be wrong as well, so maybe the following is right?
```suggestion
| PyPy 3.7 v7.3.4 | ✅ |    | ✅  |    | ✅ |    |    |   |    |
```
You can see the rendered table here, showing that both checkmarks are in the 32-bit columns: https://github.com/joerick/cibuildwheel/blob/46d81290903d0c23d7e74e487e9478ea04d21425/README.md

YannickJadoul: Oh, wait, I was looking at your suggestion! Seems I move the wrong one (manylinux instead of Windows)!

Thanks for catching!

bin/update_pythons.py
```diff
@@ -115,8 +115,8 @@ def __init__(self, arch_str: ArchStr):
         self.arch = arch_str
 
     def update_version_windows(self, spec: Specifier) -> ConfigWinCP:
```
YannickJadoul, line 117: This'll probably take some more changes. But I'm guessing we'll have to wait for a final release of 7.3.4 to fix it.

README.md
YannickJadoul: Should both be fixed, now; thanks! :-)

joerick:
I'm cool to drop 2.7 on Windows, too. We're dropping 2.7 as a whole in #596 anyway, so it probably won't be a bit issue :)

YannickJadoul:
Waiting for #16 to pull this out of "draft"

YannickJadoul:
For the record:

> It turns out there is a problem with the win64 release: the c-extension platform name is win32. It should be win_amd64.
> 
> This was reported by cgohlke in https://foss.heptapod.net/pypy/pypy/-/issues/3443.
> 
> 
> I will be releasing a 7.3.5 bugfix soon. In the meantime, please do not upload win64 wheels to PyPI built with the 7.3.4 release: they will be tagged as appropriate for win32 which will cause problems for people who have locally built a 32-bit windows pypy3.7 and use "pip install".
> 
> 
> Note we did not officially release a 32-bit windows, so the damage so far might be minimal.
> 
> 
> Matti

See https://mail.python.org/pipermail/pypy-dev/2021-April/016150.html

So we probably ought to be skipping 7.3.4 and immediately pin 7.3.5?

mattip:
That makes sense, although it may be a week or two until 7.3.5 is released.

YannickJadoul:
Thanks, @mattip! I don't think we're in a huge hurry to merge this, currently?

henryiii:
Not if it's broken. :)

YannickJadoul:
Seems I broke GitHub by pushing the Windows-only changes to a new branch. Huh. Anyway, I'll open a new draft with those leftover changes that weren't in #666

PR #641, opened 2021-04-14
Bug fix: windows does not have 'which' command

MagicJack:
Windows does not have 'which' command, it should be 'where' command.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -295,7 +295,7 @@ def build(options: BuildOptions) -> None:
                 ])
 
                 # check that we are using the Python from the virtual environment
-                call(['which', 'python'], env=virtualenv_env)
+                call(['where', 'python'], env=virtualenv_env)
 
                 if options.before_test:
                     before_test_prepared = prepare_command(


YannickJadoul:
What's the context? What's the bug, or how did this bite you?

This has been working for a long while, now, I believe, so (even though I know you're right about `which` and `where`) I'm slightly confused over why it has been working (at least +1.5 years: https://github.com/joerick/cibuildwheel/commit/40968897a8e7c7fcdc5a5ff0972894e605dc7585).

MagicJack:
This bug was found when I trying to build 'confluent-kafka-python' wheel package locally.

If I run the build batch file from windows command console, it always stop at calling command 'which python'.  So, I brows their code, and found this bug form 'cibuildwheel'.  Also, I found other lines of code in 'windows.py' use 'where' command instead of 'which'.  So, I change it, and it did work on building 'confluent-kafka-python' wheel package.

Czaki:
> This bug was found when I trying to build 'confluent-kafka-python' wheel package locally.

It is not recommended to use cibuildwheel locally on non-linux system (Linux builds use docker) as cibuildwheel install python system side. 


> And, of cause, if I run this build batch from window's git-bash, this bug no longer exist due to git for windows did have 'which' command in it.

It looks like all CI system which are tested has bash installed.

joerick:
It's interesting that we've not hit this bug before, I guess that many Windows installs have `which` available, but perhaps some don't. I think I've come across `where` before, it's almost the same, but I think that it can return multiple paths rather than just a single one. But, that doesn't matter for this usage. Actually, I see `where` is already used around line 150, so this change seems good to me.

YannickJadoul:
> Actually, I see `where` is already used around line 150, so this change seems good to me.

Nope, I agree. Just trying to understand why we haven't run into it before and why tests pass! Maybe something to do with powershell?

And we're not digesting the output from `which`, right? I think `where` has a slightly different output, but this is just to log.

YannickJadoul (approved):

YannickJadoul:
Thanks, @MagicJack!

henryiii:
Most CI Windows runners are set up to be as close to Unix as possible, even when not using bash. For example, GHA/Azure added the `python3` command when I asked for it, just to improve the similarity between Windows and Unix runners. But agree, this is more technically correct, and a good change.

PR #649, opened 2021-04-22
ci: fix choco usage

henryiii:
Hoping to fix the issues seen in current CI runs. At the very least, we will follow the style guide linked below.

Following https://docs.chocolatey.org/en-us/choco/commands/#scripting-integration-best-practices-style-guide

* Avoid `-f` in scripts
* Always use `-y` in scripts
* Use `upgrade` instead of `install` in scripts
* Use `--limit-output` for more parsable output (should turn off progress bar as well, hopefully)

The issue with vcpython27 might be temporary, caused by the -f, or possibly permanent since it has been untouched since 2016.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,11 +39,6 @@ jobs:
       run: |
         python -m pip install -e ".[dev]"
 
-    - name: Install Visual C++ for Python 2.7
-      if: runner.os == 'Windows'
-      run: |
-        choco upgrade vcpython27 -y
-
     - name: Get PR labels
       id: pr-labels
       uses: joerick/pr-labels-action@v1.0.6
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Install Visual C++ for Python 2.7
       if: runner.os == 'Windows'
       run: |
-        choco install vcpython27 -f -y
+        choco upgrade vcpython27 -y
 
     - name: Get PR labels
       id: pr-labels
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,9 @@ jobs:
       os: windows
       language: shell
       before_install:
-       - choco install python3 --version 3.6.8 --no-progress -y
-       # Update root certificates to fix SSL error; see http://www.chawn.com/RootCerts.htm
-       - powershell "md C:\temp\certs; CertUtil -generateSSTFromWU C:\temp\certs\RootStore.sst; Get-ChildItem -Path C:\\temp\certs\Rootstore.sst | Import-Certificate -CertStoreLocation Cert:\\LocalMachine\\Root\\ | out-null"
+        - choco upgrade python3 -y --version 3.6.8 --limit-output
+        # Update root certificates to fix SSL error; see http://www.chawn.com/RootCerts.htm
+        - powershell "md C:\temp\certs; CertUtil -generateSSTFromWU C:\temp\certs\RootStore.sst; Get-ChildItem -Path C:\\temp\certs\Rootstore.sst | Import-Certificate -CertStoreLocation Cert:\\LocalMachine\\Root\\ | out-null"
       env:
         - PYTHON=C:\\Python36\\python
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ What does it do?
 | PyPy 3.6 v7.3.3 | ✅ |    |    | ✅  | ✅ |    |    |    |   |
 | PyPy 3.7 (beta) v7.3.3 | ✅ | |  | ✅  | ✅ |    |    |   |    |
 
-<sup>¹ Not supported on Travis</sup><br>
+<sup>¹ Only using a workaround with a newer compiler; Microsoft removed the 2008 compiler for Python 2.7 in April 2021.</sup><br>
 
 - Builds manylinux, macOS 10.9+, and Windows wheels for CPython and PyPy
 - Works on GitHub Actions, Azure Pipelines, Travis CI, AppVeyor, CircleCI, and GitLab CI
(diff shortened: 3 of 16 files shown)


henryiii:
Looks to me that Microsoft has finally killed of VS for Python 2.7?

https://www.microsoft.com/en-us/download/404Error.aspx

joerick:
Wow, yeah you might be right! But it's very abrupt... also the package is still in the chocolately 
directory which is confusing: https://community.chocolatey.org/packages/vcpython27 . It might be a mistake / temporary issue.

By the way @henryiii, how was your style normalisation work coming along? This might be a nice time to merge the drop 2.7 branch #596 ! :)

henryiii:
Let's give it a little time; not sure who to contact about it but it would be nice to get confirmation if this was intentional. The chocolaty package is not "official", so it isn't likely to be removed in sync with the old download. I can't find any statement saying they planned to drop the download, but the download is still missing.

If we do have to drop this old workaround, we need to drop it on the v1 branch anyway so that our CI isn't dead for supporting the LTS version. My secret target date for 596 is when [3.5 gets dropped from manylinux (May 1)](https://github.com/pypa/manylinux/issues/995) so we can remove both at the same time. Sadly, manylinux1 will keep supporting both, so v1 can stay as is.

First batch (style in bin) is either done or almost done, then I'll do `cibuildwheel`. Assume we really don't care that much about test/unit_test style and are fine to accept whatever gets produced there.

henryiii:
I've contacted Microsoft support, and:

> I'm very sorry Henry, but the Visual Studio for Python 2.7 is not available for download since it is currently not supported anymore.

It also says on the VS page:

> Visual Studio 2008 reached end of support on April 10, 2018. To aid the discovery of the latest downloads, the links are retained currently, but may be removed in the future.

So maybe someone will put pressure on them to restore it and it might show up again (I'm not going to do that!), but currently the plan is that it's gone and will stay that way.

joerick:
No idea what happened with the Azure run, it said 'Cancelled'. I've restarted it...

henryiii:
Passing. I think this should be rebased and merged, just in case this is restored or fixed and someone wants to revert just the second commit some day.

joerick (approved):
Yep. Let's do it. So we're saying here that you _need_ https://github.com/ilammy/msvc-dev-cmd to build 2.7 on Windows?

henryiii:
You need to "override" the default compiler in order to build for 2.7 on Windows. If you override with the original 2.7 2008 compiler, you can still build 2008 extensions (of course, you need to find a copy somewhere). If you choose a newer one, you have to be a little careful in how you use Python (tools like pybind11 are), and you have to have the 2015 redistributable. Since the 2008 compiler is really, really old, I think pretty much the norm on Windows is to override and use the 2015 compiler, or just to not support 2.7 on Windows at all.

joerick:
Gotcha. Just figuring out how we might document this. I'm thinking an FAQ entry like this. Any thoughts?

---

## Windows and Python 2.7

Building 2.7 extensions on Windows is difficult, because the VS2008 compiler that was used for the original Python compilation is difficult to find these days. Most people choose to not build Windows 2.7 wheels, or to override the compiler to something more modern.

To override, you need to have a modern compiler toolchain activated, and set `DISTUTILS_USE_SDK=1` and `MSSdk=1`. For example, on Github Actions, you would add these steps:

```yaml
    - name: Prepare compiler environment for Windows
      if: runner.os == 'Windows'
      uses: ilammy/msvc-dev-cmd@v1
      with:
        arch: x64

    - name: Set Windows environment variables
      if: runner.os == 'Windows'
      shell: bash
      run: |
        echo "DISTUTILS_USE_SDK=1" >> $GITHUB_ENV
        echo "MSSdk=1" >> $GITHUB_ENV
```

Note that changing to a more modern compiler will mean your wheel picks up a runtime dependency to a different [Visual C++ Redistributable](https://support.microsoft.com/en-us/topic/the-latest-supported-visual-c-downloads-2647da03-1eea-4433-9aff-95f26a218cc0).

henryiii:
Building 2.7 extensions on Windows is difficult, because the VS 2008 compiler that was used for the original Python compilation was discontinued in 2018, and has since been removed from Microsoft's downloads. Most people choose to not build Windows 2.7 wheels, or to override the compiler to something more modern.

...

cibuildwheel will not try to build 2.7 on Windows unless it detects that the above two variables are set. Note that changing to a more modern compiler will mean your wheel picks up a runtime dependency to a different [Visual C++ Redistributable](https://support.microsoft.com/en-us/topic/the-latest-supported-visual-c-downloads-2647da03-1eea-4433-9aff-95f26a218cc0). You also need to be a bit careful in designing your extension; some major binding tools like pybind11 do this for you.

joerick:
Thanks @henryiii , I'm adding this to #650 now.

> You also need to be a bit careful in designing your extension

Any hints on this? Or maybe a link? I'm wondering what a reader would do with this tantalising quantum of wisdom ;)

henryiii:
https://pybind11.readthedocs.io/en/stable/faq.html#working-with-ancient-visual-studio-2008-builds-on-windows

joerick:
🙌 perfect

YannickJadoul:
> https://pybind11.readthedocs.io/en/stable/faq.html#working-with-ancient-visual-studio-2008-builds-on-windows

Didn't we already have a link to that, somewhere? Probably in the "Python 2.7 on Windows" section.

joerick:
You are right - https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#windows-and-python-27 

The section in cpp-standards goes into a bit more depth, I'll link there from the FAQ.

YannickJadoul:
> The section in cpp-standards goes into a bit more depth, I'll link there from the FAQ.

Can't have too many links from different places for this, indeed. Apparently it's hidden enough for the maintainers themselves to miss it ;-)

PR #650, opened 2021-04-23
Docs updates

joerick:
A few little updates and tweaks to docs, to help readability and keep things up-to-date.


docs/options.md
```diff
@@ -90,20 +90,31 @@ Default: `auto`
 
 `auto` will auto-detect platform using environment variables, such as `TRAVIS_OS_NAME`/`APPVEYOR`/`CIRCLECI`.
 
-For `linux` you need Docker running, on macOS or Linux. For `macos`, you need a Mac machine, and note that this script is going to automatically install MacPython on your system, so don't run on your development machine. For `windows`, you need to run in Windows, and `cibuildwheel` will install required versions of Python to `C:\cibw\python` using NuGet.
+- For `linux`, you need Docker running, on macOS or Linux.
+- For `macos`, you need a Mac machine. Note that cibuildwheel is going to install MacPython on your system, so you probably don't want to run this on your development machine.
+- For `windows`, you need to run in Windows. cibuildwheel will install required versions of Python to `C:\cibw\python` using NuGet.
```
henryiii, line 95: I didn't realize this was scoped like this, nice. :) So it's really only macOS that gets true "global" installs.

henryiii (approved):
Looks like nice cleanups. :)

docs/options.md
```diff
@@ -90,20 +90,31 @@ Default: `auto`
...
 
 This option can also be set using the [command-line option](#command-line) `--platform`.
 
+!!! tip
+    If you have Docker installed on macOS or Linux, you can locally debug your cibuildwheel Linux config, instead of pushing to CI to test every change. For example:
```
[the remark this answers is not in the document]
joerick: Good question! It looks like in fact, these days you can! 

https://docs.microsoft.com/en-us/virtualization/windowscontainers/quick-start/quick-start-windows-10-linux

I'll update it.

docs/options.md
joerick: Yeah. We could even install the Pythons somewhere nice from [appdirs], then maybe we could drop the 'only on CI' warning. 
[appdirs]: https://pypi.org/project/appdirs/

On Mac, I seem to remember that you couldn't install it to a different location for some reason. The `installer` command supports it, but something in Python wasn't happy.

docs/options.md
```diff
@@ -539,8 +573,7 @@ CIBW_DEPENDENCY_VERSIONS: ./constraints.txt
 
 Shell command to run tests after the build. The wheel will be installed
 automatically and available for import from the tests. To ensure the wheel is
-imported by your tests (instead of your source copy), tests are run from a
-different directory. Use the placeholders `{project}` and `{package}` when
+imported by your tests (instead of your source copy), **tests are not run from your project directory**. Use the placeholders `{project}` and `{package}` when
```
YannickJadoul, line 576: Line breaks are slightly inconsistent here, now, if you care about that.

docs/options.md
```diff
@@ -432,31 +443,54 @@ The command is run in a shell, so you can run multiple commands like `cmd1 && cm
...
+than x86 (x86\_64 and i686) `manylinux2014` or `manylinux_2_24` must be used, because the first version of the manylinux specification that supports additional architectures is `manylinux2014`.
 
 Note that `manylinux2014`/`manylinux_2_24` don't support builds with Python 2.7 - when building with `manylinux2014`/`manylinux_2_24`, skip Python 2.7 using `CIBW_SKIP` (see example below).
 
+If setting a custom Docker image, you'll need to make sure it can be used in the same way as the official, default Docker images: all necessary Python and pip versions need to be present in `/opt/python/`, and the auditwheel tool needs to be present for cibuildwheel to work. Apart from that, the architecture and relevant shared system libraries need to be compatible to the relevant standard to produce valid manylinux1/manylinux2010/manylinux2014/manylinux_2_24 wheels (see [pypa/manylinux on GitHub](https://github.com/pypa/manylinux), [PEP 513](https://www.python.org/dev/peps/pep-0513/), [PEP 571](https://www.python.org/dev/peps/pep-0571/), [PEP 599](https://www.python.org/dev/peps/pep-0599/) and  [PEP 600](https://www.python.org/dev/peps/pep-0600/)  for more details).
```
YannickJadoul, line 490: If you really want to have some microscopic nits picked:
```suggestion
If setting a custom Docker image, you'll need to make sure it can be used in the same way as the official, default Docker images: all necessary Python and pip versions need to be present in `/opt/python/`, and the auditwheel tool needs to be present for cibuildwheel to work. Apart from that, the architecture and relevant shared system libraries need to be compatible to the relevant standard to produce valid manylinux1/manylinux2010/manylinux2014/manylinux_2_24 wheels (see [pypa/manylinux on GitHub](https://github.com/pypa/manylinux), [PEP 513](https://www.python.org/dev/peps/pep-0513/), [PEP 571](https://www.python.org/dev/peps/pep-0571/), [PEP 599](https://www.python.org/dev/peps/pep-0599/) and [PEP 600](https://www.python.org/dev/peps/pep-0600/) for more details).
```

YannickJadoul: (Yes, those are two spaces I removed. I don't care, but I noticed.)

YannickJadoul (approved):
Nice, looks great! :-) Seems not a huge update, but I think it's actually a nice sign of the project getting more mature!

Quick thing I noticed: until now, I've been using `cibuildweel` as stylized version, but so now you're switching to just cibuildweel? (I don't really mind; just thought I'd double check and adapt my habits ;-) )

docs/options.md
joerick: I've long been a hard-wrapper, but lately I've been wondering about stripping out the manual line breaks, and turning on Soft Wrap in my editor. It's kinda tedious to have to manually reformat all the time. And it would be a bit nicer to have diff see each paragraph as a single instance. What do you think?

docs/options.md
joerick: haha. can't unsee :)

docs/options.md
joerick, line 576:
```suggestion
imported by your tests (instead of your source copy), **tests are not run from
your project directory**. Use the placeholders `{project}` and `{package}` when
```

joerick:
> Quick thing I noticed: until now, I've been using `cibuildweel` as stylized version, but so now you're switching to just cibuildwheel? (I don't really mind; just thought I'd double check and adapt my habits ;-) )

Yes, good question. Some of the docs entries were getting very backtick heavy, so I decided to remove some of them when they're referring to a project, rather than a `command` or something the user would type.

Here's what I wrote in the commit message at the time:

> Using the backtick too much breaks up the flow of the text and makes it
> hard to read. Here, I'm deciding that project names are not backtick'd,
> unless they're referring to something the user would type. e.g.
> 
> ❌ the `manylinux` project
> ✅ the manylinux project
> 
> ❌ set `CIBW_MANYLINUX_IMAGE` to manylinux2010
> ✅ set `CIBW_MANYLINUX_IMAGE` to `manylinux2010`
> 

Of course, we then hit up the capitalisation problem. All this would read better if we could capitalise the proper nouns. But cibuildwheel doesn't really have a capitalised version that works. Cibuildwheel (eh - Ci?), CIbuildwheel (what), CIBuildWheel (quite ugly IMO). So I'm thinking we stick to lowercase cibuildwheel.

docs/options.md
YannickJadoul: I don't really mind either way. I'm never a fan of imposed line limits (though @henryiii is, I believe), so soft-wrapping is perfect for me.

YannickJadoul: Does this keep the **markdown** markup? It's not highlighted anymore by GitHub, now?

YannickJadoul:
> Yes, good question. Some of the docs entries were getting very backtick heavy, so I decided to remove some of them when they're referring to a project, rather than a `command` or something the user would type.

That makes sense to me, yes. It might also make other people typing "cibuildwheel" without the backticks more consistent (not sure if that's a good reason, but it's a nice consequence).
I don't really mind, but it's good to know, because I've been doing an effort on having backticks around cibuildwheel, up until now :-)

> Here's what I wrote in the commit message at the time [c6ede9b](https://github.com/joerick/cibuildwheel/commit/c6ede9b542ae6becaa7482016f7d9b3c37662b26):

Right, sorry, didn't see that!

> Of course, we then hit up the capitalisation problem. All this would read better if we could capitalise the proper nouns. But cibuildwheel doesn't really have a capitalised version that works. Cibuildwheel (eh - Ci?), CIbuildwheel (what), CIBuildWheel (quite ugly IMO). So I'm thinking we stick to lowercase cibuildwheel.

Also fully agree on those three assessments, yes ;-)

henryiii:
I'm a mild fan of hard wrapping, though I only manually do it (in vi with <kbd>Shift</kbd>+<kbd>V</kbd>,<kbd>G</kbd>,<kbd>Q</kbd>), though I didn't used to be. It makes the diffs easier to read, as small one change in a line becomes much harder to spot if that lines is wider than your screen / diff view in GitHub. Large changes "reflow" the whole paragraph, but that's not any worse than the unreadable one-line diff if it's all on a single line. You don't have to be perfect about it, though, I only reflow when it starts looking bad.

> Also fully agree on those three assessments

I would only add that it then should be "Cibuildwheel" at the beginning of a sentence. Which I probably have been doing once in a while anyway.

docs/options.md
henryiii: That's just the "suggestion" view, which is some sort of diff without parent language (guessing it guesses Python here?)

joerick:
Travis failure unrelated, ofc.

PR #651, opened 2021-04-25
style: second set of updates

henryiii:
Adding a few more stylistic updates.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -53,6 +53,7 @@ def get_option_from_environment(option_name: str, *, platform: Optional[str] = N
 
 def main() -> None:
     platform: PlatformName
+    arch_list_str = ", ".join(a.name for a in Architecture)
 
     parser = argparse.ArgumentParser(
         description='Build wheels for all the platforms.',
@@ -75,17 +76,19 @@ def main() -> None:
 
     parser.add_argument('--archs',
                         default=None,
-                        help='''
+                        help=f'''
                             Comma-separated list of CPU architectures to build for.
                             When set to 'auto', builds the architectures natively supported
                             on this machine. Set this option to build an architecture
                             via emulation, for example, using binfmt_misc and QEMU.
                             Default: auto.
-                            Choices: auto, auto64, auto32, native, all, {}
-                        '''.format(", ".join(a.name for a in Architecture)))
+                            Choices: auto, auto64, auto32, native, all, {arch_list_str}
+                        ''')
+
     parser.add_argument('--output-dir',
                         default=os.environ.get('CIBW_OUTPUT_DIR', 'wheelhouse'),
                         help='Destination folder for the wheels.')
+
     parser.add_argument('package_dir',
                         default='.',
                         nargs='?',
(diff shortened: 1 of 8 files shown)


cibuildwheel/bashlex_eval.py
```diff
@@ -58,9 +58,8 @@ def evaluate_word_node(node: bashlex.ast.node, context: NodeExecutionContext) ->
 
         if part_string not in value:
             raise RuntimeError(
-                'bash parse failed. part "{}" not found in "{}". Word was "{}". Full input was "{}"'.format(
-                    part_string, value, node.word, context.input,
-                )
+                f'bash parse failed. part "{part_string}" not found in "{value}". '
+                'Word was "{node.word}". Full input was "{context.input}"'
```
henryiii, line 62:
```suggestion
                f'Word was "{node.word}". Full input was "{context.input}"'
```

cibuildwheel/__main__.py
```diff
@@ -53,6 +53,7 @@ def get_option_from_environment(option_name: str, *, platform: Optional[str] = N
 
 def main() -> None:
     platform: PlatformName
+    arch_list_str = ", ".join(a.name for a in Architecture)
```
joerick, line 56: Maybe this could be moved closer to the place it's used?

cibuildwheel/__main__.py
```diff
@@ -368,7 +372,8 @@ def detect_warnings(platform: str, build_options: BuildOptions) -> List[str]:
         option_value = getattr(build_options, option_name)
 
         if option_value and ('{python}' in option_value or '{pip}' in option_value):
-            warnings.append(option_name + ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
+            msg = ": '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead."
+            warnings.append(option_name + msg)
```
joerick, line 376: I wonder if this would be better as f-string?

```suggestion
            # p.s. in an f-string, double braces means literal single brace
            warnings.append(f"{option_name}: '{{python}}' and '{{pip}}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.")
```

cibuildwheel/macos.py
```diff
@@ -233,6 +233,9 @@ def setup_python(python_configuration: PythonConfiguration,
     # inconsistencies if it's left unset.
     env.setdefault('MACOSX_DEPLOYMENT_TARGET', '10.9')
 
+    is_arm64 = python_configuration.identifier.endswith('arm64')
+    is_universal2 = python_configuration.identifier.endswith('universal2')
```
joerick, line 237: I'm not sure about pulling these out... I kinda like how explicit/direct the code is currently. Unless there's a formatting reason why you'd like to do it?

joerick (commented):
I'm pretty happy with these, too. :) I made a note on the ones I'm not sure about

cibuildwheel/macos.py
henryiii: The reason is that when combined, they are really long and don't format well (and they are also used often).
```python
building_arm64 = is_arm64 or is_universal2
```
reads better than
```python
building_arm64 = python_configuration.identifier.endswith('arm64') or python_configuration.identifier.endswith('universal2') 
```

What about `config_is_*`?

henryiii: Pushed a version that uses config_is_arm64 and such.

henryiii:
@joerick Does this look okay now?

joerick (approved):
Looks good! 👍

PR #653, opened 2021-04-30
Fix nuget error on Travis

joerick:
Fix this error on Travis:

> Package 'pythonx86 3.5.4' is not found in the following primary source(s): 'C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\'. Please verify all your online package sources are available (OR) package id, version are specified correctly

The problem seems to be that Travis added a blank nuget config file at `%AppData%\NuGet\NuGet.Config`, so nuget didn't know where to fetch its packages from. This PR adds a `FallbackSource` via the command line, so regardless of config files, nuget can find the registry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,4 @@ jobs:
 
 install: $PYTHON -m pip install -e ".[dev]" pytest-custom-exit-code
 
-script: |
-  $PYTHON -c "import os, pprint; pprint.pprint(os.environ)"
-  $PYTHON -c "import os; config_path = os.environ['APPDATA']+'/NuGet/NuGet.Config'; print(open(config_path).read())"
-  ls $APPDATA
-  ls $APPDATA\NuGet
-  $PYTHON ./bin/run_tests.py
+script: $PYTHON ./bin/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install: $PYTHON -m pip install -e ".[dev]" pytest-custom-exit-code
 
 script: |
   $PYTHON -c "import os, pprint; pprint.pprint(os.environ)"
-  $PYTHON -c "import os; config_path = os.environ['APPDATA']+'\NuGet\NuGet.Config'; print(open(config_path).read())"
-  dir %AppData%
-  dir %AppData%\NuGet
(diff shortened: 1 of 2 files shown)


YannickJadoul:
Before you put too much time into this, @joerick: is it worth it? travis-ci.org is still saying it's soon shutting down and we should all switch to www.travis-ci.com and beg for free credits as open source project.

henryiii:
See my email. Also, about 13 or so of our known users (still) use travis (may need an update, though, I know boost-histogram has dropped it ;) ) Travis Windows is pretty rare, though.

YannickJadoul:
> I'd be happy to drop Travis support, I don't use it and it's a disproportionate pain to maintain.

Well, to be fair, my take on this is: they did that to themselves.

I'm mostly worried that even if we keep supporting Travis CI, we'd need to switch to travis-ci.com anyway, which might have a slightly different way of configuring things.
From that perspective, we could drop travis-ci.org until someone picks up the courage to transfer to and set up travis-ci.com?

YannickJadoul:
(I'm also not stopping you, btw. If you have time and want to get this working, be my guest, but I thought it was worth mentioning :-) )

joerick:
I think I figured this one out already :)

But thank you for bringing it up, I've been meaning to do the same for a while. On top of changes like these that need workarounds, it does seem that it's the most flaky of the CIs that we support.

Additionally, there's the business-level stuff. They seem to be delaying on actually flipping the switch on travis-ci.org changeover, I'm not sure why. I'm not sure how things will work once we make the switch, but I've little interest in doing open-source work to support them if they're not supporting open-source with their service. So if that's how things look after the switch, I'd be in favour of dropping our support on principle.

YannickJadoul:
> Additionally, there's the business-level stuff. They seem to be delaying on actually flipping the switch on travis-ci.org changeover, I'm not sure why. I'm not sure how things will work once we make the switch, but I've little interest in doing open-source work to support them if they're not supporting open-source with their service. So if that's how things look after the switch, I'd be in favour of dropping our support on principle.

I entirely, full-heartedly agree here! In some way, we'd be doing them a favor by keeping up support, so I don't feel like jumping through a bunch of hoops to keep up support.

henryiii:
Technically, this helped our GHA support too. ;)

henryiii (approved):

PR #659, opened 2021-05-07
fix: missing f for fstrings in bin

henryiii:
Recent typo.


henryiii:
#633 discovered the same bug too, it appears.

mayeut:
@henryiii, 

> #633 discovered the same bug too, it appears.

I thought I'd do that one on its own today but you beat me to it :-)

It was too late yesterday evening for me and that's now one item less on the to-do list, thanks.

PR #660, opened 2021-05-08
ci: fix auto-update by blocking PyPy 7.3.4

henryiii:
Overlaps a bit with #629; this currently supports updating 32-bit to 32-bit PyPy, or 64-bit to 64-bit PyPy on Windows, like CPython. Also blocks PyPy 7.3.4, since that one is broken.


henryiii:
The real test of this is if the update deps workflow runs, so I'll go ahead and merge, then trigger that run.

PR #662, opened 2021-05-08
ci: use native actions syntax

henryiii:
I think the custom action predates the native syntax? Besides removing a requirement, this also should get around the "snapshot" issue seen in #541; the native syntax should allow you to add the label and retrigger the build, while the custom action doesn't let you retrigger with new labels.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,12 +39,8 @@ jobs:
       run: |
         python -m pip install -e ".[dev]"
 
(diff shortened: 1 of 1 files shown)


joerick (approved):
Yeah, I think they might have added the native ability to do this after I created the custom action. Or maybe I just missed it ;) . I'm not sure this would get us around the 'snapshot' problem as you put it, but simpler = better.

PR #664, opened 2021-05-08
pin CPython 3.8 installer on macosx10.9

mayeut:
macosx11 installer requires macOS 11+ for CPython 3.8 while macosx11 installer for CPython 3.9+ only requires macOS 10.9+ on x86_64.

c.f. comments in #663

### DIFF
--- a/bin/update_pythons.py
+++ b/bin/update_pythons.py
@@ -196,27 +196,32 @@ class CPythonVersions:
                 uri = int(release["resource_uri"].rstrip("/").split("/")[-1])
                 self.versions_dict[version] = uri
 
-    def update_version_macos(self, identifier: str, spec: Specifier) -> ConfigMacOS | None:
-        file_idents = ("macos11.pkg", "macosx10.9.pkg", "macosx10.6.pkg")
+    def update_version_macos(
+        self, identifier: str, version: Version, spec: Specifier
+    ) -> ConfigMacOS | None:
         sorted_versions = sorted(v for v in self.versions_dict if spec.contains(v))
 
-        for version in reversed(sorted_versions):
+        if version <= Version("3.8.9999"):
+            file_ident = "macosx10.9.pkg"
+        else:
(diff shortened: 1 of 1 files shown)


bin/update_pythons.py
```diff
@@ -196,13 +196,19 @@ def __init__(self) -> None:
...
+    def update_version_macos(
+        self, identifier: str, version: Version, spec: Specifier
+    ) -> ConfigMacOS | None:
         sorted_versions = sorted(v for v in self.versions_dict if spec.contains(v))
 
-        for version in reversed(sorted_versions):
+        if version <= Version("3.8.9999"):
+            file_idents = ("macosx10.9.pkg",)
+        else:
+            file_idents = ("macos11.pkg", "macosx10.9.pkg")
```
mayeut, line 207: Should we just drop the tuple here and just go ahead and define `file_ident` once and for all ?

mayeut: Nevermind, I went this way just to silence the MyPy issue - I guess I could have solved differently.

joerick (approved):
looks good to me but I'll defer to @henryiii as he wrote this script originally

bin/update_pythons.py
```diff
@@ -196,27 +196,32 @@ def __init__(self) -> None:
...
+    def update_version_macos(
+        self, identifier: str, version: Version, spec: Specifier
+    ) -> ConfigMacOS | None:
         sorted_versions = sorted(v for v in self.versions_dict if spec.contains(v))
 
-        for version in reversed(sorted_versions):
+        if version <= Version("3.8.9999"):
+            file_ident = "macosx10.9.pkg"
+        else:
+            file_ident = "macos11.pkg"
```
henryiii, line 207: We could instead take the old file identifier, and only look for that.

henryiii: This is fine for now, but that's a possible update later to remove the explicit version check.

henryiii (approved):

PR #630, opened 2021-03-26
style: apply black

henryiii:
This enables black. String normalization is turned off until the last commit to keep the diffs down, so the focus can be on structure rather than a bunch of quote changes (seems to change docstrings anyway, though, FYI). We also have a longer than normal line length set, compared to last time this might have been attempted.

Having a diff-optimized and highly restricted style might be especially useful before we split for 2.x development. What do others think? The auto-fixes from pre-commit are especially nice with something like black over flake8.

PR #666, opened 2021-05-09
chore: bump PyPy and drop PyPy 3.6

henryiii:
Rebased (and closes #629).

- chore: update PyPy to 7.3.4 except on Windows
- chore: bump dependencies


henryiii:
Looks like things are passing. Would like 1-2 sign-offs, as this does drop PyPy 3.6.

README.md
```diff
@@ -28,8 +28,7 @@ What does it do?
 | CPython 3.7     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.8     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.9     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| PyPy 3.6 v7.3.3 | ✅ |    |    | ✅  | ✅ |    |    |    |   |
-| PyPy 3.7 (beta) v7.3.3 | ✅ | |  | ✅  | ✅ |    |    |   |    |
+| PyPy 3.7 v7.3.4 | ✅ |   | ✅  |    | ✅ |    |    |    |   |
```
mayeut, line 31:
```suggestion
| PyPy 3.7 v7.3 | ✅ |   | ✅  |    | ✅ |    |    |    |   |
```

IMHO, the patch version is not needed because the ABI tag only relates to PyPy being 3.7 v7.3.*.
That's one less thing to remember to update and it's already what's being done for CPython.

joerick: Good idea, I hadn't fully twigged this. The Python/ABI tag is `pp36-pypy36_pp73`. We don't list patch version of CPythons, so let's remove the PyPy patch version it from here.

joerick:
Looks good to me! Let's try and get a @YannickJadoul review before merging, if possible.

YannickJadoul:
I don't get where this is coming from.
Why are we doing this, instead of just going with #629 that I've been preparing? That should be ready (especially, since the manylinux image was merged recently).

Also, I was actually still waiting with that, because PyPy is almost releasing 7.3.5, with a (reasonably urgent?) bugfix. They're already at rc2, so I'm expecting a final release soon.

YannickJadoul:
> Why are we doing this, instead of just going with #629 that I've been preparing? That should be ready (especially, since the manylinux image was merged recently).

That seems a lot better for tracability. And I just checked, no one poked me there to rebase or merge anything.

> Also, I was actually still waiting with that, because PyPy is almost releasing 7.3.5, with a (reasonably urgent?) bugfix. They're already at rc2, so I'm expecting a final release soon.

See also https://github.com/joerick/cibuildwheel/pull/629#issuecomment-824244123

henryiii:
> Why are we doing this, instead of just going with #629 that I've been preparing?

This _is_ #629; you'll see you are a coauthor to this. I felt a little weird force-pushing such a major change to your branch, but it is your work _and_ is attributed to you, if you are worried about that, it's just a rebase + rerun of the update script. I can force-push there if you want, and close this. Really all that's left is removing PyPy3.6 (and the bump to 7.3.4 on macOS, which is only a fraction of #629. The rest is done.

> Also, I was actually still waiting with that, because PyPy is almost releasing 7.3.5

Actually, most of #629 is _already_ in master, as part of #633, also #660 as well. We _had_ to update to 3.3.4 on linux, since we needed a newer image to get Auditwheel 4. It was also is blocking #521. These PRs were needed to get back to the latest versions of dependences and fix our update cycle. We were told "a week or two" for a 7.3.5 release more than two weeks ago, so (was) time to update. We pinned 7.3.3 on Windows by not changing archs automatically, that bug was about Windows. If it was a more general issue, pypy/manylinux would not have / should not have updates. Since we can't have a uniform PyPy version anyway, bumping macOS and dropping 3.6 seems reasonable.

Again, happy to force-push to the branch in #629 if you prefer that.

README.md
henryiii: Nice, thanks! Done.

henryiii:
PS: The rebase was a little harder than usual due to #630 going in in them meantime, which you were not very fond of, so I was also trying to help with that rather than ask you to do extra work.

YannickJadoul:
Right, OK, now I think I see where this is coming from! Thanks for the extra explanation (and rebasing work; something tells me that wasn't obvious?), @henryiii!

> PS: The rebase was a little harder than usual due to #630 going in in them meantime, which you were not very fond of, so I was also trying to help with that rather than ask you to do extra work.

And thanks for that, actually, now that I understand! :-) That now also makes sense.

So, two things to remark, from my side:
1. There are a few Windows-changes we might want to keep for PyPy 7.3.5 (coming soon?), so let's keep those in #629 and let's not close that? Either we repurpose that PR for the 7.3.5 testing and Windows changes, or we turn the leftovers into a new one?
2. Does this need to be backported to 1.x? Probably, because PyPy 3.6 should also be phased out there, or not? (Actually, I still need to catch up with those discussions, so maybe this is obvious; apologies, in that case).

Apart from that, good to merge, I think, yes!

P.S.: Even thought it's a bit of a bigger rebase, if you had left a comment and force-pushed, that would've also worked for me. But this is also fine, now that I understand why and how :-) Thanks again!

Issue #657, opened 2021-05-03
Plan for 2.0


henryiii:
Now that the main style merge is in, and manylinux is about to drop Python 3.5, I think we should look at the 2.0 work. This is what I'm thinking (maintainers please edit the list directly):

Version 1.x:
* [x] Create branch `1.x`
* ~Pin auditwheel < 4 for automatic updates~ ([ref](https://github.com/joerick/cibuildwheel/issues/657#issuecomment-835313619))
* [x] Enable CI on `1.x` (I have done GHA only for now, Azure, Circle and Gitlab picked it up anyway).
* [x] Add `v1` branch to the docs and allow tests to run when merging to v1. ~_Maybe_ automatic updates on this branch too, but probably better to just do it manually if needed.~

Version 2.0.0a2:
* [x] Drop Python 3.5
* [x] Drop Python 2.7 (unpin manylinux 2010)
* [x] Add links to the v1 docs for "how to build with Python 2" or something like that.
* [x] Perhaps `--pre` support for pre-releases and Python 3.10 #675
* [x] Drop warnings for old specifiers (deprecated in 1.x) #700
* [x] Produce a warning if "?p27" is detected in SKIP, and an error if "?p27" is detected in BUILD? Likewise with 3.5. #700

Version 2.0.0 final:
* [x] Add "build" as an optional build method #521
* [x] Support configuration from pyproject.toml #684
* [x] Discuss `CIBW_MANYLINUX_PYPY_*` options as mentioned in https://github.com/pypa/cibuildwheel/pull/671#discussion_r637931737 and https://github.com/pypa/cibuildwheel/pull/671#issuecomment-847348461 - #711 - no change

Some future release:
* Add nicer Limited API support #569
*  Perhaps musllinux #627 support... (currently needed: auditwheel support)

Although we developers love features like "drop X", those are not very exciting new features for users, so pyproject.toml support could be a nice shiny new feature to get users to upgrade. Better Limited API support is nice, but affects very few users. Musllinux would be nice, but don't think that auditwheel / images will be ready in time - probably will be 2.x down the line. Python 3.10 beta is nice, but not really supposed to be used by anyone besides testing yet, so not so exciting. `build` is also interesting from a back-end, but not likely to be that exciting as a reason to upgrade. And PyPy 64-bit Windows & any manylinux PyPy is also nice, honestly that might be the second most exciting thing in 2.0 for users.

joerick:
[...]

PR #671, opened 2021-05-12
Update PyPy to 7.3.5 and use PyPA's manylinux images with PyPy

YannickJadoul:
This also includes the leftover Windows changes from #629 and #666.


YannickJadoul:
@mattip, FYI, 7.3.5rc2 seems to pass all our tests, also on Windows :-)

mayeut:
@YannickJadoul, with PyPy 7.3.5 released and integrated in manylinux images, I've pushed a few commits and will push a bit more on this branch.

YannickJadoul:
Great news, @mayeut! Let me quickly rebase this onto master (with #685).

YannickJadoul:
(I hope this isn't going to cause any issues for you. If so, feel free to force-push whatever it is you currently have!)

mayeut:
rebased onto master (with #689)

mayeut:
@mattip, FYI, 7.3.5 seems to pass all our tests (including the new manylinux images with x86_64/i686/aarch64)

mattip:
Whew! Do the windows wheels look normal?

mayeut:
> Do the windows wheels look normal?

@mattip, What do you mean by normal ?

They can be built with the proper tag, they can be installed and they can be used.

mayeut:
I think this is now ready for review.

README.md
```diff
@@ -110,7 +110,7 @@ Options
...
+|   | [`CIBW_MANYLINUX_X86_64_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_I686_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_PYPY_X86_64_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_AARCH64_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_PPC64LE_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_S390X_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_PYPY_AARCH64_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  <br> [`CIBW_MANYLINUX_PYPY_I686_IMAGE`](https://cibuildwheel.readthedocs.io/en/stable/options/#manylinux-image)  | Specify alternative manylinux docker images |
```
mayeut, line 113: I will do a single comment here that might apply throughout the remaining of the changes.

Should we remove the `CIBW_MANYLINUX_PYPY_*` environment variables (adding a deprecation warning for `CIBW_MANYLINUX_PYPY_X86_64_IMAGE `) ?

I went with "we should keep it" and thus "them" for consistency's sake.

The only thing remaining would thus be how to best order those options in docs (either group by CPython/PyPy or by arch ? The current layout is by age).

YannickJadoul: I was thinking something similar, but seems you beat me :-) If manylinux now "compresses" the two into one, then shouldn't we do so as well?

The more I think about it, the more I think we should do this, if we can deprecate `CIBW_MANYLINUX_X86_64_IMAGE`. Extra argument: the PyPy manylinux images also include everything from manylinux itself, as a base. So if someone needs to use these old ones, it should still be possible?

YannickJadoul: About deprecating `CIBW_MANYLINUX_PYPY_X86_64_IMAGE`: this is slightly more tricky. Before, when we did so, we basically added it to the new setting. But now, if both `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_PYPY_X86_64_IMAGE` are set, it's not clear what the new value of `CIBW_MANYLINUX_X86_64_IMAGE` should be?

Given that I'm not expecting a lot of users to actually set `CIBW_MANYLINUX_PYPY_X86_64_IMAGE`, would it be possible to just raise an error and break the config?

YannickJadoul: > The only thing remaining would thus be how to best order those options in docs (either group by CPython/PyPy or by arch ? The current layout is by age).

That's the other thing I noticed, yes. If we would keep them, I think maybe just lsting all PyPy things at the end in the same order as the default ones would be best, as we also e.g. do this for the Python versions.

mayeut: > About deprecating CIBW_MANYLINUX_PYPY_X86_64_IMAGE: this is slightly more tricky. Before, when we did so, we basically added it to the new setting. But now, if both CIBW_MANYLINUX_X86_64_IMAGE and CIBW_MANYLINUX_PYPY_X86_64_IMAGE are set, it's not clear what the new value of CIBW_MANYLINUX_X86_64_IMAGE should be?

> Given that I'm not expecting a lot of users to actually set CIBW_MANYLINUX_PYPY_X86_64_IMAGE, would it be possible to just raise an error and break the config?

You're right, I didn't fully thought this through.

One benefit of keeping a separate key for PyPy is that it allows the following config to be valid:
```
# Use manylinux1 for CPython because pip isn't always updated...
CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
CIBW_MANYLINUX_I686_IMAGE: manylinux1
# Use the default for PyPy, do not skip it
```

henryiii:
This is fine to merge, right? Then we could eventually consolidate the manylinux images, but for now, it's rather handy to have both. One idea could be to leave the pypy one blank, and then use the regular one if it's blank. (But than can be a separate discussion and PR, probably after #684 - it's also slightly non-backwards compatible, so a good idea to decide before 2.0)

henryiii (approved):

mayeut:
Let's merge this as proposed by @henryiii, I will add an item to #657 for the remaining tasks that should be done before 2.0

PR #675, opened 2021-05-14
Add CPython 3.10 support

mayeut:
Just checking if we should expect any surprise with this.
No reflection on the `--pre` proposal in #657 yet


henryiii:
I've added a basic implementation of a `--pre` flag.

henryiii:
Should we un-draft this? It shouldn't build Python 3.10 wheels unless someone explicitly passes `--pre`.

README.md
```diff
@@ -28,13 +28,16 @@ What does it do?
 | CPython 3.7     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.8     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.9     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.10¹     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | PyPy 3.7 v7.3   | ✅ |   | ✅  |    | ✅ |    |    |    |   |
+<sup>¹ Requires `--pre` flag to be passed - Python Beta candidates not recommended for upload to PyPI. </sup><br>
```
mayeut, line 33: Should we expend a bit more on ABI compatibility here ? It might be too verbose for the README, maybe a link to an issue that's kept open until RC1 or Final Release with a bit more information ?

cibuildwheel/util.py
```diff
@@ -59,17 +59,29 @@ class IdentifierSelector:
...
+        filter_prerelease: str = "",
     ):
         self.build_patterns = build_config.split()
         self.skip_patterns = skip_config.split()
         self.requires_python = requires_python
+        self.filter_prerelease = filter_prerelease
 
     def __call__(self, build_id: str) -> bool:
+        # hardcode pre-releases here
+        # Current: Python 3.10
```
mayeut, line 81:
```suggestion
```
Not hardcoded here anymore.

cibuildwheel/__main__.py
```diff
@@ -222,8 +228,12 @@ def main() -> None:
     ) or get_requires_python_str(package_dir)
     requires_python = None if requires_python_str is None else SpecifierSet(requires_python_str)
 
+    # Filters Python 3.10 unless --pre is passed
     build_selector = BuildSelector(
-        build_config=build_config, skip_config=skip_config, requires_python=requires_python
+        build_config=build_config,
+        skip_config=skip_config,
+        requires_python=requires_python,
+        filter_prerelease="" if args.pre else "310",
```
mayeut, line 236:
```suggestion
        filter_prerelease="" if args.pre else "310",  # hardcode pre-releases here, current: Python 3.10
```

mayeut:
> Should we un-draft this?

The dependency update script might need to be updated first, It filters out pre-releases and beta2.../rc1/... would not be picked up. I might just quote you on this:
> Wouldn't this be better than updating it every time?

bin/update_pythons.py
```diff
@@ -190,21 +190,22 @@ def __init__(self) -> None:
...
-                self.versions_dict[version] = uri
+            uri = int(release["resource_uri"].rstrip("/").split("/")[-1])
+            self.versions_dict[version] = uri
 
     def update_version_macos(
         self, identifier: str, version: Version, spec: Specifier
     ) -> ConfigMacOS | None:
-        sorted_versions = sorted(v for v in self.versions_dict if spec.contains(v))
+
+        unsorted_versions = cast(Iterable[Version], spec.filter(self.versions_dict))
```
henryiii, line 200: This is really powerful; it gets the non-pre-releases unless it's all pre-releases; but had to mess with types a bit and manually cast, fixed in https://github.com/pypa/packaging/pull/430

mayeut: I've not taken the time to understand what's going on in there yet. All I can say is that it works !

joerick (commented):
Nice. My only comment is that I think we should keep to the convention and have an environment variable to configure this, rather than just a command line flag. I'd suggest `CIBW_PRERELEASE_PYTHONS`. Then we can add a little more documentation about ABI stability etc. in that option's documentation. 

That then suggests the command-line option should be `--prerelease-pythons`. I know somebody was keen on matching pip's `--pre` flag, but I think our consistency is more important (`CIBW_PLATFORM` <-> `--platform`, `CIBW_ARCHS` <-> `--archs`, `CIBW_OUTPUT_DIR` <-> `--output-dir`).

henryiii:
I'm not sure we should have an environment variable for this. It's a special testing flag, and not something that users should do without understanding that the wheels generated should not be uploaded to PyPI; if it's a general listed option, it might be more likely to be misused? However, having the OS variants could be useful, so I'm not against it either.

henryiii:
It could be useful in a matrix to run a test, so I'm +0.1.

joerick:
I'd prefer it was a sibling to CIBW_BUILD, CIBW_SKIP, CIBW_REQUIRES_PYTHON, as it's doing a very similar job to those options - build selection.

henryiii:
I've added CIBW_PRERELEASE_PYTHONS.

One argument against it is that I think it's the first and only boolean option in our environment variables. It's also the only one that explicitly makes invalid-to-upload wheels. And it can be completely controlled via BUILD/SKIP if you enable it.

One argument for it is that then we'd have to add a way to use it with the GitHub action if there's no env-var (though it could just be unavailable with the action, as it's a special setting for testing only, I guess).

bin/update_pythons.py
henryiii: `spec.filter(x) is like `v for v in self.versions_dict if spec.contains(v)` _except_ that if _no_ versions match, it tries again with pre-releases enabled (just like pip would). Exactly what we want here. :)

cibuildwheel/macos.py
```diff
@@ -270,7 +270,7 @@ def setup_python(
     config_is_arm64 = python_configuration.identifier.endswith("arm64")
     config_is_universal2 = python_configuration.identifier.endswith("universal2")
 
-    if python_configuration.version == "3.9":
+    if python_configuration.version in {"3.9", "3.10"}:
```
henryiii, line 273: Should we just make this
```suggestion
    if float(python_configuration.version) >= 3.9:
```
? Or change it for an inverse not so that it doesn't grow?

docs/options.md
```diff
@@ -299,6 +299,30 @@ the package is compatible with all versions of Python that it can build.
...
+Default: Off (0) if Python is available in beta phase. No effect otherwise.
+
+Platform-specific variants also available:<br/>
+`CIBW_PRERELEASE_PYTHONS_MACOS` | `CIBW_PRERELEASE_PYTHONS_WINDOWS` | `CIBW_PRERELEASE_PYTHONS_LINUX`
+
+#### Examples
+
+```yaml
+# Include latest Python beta
+CIBW_PRERELEASE_PYTHONS=1
```
henryiii, line 323:
```suggestion
CIBW_PRERELEASE_PYTHONS: 1
```

henryiii: True/False might work here?

test/test_docker_images.py
```diff
@@ -49,6 +49,6 @@ def test(tmp_path):
     expected_wheels = [
         w
         for w in utils.expected_wheels("spam", "0.1.0", manylinux_versions=["manylinux2010"])
-        if "-pp" not in w and "-cp39-" not in w
+        if "-pp" not in w and "-cp39-" not in w and "-cp310-" not in w
```
henryiii, line 52: Same comment about finding a way to not grow this. Maybe `"-cp31" not in w`? That will hold us until it doesn't matter anymore.

cibuildwheel/macos.py
mayeut: I'd rather not rely on float here.
Either cast to a packaging.Version and compare that or, as you suggest, reverse to `not in {"3.6", "3.7", "3.8"}`

test/test_docker_images.py
mayeut: Your suggestion works for me. It's not intuitive though.

```suggestion
        if "-pp" not in w and "-cp39-" not in w and "-cp31" not in w
```

test/test_manylinuxXXXX_only.py
```diff
@@ -80,6 +82,8 @@ def test(manylinux_image, tmp_path):
     expected_wheels = utils.expected_wheels(
         "spam", "0.1.0", manylinux_versions=platform_tag_map.get(manylinux_image, [manylinux_image])
     )
+    if manylinux_image in {"manylinux1"}:
+        expected_wheels = [w for w in expected_wheels if "-cp310-" not in w]
```
mayeut, line 86: Same as previous suggestion.

```suggestion
        expected_wheels = [w for w in expected_wheels if "-cp31" not in w]
```

bin/update_pythons.py
joerick: This is super handy, indeed! Perhaps worth a comment to that effect?

bin/update_pythons.py
```diff
@@ -190,21 +189,22 @@ def __init__(self) -> None:
...
```
joerick, line 199:
```suggestion
        # spec.filter(x) is like [v for v in x if spec.contains(v)] except that if no
        # versions match, it tries again with pre-releases enabled
        unsorted_versions = cast(Iterable[Version], spec.filter(self.versions_dict))
```

cibuildwheel/__main__.py
```diff
@@ -206,6 +212,9 @@ def main() -> None:
     build_verbosity_str = get_option_from_environment(
         "CIBW_BUILD_VERBOSITY", platform=platform, default=""
     )
+    prerelease_pythons = args.prerelease_pythons or get_option_from_environment(
+        "CIBW_PRERELEASE_PYTHONS", platform=platform, default="0"
+    ) not in {"", "0"}
```
joerick, line 217: There is a helper `strtobool` in `cibuildwheel.util` that can parse this. It would also give us case-insensitive true/false matching, which is a bit more friendly to a Python crowd I think.

cibuildwheel/macos.py
joerick: mayeut's suggestion (`not in`) looks good to me, also.

cibuildwheel/__main__.py
```diff
@@ -222,8 +231,12 @@ def main() -> None:
     ) or get_requires_python_str(package_dir)
     requires_python = None if requires_python_str is None else SpecifierSet(requires_python_str)
 
+    # Hardcode pre-releases here, current: Python 3.10
     build_selector = BuildSelector(
-        build_config=build_config, skip_config=skip_config, requires_python=requires_python
+        build_config=build_config,
+        skip_config=skip_config,
+        requires_python=requires_python,
+        filter_prerelease="" if prerelease_pythons else "310",
```
joerick, line 239: How about, instead of matching a little bit of the string, we do instead `exclude_prerelease="" if prerelease_pythons else "cp310-*",`, and match that in BuildSelector using wcmatch/fnmatch?

joerick: That would also allow us to do a pypy release in the future, or a prerelease of a new arch or something.

docs/options.md
joerick: Yeah, I like case-insensitive True/False too.

test/test_docker_images.py
joerick: I don't really mind about the growing, but I'd like the logic to match the CIBW_SKIP above. Or, could we flip it instead?

So we'd do `"CIBW_BUILD": "cp{36,37,38}-*"`, and

```suggestion
        if "-cp36-" in w or "-cp37-" in w or "-cp38-" in w
```

test/test_manylinuxXXXX_only.py
joerick: This one is a bit fiddlier, maybe mayeut's suggestion is fine, with a comment `# remove CPython 3.10 and above`

test/utils.py
```diff
@@ -24,13 +24,17 @@
     raise Exception("Unsupported platform")
 
 
-def cibuildwheel_get_build_identifiers(project_path, env=None):
+def cibuildwheel_get_build_identifiers(project_path, env=None, *, pre=False):
```
joerick, line 27: For consistency, renaming `pre` to `prerelease_pythons` would be good I think.

docs/options.md
```diff
@@ -299,6 +299,30 @@ the package is compatible with all versions of Python that it can build.
...
+> Enable building with pre-release versions of Python
+
+By default, cibuildwheel only builds wheels compatible with minor versions of
+Python. However, during the Beta period when new versions of Python are being
+tested, cibuildwheel will often gain early support for beta releases. If you
+would like to test wheel building with these versions, you can enable this
+flag.  Please _do not_ upload these wheels to PyPI, as they are not guaranteed
+to work with the final Python release. Once Python is ABI stable and enters the
+release candidate phase, that version of Python will become available without
+this flag.
```
joerick, line 312: Nice job with this, it looks good.

A couple notes-

- The first sentence is a little confusing "compatible with minor versions of Python" Actually do we need the first sentence at all? We could just start with "During the beta period when new versions of Python are being tested..."
- Adding an admonition would be nice here, to draw attention to the ABI compatibility note. Perhaps:

  ```
  !!! caution
      Please _do not_ upload these wheels to PyPI, as they are not guaranteed... 
  ```

This is my suggestion, if you want to copy/paste:

```
During the beta period, when new versions of Python are being tested,
cibuildwheel will often gain early support for beta releases. If you would
like to test wheel building with these versions, you can enable this flag.

!!! caution
    Please _do not_ upload these wheels to PyPI, as they are not guaranteed to
    work with the final Python release. Once Python is ABI stable and enters
    the release candidate phase, that version of Python will become available
    without this flag.
```

joerick (commented):
Nice job. Just a few minor changes to go I think.

mayeut:
rebased on master to get CI working.

cibuildwheel/macos.py
```diff
@@ -270,7 +270,7 @@ def setup_python(
     config_is_arm64 = python_configuration.identifier.endswith("arm64")
     config_is_universal2 = python_configuration.identifier.endswith("universal2")
 
-    if python_configuration.version == "3.9":
+    if python_configuration.version not in {"3.6", "3.7", "3.8"}:
```
mayeut, line 273: I've updated this but I can also think of the following even though it doesn't read as easily:
```suggestion
    if tuple(map(int, python_configuration.version.split("."))) >= (3, 9):
```

cibuildwheel/__main__.py
mayeut: I've reworked this with `skip_prerelease` and expanded the skip_patterns of the build selector with that.

docs/options.md
mayeut: I copy/pasted @joerick suggestion.
I think I understand the intent of @henryiii and that's still missing from the docs:
"you won't get pre-release for 3.7.y, 3.8.y, 3.9.y. You'll only get pre-release for a minor version 3.x if 3.x.0 isn't there yet"
If that was the intent, I don't really know  how to phrase that properly (and given it acts on non ABI stable versions, "3.x.0" isn't right either)

cibuildwheel/macos.py
joerick: I prefer the `not in`, personally, I think it reads better.

henryiii:
I'm still not really fond of making this an official, documented environment variable option - it's very much a testing-only thing you might want to try out but should never upload. I'd even be okay with not allowing it with the GitHub Action form, but only with `--pre`. But I guess I'm outnumbered.

I think we need to update the `cibuildwheel --help` output that's in the docs. @mayeut what's the status of this? Are all the review comments addressed or should I do something?

mayeut:
>  @mayeut what's the status of this? Are all the review comments addressed or should I do something?

@henryiii , I will mention you at the appropriate places.

bin/update_pythons.py
mayeut: @henryiii, could you do this one please ?

cibuildwheel/__main__.py
mayeut: @henryiii, @joerick, does the rework work for you ?

docs/options.md
mayeut: @henryiii, did I understand correctly your intent ? If so, do you or @joerick have something to propose ?

docs/options.md
mayeut: done

mayeut:
> I think we need to update the cibuildwheel --help output that's in the docs.

done.

cibuildwheel/__main__.py
joerick: Hey @mayeut, thanks for doing that! I think this works for me. One side effect of this implementation is that it shows up at the start of the build in the build options printout. What do you think, is this a little confusing for users? 

https://github.com/pypa/cibuildwheel/pull/675/checks?check_run_id=2662524426#step:6:25

I suppose it should show up somewhere in the build options... this printout isn't that intuitive for users anyway, because they set 'build', 'skip', and it's printed here as `build_selector`. And I like how simple this implementation is. So I think this is okay.

docs/options.md
joerick: Yes, I understand. But it's quite hard to write... I also think that anyone who might want to use this option would probably understand that they don't care about the ABI-compatible patch betas. Given the warning below, it should be clear that we're talking about the ABI-incompatible ones, I think?

cibuildwheel/__main__.py
mayeut: > One side effect of this implementation is that it shows up at the start of the build in the build options printout. What do you think, is this a little confusing for users?

That's the only thing I don't like with this implementation. As you said, the current printout was already non intuitive so went with that anyway because it keeps the implementation real simple.
We could get the printout to say something like: `build_selector: BuildSelector('*' - '' - 'cp310-*')` but I'm not sure it's better (in fact I'm pretty sure it's worse).

joerick: I had a go at some refactoring to neaten this up - let me know what you think:

https://github.com/pypa/cibuildwheel/commit/97be46543c6b2fb7baaf67b9eeeb5d02f7b95cca

You can cherry pick it onto this branch if you like.

---

(aside - does Github need a PR-for-a-PR feature? Or should I just push to a PR and say 'feel free to revert'?)

henryiii: You can make a PR to a PR branch. But it's easier (and more visible) if it's in the same repo, since the PR to a PR will show up in the repo it is targeting.

docs/options.md
henryiii: I think that's fine.

mayeut: ok, let's stick with this then.

cibuildwheel/__main__.py
joerick: Ah, yes, good idea. In doing so I just discovered there's a bug in it actually!

joerick: ok, so the new commit is 219ead542f5bd54d08ea894db68ab77add151359 🤭

Before:

```

cibuildwheel version 1.11.0

Build options:
  platform: 'linux'
  architectures: {<Architecture.i686: 'i686'>, <Architecture.x86_64: 'x86_64'>}
  before_all: ''
  before_build: None
  before_test: None
  build_selector: BuildSelector('*' - 'cp310-*')
  build_verbosity: 0

```

After:

```

cibuildwheel version 1.11.0

Build options:
  platform: 'linux'
  architectures: {<Architecture.i686: 'i686'>, <Architecture.x86_64: 'x86_64'>}
  before_all: ''
  before_build: None
  before_test: None
  build_selector: BuildSelector(build_config='*')
  build_verbosity: 0
```

After with pre-release enabled:

```

cibuildwheel version 1.11.0

Build options:
  platform: 'linux'
  architectures: {<Architecture.i686: 'i686'>, <Architecture.x86_64: 'x86_64'>}
  before_all: ''
  before_build: None
  before_test: None
  build_selector: BuildSelector(build_config='*', prerelease_pythons=True)
  build_verbosity: 0
```

bin/update_pythons.py
henryiii: Ahh, missed this. Thought I saw it in the email, then couldn't find it in the discussion. Now see it in the code review.

henryiii (approved):

cibuildwheel/__main__.py
mayeut: @joerick, This seems much better, will look at integrating this in this branch.

mayeut:
I think all comments have been addressed now.
@henryiii, re-requested review from your previous approval since I included the commit from @joerick

cibuildwheel/__main__.py
```diff
@@ -206,6 +212,9 @@ def main() -> None:
     build_verbosity_str = get_option_from_environment(
         "CIBW_BUILD_VERBOSITY", platform=platform, default=""
     )
+    prerelease_pythons = args.prerelease_pythons or cibuildwheel.util.strtobool(
+        get_option_from_environment("CIBW_PRERELEASE_PYTHONS", platform=platform, default="0")
```
henryiii, line 216: Grrr, this will either require it be added to the pyproject.toml (yuck, not at all a good idea for a flag like this), or removing the platform specific choices, or adding custom code back to handle platform choices but no pyproject.toml support for #684. Does it have to support platform specific variants?

henryiii (approved):

cibuildwheel/__main__.py
mayeut: I guess no platform specific variant is not an issue for this one (and shall definitely not be in `pyproject.toml`)

joerick: I hadn't fully considered the pyproject.toml angle, but I think you're right, it would be kinda crazy to write this into a `pyproject.toml`! I could imagine that perhaps this option would only be specifiable at runtime from flag or env var, if that makes you feel better henryiii? I could imagine in the docs something like-

> Only configurable at runtime via `--prerelease-pythons` or `CIBW_PRERELEASE_PYTHONS`. Cannot be set in `pyproject.toml`.

It's interesting that `CIBW_OUTPUT_DIR` is another one of these runtime-only options? (maybe you've discussed this in #547, apologies, I'm behind!)

I'm not so worried about the platform-specific thing here, it seems pretty harmless.

henryiii: Okay, I'll remove the variants for now.

> `CIBW_OUTPUT_DIR`

This is undocumented currently, actually. It's discussed in #547, though it's a bit of an open question. I had it in the toml then I took it out, but could put it back in.

henryiii:
I'd love to get #684 into an alpha release too, so it's easier to start testing it in some projects. Though using GHA you can pin a commit, so can still test if there's no alpha for it. And I guess there could be an alpha 1 and an alpha 2, there's no cost for alphas...

PR #682, opened 2021-05-17
chore: rename joerick/cibuildwheel -> pypa/cibuildwheel

henryiii:
Doesn't change AppVeyor, might need to be special cased.

PR #683, opened 2021-05-19
Empty PR to test CI

joerick:


joerick:
#682 is a better place for me to track this, because it has code changes.

PR #684, opened 2021-05-22
feat: configuration support

henryiii:
Closes #547.

Design:

`tool.cibuildwheel` holds options with a structure that exactly matches the environment variables, just with dashes used instead of `_` to match TOML style. `tool.cibuildwheel.<platform>` holds the platform specific commands; these are identical but override the global value if defined and running on that platform. The defaults are defined in `cibuildwheel/resources/defaults.toml`. Environment variables override the pyproject.toml values, which override the defaults.toml values. An error is thrown if an unexpected option is found inside `tool.cibuildwheel` or a section based on what was found in the defaults.toml file.

`CIBW_PROJECT_REQUIRES_PYTHON` does not participate, since it has a canonical location in the pyproject.toml file already, and `CIBW_PLATFORM` really is only intended to be non-static, so it also does not participate. I have also removed `CIBW_OUTPUT_DIR`; having that "hidden" in a config file might be confusing? It's already a command line option, an action option, and an environment variable. Also this avoids having a way to specify it per platform.

I have not added a check to see if `manylinux-*` is in windows/macOS sections (it would do nothing if placed there), or `dependency-versions` in linux. Those could be added later.

Tables and arrays automatically stringify using `sep`, so those are recommended but not required for use for things like environment and extras. Strings always behave exactly the same as in the environment variables.

`defaults.toml`, which serves as the canonical source for defaults, and an example:

```toml
[tool.cibuildwheel]
build = "*"
skip = ""
test-skip = ""

archs = ["auto"]
dependency-versions = "pinned"
environment = {}
build-verbosity = ""

before-all = ""
before-build = ""
repair-wheel-command = ""

test-command = ""
before-test = ""
test-requires = []
test-extras = []

manylinux-x86_64-image = "manylinux2010"
manylinux-i686-image = "manylinux2010"
manylinux-aarch64-image = "manylinux2014"
manylinux-ppc64le-image = "manylinux2014"
manylinux-s390x-image = "manylinux2014"
manylinux-pypy_x86_64-image = "manylinux2010"
manylinux-pypy_i686-image = "manylinux2010"
manylinux-pypy_aarch64-image = "manylinux2014"


[tool.cibuildwheel.linux]
repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"

[tool.cibuildwheel.macos]
repair-wheel-command = [
  "delocate-listdeps {wheel}",
  "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}",
]

[tool.cibuildwheel.windows]
```

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -16,7 +16,7 @@ import cibuildwheel.util
 import cibuildwheel.windows
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.environment import EnvironmentParseError, parse_environment
-from cibuildwheel.options import ConfigNamespace, ConfigOptions
+from cibuildwheel.options import ConfigOptions
 from cibuildwheel.projectfiles import get_requires_python_str
 from cibuildwheel.typing import PLATFORMS, PlatformName, assert_never
 from cibuildwheel.util import (
@@ -147,12 +147,12 @@ def main() -> None:
     output_dir = Path(
         args.output_dir
         if args.output_dir is not None
-        else options("output-dir", namespace=ConfigNamespace.MAIN)
+        else os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse")
     )
 
-    build_config = options("build", namespace=ConfigNamespace.MAIN) or "*"
-    skip_config = options("skip", namespace=ConfigNamespace.MAIN)
-    test_skip = options("test-skip", namespace=ConfigNamespace.MAIN)
+    build_config = options("build", env_plat=False) or "*"
+    skip_config = options("skip", env_plat=False)
+    test_skip = options("test-skip", env_plat=False)
 
     archs_config_str = options("archs") if args.archs is None else args.archs
 
@@ -262,7 +262,7 @@ def main() -> None:
         ]:
             pinned_images = all_pinned_docker_images[build_platform]
 
-            config_value = options(f"{build_platform}-image", namespace=ConfigNamespace.MANYLINUX)
+            config_value = options(f"manylinux.{build_platform}-image")
 
             if config_value is None:
                 # default to manylinux2010 if it's available, otherwise manylinux2014
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -16,7 +16,7 @@ import cibuildwheel.util
 import cibuildwheel.windows
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.environment import EnvironmentParseError, parse_environment
-from cibuildwheel.options import ConfigOptions
+from cibuildwheel.options import ConfigNamespace, ConfigOptions
 from cibuildwheel.projectfiles import get_requires_python_str
 from cibuildwheel.typing import PLATFORMS, PlatformName, assert_never
 from cibuildwheel.util import (
@@ -71,7 +71,6 @@ def main() -> None:
 
     parser.add_argument(
         "--output-dir",
-        default=os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse"),
         help="Destination folder for the wheels.",
     )
 
@@ -143,20 +142,24 @@ def main() -> None:
         sys.exit(2)
 
     package_dir = Path(args.package_dir)
-    output_dir = Path(args.output_dir)
 
     options = ConfigOptions(package_dir, platform=platform)
+    output_dir = Path(
+        args.output_dir
+        if args.output_dir is not None
+        else options("output-dir", namespace=ConfigNamespace.MAIN)
+    )
 
-    build_config = options("build", platform_variants=False) or "*"
-    skip_config = options("skip", platform_variants=False)
-    test_skip = options("test-skip", platform_variants=False)
+    build_config = options("build", namespace=ConfigNamespace.MAIN) or "*"
+    skip_config = options("skip", namespace=ConfigNamespace.MAIN)
+    test_skip = options("test-skip", namespace=ConfigNamespace.MAIN)
 
     archs_config_str = options("archs") if args.archs is None else args.archs
 
     environment_config = options("environment")
     before_all = options("before-all")
     before_build = options("before-build")
-    repair_command = options("repair-command")
+    repair_command = options("repair-wheel-command")
 
     dependency_versions = options("dependency-versions")
     test_command = options("test-command")
@@ -259,8 +262,7 @@ def main() -> None:
         ]:
             pinned_images = all_pinned_docker_images[build_platform]
 
-            config_name = f"CIBW_MANYLINUX_{build_platform.upper()}_IMAGE"
-            config_value = os.environ.get(config_name)
+            config_value = options(f"{build_platform}-image", namespace=ConfigNamespace.MANYLINUX)
 
             if config_value is None:
                 # default to manylinux2010 if it's available, otherwise manylinux2014
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -262,7 +262,7 @@ def main() -> None:
         ]:
             pinned_images = all_pinned_docker_images[build_platform]
 
-            config_value = options(f"manylinux.{build_platform}-image")
+            config_value = options(f"manylinux-{build_platform}-image")
 
             if config_value is None:
                 # default to manylinux2010 if it's available, otherwise manylinux2014
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -185,7 +185,7 @@ def main() -> None:
 
     archs_config_str = args.archs or options("archs", sep=" ")
 
-    environment_config = options("environment", sep=" ")
+    environment_config = options("environment", table={"item": '{k}="{v}"', "sep": " "})
     before_all = options("before-all", sep=" && ")
     before_build = options("before-build", sep=" && ")
     repair_command = options("repair-wheel-command", sep=" && ")
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -5,7 +5,7 @@ import textwrap
 import traceback
 from configparser import ConfigParser
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Union, overload
+from typing import Dict, List, Optional, Set, Union
 
 from packaging.specifiers import SpecifierSet
 
@@ -16,6 +16,7 @@ import cibuildwheel.util
 import cibuildwheel.windows
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.environment import EnvironmentParseError, parse_environment
+from cibuildwheel.options import ConfigOptions
 from cibuildwheel.projectfiles import get_requires_python_str
 from cibuildwheel.typing import PLATFORMS, PlatformName, assert_never
 from cibuildwheel.util import (
@@ -27,40 +28,6 @@ from cibuildwheel.util import (
     detect_ci_provider,
     resources_dir,
 )
-
-
-@overload
-def get_option_from_environment(
-    option_name: str, *, platform: Optional[str] = None, default: str
-) -> str:
-    ...  # noqa: E704
-
-
-@overload
-def get_option_from_environment(
-    option_name: str, *, platform: Optional[str] = None, default: None = None
-) -> Optional[str]:
-    ...  # noqa: E704 E302
-
-
-def get_option_from_environment(
-    option_name: str, *, platform: Optional[str] = None, default: Optional[str] = None
-) -> Optional[str]:  # noqa: E302
-    """
-    Returns an option from the environment, optionally scoped by the platform.
-
-    Example:
-      get_option_from_environment('CIBW_COLOR', platform='macos')
-
-      This will return the value of CIBW_COLOR_MACOS if it exists, otherwise the value of
-      CIBW_COLOR.
-    """
-    if platform:
-        option = os.environ.get(f"{option_name}_{platform.upper()}")
-        if option is not None:
-            return option
-
-    return os.environ.get(option_name, default)
 
 
 def main() -> None:
@@ -178,38 +145,26 @@ def main() -> None:
     package_dir = Path(args.package_dir)
     output_dir = Path(args.output_dir)
 
-    if platform == "linux":
-        repair_command_default = "auditwheel repair -w {dest_dir} {wheel}"
-    elif platform == "macos":
-        repair_command_default = "delocate-listdeps {wheel} && delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}"
-    elif platform == "windows":
-        repair_command_default = ""
-    else:
-        assert_never(platform)
+    options = ConfigOptions(package_dir, platform=platform)
 
-    build_config = os.environ.get("CIBW_BUILD") or "*"
-    skip_config = os.environ.get("CIBW_SKIP", "")
-    test_skip = os.environ.get("CIBW_TEST_SKIP", "")
-    environment_config = get_option_from_environment(
-        "CIBW_ENVIRONMENT", platform=platform, default=""
-    )
-    before_all = get_option_from_environment("CIBW_BEFORE_ALL", platform=platform, default="")
-    before_build = get_option_from_environment("CIBW_BEFORE_BUILD", platform=platform)
-    repair_command = get_option_from_environment(
-        "CIBW_REPAIR_WHEEL_COMMAND", platform=platform, default=repair_command_default
-    )
-    dependency_versions = get_option_from_environment(
-        "CIBW_DEPENDENCY_VERSIONS", platform=platform, default="pinned"
-    )
-    test_command = get_option_from_environment("CIBW_TEST_COMMAND", platform=platform)
-    before_test = get_option_from_environment("CIBW_BEFORE_TEST", platform=platform)
-    test_requires = get_option_from_environment(
-        "CIBW_TEST_REQUIRES", platform=platform, default=""
-    ).split()
-    test_extras = get_option_from_environment("CIBW_TEST_EXTRAS", platform=platform, default="")
-    build_verbosity_str = get_option_from_environment(
-        "CIBW_BUILD_VERBOSITY", platform=platform, default=""
-    )
+    build_config = options("build", platform_variants=False) or "*"
+    skip_config = options("skip", platform_variants=False)
+    test_skip = options("test-skip", platform_variants=False)
+
+    archs_config_str = options("archs") if args.archs is None else args.archs
+
+    environment_config = options("environment")
+    before_all = options("before-all")
+    before_build = options("before-build")
+    repair_command = options("repair-command")
+
+    dependency_versions = options("dependency-versions")
+    test_command = options("test-command")
+    before_test = options("before-test")
+    test_requires = options("test-requires").split()
+    test_extras = options("test-extras")
+    build_verbosity_str = options("build-verbosity")
+
     prerelease_pythons = args.prerelease_pythons or cibuildwheel.util.strtobool(
         os.environ.get("CIBW_PRERELEASE_PYTHONS", "0")
     )
@@ -218,11 +173,11 @@ def main() -> None:
 
     if not any(package_dir.joinpath(name).exists() for name in package_files):
         names = ", ".join(sorted(package_files, reverse=True))
-        print(
-            f"cibuildwheel: Could not find any of {{{names}}} at root of package", file=sys.stderr
-        )
+        msg = f"cibuildwheel: Could not find any of {{{names}}} at root of package"
+        print(msg, file=sys.stderr)
         sys.exit(2)
 
+    # This is not supported in tool.cibuildwheel, as it comes from a standard location.
     # Passing this in as an environment variable will override pyproject.toml, setup.cfg, or setup.py
     requires_python_str: Optional[str] = os.environ.get(
         "CIBW_PROJECT_REQUIRES_PYTHON"
@@ -269,13 +224,6 @@ def main() -> None:
     # Add CIBUILDWHEEL environment variable
     # This needs to be passed on to the docker container in linux.py
     os.environ["CIBUILDWHEEL"] = "1"
-
-    if args.archs is not None:
-        archs_config_str = args.archs
-    else:
-        archs_config_str = get_option_from_environment(
-            "CIBW_ARCHS", platform=platform, default="auto"
-        )
 
     archs = Architecture.parse_config(archs_config_str, platform=platform)
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -157,15 +157,15 @@ def main() -> None:
     archs_config_str = options("archs") if args.archs is None else args.archs
 
     environment_config = options("environment")
-    before_all = options("before-all")
-    before_build = options("before-build")
-    repair_command = options("repair-wheel-command")
+    before_all = options("before-all", sep=" && ")
+    before_build = options("before-build", sep=" && ")
+    repair_command = options("repair-wheel-command", sep=" && ")
 
     dependency_versions = options("dependency-versions")
-    test_command = options("test-command")
-    before_test = options("before-test")
+    test_command = options("test-command", sep=" && ")
+    before_test = options("before-test", sep=" && ")
     test_requires = options("test-requires").split()
-    test_extras = options("test-extras")
+    test_extras = options("test-extras", sep=",")
     build_verbosity_str = options("build-verbosity")
 
     prerelease_pythons = args.prerelease_pythons or cibuildwheel.util.strtobool(
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -36,8 +36,9 @@ def main() -> None:
     parser = argparse.ArgumentParser(
         description="Build wheels for all the platforms.",
         epilog="""
-            Most options are supplied via environment variables.
-            See https://github.com/pypa/cibuildwheel#options for info.
+            Most options are supplied via environment variables or in
+            --config-file (pyproject.toml usually). See
+            https://github.com/pypa/cibuildwheel#options for info.
         """,
     )
 
@@ -76,8 +77,11 @@ def main() -> None:
 
     parser.add_argument(
         "--config-file",
-        help="TOML config file for cibuildwheel; usually pyproject.toml, but can be overridden with this option. Use {project} for the project directory",
-        default="{project}/pyproject.toml",
+        help="""
+            TOML config file for cibuildwheel; usually pyproject.toml, but
+            can be overridden with this option. Use {package} for the package
+            directory.""",
+        default="{package}/pyproject.toml",
     )
 
     parser.add_argument(
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -89,10 +89,9 @@ def main() -> None:
     parser.add_argument(
         "--config-file",
         help="""
-            TOML config file for cibuildwheel; usually pyproject.toml, but
-            can be overridden with this option. Use {package} for the package
-            directory.""",
-        default="{package}/pyproject.toml",
+            TOML config file for cibuildwheel. Defaults to pyproject.toml, but
+            can be overridden with this option.
+        """,
     )
 
     parser.add_argument(
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -29,6 +29,17 @@ from cibuildwheel.util import (
     resources_dir,
 )
 
+MANYLINUX_ARCHS = (
+    "x86_64",
+    "i686",
+    "pypy_x86_64",
+    "aarch64",
+    "ppc64le",
+    "s390x",
+    "pypy_aarch64",
+    "pypy_i686",
+)
+
 
 def main() -> None:
     platform: PlatformName
@@ -153,7 +164,15 @@ def main() -> None:
 
     package_dir = Path(args.package_dir)
 
-    options = ConfigOptions(package_dir, args.config_file, platform=platform)
+    manylinux_identifiers = {
+        f"manylinux-{build_platform}-image" for build_platform in MANYLINUX_ARCHS
+    }
+    disallow = {
+        "linux": {"dependency-versions"},
+        "macos": manylinux_identifiers,
+        "windows": manylinux_identifiers,
+    }
+    options = ConfigOptions(package_dir, args.config_file, platform=platform, disallow=disallow)
     output_dir = Path(
         args.output_dir
         if args.output_dir is not None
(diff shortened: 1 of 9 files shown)


henryiii:
@pypa/cibuildwheel-team Before I start working on docs, maybe would be a good idea to look this over?

henryiii:
Also, not sure why @mayeut is not listed in the above team? @mayeut [are you on discord](https://discuss.python.org/t/theres-a-pypa-discord-now/8515)? There's a \#cibuildwheel channel there.

mayeut:
> @mayeut are you on discord?

I connected once when I saw the thread on https://discuss.python.org. I've just spin it up again, seems there's quite some activity there, I'll try to connect more often.

mayeut:
@henryiii,
Thanks for drafting this. I'm still playing a bit with this, I will follow up here with remarks shortly.

mayeut:
Looks good overall but I do have 1 major concern for now: a definition in the `tool.cibuildwheel.global` section of `pyproject.toml` does not override the definition in `tool.cibuildwheel.{platform}` section of `default.toml` (`repair-wheel-command` is the only one for now).

Minor concern: I find the `tool.cibuildwheel.manylinux` section odd given a `tool.cibuildwheel.linux` section exists. I understand it would not be 1<->1 with how env vars are working now but looking at a configuration file, I'd expect to find them in `tool.cibuildwheel.linux` section.

Here are some tests I added to check how things behaved, the `test_project_global_override_default_platform` is not passing (my major concern), I'll check if this can be fixed quickly and add a commit with those tests if successful.
```python
def test_project_global_override_default_platform(tmp_path, platform):
    tmp_path.joinpath("pyproject.toml").write_text(
        """
[tool.cibuildwheel.global]
repair-wheel-command = "repair-project-global"
"""
    )
    options = ConfigOptions(tmp_path, platform=platform)
    assert options("repair-wheel-command") == "repair-project-global"


def test_env_global_override_default_platform(tmp_path, platform, monkeypatch):
    monkeypatch.setenv("CIBW_REPAIR_WHEEL_COMMAND", "repair-env-global")
    options = ConfigOptions(tmp_path, platform=platform)
    assert options("repair-wheel-command") == "repair-env-global"


def test_env_global_override_project_platform(tmp_path, platform, monkeypatch):
    monkeypatch.setenv("CIBW_REPAIR_WHEEL_COMMAND", "repair-env-global")
    tmp_path.joinpath("pyproject.toml").write_text(
        """
[tool.cibuildwheel.linux]
repair-wheel-command = "repair-project-linux"
[tool.cibuildwheel.windows]
repair-wheel-command = "repair-project-windows"
[tool.cibuildwheel.macos]
repair-wheel-command = "repair-project-macos"
"""
    )
    options = ConfigOptions(tmp_path, platform=platform)
    assert options("repair-wheel-command") == "repair-env-global"

```

cibuildwheel/options.py
```diff
@@ -0,0 +1,159 @@
...
+
+            # This is recursive; update dicts (subsections) if needed. Only handles one level.
+            if isinstance(new_dict[key], dict):
+                if path:
+                    raise ConfigOptionError(
+                        f"Nested keys not supported, {key} should not be in {path}"
+                    )
+
+                if key not in old_dict:
+                    old_dict[key] = {}
```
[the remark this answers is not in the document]
mayeut: This might not be needed with the check done a bit earlier but, better safe than sorry ?

henryiii:
Ideally, if we didn't match the environment variable exactly (I plan to list the configuration option along with the environment variable in the help), here are some ideas:

```toml
# Current (exact match):
[tool.cibuildwheel.manylinux]
x86_64-image = "manylinux2010"
i686-image = "manylinux2010"

# Idea 1: avoid repeating "image":
[tool.cibuildwheel.linux.image]
x86_64 = "manylinux2010"
i686 = "manylinux2010"

# Idea 2: Inside linux directly:
[tool.cibuildwheel.linux]
x86_64-image = "manylinux2010"
i686-image = "manylinux2010"
# Options are in the same "level"
repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"<br class="Apple-interchange-newline"
```

Thoughts, @pypa/cibuildwheel-team ?

cibuildwheel/options.py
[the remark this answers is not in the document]
henryiii: I _think_ this is covered, but we need to add some "this should throw an error" type tests to be sure.

cibuildwheel/resources/defaults.toml
```diff
@@ -0,0 +1,39 @@
...
+build = "*"
+skip = ""
+test-skip = ""
+output-dir = "wheelhouse"
+
+
+[tool.cibuildwheel.manylinux]
+x86_64-image = "manylinux2010"
+i686-image = "manylinux2010"
+pypy_x86_64-image = "manylinux2010"
```
henryiii, line 11:
```suggestion
pypy_x86_64-image = "manylinux2010"
pypy_aarch64-image = "manylinux2010"
```

henryiii:
Actually, thinking about musllinux, this might be best:

```
[tool.cibuildwheel]
manylinux-x86_64-image = "manylinux2010"
manylinux-i686-image = "manylinux2010"
```

That leaves room for musllinux with overlapping archs.

henryiii:
Also, is "global" best, or would "all" be better?

henryiii:
What if we just allowed every option to be platform specific, including build and skip and such? Then we could just drop "all/global" and `tool.cibuildwheel.<plat>.X` would always override `tool.cibuildwheel.X` for all `X`. It might be a little silly to customize the output directory per OS, but it would be simple and consistent.

henryiii:
Another thought: environment should be a toml dict.

joerick:
Sorry for lack of comms on this, I won't be able to give this proper time for the next few days - apologies!

mayeut:
> This whole thing could go into Linux, though

@henryiii, I like you're proposal for `manylinux`. I was going to bring up `musllinux` but you beat me to it.

> Also, is "global" best, or would "all" be better?
> What if we just allowed every option to be platform specific, including build, skip, test-skip, and output-dir? Then we could just drop "all/global" and tool.cibuildwheel.<plat>.X would always override tool.cibuildwheel.X for all X. It might be a little silly to customize the output directory per OS, but it would be a simple and consistent rule with no surprises. (And we could also not allow output dir to be in the toml)

I think not having "global" or "all" would be better: it seems implied that options in `tool.cibuildwheel` would be global when reading a configuration file.
Having said that, your comment about allowing every option to be platform specific raises a few questions:
- would it change the current environment variables (probably not, it would just extend the ones already supported ?)
- how to disallow specifying a manylinux image on windows/macOS (i.e. have some options really specific to a platform) ?
- I've not checked the relation with command line args yet, I would assume command line arg overrides env var overrides config file ?

henryiii:
> Sorry for lack of comms on this, I won't be able to give this proper time for the next few days - apologies!

No problem @joerick , it just means you'll get a more polished version to review when you do! :)

> would it change the current environment variables (probably not, it would just extend the ones already supported ?)

It wouldn't have to, we could leave `CIBW_SKIP_LINUX` from working if we want; but `tool.cibuildwheel.linux.skip` would work. It is much easier to explain if anything can be placed in "<plat>", it a bit more useful for static configs, and while people usually look up env vars that are available, for static configs, they usually go from an example.

> how to disallow specifying a manylinux image on windows/macOS (i.e. have some options really specific to a platform) ?

Probably could just be a static check to disallow setting these? Honestly, if someone did set `tool.cilbuildwheel.windows.manylinux.x86_64-image`, I would also be okay with just considering it unused and ignoring it. 🤦 

> I've not checked the relation with command line args yet, I would assume command line arg overrides env var overrides config file ?

Yes (I'm not sure about cl args vs env var, but I think I do whatever we currently do there).

henryiii:
The table/list implementation is quite simple. :) For each option, you just describe how to merge a list into a string (and tables are slightly hard coded when formatting, but I think that's fine, as they are only for environments). We could possibly even expand this to make nice lists for commands, `sep=";"` or `sep="&&"`, perhaps, depending on the OS too?

henryiii:
Docs at https://cibuildwheel--684.org.readthedocs.build/en/684/options/. @joerick can we set up synced tabs? Like seen on https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives , where if you select one item, all matching tabs also change.

henryiii:
Remaining items:

* Probably should have more tests. Happy to accept help with making good tests. ;P
* Should we explicitly disallow joining arrays for fields that don't need it? All fields can be written as an array currently, and will be joined with a space if not requested otherwise. It could be handy in some of the cases for readability, like `build = ["cp*", "pp*"]`; you could have one per line, etc.
* Should we disallow "bad" platform settings? manylinux in windows/macOS, dependency-versions in linux?
* Is it best to keep `tool.cibuildwheel.output-dir` out? I don't think it's even in our help file already in environment form.
* Maybe we could put a mention of the config in a few more places.

henryiii:
1.x:

```console
CIBW_BUILD='cp39-*' \
CIBW_MANYLINUX_X86_64_IMAGE=dockcross/manylinux1-x64 \
CIBW_MANYLINUX_I686_IMAGE=dockcross/manylinux1-x86 \
CIBW_BEFORE_BUILD="pip install -r requirements-repair.txt" \
CIBW_REPAIR_WHEEL_COMMAND="python scripts/repair_wheel.py -w {dest_dir} {wheel}" \
CIBW_TEST_REQUIRES="-r requirements-test.txt" \
CIBW_TEST_COMMAND="pytest {project}/tests" \
CIBW_ENVIRONMENT_LINUX="SKBUILD_CONFIGURE_OPTIONS='-DOPENSSL_ROOT_DIR:PATH=/usr/local/ssl -DCMAKE_JOB_POOL_COMPILE:STRING=compile -DCMAKE_JOB_POOL_LINK:STRING=link -DCMAKE_JOB_POOLS:STRING=compile=2;link=1'" \
pipx run cibuildwheel --platform linux
```

Current PR:

```toml
[tool.cibuildwheel]
build = "cp39-*"
manylinux.x86_64-image = "dockcross/manylinux1-x64"
manylinux-i686-image = "dockcross/manylinux1-x86"
before-build = "pip install -r requirements-repair.txt"
repair-wheel-command = "python scripts/repair_wheel.py -w {dest_dir} {wheel}"
test-requires = "-r requirements-test.txt"
test-command = "pytest {project}/tests"

[tool.cibuildwheel.linux.environment]
SKBUILD_CONFIGURE_OPTIONS = "-DOPENSSL_ROOT_DIR:PATH=/usr/local/ssl -DCMAKE_JOB_POOL_COMPILE:STRING=compile -DCMAKE_JOB_POOL_LINK:STRING=link -DCMAKE_JOB_POOLS:STRING=compile=2;link=1" 
```

```console
pipx run cibuildwheel --platform linux
```

mayeut:
I'll try to look at this a bit today. At least, I'll try to get some more tests in there.

joerick:
I haven't gotten into the implementation yet, just the design and the documentation, as I think these are the most important parts to get right, as the implementation can flow from decisions made at that level.

The biggest challenge for this is going to be the documentation. We don't want the addition of the pyproject.toml to make the docs any more confusing, either for new users or existing users of environment variables.

In general, I think this is looking pretty good, design-wise. Putting the global/all-platform options under `[tool.cibuildwheel]` and platform ones under `[tool.cibuildwheel.linux]` makes lots of sense, and is what I'd expect, I think. 

Here are some things on my mind, having thought about this for a couple hours...

- (design) The manylinux images option... I think either we should leave it more-or-less as-is (just strings called `manylinux-x86_64-image = "manylinux2010"`) or change it completely, to be `manylinux-image = { x86_64: "manylinux2010", ... }`. I'd lean toward the former, just leaving as a collection of simple string options, because I think it's easier to understand the duality between the env vars and the pyproject.toml options, especially in the documentation.
- (docs) ~Putting manylinux images into their own section of the pyproject.toml is also a little weird, as they only affect the linux build, so they should be either globally scoped, or within `[tool.cibuildwheel.linux]` - probably both should work.~ Ah, I see that this isn't a section, it's a toml dict. Hmm. Even though I understand how TOML works, this is very confusing for me, because `manylinux` looks like a sibling of `linux` when written like that i.e. a different platform. I think we should avoid using any TOML headers other than `[tool.cibuildwheel]` and `[tool.cibuildwheel.platform]` in the documentation. Pro TOML users would still be able to use these other dict forms, but most just want to write ini-style configs I think.
- (docs) on that note, I noticed that the `archs` docs is using another TOML feature that confused me a little - `macos.archs: ["x86_64", "universal2", "arm64"]`. The dotted keypath is neat, but I don't think it belongs in  reference documentation. The name of the option is `archs`, so we should have examples that say `archs = ...`. Nesting them within a `[tool.cibuildwheel.<platform>]` will help reinforce that concept for users, too.
- (design) I like the list variant of some of the options - `archs` is a good example. `environment` also works great as an inline dict. Maybe we should also include `test_requires` in that. But I'm less sure about the commands being lists, like `repair_command`. This was surprising, because we don't think about 'commands' being 'lists'. And I don't think it's worth the extra headscratching when users first come across it.
- (docs) on the topic of the list-style configs, lots of options are already documented as 'a space-separated list of…'. Maybe that should be changed to 'a list of…' and we can lean on the examples to show how that differs between environment variables and pyproject.toml.
- (docs) I think it's _probably_ the right approach to keep env var as primary (in the titles and the nav), as you have done. 
- (docs) I can't exactly put my finger on it, but the examples aren't as easy to read in this new form. Maybe it's a visual design thing. One minor thing that might improve this would be to restore the "Examples" header and then make the tabs called "Environment variables" and "pyproject.toml".

Hope this is helpful. Sorry I haven't had more time to dig into this properly!

henryiii:
> The manylinux images option

Yes, would (slightly reluctantly) be okay with that. When we add musllinux, that would be a different "section", even though it's really very similar. I like "." better than "-" if something is really "part of" something - the different images are logically part of manylinux. Note that they can be placed in "linux" as well, we could always document them that way.

> But I'm less sure about the commands being lists

Multiline commands are currently rather irritating to work with, requiring manual chaining - I thought this seemed like a good way to write them - and it has a strong precedent in YAML, where Travis and others allows a list of commands.

I intentionally showed different ways to write things in the examples, rather than keeping a consistent "recommended" style. Though we can converge on one.

One key "quirk" of the pyproject.toml examples is that they have a single header at the top, but then show lots of examples of a value being set without putting a header on each one. It makes it look a bit like it's a complete file, rather than multiple examples.

henryiii:
I'm likely not going to be able to do much today, but my next target will be to work on the docs items above, standardizing to something like this:

```toml
[tool.cibuildwheel]
build = "*"
environment = { MY_ENV_VAR = "hi" }


[tool.cibuildwheel.linux]
manylinux.x86_64-image = "manylinux2010" # could change to dash?
repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"

[tool.cibuildwheel.macos]
repair-wheel-command = [ # Could use && and drop the list?
  "delocate-listdeps {wheel}",
  "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}",
]
```

Leaving the questions:
* [ ] manylinux. or manylinux-? Include the -image?
* [ ] Are you still against showing list commands? Or perhaps even allowing them?

henryiii:
On `manylinux.` vs. `manylinux-`, it's still a 1-1 mapping to go from config to environment variables (dots and dashes all go to underscores, that's literally how it's programmed), and the "." gives pro users the ability to write this:

```toml

[tool.cibuildwheel]
build = "*"

[tool.cibuildwheel.linux.manylinux]
x86_64-image = "manylinux2010"
i686-image = "manylinux2010"
pypy_x86_64-image = "manylinux2010"
pypy_i686-image = "manylinux2010"
pypy_aarch64-image = "manylinux2010"
aarch64-image = "manylinux2014"
ppc64le-image = "manylinux2014"
s390x-image = "manylinux2014"

[tool.cibuildwheel.linux.musllinux]
x86_64-image = "musllinux_x_y"
i686-image = "musllinux_x_y"
```

I'm not worried about the extra characters (writing manylinux many times, or -image), but the grouping here is quite nice and elegant "ini" style, IMO.

henryiii:
Did a docs update, and now only allow joining if requested. Marked the things (hopefully) addressed above.

joerick:
> Leaving the questions:
> 
> * [ ]  manylinux. or manylinux-? Include the -image?

Having thought more on this, I'm more sure that keeping it flat, like `manylinux-x86_64-image`, is the right way to go. It's the only example of our options 'changing shape' in a non-trivial way. I think that the 'pro TOML user' advantage is slightly cool, but I don't think it's worth making the manylinux images a different 'shape'. The problems with different 'shapes' are:
- it's harder to document- the name of that section `CIBW_MANYLINUX_*_IMAGE` doesn't correlate with how it looks in pyproject.toml. 
- it's harder for a user to move config between pyproject.toml and env vars.

Or, to put it another way, if we were going to change the option to be e.g 

```toml
manylinux-images = { x86_64: "manylinux2010", aarch64: "manylinux2014" }
```
then I think we should be changing the environment variable also to be like:
```yaml
CIBW_MANYLINUX_IMAGES: "x86_64:manylinux2010 aarch64:manylinux2014"
```

...which might be a good idea, but it's probably not worth the interface churn, especially for an option that isn't all that widely used.

> * [ ]  Are you still against showing list commands? Or perhaps even allowing them?

To be honest, I'm pretty torn on this one. I agree that it would be lovely to have a good, documented way to do multiline commands.

Perhaps another way would be to document something like:

```toml
before_build = """
  set -o errexit
  command_1
  command_2
"""
```

(That's probably nicer to write, too, since there's no "-quotes and commas getting in the way.)

After all, it is a just a shell script that's passed to /bin/sh. We could even add `-o errexit` when calling the commands, so the user doesn't have to. Not sure what to do on Windows, though. Perhaps there's a CMD feature that can exit on first error? Or on Windows we could munge the script to insert `&&`s on newlines?

> * [ ]  Any ideas for making the examples nicer? Should maybe the headers be included each time, or removed, for example?

Ah, sorry, I guess I wasn't clear - I wasn't referring to the contents of the TOML examples. I think it's just a visual design thing, actually. I'd say we want the h4 "examples" header to still be there, and then the tabs shouldn't indent the examples code block. But it's something that I can fix later on, if you like.

For the TOML examples, I think including the headers each time is good, yeah. I like that. It reinforces that you're in pyproject.toml, and that those headers are required for the option to work.

---

Another high-level design idea - would it be useful for the user to be able to store the options in a file other than pyproject.toml, and do `pipx run cibuildwheel -c path/to/config-file.toml`? I suppose I'm thinking about projects where the build config lives separately from the code, sometimes the builder can't modify the code itself. I'd say we'd keep the `[tool.cibuildwheel]` headers for consistency of documentation.

henryiii:
> then I think we should be changing the environment variable also to be like

Environment variables are flat. There's no correlation to a nested variable in TOML - and in fact, toml _forces_ at least one level of nesting (you can't have a raw value outside of a section, more on that later). Smashing all these into a single environment variable would be awful, I think, while it's fine with TOML, since they are not forced to be defined in one go, but can be separately set items with structure. That's a benefit of TOML, as it lets you set items one at a time, "INI" style.

I don't see any harm with A_B_C -> a.b.c; that is clearly the same option and reads the same. In fact, I was considering using `test.command`, etc., to provide more structure; then "_" would almost always translate to a ".". The problem was before-build, which really isn't "before.build"; some work would have to be done to make these logically nest so that it is always "CIBW_WHEN_WHAT" (`CIBW_BEFORE_BUILD_COMMAND` -> `before-build.command` in this case, I think), so it's not worth the change.

Also, I expect most users to look up values in either the "defaults" example or in the examples, and not worry that much about remembering the exact spelling - that only happens once, when the file is first written. The most important thing is in making the file readable and maintainable, and giving pro users more options would help there. I would be rather picky about styling in my pyproject.toml, as that's becoming the "all-in-one" config for the entire package, and readability matters as the file grows, not whether I had to look up if an option has a dash or a dot. You'll get a clear error if there are unexpected keys.

But, as I said, I'm fine to go with whatever you want here, so I can change it to `manylinux-`.

> I'm pretty torn on this one

To be clear, there's an absolute rule here I do not want to give up: if you set a string in the TOML, it's _EXACTLY THE SAME_ as in the environment variable form. If you copy-paste the config values from your YAML in to the correct keys, it will work identically. You don't have to use lists or tables for any of the values, they all can be written as strings. Tables and strings are more natural, and always shown in the examples, but not required. So a multiline string in TOML is identical to the current multiline string in YAML, and should remain that way. ~~I've seen multiline strings in people's YAML for commands before in the wild, but I forget how it's handled, but that should remain the same.~~ 
Ahh, [found one](https://github.com/dib-lab/kProcessor/blob/e35156855257d63113223e37b223b14a42e975b1/.github/workflows/cibuildwheel.yml#L39)

So it's supported, I'm just not sure how errors are handled.

PS: I checked for `CIBW_BEFORE_BUILD`; scrolling through a few of the 500 uses on GitHub I saw lots of `&&` and some `;`, but no multiline strings. People used the line chomping feature a bit to chain them together if they did multiline, which TOML doesn't have. I feel that, taking nabla-c0d3/nassl, for example, that

```yaml
CIBW_BEFORE_ALL: "python -m pip install invoke &amp;&amp; invoke build.deps"
CIBW_BEFORE_ALL_WINDOWS: '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvars64.bat" &amp;&amp; python -m pip install invoke &amp;&amp; invoke build.deps'
```

Would read much better as:

```toml
[tool.cibuildwheel]
before-all = [
  "python -m pip install invoke",
  "invoke build.deps"
]
[tool.cibuildwheel.windows]
before-all = [
  "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvars64.bat",
  "python -m pip install invoke",
  "invoke build.deps",
]
```

Than this:

```toml
[tool.cibuildwheel]
before-all = """
  python -m pip install invoke &&
  invoke build.deps
"""

[tool.cibuildwheel.windows]
before-all = """
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvars64.bat &&
  python -m pip install invoke &&
  invoke build.deps
"""
```

But if we can make sure that "&&" is not needed (possibly by just adding `set -o errexit` or whatever Windows needs if more than one line is detected?), and make the change for YAML too (if it is a change), then only using multiline strings is fine with me.

> But it's something that I can fix later on, if you like.

Yes, I can add the examples headers and add `[tool.cibuildwheel]` above each line, but if you could fix the nesting and make the tabs all change in sync, that would be great. (And could be done later).

---

> Another high-level design idea

Ohh, I originally had something very similar, but this is slightly different. In the first version (and intentionally easy to re-add), there was one more level of nesting. So it was default.toml -> pyproject.toml -> cibuildwheel.toml -> ENVIRIONMENT_VARIABLES. The cibuildwheel.toml was looked for in the project directory, and would override the pyproject.toml file. It had two purposes; some people might prefer not to pollute the pyproject.toml (though that's kind of what it was for), and sometimes you might build someone else's wheels and not be able to change the existing files. It wasn't a great solution for that later one, though; environment variables were better. And unlike "classic" packages (pytest, tox, etc), black does not provide a "black.ini" or "black.toml", and only supports pyproject.toml, so I left that off for now.

But your idea is slightly different. I have two questions: a) Does pyproject.toml get read if you pass `-c file.toml`? That is, would it only override/merge with pyproject.toml, or would it cause the pyproject.toml to be ignored? b) Would "cibuildwheel.toml" be the default? Or nothing? Or, perhaps, "pyproject.toml"? (likely depends on the answer to a).

I'd very strongly recommend the structure exactly match. You can't have raw values in TOML, it needs a heading, so you'd have `[cibuildwheel]` at least, then it would be _really_ confusing as to whether the 'tool.` was present - so always having `[tool.cibuildwheel]` is much simpler (besides making our parsing, and any automated parsing, much simpler).

henryiii:
I've pushed three changes:
* Changed `manylinux.` to `manylinux-`
* Made the above-listed changes to the docs, like "Examples" headers and the TOML section on each example line.
* Support setting the input file with `-c / --config-file`, defaults to `{project}/pyproject.toml`.

joerick:
I'm doing a bit of work at the moment to improve the styling of docs. Should get it done this week !

henryiii:
This is almost ready, remaining tasks:

- [ ] Check to see what multiline strings do, especially on Windows. See if lists for commands have a better solution. (Since we don't have line comping in TOML, common YAML tricks for this will not work)
- [ ] Add more tests
- [ ] Should we disallow "bad" platform settings? manylinux-* in windows/macOS, dependency-versions in linux?
- [ ] Maybe we could put a mention of the config in a few more places.

There was actually a feature in TOML 1 that allows `tool.cibuildwheel.*` to be a section title, which might be nice, but it's not supported in the `toml` library, which is stuck on TOML 0.5. Even if we moved to a different library, pip uses `toml`, so it would break pip to add such a section title. That might change soon, but older pips will still break: https://github.com/pypa/pip/issues/10034

joerick:
Just pushed some docs updates. The tabs now look like this-

image

Which is a bit neater and easier to read I think. 

> * Check to see what multiline strings do, especially on Windows. See if lists for commands have a better solution

We landed on `>` and `&&` for YAML and the line-style TOML variant of this, after a discussion on the Discord. I've included a reference to this in before_all, but we should probably include it in a few more examples.

mayeut:
Catching-up on all that was done (including reading discussion on discourse), one remaining question is:

> Should we disallow "bad" platform settings? manylinux-* in windows/macOS, dependency-versions in linux?

Given those are not platform specific at the moment using environment variables, I'd say it can come as a further enhancement when deemed necessary as it would probably add some complexity to  the option parsing.

Getting a generic/reusable way of doing that would be to depend on where options appear in the `default.toml` I guess ? Everything that appears only in "platform" section but not in the "generic / all" section would be specific to the platform it appears in ?

henryiii:
I'm happy to leave it for a future enhancement. Due to the implementation, it either would be a complete hack, or would add complexity. What I'd like to do is add an `allowed = {"linux", "macos", "windows"}` optional argument when you access these, and an option can specify if it only allows a smaller subset. However, since we merge the global / platform keys, we don't know where a key came from, so we can't tell if one of these came from the global location (which is fine) or from a disallowed platform (that would also only check the platform you were running on, but that's fine). So not worth the trouble now, I think. We don't have similar checking for environment variables, either.

The "hack" method would be to just hard wire it into to the constructor when it reads the file (possibly as an argument, which would be less hacky). `disallow = {'linux': {'dependency-versions'}, ...}` (or inverted), something like that. That's not that bad, so might make a quick attempt on that. But don't think it's critical.

henryiii:
@joerick The tabs are nice, but rather hard to see - if you didn't know they were tabs, it is very easy to miss. Maybe the space to the left of the tabs could be clear/bg color so they faintly look like they stick out?

Also, can we have all the tabs change in sync? If someone clicks "pyproject.toml", they shouldn't have to change it on every single box, once should be enough. Keeping the page steady during the change might be tricky?

Both things could be a later improvement.

joerick:
> if you didn't know they were tabs, it is very easy to miss. Maybe the space to the left of the tabs could be clear/bg color so they faintly look like they stick out?

yeah.... it's a delicate act, balancing interface and content... my feeling before was that the tabs were too prominent, distracting from the content, so I knocked them back. but maybe we could do something a touch more 'buttony' somehow.

> Also, can we have all the tabs change in sync? If someone clicks "pyproject.toml", they shouldn't have to change it on every single box, once should be enough. Keeping the page steady during the change might be tricky?

yeah... it would change the height of elements up the page, meaning the content is gonna jump around when they are toggled :frowning:

To test it for yourself, you can scroll to the bottom of the page and paste these commands into the JS console:
```js
$("div.tabs-header > button:nth-child(1)").click()
$("div.tabs-header > button:nth-child(2)").click()
```
The content jumps around... it does get a bit disorienting I think.

henryiii:
I've added the second idea, what I called the "hack" method - it's pretty clean, really, and fixes the last remaining hole in "strictness" of the pyproject.toml configuration. You can now only specify things that make sense in tool.cibuildwheel.

I've made the examples a _tiny_ bit more tabby, maybe that will be enough?

Let's not worry about syncing the tabs just yet, then - Eventually, we could either match the number of lines on both options, or maybe there's a trick for pinning the page at the current content when changing tabs. But not something to worry about now.

cibuildwheel/__main__.py
```diff
@@ -300,7 +308,7 @@ def main() -> None:
         repair_command=repair_command,
         environment=environment,
         dependency_constraints=dependency_constraints,
-        manylinux_images=manylinux_images,
+        manylinux_images=manylinux_images or None,
```
henryiii, line 311: Note for later: some tests assume this is none, but it probably is simpler type-wise just to leave it an empty set. Potential cleanup unrelated to this PR, though.

cibuildwheel/__main__.py
```diff
@@ -164,7 +183,7 @@ def main() -> None:
     skip_config = options("skip", env_plat=False, sep=" ")
     test_skip = options("test-skip", env_plat=False, sep=" ")
 
-    archs_config_str = options("archs", sep=" ") if args.archs is None else args.archs
+    archs_config_str = args.archs or options("archs", sep=" ")
```
henryiii, line 186: Tiny unrelated cleanup - this reads more like the other options, and is clearer - args.archs overrides options, which is rather hidden the way it was written before.

cibuildwheel/__main__.py
```diff
@@ -29,6 +29,17 @@
     resources_dir,
 )
 
+MANYLINUX_ARCHS = (
```
henryiii, line 32: I expect order doesn't matter and this could be a set, but I left it as is, so the ordering would match in the dict it creates.

henryiii:
Doccross update now includes auditwheel 4, perhaps? Can’t look into it at the moment.

henryiii:
Note for anyone watching - two maintainers are out for 1-2 weeks, so this will likely sit around for a bit, which is fine. :)

henryiii:
@mayeut is already using this in https://github.com/scikit-build/cmake-python-distributions/pull/139/commits/f5223b1694642250d6124402c08f2c93c8c30e7e :)

henryiii:
I chose "-" over "_" because it was more natural in pyproject.toml; build-backend uses dashes, for example. However, looking at #717, I see mypy, pytest, and several others are using "_". Now, they had a pre-existing ini config and it's directly a 1:1 adaptation, so it's not the same as our environment variable overrides, but I'd be okay with "_" if you think it would be easier to convert/remember. I think I still do mildly prefer the dashes because they look better.

joerick:
> I chose `"-"` over `"_"` because it was more natural in pyproject.toml

I think `kebab-case` is slightly better, yes, as it's the form used in the PEPs.

henryiii:
Approve when you are ready. I'd like to see a "alpha/beta" release with this in to to start playing with it. :) (build support too, #521)

joerick:
Did you see my PR to this PR? 😛 haha maybe github notification aren't quite set up for it. maybe I should have pinged you. https://github.com/henryiii/cibuildwheel/pull/2

joerick (approved):
Looks good to go for me!

joerick:
🎉  woot! great work @henryiii  and @mayeut !

PR #686, opened 2021-05-22
Remove macOS build on travis-ci

mayeut:
Leaving macOS results in an "Not enough credits" on travis-ci.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,6 @@ jobs:
       arch: ppc64le
       env: PYTHON=python
 
-    - name: macOS | x86_64 | Python 3.x
-      os: osx
-      env:
-        - PYTHON=python3
-        - PYTEST_ADDOPTS='-k "unit_test or test_0_basic" --suppress-no-test-exit-code'
-
     - name: Windows | x86_64 | Python 3.6
       os: windows
       language: shell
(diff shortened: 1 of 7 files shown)


joerick (approved):

mayeut:
@joerick, now that I'm seeing the builds triggered, there's one question left:
should we remove it from the doc or just say it's untested ?

pradyunsg:
My 2 cents: "it's untested because Travis CI doesn't provide us free credits to test it anymore" 🙃

joerick:
heheh! good point. I suppose the 'correct' thing to do would be to remove it from the docs - can we claim that we support it if we don't actually test it? Probably not. Plus, how would we fix it if somebody did report a problem?

Yeah I'd say let's remove it from the docs too.

mayeut:
> Yeah I'd say let's remove it from the docs too.

Done

YannickJadoul:
Do we also want to (temporarily) disable macOS on CircleCI, until they - hopefully - give us the same treatment as joerick/cibuildwheel got?

mayeut:
> Do we also want to (temporarily) disable macOS on CircleCI, until they - hopefully - give us the same treatment as joerick/cibuildwheel got?

That's being handled in #685
I will merge that one & open a new one not to forget it's a temporary state.

YannickJadoul:
> That's being handled in #685

Ugh. Sorry. I need to catch up with my list of unread cibuildwheel emails...

joerick (approved):

PR #689, opened 2021-05-23
Fix Travis CI tests on ppc64le/s390x

mayeut:
- Disable manylinux only tests on Travis CI ppc64le
Those tests are taking too much disk space for the Travis CI runner right now.
c.f. https://travis-ci.community/t/running-out-of-disk-space-quota-when-using-docker-on-ppc64le/11634

- Use focal to run aarch64/ppc64le/s390x jobs on Travis CI
This removes allow_failures from the Travis CI s390x job which used a buggy xenial image.

### DIFF
--- a/test/test_manylinuxXXXX_only.py
+++ b/test/test_manylinuxXXXX_only.py
@@ -1,3 +1,4 @@
+import os
 import platform
 import textwrap
 
@@ -44,9 +45,12 @@ project_with_manylinux_symbols = test_projects.new_c_project(
     "manylinux_image", ["manylinux1", "manylinux2010", "manylinux2014", "manylinux_2_24"]
 )
 def test(manylinux_image, tmp_path):
+    pm = platform.machine()
     if utils.platform != "linux":
         pytest.skip("the docker test is only relevant to the linux build")
-    elif platform.machine() not in ["x86_64", "i686"]:
+    elif pm == "ppc64le" and "TRAVIS" in os.environ:
+        pytest.skip("this test requires too much disk space for this runner")
+    elif pm not in ["x86_64", "i686"]:
         if manylinux_image in ["manylinux1", "manylinux2010"]:
             pytest.skip("manylinux1 and 2010 doesn't exist for non-x86 architectures")
 
--- a/test/test_manylinuxXXXX_only.py
+++ b/test/test_manylinuxXXXX_only.py
@@ -1,4 +1,3 @@
-import os
 import platform
 import textwrap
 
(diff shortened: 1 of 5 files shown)


joerick (approved):
Nice! Should we update the example configs with `dist: focal`, too?

joerick:
BTW, another idea to skip the test could be something like `PYTEST_ADDOPTS='-k "not test_manylinuxXXXX_only"'` in the Travis config? I haven't tested that but it might work...? but this is fine too

test/test_manylinuxXXXX_only.py
```diff
@@ -44,9 +45,12 @@
     "manylinux_image", ["manylinux1", "manylinux2010", "manylinux2014", "manylinux_2_24"]
 )
 def test(manylinux_image, tmp_path):
+    pm = platform.machine()
     if utils.platform != "linux":
         pytest.skip("the docker test is only relevant to the linux build")
-    elif platform.machine() not in ["x86_64", "i686"]:
+    elif pm == "ppc64le" and "TRAVIS" in os.environ:
+        pytest.skip("this test requires too much disk space for this runner")
```
YannickJadoul, line 52: Alternatively, we could always add this with `-k "not manylinuxXXXX_only` in the actual invocation of pytest on that specific CI (as it feels slightly weird to have CI-specific things in the tests). But this also works perfectly for me, and is probably easier.

mayeut:
> Should we update the example configs with dist: focal, too?

You're right.
On a side note, I only updated the alternate archs but my recommendation would always be to run the linux axis on the most recent stable distro provided by any CI provider.

test/test_manylinuxXXXX_only.py
YannickJadoul: OK, seems @joerick had the same idea ;-)

mayeut: I did not think of this. This felt natural (maybe it shouldn't) and I think we are more likely to bump into this code now & then rather than just forget we added a specific option in the CI configuration.
I can still try the `PYTEST_ADDOPTS='-k "not test_manylinuxXXXX_only"'`  option, my previous feeling is not a strong one.

YannickJadoul: Also true. As you prefer; it wasn't a very strong, need-to-have suggestion either!

mayeut: > OK, seems @joerick had the same idea ;-)

That seems to say that it feels more natural to both of you. I'll update the config.

joerick (approved):
Looks good!

henryiii (approved):
I like the setting in the config, too.

YannickJadoul (approved):
Same for me! I'll go ahead and merge this, then :-)

PR #692, opened 2021-05-24
[1.x] Backport recent changes

henryiii:
Putting in cherry-picked changes to get the 1.x branch working with the new name and CI changes.

(rebase and merge)


joerick:
Thanks @henryiii!

I just reverted the docs changes that removed Travis support - I figured that we're not intending to drop anything in 1.x, even if it's not tested, there aren't any code changes anyway.

Hope that makes sense :)

henryiii:
Can we promise 1.x works with Travis macOS when we can't test it anymore? What happens when Travis updates their runners, etc?

joerick:
I guess, no, we can't, but we're not formally maintaining 1.x anyway, it's more of an if-and-when thing. I just wasn't intending to drop support for anything in the 1.x branch, that was kinda the point of 2.0.

henryiii:
I think there will be things we have to drop support for because it's died (think Windows 2.7, for example). I would have bundled Travis in here too - if we can't support travis, we shouldn't list it. I didn't want to remove it from the examples, though, so I'm fine with it as is. I might FP to drop the commit instead of reverting it, though. :)

henryiii:
Oh, never mind, it's only reverting part of the commit. I think it's fine as is.

joerick:
Do you think 1.x is ready for a release now? I think it is?

henryiii:
I think so - looking at https://github.com/pypa/cibuildwheel/compare/v1.11.0...1.x , it looks like the only real change is the move to the pypa name, along with the CI changes to make that happen (we can re-enable Circle CI macOS if you want). Guessing this would be 1.11.1?

joerick:
Yeah, agreed. I'll backport the CircleCI changes and release that soon.

joerick:
v1.11.1 released! And then I realised I had forgotten the edit-in-github fix, so theres a v1.11.1.post1 too 😅

PR #694, opened 2021-05-25
fix: docker run example and requirements paths

henryiii:
The current docker command doesn't work.

For Python 3.6, the `manylinux{1,2010,2014}` images have an ~~incorrectly configured encoding~~, so click just crashes with complaints about the encoding. `manylinux_2_24` doesn't have this issue. (It's also more compatible == better for _running_ wheels) This was in #661 but got changed to manylinux2010 before merging. EDIT: No, it was forcing the locale to one that was missing; removing that fixes it. I still think this might be a slightly better image?

Second, the paths printed out in the constraint files are wrong - they have a `/src` added to them. Fixed, and setting the working directory makes the docker line a little simpler, I think.

### DIFF
--- a/bin/update_dependencies.py
+++ b/bin/update_dependencies.py
@@ -18,18 +18,15 @@ env = os.environ.copy()
 # regenerate the constraints files
 env["CUSTOM_COMPILE_COMMAND"] = "bin/update_dependencies.py"
 
-if python_version == "36":
-    # Bug with click and Python 3.6
-    env["LC_ALL"] = "C.UTF-8"
-    env["LANG"] = "C.UTF-8"
+os.chdir(DIR.parent)
 
 subprocess.run(
     [
         "pip-compile",
         "--allow-unsafe",
         "--upgrade",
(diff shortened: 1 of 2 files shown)


joerick (approved):

mayeut (approved):
> For Python 3.6, the manylinux{1,2010,2014} images have an incorrectly configured encoding, so click just crashes with complaints about the encoding. manylinux_2_24 doesn't have this issue. (It's also more compatible == better for running wheels) This was in #661 but got changed to manylinux2010 before merging. EDIT: No, it was forcing the locale to one that was missing; removing that fixes it. I still think this might be a slightly better image?

I'll probably think about also adding the C.UTF-8 locale in manylinux images, it seems like it should be there. At least, it should be consistent across images.

As you mentioned, the manylinux_2_24 being the most recent one, it should have no issues pulling recent manylinux wheels (if there are some and someone drops manylinux1 & manylinux2010) so seems like a good default.

PR #695, opened 2021-05-25
fix: `bin/update_pythons.py` to work properly with PyPy `win_amd64`

mayeut:


bin/update_pythons.py
```diff
@@ -262,9 +262,12 @@ def update_config(self, config: dict[str, str]) -> None:
...
         elif "win_amd64" in identifier:
-            config_update = self.windows_64.update_version_windows(spec)
-            if config_update:
-                config.update(**config_update)
+            if identifier.startswith("pp"):
+                config.update(**self.windows_pypy.update_version_windows(spec))
+            else:
+                config_update = self.windows_64.update_version_windows(spec)
+                if config_update:
+                    config.update(**config_update)
```
henryiii, line 270:
```suggestion
            config_update: ConfigWinCP | ConfigWinPP | None
            if identifier.startswith("pp"):
                config_update = self.windows_pypy.update_version_windows(spec)
            else:
                config_update = self.windows_64.update_version_windows(spec)
            if config_update:
                config.update(**config_update)
```

Not sure this is any better, though.

henryiii:
```suggestion
config_update = (
                    self.windows_pypy.update_version_windows(spec)
                    if identifier.startswith("pp")
                    else self.windows_64.update_version_windows(spec)
                )
                if config_update:
                    config.update(**config_update)
```
Assuming this passes MyPy, maybe it's better?

henryiii: Technically, the current form looks just like the win32 version above, so it's also fine (and if changed, the one above should follow too).

henryiii (approved):

bin/update_pythons.py
mayeut: I reworked this a bit. Just fail if `config_update is None` as was done on macOS, it shall never happen.

henryiii: Okay, sounds good.

henryiii (approved):

henryiii:
Follow up to #671, for the record.

Issue #104, opened 2018-09-23
Optionally use existing installations of tools


altendky:
https://github.com/joerick/cibuildwheel/blob/9372e37b1886be448ca6b67a14dc6708657ac4a8/README.md#L35
> It will try to install packages globally; this is no good.

I haven't looked and this isn't any priority for me but...  couldn't this installation be optional?  I might tend towards having it disabled by default but I get there will be arguments both ways.

joerick:
Well, on Mac, cibuildwheel installs MacPython, and upgrades setuptools, wheel, delocate globally (not in a virtualenv). On Windows it assumes the existence of the interpreters, but still installs things like pip, setuptools globally.

What would the use-case be for disabling these installs? running a test build on your mac/windows dev machine?

altendky:
I didn't dig into the details, just read the docs, but yes.  As I said this is of no urgency to me to me but I think software is generally better when it doesn't need warnings like this. Is there some need to install these globally rather than in an env?

joerick:
> Is there some need to install these globally rather than in an env?

MacPython can only be installed globally. But other than that, the python bits could install into a venv. I guess that running cibuildwheel locally could be helpful for debugging. In which case, maybe we could have something like `cibuildwheel --local`, which would look for existing installs of MacPython (or Windows local installs) and use a virtualenv for the python dependencies.

altendky:
I would tend to default to not trying to fuss with the system.  Requiring that people type `--install-deps` or whatever once in a tox.ini doesn't seem particularly onerous.  or even a separate subcommand (are there any?  i forget...).

HDembinski:
I am trying to use cibuildwheels for a non-trivial project, and - of course - it does not work, because I need to install certain dependencies that are missing in some docker images or not found by the CMake scripts in my project. Debugging these issues by adding small commits on a branch until the CI works is an incredibly painful process.

Long story short: it would be great if one could run cibuildwheels locally for testing/debugging of the build.

YannickJadoul:
> Long story short: it would be great if one could run cibuildwheels locally for testing/debugging of the build.

It can be, right? If you're happy to have docker images downloaded and run, or Python versions installed on macOS/Windows, you can perfectly run cibuildwheel locally. Just skip the platform checks/detection by adding the correct `--platform` flag.
It's just really not meant to run locally, so we're not enabling that by default.

hoffmang9:
[...]

PR #697, opened 2021-05-25
Clarify Apple Silicon support on various Python versions

pradyunsg:
This changes the blank cell to N/A, since arm64 is not natively supported on
those versions of Python.

Shoutout to @pganssle for noting that this would be useful in https://twitter.com/pganssle/status/1397271016536252421, and @FFY00 for coming up with "N/A" as a choice for phrasing that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,11 @@ What does it do?
 |---|---|---|---|---|---|---|---|---|---|
 | CPython 3.6     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.7     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| CPython 3.8     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.8     | ✅ | N/A¹ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.9     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| PyPy 3.7 v7.3   | ✅ | N/A | ✅  |    | ✅ | ✅  | ✅  |    |   |
+| PyPy 3.7 v7.3   | ✅ | N/A | ✅  | N/A | ✅ | ✅  | ✅  | N/A | N/A |
+
+<sup>¹ CPython 3.8's final binary release has experimental Universal2 support, but does not support macOS 10.x, so this is not currently available.</sup><br>
 
 - Builds manylinux, macOS 10.9+, and Windows wheels for CPython and PyPy
 - Works on GitHub Actions, Azure Pipelines, Travis CI, AppVeyor, CircleCI, and GitLab CI
--- a/README.md
+++ b/README.md
@@ -24,11 +24,11 @@ What does it do?
 
 |   | macOS Intel | macOS Apple Silicon | Windows 64bit | Windows 32bit | manylinux x86_64 | manylinux i686 | manylinux aarch64 | manylinux ppc64le | manylinux s390x |
 |---|---|---|---|---|---|---|---|---|---|
-| CPython 3.6     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| CPython 3.7     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| CPython 3.8     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.6     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.7     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.8     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
 | CPython 3.9     | ✅ | ✅ | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| PyPy 3.7 v7.3   | ✅ |   | ✅  |    | ✅ | ✅  | ✅  |    |   |
+| PyPy 3.7 v7.3   | ✅ | N/A | ✅  |    | ✅ | ✅  | ✅  |    |   |
 
 - Builds manylinux, macOS 10.9+, and Windows wheels for CPython and PyPy
 - Works on GitHub Actions, Azure Pipelines, Travis CI, AppVeyor, CircleCI, and GitLab CI


pganssle:
To be clear, my suggestion was to have tool-neutral documentation on how to build wheels targeting various tags, not to clarify which platforms support wich tags in `cibuiildwheel` (though admittedly it doesn't hurt to do it here as well).

README.md
```diff
@@ -24,11 +24,11 @@ What does it do?
 
 |   | macOS Intel | macOS Apple Silicon | Windows 64bit | Windows 32bit | manylinux x86_64 | manylinux i686 | manylinux aarch64 | manylinux ppc64le | manylinux s390x |
 |---|---|---|---|---|---|---|---|---|---|
-| CPython 3.6     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| CPython 3.7     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
-| CPython 3.8     | ✅ |   | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.6     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.7     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
+| CPython 3.8     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
```
pganssle, line 29: Evidently it's more complicated than "N/A" here: https://twitter.com/hugovk/status/1397275649253990405

joerick: Yeah, essentially, since we only have Intel CI runners running 10.15, we can't tick this box. But the wheels could exist, if you were able to build on an Apple Silicon machine.

mayeut:
I don't have an Apple Silicon but here's the stock python from Apple on my laptop:
```
Matt$ /usr/bin/python3 -V
Python 3.8.2
Matt$ file /usr/bin/python3
/usr/bin/python3: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/usr/bin/python3 (for architecture x86_64):	Mach-O 64-bit executable x86_64
/usr/bin/python3 (for architecture arm64e):	Mach-O 64-bit executable arm64e
Matt$ /usr/bin/python2 -V
Python 2.7.16
Matt$ file /usr/bin/python2
/usr/bin/python2: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/usr/bin/python2 (for architecture x86_64):	Mach-O 64-bit executable x86_64
/usr/bin/python2 (for architecture arm64e):	Mach-O 64-bit executable arm64e
```

I also don't see why rebuilding from source any version would fail (I've not tried it, maybe it would).

README.md
mayeut: > But the wheels could exist, if you were able to build on an Apple Silicon machine.
or x86_64 runners with macOS 11+

henryiii:
Apple created universal Python 2.7 and 3.8 versions. If we could use the Xcode versions of Python, that would enable us to do 10.15+ or 11+ wheels, though it would mean that if Xcode upgraded and provide 3.9 instead, that would suddenly not work. Having a way to specify use of "built-in" Python (maybe with a warning that it won't work on older macOS's) might be useful (for this and for #104).

I've seen a project that used cibuildwheel but also had a native Python 3.8 job that built Universal or ARM wheels (don't remember which) by hand for macOS 11+. I don't remember where that was, though, does anyone remember what project that was? I'm pretty sure I posted about it somewhere here.

README.md
henryiii: GitHub Actions was working on that a while back.

henryiii: I'd put a superscript here, and have a little note. Since there is a Python 3.8 Universal release _and_ 3.8 is the default Python currently with Big Sur's XCode toolkit.

henryiii:
@pganssle if one of us can find that example of a manual Python 3.8 build, that's probably the best answer to your twitter question, and I'll reply there if it's found.

henryiii:
Yes! I found it: https://github.com/pypa/cibuildwheel/discussions/579#discussioncomment-364552

pganssle:
> @pganssle if one of us can find that example of a manual Python 3.8 build, that's probably the best answer to your twitter question, and I'll reply there if it's found.

In reality, I no longer really need the information, because it's a lot of extra work to build one Apple Silicon wheel that only works on Python 3.8.10. My point was, though, that a chart on the `cibuildwheel` documentation is not a good canonical source of information on how to build wheels for various platforms. Evidently the example you link to is fragile in some way (how would I know that?), and it builds an arm64 wheel, not a universal2 wheel (how would I build one of these without `cibuildwheel`).

My comment here was mainly to clarify that in the tweet Pradyun linked to to start this thread, I was _not_ trying to say that `cibuildwheel`'s documentation needs clarification  — it seemed fine given its proper scope. I was saying that it's unfortunate that two new wheel formats (I think) were added with, as far as I can tell, _no_ documentation about them. This is maybe nice (if accurate), but it doesn't really solve the underlying problem.

README.md
henryiii: How about:

```
CPython 3.8's final binary release has experimental Universal2 support,
but does not support macOS 10.x, so this is not currently available in cibuildwheel.
```

If we were running on macOS 11 runners, are you pretty sure we could get this to work?

henryiii: Should we add add N/A For PyPy and the special arches, as those are not provided by PyPy & manylinux?

henryiii:
I'm not against better documentation, but someone needs to write it and maintain it. One possible page would be https://packaging.python.org/guides/packaging-binary-extensions/ - notice it documents 2.7, 3.4, and 3.5. The linked page, https://github.com/MacPython/wiki/wiki/Spinning-wheels, is better, but still tops out at 3.7 and the "new" dual arch support (FAT & pure 64-bit versions).

I'd be okay to have a "theory of building wheels" or something like that in the cibuildwheel docs that describes how we build wheels - the problem would be keeping such a page up-to-date, as it would not be able to be tested. My own page (https://iscinumpy.gitlab.io/post/azure-devops-python-wheels/) is out of date; there's no automated way to test it, and cibuildwheel is always going to be better kept up-to-date than an example _every_ package has to copy. Why not centralize and use a tool designed for building wheels? That's what most major packages do already - either cibuildwheel (mypy, scikit-learn, scikit-image, matplotlib, etc.) or multibuild (numpy, pandas, etc.).

README.md
henryiii: @pypa/cibuildwheel-team opinions here on these two choices? The comment and the N/As for PyPy? (Maybe @mattip too on PyPy)

mattip: N/A would be appropriate I think.

henryiii (approved):

joerick:
I did have a play just now with hiding the footnote in a tooltip... but on balance I'm thinking the footnote is better. But I did learn how to do tooltips in Github Markdown - included below if anyone is curious!

|   | macOS Intel | macOS Apple Silicon | Windows 64bit | Windows 32bit | manylinux x86_64 | manylinux i686 | manylinux aarch64 | manylinux ppc64le | manylinux s390x |
|---|---|---|---|---|---|---|---|---|---|
| CPython 3.6     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| CPython 3.7     | ✅ | N/A | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| CPython 3.8     | ✅ | <ins title="CPython 3.8's final binary release has experimental Universal2 support, but does not support macOS 10.x, so this is not currently available.">N/A</ins> | ✅  | ✅  | ✅ | ✅ | ✅ | ✅ | ✅ |
| PyPy 3.7 v7.3   | ✅ | N/A | ✅  | N/A | ✅ | ✅  | ✅  | N/A | N/A |

joerick (approved):

joerick:
> I don't seem to be able to see it in GitHub mobile. It's underlined but not tappable.

Oh yeah, hover states. Probably not a great pattern then! It's annoying that Github doesn't support `<abbr>` anymore... I think they used to? Maybe they dropped support because they weren't mobile-accessible.

PR #698, opened 2021-05-25
Drop pre-commit from GitHub Actions

pradyunsg:
We now use pre-commit.ci for this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,17 +11,8 @@ on:
     # allow manual runs on branches without a PR
 
 jobs:
-  pre-commit:
-    name: Pre-commit checks (mypy, flake8, etc.)
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
-    - uses: pre-commit/action@v2.0.0
-
   test:
     name: Test cibuildwheel on ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


henryiii:
We can't fully drop it - the check-manifest step does not run on pre-commit.ci. We should drop the rest, though (which I thought we had, but obviously not, which turned out to be nice since pre-commit.ci was missing after the move for a bit. I can edit the PR to drop everything but that one step.

Why is this version 2.0.0 instead of 2.0.2? I would have thought that dependabot was supposed to be updating this.

henryiii:
Ahh! I remember now why we didn't drop it. The "needs:" step saves some CI time if the pre-commit step does not pass.

henryiii:
This is a duplicate of #624, where we decided to use `needs:` instead of dropping the check. `pre-commit.ci` is used to keep pre-commit up to date, and to help contributors who don't run pre-commit locally, and GHA is for running check-manifest (which can't trivially be run on pre-commit.ci, since it's not allowed to download PyPI packages), and for limiting CI use on commits that don't pass formatting.

pradyunsg:
Ah cool! So same reasons as pip. Lemme know if you want to remove pre-commit.ci then, since @mayeut had requested I install that app here. :)

henryiii:
We do use pre-commit.ci for the two reasons listed above (auto-updates and auto-fixes), so it should stay. :)

PR #699, opened 2021-05-25
ci: add dependabot

henryiii:
I thought I already had done this. Trying out dependabot's new pinning control here.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,4 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "actions/*"
-      - update-types: "version-update:semver-minor"
+        update-types: ["version-update:semver-minor", "version-update:semver-patch"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
     ignore:
       - dependency-name: "actions/*"
       - update-types: "version-update:semver-minor"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
mayeut, line 7: Should it be weekly, like the other dependencies ?

mayeut, line 7:
```suggestion
      interval: "weekly"
```

henryiii: Okay, though "daily" would give quicker feedback if it worked, and we don't depend on so many actions that it's likely to be more than once a week anyway - but I have found you can manually trigger dependabot, so weekly is fine.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "actions/*"
+      - update-types: "version-update:semver-minor"
```
henryiii, line 10: I have not tried this line before, usually I ignore `actions/*`, since GitHub is good at providing moving "simple" tags and it's pretty rare (though it has happened) that an update to these breaks things.

joerick: Ignoring minor updates... so does this mean dependabot will only update to newer patch versions? And newer major versions?

henryiii: Actually, I didn't structure it correctly anyway: https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/

joerick: Ah, I see, that's nested within the `actions/*` rule... and Github actions will move the `v2` tag to point to the latest `v2` e.g. `v2.3.4`... how strange! Well, anyway, that's cool with me.

joerick (approved):

PR #700, opened 2021-05-25
feat: update deprecated option check

henryiii:
Dropping the old check, and adding a new one for `p2` or `p35` in selectors.

Not sure if this is a "fix", "feat", or "chore". 😜

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -140,8 +140,6 @@ def main() -> None:
 
     args = parser.parse_args()
 
-    detect_obsolete_options()
-
     if args.platform != "auto":
         platform = args.platform
     else:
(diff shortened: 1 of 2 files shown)


cibuildwheel/__main__.py
```diff
@@ -361,35 +363,12 @@ def main() -> None:
...
-            ("macosx_10_6_intel", "macosx_x86_64"),
-            ("macosx_10_9_x86_64", "macosx_x86_64"),
-        ]:
-            if option in os.environ and deprecated in os.environ[option]:
-                print(
-                    f"Build identifiers with '{deprecated}' have been deprecated. Replacing all occurences of '{deprecated}' with '{alternative}' in the option '{option}'"
-                )
-                os.environ[option] = os.environ[option].replace(deprecated, alternative)
+def deprecated_selectors(name: str, selector: str, *, error: bool = False) -> None:
+    if "p27" in selector or "p35" in selector:
```
mayeut, line 367:
```suggestion
    if "p2" in selector or "p35" in selector:
```
Should it be `p2` instead of `p27` ?

henryiii: Probably could be, just in case. It just has to only match the correct place, but I'm pretty sure no other parts of our selectors have a p2?

mayeut (approved):
LGTM !

PR #701, opened 2021-05-27
Code changes to support the `master`->`main` branch switch

joerick:
As per #690.﻿

Closes #690.

PR #702, opened 2021-05-27
fix: `github-deploy.yml` example

mayeut:
Follow up on #701, missed that bit in my review & merged it...


joerick (approved):
thanks @mayeut !

PR #706, opened 2021-05-28
ci: try Travis AWS arm64-graviton2

henryiii:
These fail a lot, so maybe using the new Graviton's would be better? Just trying it out.

See https://blog.travis-ci.com/2020-09-11-arm-on-aws  and https://docs.travis-ci.com/user/multi-cpu-architectures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,4 +49,4 @@ jobs:
 
 install: $PYTHON -m pip install -e ".[dev]" pytest-custom-exit-code
 
-script: travis_wait 20 $PYTHON ./bin/run_tests.py
+script: travis_wait 60 $PYTHON ./bin/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,9 @@ jobs:
       services: docker
       arch: arm64-graviton2
       group: edge
-      virt: lxd
+      virt: vm
       env: PYTHON=python
+      allow_failure: True
 
     - name: Linux | ppc64le | Python 3.6
       python: 3.6
@@ -49,4 +50,4 @@ jobs:
 
 install: $PYTHON -m pip install -e ".[dev]" pytest-custom-exit-code
 
-script: travis_wait 60 $PYTHON ./bin/run_tests.py
+script: $PYTHON ./bin/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,4 +49,4 @@ jobs:
 
 install: $PYTHON -m pip install -e ".[dev]" pytest-custom-exit-code
 
-script: $PYTHON ./bin/run_tests.py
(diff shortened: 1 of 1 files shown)


henryiii:
This drops the build time from 23 minutes to 12 minutes for ARM. Not sure if it's more stable, but it passed at least once. :)

(The problem for ARM has been "unable to generate build script", FYI, the builds were not even starting)

henryiii:
How about making ppc an allowed error? 😠

henryiii:
Screen Shot 2021-05-28 at 8 44 22 PM

mayeut:
xref https://travis-ci.community/t/travis-arm64-a-job-does-not-start-with-an-error-occurred-while-generating-the-build-script/11629

I was waiting a bit for arm64 to see if TravisCI resolved this as the one used is the one from the Partner Queue (free plan).
I find the AWS Graviton 2 VMs really stable (both in timings and in the fact that they just work) so it might be a good idea to switch as we're on a paid plan.

> How about making ppc an allowed error?

Agreed, I tried many things especially for this one, TravisCI is looking into that according to https://travis-ci.community/t/running-out-of-disk-space-quota-when-using-docker-on-ppc64le/11634

The only thing that seems to work well on this one is spawning a rootless docker daemon. I'm not sure that we want this added complexity in cibuildwheel CI tests. I use it in manylinux because there's no other choice to get ppc64le images (GHA was really too slow with qemu with timeouts after 6 hours....).

.travis.yml
```diff
@@ -16,7 +16,9 @@ jobs:
     - name: Linux | arm64 | Python 3.6
       python: 3.6
       services: docker
-      arch: arm64
+      arch: arm64-graviton2
+      group: edge
+      virt: lxd
```
mayeut, line 21: Any reason to choose `lxd` rather than `vm` here ?
I think that `vm`, while they are a bit slower to start are more stable and are a bit more appropriate for docker usage. They're are all kinds of things going bad on ppc64 lxd, one of the reason is docker is running on top of lxd does not always agree with the lxd configuration.

.travis.yml
```diff
@@ -16,8 +16,11 @@ jobs:
     - name: Linux | arm64 | Python 3.6
       python: 3.6
       services: docker
-      arch: arm64
+      arch: arm64-graviton2
+      group: edge
+      virt: vm
       env: PYTHON=python
+      allow_failure: True
```
mayeut, line 23: Wasn't the `allow_failure` meant to be added on ppc64le ?

henryiii: 😳

mayeut:
I'll go ahead and merge this one.

The GHA "Test / Test emulated cibuildwheel using qemu" also seems to be flaky and sometimes just hangs...

henryiii:
I know, but it's a bit less often. Not sure why it times out, ideally pytest should timeout first.

PR #717, opened 2021-06-12
chore: bump mypy and update config

henryiii:
I split up the extras a little to reduce the number of packages installed.

I then bumped pre-commit, mostly mypy, and moved the config to the newly supported `pyproject.toml` location for mypy. This is simpler and prettier. MyPy 0.901 no longer installs typestubs for common third party libraries, so those have been added to pre-commit explicitly.

Make the mypy config just a bit stricter. Eventually we can likely touch up bin and remove it from the "loose" category.


henryiii:
I think @joerick might still be on vacation, and I know @mayeut is, so @YannickJadoul, could you take a quick look to make sure I didn't make a mistake?

.pre-commit-config.yaml
```diff
@@ -42,17 +42,28 @@ repos:
...
-    args: ["--python-version=3.7", "--ignore-missing-imports", "--scripts-are-modules"]
+    args: ["--python-version=3.7", "--scripts-are-modules"]
+    additional_dependencies:
+      - packaging
+      - rich
+      - types-jinja2
+      - types-pyyaml
+      - types-click
+      - types-requests
+      - types-toml
```
joerick, line 66: Should we add these to the `dev` requirements in setup.py too?

henryiii: Sure, we can. I'd normally see mypy as something you really should use either nox or pre-commit for, not something you'd set up in your development environment, but it shouldn't hurt. There's also @asottile's `types-all`.

I'm moving these to a "mypy" extra that gets included in dev.

henryiii:
Are we good to go here? I'd like to get it in before the next update pass on Monday.

.pre-commit-config.yaml
joerick: Brilliant, thanks. That's helpful to me, I do sometimes run `mypy .` at a shell, because pre-commit is more interested in the staged files than the working copy.

joerick (approved):

.pre-commit-config.yaml
henryiii:
```console
$ pre-commit run -a
```

This will run on all files and will ignore the staging area.

henryiii:
```console
$ pre-commit run -a mypy
```

If you only want the mypy check. :)

joerick: ah yeah that's pretty handy!

PR #720, opened 2021-06-20
Add a mkdocs extension to generate the --help output

joerick:
I was getting a bit tired of having to remember to update the --help output in the options docs, so I wrote a mkdocs extension that allows a snippet of python code in the docs that will run when the page is rendered. It's invoked like this:

~~~markdown
```python execute-and-replace
import subprocess

print('```text')
subprocess.run(['cibuildwheel', '--help'], check=True, stderr=subprocess.STDOUT)
print('```')
```
~~~

That is replaced at build time with the --help output from cibuildwheel.

I have a feeling it might be useful elsewhere, too, but I can't think of anything off the top of my head!

PR #721, opened 2021-06-21
[docs-execute-plugin] docs: use macro plugin instead

henryiii:
Trying the macro plugin for #720


setup.py
```diff
@@ -1,36 +1,46 @@
...
         "mkdocs-include-markdown-plugin==2.8.0",
         "mkdocs==1.0.4",
         "pymdown-extensions",
-        f"mkdocs_execute_python_plugin @ file://{DIR}/docs/mkdocs_execute_python_plugin",
+        "mkdocs-macros-plugin",
     ],
     "test": [
         "jinja2",
-        "pytest>=4",
+        "pytest>=6",
```
henryiii, line 12: Oops, I tried to grab the latest setup.py with `git checkout main -- setup.py`, forget it had changed. The extra changes where will go away if we rebase.

joerick:
oh that is quite nice actually. I don't _love_ the extra level of indirection, but I can see the advantages, e.g. code editor support `docs/main.py`. Neat trick with the jinja tags, I guess that's to make it a bit more robust?

henryiii:
A nice plugin for inline Python execution would be nice (or maybe even shell execution, too), but this seems to be pretty close and doesn't require writing and maintaining a new extension. It's much better than having to maintain an internal extension, I think.

> jinja tags, I guess that's to make it a bit more robust

It's to avoid `${{}}` from triggering as a jinga tag that happens to be right after a `$`; we have quite a lot of that in the CI examples. We probably could surround things in `{% raw %} {% endraw %}`, but that's ugly, just having a special string is nicer. You can also change block tags, but those didn't conflict with anything (and I don't think they are even used by anything in macro, much less in our usage).

henryiii:
PS: this is a PR to a PR, if the `[]` tag at the front didn't give it away.

henryiii:
Reduced the indirection, you can now see what's happening better inline.

setup.py
henryiii: Fixed before pushing the reduced indirection commit.

joerick:
Nice. thanks @henryiii !

Issue #540, opened 2021-01-15
Proposal: move changelog


henryiii:
GitHub's dependabot is not used to seeing the changelog in the README (at least when there also is a docs/CHANGELOG.md file too). If we move the changelog to be _in_ the docs/CHANGLOG.md file, and then put a link to it in the readme, at the top, that will a) fix dependabot, b) keep the README from growing forever, c) make the CHANGELOG link visible without having to scroll down, and d) keep the README closer to what you see in the docs. For PyPI, where having the CHANGELOG inline is helpful, we could just append the CHANGELOG to the README when we read it in. Or, what many project do is add a changelog link into the links list, see https://pypi.org/project/coveralls/ for example, just below home page in the sidebar is a link to the changelog. Thoughts, @joerick?

If we increase the visibility of using dependabot to keep CIBW up to date, it would be nice to have more useful changelogs in the PRs it generates!

PS: "Working examples" on PyPI have broken images.

henryiii:
This was done a bit ago.

PR #724, opened 2021-06-21
ci: also update projects displaly in workflow

henryiii:
This adds the README update step to the weekly workflow. I think it's working?


joerick:
This seems fine to me. 

Off topic, but something I've been meaning to mention, this projects list is getting quite long for the README... do you think it would be a good idea to trim these projects to e.g. top 10 in the readme, and then link to the full list in the docs? I don't know how much work that would be but it might be a nice idea?

henryiii:
Shouldn't be too hard since it's already generated - making a file in docs is easier than the current injection, so the "hard" part is already done, and as you know, I'm not a fan of growing readmes (AKA #540 ) ;)

joerick:
Okay, maybe we can strike a deal, top 10 projects and latest 3 versions in the readme? ;)

PR #731, opened 2021-06-24
docs: clarify location of dependabot file

henryiii:
Might be less confusing.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -232,14 +232,14 @@ updates:
 
 #### Option 2: Requirement files
 
-The second option, and the only one that supports other CI systems, is using a `requirements-*.txt` file. The file should have a distinct name and have only one entry:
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -232,14 +232,14 @@ updates:
 
 #### Option 2: Requirement files
 
-The second option, and the only one that supports other CI systems, is using a `requirements-*.txt` file. The file should have a distinct name and have only one entry:
+The second option, and the only one that supports other CI systems, is using a `.github/requirements-*.txt` file. The file should have a distinct name and have only one entry:
```
joerick, line 235: This one might be an error, I think? it seems to contradict the 'other CI systems' previously.

henryiii: Oops, typo, added it to the wrong spot.

henryiii: Should be moved to the right place now.

joerick (approved):

PR #732, opened 2021-06-24
Clarify docs around manylinux and c++

AllSeeingEyeTolledEweSew:
I wanted to clarify and expand on some docs here. In the current docs, I especially found it misleading that manylinux2010 and manylinux2014 "support all C++ standards (up to C++17)", since using newer standards can produce `manylinux`-incompatible builds.

### DIFF
--- a/docs/cpp_standards.md
+++ b/docs/cpp_standards.md
@@ -5,10 +5,39 @@ title: Modern C++ standards
 Building Python wheels with modern C++ standards (C++11 and later) requires a few tricks.
 
 
-## manylinux1 and C++14
-The default `manylinux1` image (based on CentOS 5) contains a version of GCC and libstdc++ that only supports C++11 and earlier standards. There are however ways to compile wheels with the C++14 standard (and later): https://github.com/pypa/manylinux/issues/118
+## manylinux and C++
+The `manylinux*` standards imply a limit on which C++ standard you can use. There are workarounds to this, and `cibuildwheel` uses some workarounds by default. You should be aware of them!
 
-`manylinux2010` and `manylinux2014` are newer and support all C++ standards (up to C++17).
+Consider that the `manylinux*` standards constrain _symbol versions_ in `libstdc++.so`. So when *dynamically linking* to `libstdc++.so`, the desired `manylinux*` standard constrains the gcc version, and thus constrains the C++ standard.
+
+Cross-referencing [current manylinux standards](https://github.com/mayeut/pep600_compliance/blob/f86a7d7c153cc45aa3f2add6ffcf610c80501657/pep600_compliance/tools/policy.json) with [gcc's symbol versions](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html) and [libstdc++'s language support](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html), we have:
+ * `manylinux1`: gcc 4.1.x
+ * `manylinux2010`: gcc 4.4.x
+ * `manylinux2014`: gcc 4.8.x
+ * `manylinux_2_24`: gcc 6.x (stable C++14 ABI)
+ * `manylinux_2_27`: gcc 7.x (stable C++14 ABI)
+
+([The first release of a complete and stable C++17 ABI is gcc 9.1](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017), but as of writing there is no official `manylinux*` standard that supports this version. Prior standards like C++11 *are* supported, but gcc doesn't document the per-version support as clearly as C++14 and later)
+
+We *can* use newer C++ standards and support older `manylinux*`, if we use *static linking*.
+
+The default `manylinux1`, `manylinux2010` and `manylinux2014` docker images include *more recent* gcc than the one matched to their standard. These CentOS-based images use automagic *selective static linking* for newer `libstdc++.so` symbols, and dynamic linking for older ones (this is done by having `libstdc++.so` from the `devtools` package be a linker script, and splitting newer symbols into a static library).
+
+The debian-based `manylinux_2_24` image *does not* do selective static linking as of writing. See https://github.com/pypa/manylinux/issues/1012
+
+As of 2021-06-24, the gcc versions on each (x86-64) image are:
+ * `manylinux1`: gcc 4.8.2 (C++11 stable)
+ * `manylinux2010`: gcc 8.3.1 (C++14 stable)
+ * `manylinux2014`: gcc 9.3.1 (C++17 stable)
+ * `manylinux_2_24`: gcc 6.3.0 (C++14 stable)
+
+Each gcc version may support later C++ standards incompletely or experimentally. In some cases gcc may support a standard with an unstable ABI, but this won't matter with static linking (so using C++17 on `manylinux2010` will probably just work). You will have to experiment for yourself to see what works.
+
+If you hit an edge case with selective static linking, or want to try to support an older `manylinux*` standard, you can use unconditional static linking with `LDFLAGS=-static-libstdc++`. Note that this will bloat your wheels, and pypi *does* have a limit on total project size!
+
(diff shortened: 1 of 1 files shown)


henryiii:
This is not correct. The newer GCC compilers on manylinux1 through manylinux2014 contain versions of GCC that have been patched by Red Hat to produce binaries that are compatible with the default libstdc++.so on that system. There might be some small issues in corner cases, but this works in most situations and is used across the ecosystem. The issue you've linked is getting manylinux1 + C++14, which doesn't work, since Red Hat dropped support for CentOS 5's back ports with GCC 4.9.

Manylinux2010 has GCC 8, and Manylinux2014 has GCC 9.

There is an open issue to find or perform something similar with Debian based manylinux_2_24 (and newer) images https://github.com/pypa/manylinux/issues/1012, but it's still unresolved.

With some edits and changing versions, this could be useful, there's some good information and links.

AllSeeingEyeTolledEweSew:
Fascinating!

Is there documentation about how `devtoolset`'s patches work? If I use a new c++17 feature from `devtoolset-9` where gcc _would_ link to a symbol with version `CXXABI_1.3.12` (from 9.x), what happens to it? Are all `libstdc++.so`-specific bugfixes and behaviors from 9.x backported?

I'm surprised and confused, so it would be nice to have a link to relevant info to include in the docs.

AllSeeingEyeTolledEweSew:
Oh, I found:
> Yes, build gcc 10 with the patches from https://git.centos.org/rpms/devtoolset-10-gcc/blob/c7/f/SOURCES.
In devtoolset, libstdc++.so is a linker script that points to a static libstdc++ for the new parts and shared libstdc++ for the older parts and the C++ ABI is set to GCC-4 ABI.

I'll try to redo this PR

henryiii:
> Is there documentation about how devtoolset's patches work? 

I would love this too, I'd like to see how hard it would be to apply for ubuntu.

> Oh, I found:

Ahh, that's great, thanks! Where did you find it, out of curiosity? That's a clever way to handle it, actually.

henryiii:
Also, currently, what you have written is completely correct for the Debian based builds, which do not do this trick.

henryiii:
(I'm making this draft until you update it and un-mark it as draft, so I'll know to look at it again)

AllSeeingEyeTolledEweSew:
https://github.com/pypa/manylinux/issues/1012#issuecomment-835461315

It is clever, but surprising to me as a user!

Since the centos images are default, `cibuildwheel`'s default behavior is really an implicit and selective version of `-static-libstdc++`.

Given that, I guess I'm leaning toward doing a full `-static-libstdc++` unless I can find a reason *not* to. Explicit is better than implicit, and all.

henryiii:
It's really important to keep wheels small, since you could end up uploading dozens of them per release. And pretty much every other compiled  package you use is likely depending on this - certainly all pybind11 packages like SciPy, PyTorch, etc. are shipping C++11 + manylinux1 wheels have have been for years.

How this will be handled in the future remains to be seen.

docs/cpp_standards.md
```diff
@@ -5,10 +5,39 @@ title: Modern C++ standards
 Building Python wheels with modern C++ standards (C++11 and later) requires a few tricks.
 
 
-## manylinux1 and C++14
-The default `manylinux1` image (based on CentOS 5) contains a version of GCC and libstdc++ that only supports C++11 and earlier standards. There are however ways to compile wheels with the C++14 standard (and later): https://github.com/pypa/manylinux/issues/118
+## manylinux and C++
+The `manylinux*` standards imply a limit on which C++ standard you can use. There are workarounds to this, and `cibuildwheel` uses some workarounds by default. You should be aware of them!
```
henryiii, line 9:
```suggestion
The `manylinux*` standards imply a limit on which C++ standard you can dynamically link to. There is a workaround built into the manylinux1, manylinux2010, and manylinux2014 compilers described below that allows partial linking to just the portions that are allowed; this workaround is not available for the `manylinux_2_*` images yet.
```

For 95%+ of users, this makes no difference - it just works. No one compiles with old GCCs - 4.8 is already _very_ old - C++11 is 10 years old this year. And it's not cibuildwheel's workaround at all, it's Red Hats - one could _maybe_ argue it's manylinux's (which is the way I've stated it above), but it's not cibuildwheel.

Pybind11 would not exist (at least on linux) if we were constrained to the original compilers. And statically linking everything would probably down PyPI.

docs/cpp_standards.md
```diff
@@ -5,10 +5,39 @@ title: Modern C++ standards
...
 
-`manylinux2010` and `manylinux2014` are newer and support all C++ standards (up to C++17).
+Consider that the `manylinux*` standards constrain _symbol versions_ in `libstdc++.so`. So when *dynamically linking* to `libstdc++.so`, the desired `manylinux*` standard constrains the gcc version, and thus constrains the C++ standard.
+
+Cross-referencing [current manylinux standards](https://github.com/mayeut/pep600_compliance/blob/f86a7d7c153cc45aa3f2add6ffcf610c80501657/pep600_compliance/tools/policy.json) with [gcc's symbol versions](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html) and [libstdc++'s language support](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html), we have:
+ * `manylinux1`: gcc 4.1.x
```
henryiii, line 14: This should probably be relegated to a note. For almost every user, this doesn't matter. IFAIK, all packages dynamically link and go about their day. This is a nice note, but what _most_ users what is the second table - what GCC version is _actually_ used. `manylinux1` is C++11 compatible, but not C++14 compatible, etc. This is good detail to have in a note just in case someone is debugging a really weird bug, or wondering how CentOS 5 has GCC 4.8.

docs/cpp_standards.md
```diff
@@ -5,10 +5,39 @@ title: Modern C++ standards
...
+ * `manylinux2010`: gcc 4.4.x
+ * `manylinux2014`: gcc 4.8.x
+ * `manylinux_2_24`: gcc 6.x (stable C++14 ABI)
+ * `manylinux_2_27`: gcc 7.x (stable C++14 ABI)
+
+([The first release of a complete and stable C++17 ABI is gcc 9.1](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017), but as of writing there is no official `manylinux*` standard that supports this version. Prior standards like C++11 *are* supported, but gcc doesn't document the per-version support as clearly as C++14 and later)
+
+We *can* use newer C++ standards and support older `manylinux*`, if we use *static linking*.
```
henryiii, line 22: I don't think we are trying to push anyone to static linking. The current system works fine. It's better if it's just mentioned as a detail.

AllSeeingEyeTolledEweSew:
Reading between the lines, it seems like you'd prefer I put the most relevant info _first_ in the section, whereas I wrote it starting from first principles and building to relevant info later.

So rather than tweak, I'll just rewrite it and we can go from there. Will un-draft when ready

henryiii:
I would ask the question: what is the purpose of this section? The current version could be "convince users that the way we currently build wheels is wrong, they should either statically link or use GCC 4.1", or at least put worries and doubts in their minds. It wasn't intentional, but it was the way it read due to the historical path it was written. The purpose should be "inform users about what they can use, as well as what is happening behind the scenes".

You want users to be assured that the version of GCC they get is just fine to use as normal. Most packages just use it and are happy. "What C++ version can I use" is the high level question, and "what GCC version" is the slightly lower level question, and most users should be happy to stop there. Just in case a really weird bug occurs, having the information like what the true version of GCC is being dynamically linked, and anything newer than than is static, or why `manylinux_2_*` is "what you see is what you get" for GCC linking is interesting for _very_ advanced users (and really hard information to find! Thanks for digging it up and collecting it!).

henryiii:
Also, I've got a long term plan to improve the packaging.python.org docs, such as in https://github.com/pypa/packaging.python.org/pull/911 . Would you be interested in helping with that? I think a version of this page would work very well there as part of a new/rewritten section on building wheels for linux.

joerick:
Closing due to inactivity. Feel free to reopen if you want to continue.

PR #735, opened 2021-06-25
Show auditwheel/delocate version in default repair command.

bdice:
I'm facing an issue right now where it would be helpful to know the auditwheel version that is used to repair the wheel. https://github.com/glotzerlab/freud/issues/803

This information might be generally helpful to show before performing repairs.

I'm happy to have this PR accepted/edited/rejected, I just wanted to offer the idea. cibuildwheel is amazing and I appreciate your hard work. 👍  I can fix the failing tests if this sounds like a helpful addition.

The expected output would be a line like this for auditwheel (tested on `manylinux2010_x86_64:latest`):
```
auditwheel 4.0.0 installed at /opt/_internal/tools/lib/python3.9/site-packages (python 3.9)
```
or for delocate:
```
delocate-wheel 0.8.2
```

### DIFF
--- a/cibuildwheel/resources/defaults.toml
+++ b/cibuildwheel/resources/defaults.toml
@@ -29,10 +29,14 @@ manylinux-pypy_aarch64-image = "manylinux2014"
 
 
 [tool.cibuildwheel.linux]
-repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"
+repair-wheel-command = [
+  "auditwheel --version",
+  "auditwheel repair -w {dest_dir} {wheel}",
+]
 
 [tool.cibuildwheel.macos]
 repair-wheel-command = [
+  "delocate-wheel --version",
   "delocate-listdeps {wheel}",
   "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}",
 ]


henryiii:
Couldn't we print all (pinned, usually) versions of things at the top somewhere? auditwheel, build, pip, etc? (Like pytest does)

bdice:
> 
> 
> Couldn't we print all (pinned, usually) versions of things at the top somewhere? auditwheel, build, pip, etc? (Like pytest does)

@henryiii This sounds even more helpful. Unfortunately I don't know enough about cibuildwheel to do this quickly and I'm currently busy writing my PhD dissertation. 🙃 One simple and cross-platform approach might be to run `pip list` and grep lines for relevant package names? I'm not sure what list of tools would need to be included beyond what you mentioned.

henryiii:
Seeing the pip version could also be handy for cases like #742.

joerick:
Printing the relevant tool versions at the start of a build would be nice. Perhaps at the end of the "Setting up build environment…" step. `pip freeze` would be a bit noisy, but if we could limit it to just the relevant ones e.g. pip, build, delocate, auditwheel, that would be good.

I'm tidying up PRs so I'll close this one.

PR #738, opened 2021-06-29
fix: support space in constraints file path

### LINKED ISSUE #740: bug: space in directory name can cause issues
When testing out build support in cibuildwheel 2.0.0a4, I run into this failure:

```pytb
python -m build . --wheel --outdir=C:\Users\RUNNER~1\AppData\Local\Temp\cibuildwheelt0jax362\built_wheel --config-setting=-v
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'C:\\Program'
  File "C:\cibw\python\python.3.8.10\tools\lib\site-packages\build\__main__.py", line 302, in main
    build_call(args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependency_check)
  File "C:\cibw\python\python.3.8.10\tools\lib\site-packages\build\__main__.py", line 145, in build_package
    _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
  File "C:\cibw\python\python.3.8.10\tools\lib\site-packages\build\__main__.py", line 101, in _build
    return _build_in_isolated_env(builder, outdir, distribution, config_settings)
  File "C:\cibw\python\python.3.8.10\tools\lib\site-packages\build\__main__.py", line 81, in _build_in_isolated_env
    env.install(builder.build_system_requires)
  File "C:\cibw\python\python.3.8.10\tools\lib\site-packages\build\env.py", line 169, in install
    subprocess.check_call(cmd)
  File "C:\cibw\python\python.3.8.10\tools\lib\subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['C:\\Users\\runneradmin\\AppData\\Local\\Temp\\build-env-s3f7jwu0\\Scripts\\python.exe', '-Im', 'pip', 'install', '--use-pep517', '--no-warn-script-location', '-r', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\build-reqs-ckde1r2k.txt']' returned non-zero exit status 1.

ERROR Command '['C:\\Users\\runneradmin\\AppData\\Local\\Temp\\build-env-s3f7jwu0\\Scripts\\python.exe', '-Im', 'pip', 'install', '--use-pep517', '--no-warn-script-location', '-r', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\build-reqs-ckde1r2k.txt']' returned non-zero exit status 1.
```

I _think_ this is a problem with quoting paths on Windows probably in `build`, or maybe `pip`? The same build works using `pip wheel` from cibuildwheel. Seen in scikit-hep/boost-histogram#583.

henryiii:
Closes #740.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -314,7 +314,7 @@ def build(options: BuildOptions) -> None:
                 build_env = env.copy()
                 if options.dependency_constraints:
                     constr = options.dependency_constraints.get_for_python_version(config.version)
-                    build_env["PIP_CONSTRAINT"] = f"'{constr}'"
+                    build_env["PIP_CONSTRAINT"] = constr.as_uri()
                 build_env["VIRTUALENV_PIP"] = get_pip_version(env)
                 call(
                     [
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -314,7 +314,7 @@ def build(options: BuildOptions) -> None:
                 build_env = env.copy()
                 if options.dependency_constraints:
                     constr = options.dependency_constraints.get_for_python_version(config.version)
-                    build_env["PIP_CONSTRAINT"] = f'"{constr}"'
+                    build_env["PIP_CONSTRAINT"] = f"'{constr}'"
                 build_env["VIRTUALENV_PIP"] = get_pip_version(env)
                 call(
                     [
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -317,20 +317,22 @@ def build(options: BuildOptions) -> None:
                 config_setting = " ".join(verbosity_flags)
                 build_env = env.copy()
                 if options.dependency_constraints:
-                    constr = options.dependency_constraints.get_for_python_version(config.version)
+                    constraints_path = options.dependency_constraints.get_for_python_version(
+                        config.version
+                    )
                     # Bug in pip <= 21.1.3 - we can't have a space in the
                     # constraints file, and pip doesn't support drive letters
                     # in uhi.  After probably pip 21.2, we can use uri. For
                     # now, use a temporary file.
-                    if " " in str(constr):
+                    if " " in str(constraints_path):
                         tmp_file = tempfile.NamedTemporaryFile(
                             "w", suffix="constraints.txt", delete=False, dir=CIBW_INSTALL_PATH
                         )
-                        with tmp_file as constr_file, open(constr) as f:
-                            constr_file.write(f.read())
-                            constr = Path(constr_file.name)
+                        with tmp_file as new_constraints_file, open(constraints_path) as f:
+                            new_constraints_file.write(f.read())
+                            constraints_path = Path(new_constraints_file.name)
 
-                    build_env["PIP_CONSTRAINT"] = str(constr)
+                    build_env["PIP_CONSTRAINT"] = str(constraints_path)
                     build_env["VIRTUALENV_PIP"] = get_pip_version(env)
                     call(
                         [
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -314,7 +314,7 @@ def build(options: BuildOptions) -> None:
                 build_env = env.copy()
                 if options.dependency_constraints:
                     constr = options.dependency_constraints.get_for_python_version(config.version)
-                    build_env["PIP_CONSTRAINT"] = f"'{constr}'"
+                    build_env["PIP_CONSTRAINT"] = f'"{constr}"'
                 build_env["VIRTUALENV_PIP"] = get_pip_version(env)
                 call(
                     [
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -314,7 +314,11 @@ def build(options: BuildOptions) -> None:
                 build_env = env.copy()
                 if options.dependency_constraints:
                     constr = options.dependency_constraints.get_for_python_version(config.version)
-                    build_env["PIP_CONSTRAINT"] = constr.as_uri()
+                    uri = constr.as_uri()
+                    # Bug in pip <= 21.1.3
+                    if uri.startswith(f"file:///{constr.drive}"):
+                        uri = f"file://{uri[8:]}"
+                    build_env["PIP_CONSTRAINT"] = uri
                 build_env["VIRTUALENV_PIP"] = get_pip_version(env)
                 call(
                     [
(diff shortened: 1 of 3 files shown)


henryiii:
For some reason this works on macOS but gives

```
ERROR: Could not open requirements file: [Errno 22] Invalid argument: '/C:/Users/VssAdministrator/AppData/Local/Temp/pytest-of-VssAdministrator/pytest-1/test_dependency_constraints_fi1/constraints file.txt'
```
On windows. Does pip not understand URI's on Windows?

cibuildwheel/windows.py
```diff
@@ -313,9 +313,8 @@ def build(options: BuildOptions) -> None:
                 config_setting = " ".join(verbosity_flags)
                 build_env = env.copy()
                 if options.dependency_constraints:
-                    build_env["PIP_CONSTRAINT"] = str(
-                        options.dependency_constraints.get_for_python_version(config.version)
-                    )
+                    constr = options.dependency_constraints.get_for_python_version(config.version)
+                    build_env["PIP_CONSTRAINT"] = f"'{constr}'"
```
henryiii, line 317:
```suggestion
                    build_env["PIP_CONSTRAINT"] = f'"{constr}"'
```

henryiii:
Both types of quotes do not work.

henryiii:
The best way to fix this would be to rely on  https://github.com/pypa/pip/pull/10116 when it becomes available, but for now, on Windows, we can make a temporary file, which is less likely to contain a space, I think.

henryiii:
Could someone take a look at this? I _think_ it will fix the bug with pip breaking build.

cibuildwheel/windows.py
```diff
@@ -313,22 +313,32 @@ def build(options: BuildOptions) -> None:
...
-                    build_env["PIP_CONSTRAINT"] = str(
-                        options.dependency_constraints.get_for_python_version(config.version)
+                    constr = options.dependency_constraints.get_for_python_version(config.version)
+                    # Bug in pip <= 21.1.3 - we can't have a space in the
+                    # constraints file, and pip doesn't support drive letters
+                    # in uhi.  After probably pip 21.2, we can use uri. For
+                    # now, use a temporary file.
+                    if " " in str(constr):
+                        with tempfile.NamedTemporaryFile(
+                            "w", suffix="constraints.txt", delete=False
```
mayeut, line 323: Why not setting `dir="C:\\cibw\\"` ?
We're already installing python in there and we know for sure there's no space in that path.

mayeut: Should the file be deleted ? or we just don't care to clean up running in CI and all.

henryiii: Well, eventually I'd like to not install to absolute paths, to make it nicer to run on local machines, and requiring install in drive letter C would also also be nice to avoid. But we could for now. I _think_ default temporary directories avoid spaces in their names?

henryiii: It's important to close the file before reading from it on Windows, which is why I have this here -  don't think I can close but not delete without it. I was trying to tell if this gets deleted anyway, but that wasn't clear from some perusal of tmpfile.py. I wasn't that worried about leaving it around, but it would be a little nicer to try to delete it if it's not deleted already.

mayeut: > I think default temporary directories avoid spaces in their names?

I wouldn't bet on that on windows.

If the test passes CI, that means it's true on those images (and it will probably remain that way for CI images, space in path is often a problem). For running on local machines, I'm not so sure it won't happen.

> Well, eventually I'd like to not install to absolute paths, to make it nicer to run on local machines, and requiring install in drive letter C would also also be nice to avoid.

Agreed, also, seeing this comment https://github.com/pypa/packaging.python.org/pull/897#discussion_r664532138, I thought we might be close to being able to run "stateless" on windows however, in this specific instance, I don't think it's a bad idea to use a hardcoded path as long as it's the same as the python install dir, maybe using a top-level "constant" to be able to get this consistently for now (and at least, it won't be forgotten if we go further the "stateless" road).

joerick: > Why not use keyword arg dir="C:\\cibw\\" ?
> We're already installing python in there and we know for sure there's no space in that path.

Sounds like a good idea to me. Looking at [the implementation of `tempfile`](https://github.com/python/cpython/blob/9761abf30617946a374886184b0e36ec050dd8d7/Lib/tempfile.py#L158), there are plenty of places that a space might appear in that code. 

> Should the file be deleted ?

I think it's okay to leave it lying around, as it makes the implementation easier, and this should hopefully be a temporary fix.

joerick (commented):
Other than the note, this looks good to me!

cibuildwheel/windows.py
henryiii: Honestly, since this installs to `C:/cibw`, it's not that bad to run it locally. It's not affecting anything but itself. It would be nicer to install somewhere more local and nicer to not require a C drive, but it's not replacing system Pythons or anything. It's really just macOS that is a very bad idea to run locally.

I've pulled this into a top level constant, and set the dir here.

henryiii:
@joerick Would it be possible to get another dev release after this to see if if fixes the problem I was seeing in the wild?

joerick:
Yep, sure! Let me have another look at this.

joerick:
I hope it's okay with you @henryiii, I just pushed a couple of tidy up commits. If that looks good to you, I should be able to get another alpha out today.

cibuildwheel/windows.py
```diff
@@ -27,15 +27,17 @@
 
 
 def call(
-    args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: Optional[str] = None
+    args: Sequence[PathOrStr], env: Optional[Dict[str, str]] = None, cwd: Optional[PathOrStr] = None
```
henryiii, line 30: Would have been nice with #672, would have fixed this for all OSs. :)

cibuildwheel/windows.py
```diff
@@ -317,20 +317,22 @@ def build(options: BuildOptions) -> None:
                 config_setting = " ".join(verbosity_flags)
                 build_env = env.copy()
                 if options.dependency_constraints:
-                    constr = options.dependency_constraints.get_for_python_version(config.version)
+                    constraints_path = options.dependency_constraints.get_for_python_version(
+                        config.version
+                    )
```
henryiii: FYI, this originally was exactly the number of characters that kept this from wrapping. ;) This is fine, though.

cibuildwheel/windows.py
joerick: haha! yeah... need to decide what to do with that...

PR #739, opened 2021-06-29
Update options table in README for CIBW_BUILD_FRONTEND

joerick:


henryiii (approved):

joerick:
I'm hesitating on merging this because the `stable` docs don't contain this option yet. Think I'll wait until v2.0 is imminent.

henryiii:
The docs will be fine, since they build from the last non-prerelease tag, I think. It would just be the readme on the GitHub page, which already is incorrect for 1.x, since it has the Python 3 only change in it.

henryiii:
I think we should mention the config mode in the readme as well - I'm really loving it so far, such as in the [scikit-build/cmake](https://github.com/scikit-build/cmake-python-distributions/blob/be7c3ff8ce4917f406d14f3e5461496dc12372ba/pyproject.toml) and [scikit-build/ninja](https://github.com/scikit-build/ninja-python-distributions/blob/a5753afc8d9bc66f4fc0835652f4cbdc7cf503eb/pyproject.toml) packages.

henryiii:
Also the current prerelease Pythons and Python 3.10 support part of the README is incorrect for 1.x.

joerick:
Yeah, I noticed that too. Maybe a better way forward would be to note on some options, like: _Availability: new in v2.0.0_. Anyway, this can go in I think.

> I think we should mention the config mode in the readme as well

That's there already! See https://github.com/pypa/cibuildwheel/pull/739/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L126

PR #742, opened 2021-06-30
feat: use feature in-tree-build for pip

### LINKED ISSUE #696: Try --use-feature=in-tree-build
We need to try `--use-feature=in-tree-build`; I think it would help with #466 and will become the default later anyway. Placeholder issue for that.

henryiii:
Closes #696


joerick:
FYI @henryiii, I was curious if this was now available with our latest dep updates, so I merged `main`.

joerick:
I think that it does work, but that dependency versions test pins pip to too old a version.

henryiii:
The latest pip (21.2) should also fix the issue with uri's not working, so we could update. Since this will be the new default, should we just force it in our next minor release or wait till we get pip 21.3?

henryiii:
@joerick, @mayeut, or @YannickJadoul should we do this, or just wait till 21.3?

mayeut:
I'm not sure we should do anything about it except maybe updating the docs.
Let's keep pip default behavior and allow users to change it by passing `PIP_USE_FEATURE` environment variable.
With this PR, we know all tests are passing with the option.

joerick:
I agree, we don't need to rush this, let's just wait for pip 21.3. I think that `PIP_USE_DEPRECATED=out-of-tree-build` will be available as a temporary fallback for users hitting this change unexpectedly.

PR #749, opened 2021-07-08
Add some docs covering initial local builds before CI

joerick:
I've been thinking that some users could use a little bit of information about local builds of their Python packages, before they get into cibuildwheel. This PR adds information about that, first using `pip wheel`, and then a local cibuildwheel run. It's a lot easier to debug problems locally, so this is how I would start setting things up myself.

Feedback welcome! I'll post as draft as I wrote some of this a while ago and I need to review it again properly.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -2,7 +2,108 @@
 title: 'Setup'
 ---
 
-# GitHub Actions [linux/mac/windows] {: #github-actions}
+# Build your wheel locally
+
+Before starting to configure cibuildwheel, it's useful to try to build a wheel
+on your local machine. It's much easier to debug problems on your machine than
+inside a CI system!
+
+## Invoking a build
+
+Start with a clean checkout of your project, inside a new clean virtual
+environment. Make sure your `pip` is up to date, and then invoke:
+
+```shell
+pip wheel -w wheelhouse .
+```
+
+If your build completes without a problem, congratulations! You can move on to
+the next step. Otherwise, you might have one of the following issues:
+
+### Missing build dependencies
+
+If your build needs Python dependencies, you can resolve this by adding
+package names to the
+[`build-system.requires`](https://www.python.org/dev/peps/pep-0518/#build-system-table)
+section of your pyproject.toml. For example, if your project requires Cython
+to build, your pyproject.toml might include a section like this:
+
+```toml
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+    "Cython",
+]
+
+build-backend = "setuptools.build_meta"
+```
+
+If you have other dependencies, you should make a note of these, you'll be
+able to install them using the cibuildwheel options [`CIBW_BEFORE_BUILD`](options.md#before-build) or
+[`CIBW_BEFORE_ALL`](options.md#before-all).
+
+### Actions you need to perform before building
+
+You might need to run some other commands before building, like running a
+script that performs codegen or downloading some data that's not stored in
+your source tree. There are a couple ways to deal with this:
+
+-   Incorporate this into your build process, by adding it to your package's
+    `setup.py`. You can add extra build steps using a structure like this:
+
+    ```python
+    import subprocess
+    import setuptools
+    import setuptools.command.build_py
+
+
+    class BuildPyCommand(setuptools.command.build_py.build_py):
+      """Custom build command."""
+
+      def run(self):
+        # your custom build steps here
+        # e.g.
+        #   subprocess.run(['python', 'scripts/my_custom_script.py'], check=True)
+        setuptools.command.build_py.build_py.run(self)
+
+
+    setuptools.setup(
+        cmdclass={
+            'build_py': BuildPyCommand,
+        },
+        # Usual setup() args.
+        # ...
+    )
+    ```
+
+    This method is usually preferred because in addition to adding being
+    included in the wheel build process, it will help users building from
+    source tarballs as well.
+
+-   Alternatively, you can instruct cibuildwheel to run commands before
+    building your wheel. Take a look at the cibuildwheel options
+    [`CIBW_BEFORE_BUILD`](options.md#before-build) or
+    [`CIBW_BEFORE_ALL`](options.md#before-all).
+
+### Environment variables
+
+Your wheel build might need some environment variables to be set. Consider
+incorporating these into setup.py to allow source tarballs to build (for
+example, using [`extra_compile_args` or
+`extra_link_args`](https://docs.python.org/3/distutils/setupscript.html#other-options)),
+but otherwise, make a note of these for inclusion in the cibuildwheel option
+[`CIBW_ENVIRONMENT`](options.md#environment).
+
+# Run cibuildwheel locally
+
+`cibuildwheel --platform linux`
+
+TODO
+
+# Configure a CI service
+
+## GitHub Actions [linux/mac/windows] {: #github-actions}
 
 To build Linux, Mac, and Windows wheels using GitHub Actions, create a `.github/workflows/build_wheels.yml` file in your repo.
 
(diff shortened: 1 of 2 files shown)


henryiii:
While this is a great idea, I think it would be better to move most of the content to a new page; the "local run of cibuildwheel" should stay on setup, but the rest of the information is optioned information that is not general to all backends, etc. I don't mean that in a bad way, just that it shouldn't be pushing the "how to setup cibuildwheel" part down on the page.

docs/setup.md
```diff
@@ -2,7 +2,146 @@
...
+
+Before starting to configure cibuildwheel, it's useful to try to build a wheel
+on your local machine. It's much easier to debug problems on your machine than
+inside a CI system!
+
+Start with a clean checkout of your project, inside a new clean virtual
+environment. Make sure your `pip` is up to date, and then invoke:
+
+```shell
+pip wheel -w wheelhouse .
```
henryiii, line 15: This really should be `pipx run build` or `pipx run build --wheel` if a valid SDist is not possible. But that gets into more details than we want here, I think?

docs/setup.md
```diff
@@ -2,7 +2,146 @@
...
+able to install them using the cibuildwheel options [`CIBW_BEFORE_BUILD`](options.md#before-build) or
+[`CIBW_BEFORE_ALL`](options.md#before-all).
+
+### Actions you need to perform before building
+
+You might need to run some other commands before building, like running a
+script that performs codegen or downloading some data that's not stored in
+your source tree. There are a couple ways to deal with this:
+
+-   Incorporate this into your build process, by adding it to your package's
```
henryiii, line 50: This is setuptools specific, and does not apply to enscons and only partially to scikit-build. I'd probably add a little "if using setuptools" or similar.

docs/setup.md
```diff
@@ -2,7 +2,146 @@
...
+### Environment variables
+
+Your wheel build might need some environment variables to be set. Consider
+incorporating these into setup.py to allow source tarballs to build (for
+example, using [`extra_compile_args` or
+`extra_link_args`](https://docs.python.org/3/distutils/setupscript.html#other-options)),
+but otherwise, make a note of these for inclusion in the cibuildwheel option
+[`CIBW_ENVIRONMENT`](options.md#environment).
+
+# Run cibuildwheel locally
```
henryiii, line 96: This section is fine on the setup page.

docs/setup.md
```diff
@@ -2,7 +2,146 @@
 title: 'Setup'
 ---
 
-# GitHub Actions [linux/mac/windows] {: #github-actions}
+# Build your wheel locally
```
henryiii, line 5: This should be a new page. This page talks about running cibuildwheel on CI (and locally), not about setuptools details.

docs/setup.md
```diff
@@ -2,7 +2,146 @@
...
+    hoops to jump through. Check [this
+    document](https://docs.microsoft.com/en-us/virtualization/windowscontainers/quick-start/quick-start-windows-10-linux)
+    for more info.
+
+This is convenient as it's quicker to iterate, and because the builds are
+happening in manylinux Docker containers, they're perfectly reproducable.
+
+Install cibuildwheel and run a build like this:
+
+```sh
```
henryiii, line 113:
```suggestion
```console
```
?

docs/setup.md
```diff
@@ -2,7 +2,146 @@
...
+pip install cibuildwheel
+cibuildwheel --platform linux
+```
+
+You should see the builds taking place. You can experiment with options by exporting environment variables, for example:
```
henryiii, line 118: I would mention the config mode here. It is _fantastic_ for local builds - in fact, I'd recommend (in general, maybe not right here) that the config should always allow a default build to work on any machine, with environment variables used to configure for a specific item in a CI matrix. That is, if you don't support pypy, I'd always add a `skip` item for it, even if in CI you never run a build where that matters by always setting `CIBW_BUILD`.

FYI, I would personally always run it this way:

```console
# CIBW_BUILD='cp39-*' pipx run cibuildwheel --platform linux
```

(the `tool.cibuildwheel.linux.before-all` setting would be in the `pyproject.toml` location)

henryiii (commented):
Several of my comments are describing what I'd do, not what I'd expect to be shown here.

henryiii:
If most of this does become a new page, I'd eventually extend it with information about how to build and include a shared library that is properly relocatable (I don't know how to do that, actually) and other "how to build a good wheel" tips.

joerick:
> While this is a great idea, I think it would be better to move most of the content to a new page; the "local run of cibuildwheel" should stay on setup, but the rest of the information is opinionated information that is not general to all backends, etc. I don't mean that in a bad way, just that it shouldn't be pushing the "how to setup cibuildwheel" part down on the page. "How do I write a good package" and "how do I use cibuildwheel" are two different topics for different pages.

Hmm. Yes. A good point. It did feel a bit strange mixing reference docs (the example configs) with this new how-to guide. But I do think that it would be valuable to have something like a 'getting started' tutorial/user guide, that might appear in the docs before the reference stuff, as that's what I'm often looking for when I find a new package. 

I suppose it's more like the distinction between:

- How do I use it? - User guide
- What does it do? - Reference

Maybe you have a suggestion of how to restructure it @henryiii ? I'm now thinking maybe renaming the how-to guide to "Getting started", and adding a new page, that's more like reference docs "CI configuration". Then we can include all the example configs on that page, rather than just the minimal ones.

joerick:
> the rest of the information is opinionated information that is not general to all backends, etc

by the way, I'm not sure what you mean by this? Setuptools is the only way to build a platform wheel, right?

henryiii:
I'm seeing to very distinct types of users. One are the people who currently package wheels, say for mypy, matplotlib, etc. They don't need to see the information in "Creating a good wheel", they just need to see the information in "Setup", to see how to control cibuildwheel itself. One is "unique" information to cibuildwheel, the other is a useful how-to guide that isn't required if you already know what you are doing.

> Setuptools is the only way to build a platform wheel, right?

I thought I just mentioned two alternatives above? Enscons is a PEP 517 builder that uses the SCONS build system and doesn't touch setuptools at all. Scikit-build is a a rather popular system that adapts CMake to setuptools - it's already used by a few projects. Hopefully over the next couple of years I can help pull the setuptools parts out and make it a PEP 517 option (with PEP 621 config? Can hope!). But setuptools is not the only method, and will hopefully be getting more competition in the near future.

Again, I think a guide for making wheels with setuptools is a great idea, but it should not interfere with the documentation for how to setup cibuildwheel, which is a key piece of information all cibuildwheel users need. The guide is only helpful to a subset of users who don't already know (and are using) setuptools.

henryiii:
Again, I don't see it as getting started vs. reference - the "how do I use cibuildwheel" part is not reference. *All* users have to see how to set up cibuildwheel. Only a subset need to learn how to build nice setuptools wheels (and that sort of information is currently part of "tips and tricks" now).

docs/setup.md
joerick: I think so. The reason for using `pip wheel` was to match the cibuildwheel mechanism. Also all users will have `pip`, some won't have `pipx`. But when we switch over to `build` as the default, we should update this.

docs/setup.md
joerick: Yeah coming back to this, without a large-ish change to docs structure, I think you're right.

docs/setup.md
joerick: https://github.com/highlightjs/highlight.js/tree/main/src/languages mkdocs uses highlightjs, so we're limited to their languages

docs/setup.md
joerick: good point, i've added this

docs/setup.md
joerick: 👍

docs/setup.md
joerick: I've reworded it and moved the section to Tips and tricks, as 'Alternatives to cibuildwheel options'. Suggestions welcome on that title, though!

joerick:
Taken another crack at this. Have moved the stuff about writing config into your package into faq.md. While doing that I noticed that file could really use a clean-up, so I've tried to structure it a bit, deleting out-of-date stuff and fixing the odd error. (Did you know we had two, almost identical FAQ entries about Apple Silicon? Must have been a bad merge.)

Anyway. Apologies for the messy diff.

joerick:
Do you want to take another look at this before merge @henryiii ?

PR #750, opened 2021-07-08
ci: fix circle ci image

### LINKED ISSUE #745: Xcode 9.4.1 image deprecated by CircleCI
It will be removed on September 14, 2021. The new minimum will be `10.3.0` .

See https://discuss.circleci.com/t/xcode-deprecation-notice-9-4-1-10-0-0-10-1-0-and-10-2-1/40515

henryiii:
This closes #745. I think we should probably recommend the latest image supported rather than the oldest image supported, similar to other platforms (and would be needed for things like Apple Silicon support anyway), but this PR has the minimum required change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   osx-python3.6:
     macos:
-      xcode: 10.3.0
+      xcode: 12.5.1
     environment:
       PYTHON: python3
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   osx-python3.6:
     macos:
-      xcode: 9.4.1
+      xcode: 10.3.0
     environment:
       PYTHON: python3
     steps:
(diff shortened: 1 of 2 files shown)


joerick:
On reflection, let's go for the latest xcode instead. Apple Silicon is a good reason, and it will also give users the longest runway before deprecation.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
   osx-python3.6:
     macos:
-      xcode: 9.4.1
+      xcode: 10.3.0
```
joerick, line 6:
```suggestion
      xcode: 12.5.1
```

examples/circleci-minimal.yml
```diff
@@ -19,7 +19,7 @@ jobs:
   osx-wheels:
     working_directory: ~/osx-wheels
     macos:
-      xcode: 10.0.0
+      xcode: 10.3.0
```
joerick, line 22:
```suggestion
      xcode: 12.5.1
```

Issue #752, opened 2021-07-09
bug: issue with 2.0.0b1 action


henryiii:
Something is wrong with 2.0.0b1. Not sure what, debugging here. https://github.com/scikit-hep/boost-histogram/pull/605 Will update when I know more.

henryiii:
It might be only affecting the action. The direct usage of pipx is working. The only thing I can think of is the `2>&1` addition, but that is [clearly supported](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_redirection?view=powershell-7.1), so don't know why that would be a problem.

henryiii:
```yaml
    - run: pipx run --spec cibuildwheel==2.0.0b1 cibuildwheel 2>&1
      shell: pwsh
```

Breaks in a really weird way (exactly the same as the action). It gets through the first 30 lines or so, then just dies with no more output and fails on all three OS's. Have no idea why.

henryiii:
I'm trying bash again just to see if they fixed the path format. Bash does handle the redirection correct AFAICT. Last time I tried it it was a Windows style path which didn't work in bash for this command, but did on powershell.

henryiii:
Ahh, I think I see how to fix the path problem on bash. Let's try that.

joerick:
I think this is resolved by v2.0.0b2?

henryiii:
Yes, it is! https://github.com/scikit-hep/boost-histogram/pull/605 is merged now that I modified the bump to b2. Actually rather impressed that dependabot seems to understand it's okay to bump a dev release to another dev release.

PR #753, opened 2021-07-09
fix(action): use bash instead of powershell

henryiii:
Fixing #752. Have no idea why this is broken, but it seems to be broken. Switching to bash fixes it. I'm not sure if we need the `2>&1` in bash;  was the interleaving only showing up in the action (powershell)? Given that redirection was buggy, maybe the output was too?

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -17,14 +17,12 @@ runs:
   using: composite
   steps:
 
-    # This needs powershell-core because github.action path is a Windows-style
-    # path on Windows.  Powershell-core understands both types of paths.
     # Redirecting stderr to stdout to fix interleaving issue in Actions.
     - run: >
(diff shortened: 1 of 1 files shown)


henryiii:
@joerick this is pretty high priority IMO, as the action is broken without it. I'd like a 2.0.0b2 for it. :) (Don't think yanking the pip release is needed, as that's not affected).

Also, I think build is now working correctly.

joerick:
I thiiink we do still need the `2>&1` in bash, I've seen it outside the action too. LGTM.

joerick:
Travis gonna Travis, but I'll get another release out for ya anyway

joerick:
Released as v2.0.0b2

PR #759, opened 2021-07-16
docs: multiline examples

### LINKED ISSUE #755: docs: add multiline examples to each command
All commands should have a multiline example, this came up a few weeks ago when someone hadn't notices the one example we do have. This is a reminder in hopes we add this before 2.0 final. :)

henryiii:
Also includes a typo fix.

Closes #755.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -738,6 +738,11 @@ Platform-specific environment variables are also available:<br/>
 
     # Pass the `--lib-sdir .` flag to auditwheel on Linux
     CIBW_REPAIR_WHEEL_COMMAND_LINUX: "auditwheel repair --lib-sdir . -w {dest_dir} {wheel}"
+
+    # Multi-line example - use && to join on all platforms
+    CIBW_REPAIR_WHEEL_COMMAND: >
+      python scripts/repair_wheel.py -w {dest_dir} {wheel} &&
+      python scripts/check_repaired_wheel.py -w {dest_dir} {wheel}
     ```
 
 !!! tab examples "pyproject.toml"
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -738,23 +738,35 @@ Platform-specific environment variables are also available:<br/>
...
 
 !!! tab examples "pyproject.toml"
 
     ```toml
     # Use delvewheel on windows
     [tool.cibuildwheel.windows]
-    before_build = "pip install delvewheel"
-    repair_wheel_command = "delvewheel repair -w {dest_dir} {wheel}"
+    before-build = "pip install delvewheel"
+    repair-wheel-command = "delvewheel repair -w {dest_dir} {wheel}"
```
henryiii, line 754: These were misspelled.

docs/options.md
```diff
@@ -987,7 +1010,11 @@ Platform-specific environment variables are also available:<br/>
     CIBW_BEFORE_TEST: rm -rf ./data/cache && mkdir -p ./data/cache
 
     # Install non pip python package
-    CIBW_BEFORE_TEST: cd some_dir; ./configure; make; make install
+    CIBW_BEFORE_TEST: >
+      cd some_dir &&
+      ./configure &&
+      make &&
+      make install
```
henryiii, line 1017: This is a bad example - if configure or make broke, it would have possibly still passed, as long as make install completed. Fixed.

joerick (approved):
Great! I'll merge this prior to release

docs/options.md
joerick: Good catch. It's very easy to do, as a Python programmer. I wonder if it would be nice to do `option_name.replace("_","-")`, after we read these. pip does something similar - if you do e.g. `pip install python_dateutil`, it will still find `python-dateutil`, because underscores are invalid.

Filed for discussion #761

PR #784, opened 2021-08-04
fix: fix the libffi package name in doc

Hanaasagi:
The package name is `libffi-devel` in yum.
https://centos.pkgs.org/7/centos-x86_64/libffi-devel-3.0.13-19.el7.x86_64.rpm.html


joerick:
Thanks!

PR #786, opened 2021-08-07
Bring 3.10 out of prerelease

joerick:
Python 3.10 is [now ABI stable](https://www.python.org/downloads/release/python-3100rc1/), and the core devs are calling for packagers to prepare for 3.10 compatibility, so we're bringing 3.10 out from behind the PRERELEASE_PYTHONS flag. Let see if the tests still pass!

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -67,7 +67,7 @@ class IdentifierSelector:
     """
 
     # a pattern that skips prerelease versions, when include_prereleases is False.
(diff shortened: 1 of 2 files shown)


joerick:
> Let see if the tests still pass!

I now see that we have been testing 3.10 on the whole test suite since its introduction! https://github.com/pypa/cibuildwheel/blob/main/test/utils.py#L76

joerick:
Gonna get this in and release v2.1.1 today.

PR #787, opened 2021-08-08
docs: Python 3.10 is not under a pre-release flag anymore

henryiii:


joerick:
Ah yes, thank you Henry!

PR #772, opened 2021-07-20
Tidy up the readme, pushing working examples and changelog into the docs

joerick:
The README has been getting pretty long, especially with the changelog and working examples section. This change makes those sections generated by scripts - the latest 5 releases and top 10 projects - while keeping the full list in the docs. This should also fix the issue with dependabot not finding the release notes for each release  !

PR #811, opened 2021-08-31
fix: nox `update_proj` session

mayeut:
follow-up on #772


mayeut:
Trivial so merging without status.

joerick:
Thanks @mayeut , missed that one!

Issue #818, opened 2021-09-08
macOS: wheels names disregarding deployment target for cp38+


boschmitt:
I have a Github workflow that builds wheels for macOS. Since my library depends on C++17 for both core language features and library features, I set `MACOSX_DEPLOYMENT_TARGET` to be fairly high, i.e., `10.15`.

When building the wheels, I get wheels with the correct name for CPython versions `3.6` and `3.7`. However, the other versions are named `*10.9*`:

```
5 wheels produced in 30 minutes:
  tweedledum-1.1.1-cp310-cp310-macosx_10_9_x86_64.whl
  tweedledum-1.1.1-cp36-cp36m-macosx_10_15_x86_64.whl
  tweedledum-1.1.1-cp37-cp37m-macosx_10_15_x86_64.whl
  tweedledum-1.1.1-cp38-cp38-macosx_10_9_x86_64.whl
  tweedledum-1.1.1-cp39-cp39-macosx_10_9_x86_64.whl
```

I believe the culprit code is:

https://github.com/pypa/cibuildwheel/blob/2709a49715535c71a85aed5b5dedde28fb0685e0/cibuildwheel/macos.py#L286-L290

As a workaround, I have manually set `_PYTHON_HOST_PLATFORM` to be `macosx-10.15-x86_64`, but I'm not sure if there might be other implications.

henryiii:
I feel like this was a bug that we were going to fix, but seems to have not been fixed. It should be fixed, or we should use https://github.com/pypa/wheel/issues/407 which I haven't had a chance to implement yet.

joerick:
Agreed, this looks like a bug to me. I think that we should be setting `_PYTHON_HOST_PLATFORM` to the maximum of our known-good values for the arch and the user supplied MACOSX_DEPLOYMENT_TARGET. In fact, we already have something like that implemented in the tests...

https://github.com/pypa/cibuildwheel/blob/02652e662e1c88a20821273926b76150e3441531/test/utils.py#L89-L99

So I think we should bring the same logic into the code, too. 

Do you have any opinion on this @mayeut? I seem to remember discussing `_PYTHON_HOST_PLATFORM` and `MACOSX_DEPLOYMENT_TARGET`, and I think you added the original `_PYTHON_HOST_PLATFORM` override a few years back. Was there a particular reason that it was static and not tied to `MACOSX_DEPLOYMENT_TARGET`?

henryiii:
[...]

PR #819, opened 2021-09-08
fix: correct the names if a higher MACOS dep target set

henryiii:
Should address #818

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -9,6 +9,8 @@ import tempfile
 from pathlib import Path
 from typing import Any, Dict, List, NamedTuple, Optional, Sequence, Set, Tuple, cast
 
+from packaging.version import Version
+
 from .architecture import Architecture
 from .environment import ParsedEnvironment
 from .logger import log
@@ -176,6 +178,20 @@ def install_pypy(version: str, url: str) -> Path:
     return installation_bin_path
 
 
+def macos_max(*versions: str) -> str:
+    """
+    Return the current target, or the system minimum, whichever is higher.
+    Forces 11+ to always return 0 for the final digit, due to macOS
+    changes.
+    """
+
+    best_version = max(Version(s) for s in versions)
+    if best_version.major >= 11:
+        best_version = Version(f"{best_version.major}.0")
+
+    return f"{best_version.major}.{best_version.minor}"
+
+
 def setup_python(
     python_configuration: PythonConfiguration,
     dependency_constraint_flags: Sequence[PathOrStr],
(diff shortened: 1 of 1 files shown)


cibuildwheel/macos.py
```diff
@@ -271,22 +271,29 @@ def setup_python(
     # PyPy defaults to 10.7, causing inconsistencies if it's left unset.
     env.setdefault("MACOSX_DEPLOYMENT_TARGET", "10.9")
 
+    def macos_depl(sys_value: str) -> str:
+        "Return the current target, or the system minimum, whichever is higher"
+        cur_value = env["MACOSX_DEPLOYMENT_TARGET"]
+        cur_value_tup = tuple(int(v) for v in cur_value.split("."))
+        sys_value_tup = tuple(int(v) for v in sys_value.split("."))
+        return cur_value if cur_value_tup > sys_value_tup else sys_value
```
mayeut, line 279: Should we force `{major}.{minor}` to be returned ? I think so.
If major >= 11, we should force minor to 0 given the new versioning scheme of macOS

henryiii: Yes, you are correct. I thought we did this, I'm remembering the "force to 0 part" - must have been in scikit-build or elsewhere.

cibuildwheel/macos.py
```diff
@@ -271,22 +271,29 @@ def setup_python(
     # PyPy defaults to 10.7, causing inconsistencies if it's left unset.
     env.setdefault("MACOSX_DEPLOYMENT_TARGET", "10.9")
```
mayeut, line 272: should the default deployment target be set to "11.0" if config_is_arm64 ?

henryiii: I don't think it really matters, but we can make sure it's set to 11. I don't think macOS on Apple Silicon does anything different with this set less than 11.

henryiii: Actually, it might be confusing with BEFORE_ALL, which always sets this to 10.9. I think it might be better to leave as is?

mayeut: > I don't think it really matters, but we can make sure it's set to 11. I don't think macOS on Apple Silicon does anything different with this set less than 11.

> Actually, it might be confusing with BEFORE_ALL, which always sets this to 10.9. I think it might be better to leave as is?

You're probably right, it probably doesn't matter much & respects the user override, we can leave it as is.

mayeut (approved):

joerick (commented):
Nice. I wonder if there's a test we can write that would fail in a similar manner to #818 so that we can verify this. Gimme a few mins...

joerick:
I've pushed #820 to add a test for this.

joerick:
I think we can close this @henryiii, if you agree with my assessment in #818?

henryiii:
I think so, will look into fixing it in Scikit-Build.

PR #820, opened 2021-09-09
Add test for #818

joerick:


joerick:
Here's a test for #818. It's currently marked as xfail, because #819 hasn't been applied yet. But on my machine it seems to be passing, which is strange, and indicates that we haven't been able to recreate this bug.

joerick:
I think we can keep this test, it might be useful in the future if we ever get issues like this again. It seems like MACOSX_DEPLOYMENT_TARGET is important enough to a build that it should be included in our test suite. I've removed the xfail, as it looks like we don't need #819.

henryiii (approved):
Okay, but does this slow down the build? Not sure it's worth increasing our test time over? Would be better as a unit test.

joerick:
It's just one python version build, so 

    5.20s call     test/test_wheel_tag.py::test_wheel_tag_is_correct_when_using_macosx_deployment_target

only 5 seconds. Not too bad within a 25minute build!

PR #822, opened 2021-09-10
Create a bug report issue template

joerick:


joerick:
I'm wondering if an issue template might be useful for us. I sometimes find myself asking reporters for a link to their config or a build log, so putting that in the template might help?

henryiii:
I would use issue forms over templates, they are a bit nicer and can be enforced to a degree. See https://github.com/pypa/packaging-problems/issues/new?assignees=&labels=&template=packaging_tutorial.yml&title=Trouble+with+a+packaging+tutorial or pip for example.

joerick:
Huh, I didn't know about issue forms. I suppose they're still in beta so Github didn't recommend them to me. But yeah, they seem potentially nicer. I'll look at using them instead...

henryiii:
Maybe you could add an optional spot for raw code, like a traceback or code sample? Either way, fine with me.

joerick:
Cool, thanks for taking a look. Let's see how we go - I err towards less structure I think.

PR #829, opened 2021-09-17
fix: fallthrough empty manylinux images

### LINKED ISSUE #827: Empty MANYLINUX env variable produces invalid result
If the MANYLINUX environment variables are empty, then cibuildwheel fails, rather than falling back to either the value in the config or the default value. This means code like this is broken:

```yaml
  build_wheels:
    strategy:
      fail-fast: false
      matrix:
        os: [windows-latest, macos-latest, ubuntu-latest]
        arch: [auto64]
        build: ["*"]

        include:
          - os: ubuntu-latest
            arch: auto32
            build: "*"

          - os: ubuntu-latest
            type: ManyLinux1
            arch: auto
            build: "cp{36,37,38}-*"
            CIBW_MANYLINUX_X86_64_IMAGE: skhep/manylinuxgcc-x86_64
            CIBW_MANYLINUX_I686_IMAGE: skhep/manylinuxgcc-i686

    steps:
    - uses: actions/checkout@v2

    - uses: pypa/cibuildwheel@v2.1.2
      env:
        CIBW_BUILD: ${{ matrix.build }}
        CIBW_MANYLINUX_I686_IMAGE: ${{ matrix.CIBW_MANYLINUX_I686_IMAGE }}
        CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.CIBW_MANYLINUX_X86_64_IMAGE }}
        CIBW_ARCHS: ${{ matrix.arch }}
```

This sets `CIBW_MANYLINUX_I686_IMAGE` to an empty string on most jobs, which is then an invalid docker container. This should either be the default, or better yet, the pyproject.toml file value or the default.

I don't think this generalizes to all parameters, though; setting things like the repair command to an empty string is valid.

henryiii:
Fixes #827. This was a regression in 2.0. https://github.com/pypa/cibuildwheel/blob/da84127892e79fab70b0ca4a4d22a23c1db53116/cibuildwheel/__main__.py#L292-L294


cibuildwheel/options.py
```diff
@@ -19,15 +19,26 @@ class ConfigOptionError(KeyError):
...
+            (dict_like.get(key, "") or _dig_first(*others, ignore_empty=ignore_empty))
+            if others
+            else dict_like[key]
+        )
+    else:
+        return (
+            dict_like.get(key, _dig_first(*others, ignore_empty=ignore_empty))
+            if others
+            else dict_like[key]
+        )
```
joerick, line 41: This is getting a little convoluted... maybe this is neater?

```python
def _dig_first(*pairs: Tuple[Mapping[str, Any], str], ignore_empty: bool = False) -> Setting:
    """
    Return the first dict item that matches from pairs of dicts and keys.
    Will throw a KeyError if missing.

    _dig_first((dict1, "key1"), (dict2, "key2"), ...)
    """
    if not pairs:
        raise ValueError('pairs cannot be empty')

    for dict_like, key in pairs:
        if key in dict_like:
            value = dict_like[key]

            if ignore_empty and value == "":
                continue

            return value

    raise KeyError(key)  # type: ignore
```

cibuildwheel/options.py
```diff
@@ -129,14 +140,16 @@ def __call__(
...
+        ignore_empty: bool = False,
     ) -> str:
         """
         Get and return the value for the named option from environment,
         configuration file, or the default. If env_plat is False, then don't
         accept platform versions of the environment variable. If this is an
         array it will be merged with "sep" before returning. If it is a table,
         it will be formatted with "table['item']" using {k} and {v} and merged
-        with "table['sep']".
+        with "table['sep']". Empty variables will not override if empty_invalid
```
joerick, line 151:
```suggestion
        with "table['sep']". Empty variables will not override if ignore_empty
```

joerick (commented):
Thanks @henryiii , just a couple comments

joerick (approved):

nitzmahone:
Nice- just hit a need for this today and was about to submit a PR for it... Thanks!

PR #768, opened 2021-07-18
feat: add musllinux support (PEP 656)

mayeut:
Add musllinux support ([PEP 656](https://www.python.org/dev/peps/pep-0656/)) allowing to build wheels for musl libc based distros (e.g. Alpine)

Fixes #627 

This will remain a draft as long as the following is not available:
- [x] pip (https://github.com/pypa/pip/pull/10144, https://github.com/pypa/pip/issues/10106, #771)
- [x] virtualenv with up-to-date pip (#778)
- [x] auditwheel (https://github.com/pypa/auditwheel/issues/305)
- [x] musllinux images (https://github.com/pypa/manylinux/pull/1135)
- [x] PyPI accepting musllinux wheels (https://github.com/pypa/warehouse/pull/9982) 

In the meantime, some workarounds are used to get this started.

- [x] cibuildwheel documentation will need some update

cc @lkollar

PR #831, opened 2021-09-19
feat(CI): run linux tests with pytest-xdist

mayeut:
This revisits #412.
With musllinux around the corner, CI times are growing quite a bit.

This does not apply to Windows/macOS (yet) for the reasons outlined in #412.

### DIFF
--- a/bin/run_tests.py
+++ b/bin/run_tests.py
@@ -17,7 +17,20 @@ if __name__ == "__main__":
     subprocess.run(unit_test_args, check=True)
 
     # run the integration tests
+    xdist_test_args = []
+    if sys.platform.startswith("linux"):
+        xdist_test_args = ["-n", "2"]
     subprocess.run(
-        [sys.executable, "-m", "pytest", "-x", "--durations", "0", "--timeout=2400", "test"],
+        [
+            sys.executable,
+            "-m",
+            "pytest",
+            *xdist_test_args,
+            "-x",
+            "--durations",
+            "0",
+            "--timeout=2400",
+            "test",
+        ],
         check=True,
     )
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ extras = {
         "jinja2",
         "pytest>=6",
         "pytest-timeout",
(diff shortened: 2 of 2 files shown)


joerick:
So the GHA timings from `main` are currently 37m, this is 23m. But, yeah, #768 takes the linux test time up to 60-70mins, so this is probably gonna be pretty handy.

I am considering if it would be worth being a bit more selective in #768 about which tests we run on muslinux, but it is very reassuring to know that _all_ our tests pass on musllinux, just like manylinux. Maybe we wait and see how performance is in #768 after this is applied.

Czaki:
for windows, it should be possible to achieve by monkeypatching installation path 

https://github.com/pypa/cibuildwheel/blob/d5add4261f58bb3553e25bc9ab6b7d7c7537ead2/cibuildwheel/windows.py#L98

To od this I suggest changing hardcoded "tools" on a global contact. For macOS I have no idea.

mayeut:
> I am considering if it would be worth being a bit more selective in #768 about which tests we run on muslinux, but it is very reassuring to know that all our tests pass on musllinux, just like manylinux. Maybe we wait and see how performance is in #768 after this is applied.

Agreed on selectivity of the tests. This might be something that can have a broader scope than just musllinux IMHO.

> for windows, it should be possible to achieve by monkeypatching installation path

Eventually, I'd like to be able to run cibuildwheel locally on all 3 platforms with no unknown side-effects on all 3 platforms. This has already been mentioned a few times & will require tweaking the Windows & macOS code a bit (Linux running in docker already has only known side-effects:  images being pulled). It will probably be a good idea to keep this use-case in mind when doing that.
I have some ideas for both OS but let's keep this linux only for now.

Czaki:
The biggest problem for me is to install python locally on macOS (other than compile from source). For windows, the only side effect may be NuGet installation.

joerick:
This is bringing the linux times back down to 34m, which is pretty good!

joerick (approved):

PR #832, opened 2021-09-19
Add CPython 3.10 to docs build id table

joerick:
Discovered this was missing in #768


mayeut (approved):
LGTM, let's not wait for Travis-CI

PR #835, opened 2021-09-20
refactor: better code factorization

henryiii:
In working on #799, it is very important to have the code better factorized so that I can apply the changes needed there. I've pulled out just the cleanup (so far) and will present that to keep the changes from being too large, and to allow this to be evaluated on its own.

Features:

* Pulled most of the options configuration out of `__main__.py` and placed it in a function in `options.py:compute_options`. This will be refactored in the future to be run on each `override` (which explains why some non-overridable options are not part of this function). This also improves the separation of the code, "main" stuff like paths is better separated from options handling.
* Architecture lists are placed in `util.py` instead of `__main__.py`.
* Stringization is moved to `BuildOptions`. Eventually, it will be swapped for `BuildOptionsContainer`, which will handle the printing of the overrides.
* For linux.py, pulled out computation of images and build configurations into `get_linux_platforms`, and the core docker run into `build_on_docker`. This will be refactored in the future to compute `before-all` and the docker image selections and break up the computation runs correctly based on that, or to at least verify that `before-all` is not set differently within the same image (TBD, but either case is easier with better code factorization). This refactorization also dramatically reduces the nesting and whitespace usage :heart:. Also makes the try/except much easier to read.
* Also fixed a typing bug with mypy failing on `bin/run_tests.py`. Not sure why it was not failing on CI?

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -2,10 +2,8 @@ import argparse
 import os
 import sys
 import textwrap
-import traceback
-from configparser import ConfigParser
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Union
+from typing import List, Optional, Set, Union
 
 from packaging.specifiers import SpecifierSet
 
@@ -15,38 +13,17 @@ import cibuildwheel.macos
 import cibuildwheel.util
 import cibuildwheel.windows
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
-from cibuildwheel.environment import EnvironmentParseError, parse_environment
-from cibuildwheel.options import ConfigOptions
+from cibuildwheel.options import ConfigOptions, compute_options
 from cibuildwheel.projectfiles import get_requires_python_str
 from cibuildwheel.typing import PLATFORMS, PlatformName, assert_never
 from cibuildwheel.util import (
-    BuildFrontend,
+    MANYLINUX_ARCHS,
+    MUSLLINUX_ARCHS,
     BuildOptions,
     BuildSelector,
-    DependencyConstraints,
     TestSelector,
     Unbuffered,
     detect_ci_provider,
-    resources_dir,
-)
-
-MANYLINUX_ARCHS = (
-    "x86_64",
-    "i686",
-    "pypy_x86_64",
-    "aarch64",
-    "ppc64le",
-    "s390x",
-    "pypy_aarch64",
-    "pypy_i686",
-)
-
-MUSLLINUX_ARCHS = (
-    "x86_64",
-    "i686",
-    "aarch64",
-    "ppc64le",
-    "s390x",
 )
 
 
@@ -171,6 +148,11 @@ def main() -> None:
         sys.exit(2)
 
     package_dir = Path(args.package_dir)
+    output_dir = Path(
+        args.output_dir
+        if args.output_dir is not None
+        else os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse")
+    )
 
     manylinux_identifiers = {
         f"manylinux-{build_platform}-image" for build_platform in MANYLINUX_ARCHS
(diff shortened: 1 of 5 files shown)


mayeut:
I had a quick look. Everything seems good.
I'll have another look tomorrow.

cibuildwheel/linux.py
```diff
@@ -47,6 +47,241 @@ def get_python_configurations(
     ]
 
 
+def get_linux_platforms(
+    options: BuildOptions, python_configurations: List[PythonConfiguration]
+) -> Iterator[Tuple[List[PythonConfiguration], str, str]]:
```
joerick, line 52: I'm pondering naming… so this function is returning a list of Docker containers and the pythons to build within them. So would `get_build_containers` (or `get_configs_by_container`) be a good name? And return a NamedTuple `BuildContainer` with attributes `docker_image`, `python_configurations`, `platform_tag`?

joerick (commented):
I haven't had a huge amount of time to review this, but it looks pretty good to me so far... the linux build refactor has been a long time coming :)

cibuildwheel/linux.py
henryiii: +1 on the NamedTuple, return values are a great place to use them. Some of our NamedTuple's really should be dataclasses, but return values are a place where NamedTuple is correct.

I didn't think much about naming. I thought about it just enough to get GitHub Copilot to guess what I meant and write the functions for me because I'm lazy. ;)

henryiii: (Since I'm mostly moving things, it wasn't that helpful; in Hist, it's been pretty impressive to watch it learn about what I'm doing, and then start writing `from_*` methods, special methods, and tests!)

henryiii: I've been in for a while, but haven't started playing with it until last week or so. Had to get into VSCode (which I'm actually really liking, now that I've got the VI extension & with the terminal enabled - and the extension even works in GitHub's "." key editor, which is great).

cibuildwheel/linux.py
```diff
@@ -28,6 +28,12 @@ def path(self) -> PurePath:
         return PurePath(self.path_str)
 
 
+class BuildConfig(NamedTuple):
+    platform_configs: List[PythonConfiguration]
+    platform_tag: str
+    docker_image: str
```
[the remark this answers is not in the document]
henryiii: Just pick a name and I'll refactor.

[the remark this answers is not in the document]
joerick: `BuildStep` seems good to me, with the function `get_build_steps()`.

cibuildwheel/__main__.py
```diff
@@ -250,105 +202,20 @@ def main() -> None:
...
-        build_verbosity = 0
-
-    # Add CIBUILDWHEEL environment variable
-    # This needs to be passed on to the docker container in linux.py
-    os.environ["CIBUILDWHEEL"] = "1"
-
-    archs = Architecture.parse_config(archs_config_str, platform=platform)
+    build_options = compute_options(
+        options, args.archs, build_selector, test_selector, platform, package_dir, output_dir
+    )
```
joerick, line 207: Hmm, this signature is a bit long. And now we have options parsing in two places, some in `__main__`, and some in `compute_options`. If we're moving some options computation out of `__main__`, we'd be best off moving it all, I think? So this function `compute_options` would take no arguments.

Then the use of these values in `__main__` would be e.g. `build_options.platform`. Which seems fine to me.

joerick (commented):
My time for reviewing is quite limited at the moment, so sorry for the dribs and drabs!

cibuildwheel/__main__.py
henryiii: The function compute_options needs several things from `args`. I chose this split to compute the parts of args that were unrelated to the build_options: the option file, the paths, the selectors, and args-specifics parts (platform and archs). These also happen to be the things that are not specific to a build selector - they will not be able to be overridden, for example. I can revisit it to see if any more could be moved in, but I don't think we should be passing in `args` - command line argument parsing should be done in `__main__`, this function is about reading the environment variables and the (previously opened) options file.

(I've already written `compute_all_options`, which runs this once per override.)

henryiii: Also, I'm intentionally avoiding multi-platform code (importing linux, macOS, windows) outside of __main__. Though I think I might be able to do a bit more, some parts I didn't want to move into a function (identifiers) aren't actually used in this function, they just happen to happen in the middle of the current code block. I'll take a look.

henryiii: Pulled things out further into two functions, take another look.

cibuildwheel/options.py
```diff
@@ -183,3 +200,205 @@ def __call__(
...
+    options: ConfigOptions,
+    args_archs: Optional[str],
+    build_selector: BuildSelector,
+    test_selector: TestSelector,
+    platform: PlatformName,
+    package_dir: Path,
+    output_dir: Path,
+) -> BuildOptions:
+    """
+    Gather options from the command line, environment, and configuration file.
```
henryiii, line 280: Not specific to this function now that I refactored (and this happened to have be written by Copilot, FYI).

```suggestion
    Compute BuildOptions for a single run configuration.
```

Something like that. Will make more since once overrides can run this multiple times.

henryiii:
Anyone had a chance to look at the latest version yet?

cibuildwheel/options.py
joerick: > and this happened to have be written by Copilot, FYI 

This was confusing me quite a bit

cibuildwheel/linux.py
```diff
@@ -47,6 +53,241 @@ def get_python_configurations(
     ]
 
 
+def get_build_step(
```
joerick, line 56:
```suggestion
def get_build_steps(
```

cibuildwheel/__main__.py
joerick: To be honest, these options refactors still aren't really making sense to me... I think it's because they're actually motivated by the change to per-build options. But I don't want to hold you up with your work on TOML overrides. Would you be able to remove the options parsing changes from this PR and we can discuss them in the TOML override PR? I think they will make more sense in that context, so it will be easier to review.

henryiii: All the option parsing is happening in options.py now. The args parsing is the only thing left in `__main__`, because it is main's job to parse `args`, and it needs several things from args (like selecting the platform, and the package_dir).

I think this is a clear step forward without the overrides; now the `__main__.py` is shorter and imports much less stuff, and options handling is in options.py. If you wanted to write a new interface to cibuildwheel with different command line arguments, you could do it and still reuse the rest of the infrastructure. There's much less coupling between command line parsing and reading the options.

henryiii: (The overrides part is really just the breaking of `compute_options` into a sub function `_compute_single_options`, but that's internal, and still improves the grouping and readability a bit by localizing the variables.)

henryiii: (If you still want it out, I can see if I can reorder the commits and pull it out, looks doable, maybe)

henryiii: PS: I also recommend looking at the branch, rather than the diff, to see how the new `__main__.py` and `options.py` flow. I think they read better with better grouping, which is not as obvious from the diff.

henryiii: Press the "." key to drop into GitHub.dev if you want. ;)

joerick: Splitting the args parsing out from compute_options is the thing that's giving me pause, because some of the defaults for the arguments are themselves environment variables. So it's not clear when we call compute_options, if those function arguments are supposed to act like command line arguments (and therefore override config/env vars) or if they already have those built in. But otherwise, yeah, I agree, `__main__` is looking much neater this way. 

But yeah, like I say, I don't want to hold you up, so if it's easier, I'm happy to go with this for now, and we can revisit it later if we need to.

henryiii: Okay, We can put this in, then change it as needed, maybe after finishing the overrides. I have a first version ready, but it needs quite a bit of refactoring to simplify it, and lots of tests (it passes mypy and the original unit tests so far). I'd like to merge this before making that PR to have CI run on it.

mayeut (approved):
Let's not hold off on this.
`linux.py` & `__main__.py` are much more readable with this.
I do like the way argument parsing & option computation are split now.
The `options.py` will be modified again short after getting this in so let's not waste too much time on it (given it's already in an overall good shape IMHO).

PR #854, opened 2021-09-30
feat: override

henryiii:
This is _very much_ a WIP, I just want to see if it breaks anything I'm not running locally, and have a placeholder for the work I did last week. I want to add tests (the new behavior is not tested at all yet), and I also want to look at one major refactor. Currently, the structure attempts to iteratively build on what is (now) there, but I think one more change might make this much simpler (and easier to test). If the "batching" of runs happens before build is called, then this would be able to pull out and generalize consistency checking. Before-all could be simpler too, because it would always happen once, at the beginning of build, because build would be called once per "batch", which is already split correctly. Linux batches will run per docker image, which will solve the docker image issue (and arches is handled for free with that, too). "Before all" will not be allowed to change inside a batch. Batching is tied to overrides, so will probably not try to pull that out into a separate PR.

Those are current thoughts and plans, anyway.


joerick:
Looking pretty good so far! The general structure seems good to me. Would you like a more thorough review now or should I hold off until you're further along?

henryiii:
If you see anything obvious, but I was hoping to get a restructured version in today. I'm also about out of time today, so it might be stable for a little bit. I've started writing `compute_build_batches`, which produce a list of identifiers for a batch, and matching BuildOptionsContainers. They still need to be Containers, since only architectures and docker images force new batches. before-all must be consistent across a batch. The environment is going to be fun for before-all...

henryiii:
Still under some flux. I realize that only linux does batches - there's no splitting by arch on other platforms, before-all really is before _all_. So I'm applying that refactor to Linux only, then I'm going to try to improve on the container concept - if we can "slim-down" a container to be just a subset, then correctness checking is much easier. Also might be handy if the container also is used to pass identifiers, possibly. Still in flux.

Also my naming is atrocious. I can't even read it. I'll pick some standard and clean up so a consistent concept has a consistent name.

joerick:
Ok, cool, let me know when you've got the most important parts done and I'll tag in. Or I can have a look now, if you like. But before that, if you have any tests in your head that are important, it would be great to get them in and then we can iterate on the implementation.

joerick:
One thing I was wondering - do env vars still override TOML overrides? I suppose they should?

joerick:
In terms of naming, yeah, the 'container' word is tricky because we already use it to mean Docker container.

I was wondering about something like this...?

```

           compute_options(...)

                   │
                   │
                   ▼
┌────────────────────────────────────┐
│                                    │
│              Options               │
│                                    │
├────────────────────────────────────┤
│ options_for_build(identifier: str) ├──┐   ┌─────────────────────┐
└────────────────────────────────────┘  │   │                     │
                                        └──▶│    BuildOptions     │
                                            │                     │
                                            └─────────────────────┘
```

So `Options` would be the object that contained everything, and `BuildOptions` refers to options resolved to just one build identifier.

henryiii:
Okay, this is probably ready for a first look. I still need some tests, and it probably could be simplified further, but probably worth a look. @joerick @mayeut @YannickJadoul.

henryiii:
We should make sure we get a nice error like this:
```
pyproject.toml: tool.mypy.overrides sections must be an array. Please make sure you are using double brackets like so: [[tool.mypy.overrides]]
```

henryiii:
Mostly on hold here until I get a review or two. I probably will start the docs if I get a chance, though.

joerick:
Sorry @henryiii , I'm a little busy but I will get a chance to review this at some point this week! Friday at the latest.

joerick:
I have a couple ideas of how the implementation of this might be simplified, but I'm going to have to work them through a bit to be sure. depending on the scale of the changes, I might send a PR to your branch @henryiii

joerick:
These docs look good @henryiii ! I suppose this feature is quite close to release? Do you think we need a few more tests for the new functionality on splitting linux runs based on image/before-all?

henryiii:
Great! I'll see if there's any extra test builds we can do once there's a beta, such as in a couple of repos that look like they would benefit, and it would be nice to get the two dependency updates in (pip 21.3 and pypy 7.3.6), but Travis's broken Sao Paulo workers are really throwing a wrench in manylinux (as well as some other projects).

PR #857, opened 2021-10-06
2.1.x Update deps

joerick:
I'm going to do a 2.1.3 release to get Python 3.10.0 out, so we don't have to wait for #854 to get that out there.


joerick:
Yeah, I'm not sure what is going on... this branch doesn't have the musllinux builds in so it shouldn't be slower 🤔

joerick:
I think I'd like to see a green test if possible, just in case this is a 3.10 issue...

joerick:
Aha, we're green.

PR #867, opened 2021-10-11
chore: adding extra checks, more scoping

henryiii:


henryiii:
manual checks are not failed in CI. pre-commit.ci does not run manual checks. And it's bad if I commit something I think was passing but then it fails, requiring me to edit and repush. Instead, you can just use `git commit -n` (or don't run `pre-commit install`), and intentionally ignore the failure. You don't need to address every failure unless you are ready to have a PR reviewed.

henryiii:
If you really don't like it, we could make it manual, and then it could be run explicitly with `--hook-stage manual`), but never automatically (locally or on CI). I'd prefer not - strictness _helps_ development, not hurts. I would much rather be told immediately to fix something, while it's easy, than to have it hurt later down the road and require a much harder fix. And type ignore / noqa should be rare, and if you are doing WIP comments, you shouldn't be working about passing any checks.

henryiii:
I've added manual staging to the two contentious checks, and the GHA style checking now runs this (though the days for a custom action sadly may be numbered). I'd much rather have them on by default - the point of running pre-commit locally is so that you see what the CI is going to complain about before submitting it to the CI, and "high-speed" local development shouldn't worry about appeasing any of the checks, but I could go with this tradeoff for now, as I'd really like this in - the mypy change alone would have sped up development.

PR #870, opened 2021-10-13
chore(setup): use cfg for extras

henryiii:
I believe I tried this in https://discuss.python.org/t/pyproject-toml-optional-dependencies-redundancy-aka-dry-extras/8428/9 and https://github.com/pypa/pip/issues/10393, but it's supposed to work, and I haven't been able to reproduce the error.


henryiii:
I've tried adding a new extra then depending on it, but it still seems to work. I'm not sure what was wrong before, though I haven't gone back to try the exact version of pip, etc.

Also, we can delete the setup.py if you want. There are two benefits to keeping it: first, GitHub hasn't supported dependency graph with pyproject.toml/setup.cfg, so that won't work. But this isn't a package that shows up in that anyway. Second, much older versions of pip can't do an install from source, and mildly older versions of pip can't do an editable install. I think the only place where the former would be a problem is in the Action, but there we should have a newer version of pip always, and the later only affects developers, and I assume we don't mind updating pip. For now, I've left setup.py there.

PR #872, opened 2021-10-14
Add --print-build-identifiers to GitHub Action

abitrolly:
For troubleshooting

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -22,6 +22,7 @@ runs:
         pipx run
         --spec '${{ github.action_path }}'
         cibuildwheel
+        --print-build-identifiers
         ${{ inputs.package-dir }}
         --output-dir ${{ inputs.output-dir }}
         2>&1


abitrolly:
Ideally the info that helps is a build matrix as a table, which shows both builds scheduled and filtered out, with a reason why they are filtered.

henryiii:
That's basically what I meant; I was thinking:

```
Starting wheel builder image for cp36-manylinux_x86_64,
                                 cp37-manylinux_x86_64,
                                 cp38-manylinux_x86_64,
                                 cp39-manylinux_x86_64,
                                 cp310-manylinux_x86_64:
  docker://quay.io/pypa/manylinux_2_24_x86_64:2021-10-06-94da8f1
```

henryiii:
Correction as I'm adding tests:

> It isn't fixed in the current implementation

Yes, it is. If you set `manylinux2014`, the runs do get merged with `pp37`. I would assume if you set pypy to 2010, it would get merged that way too.

henryiii:
We are going off on a tangent about #854. Ignore us. :)

Basically, there's a good opportunity here to add this - for each scheduled Docker launch, we can print the identifiers that are going to be built in that image. We are discussing the new details about how cibuildwheel groups identifiers in a launch.

In fact, I've got the change here for the new code:

```diff
-            log.step(f"Starting Docker image {build_step.docker_image}...")
+            ids_to_build = [x.identifier for x in build_step.platform_configs]
+            log.step(f"Starting Docker image {build_step.docker_image} for {', '.join(ids_to_build)}...")
```

abitrolly:
> thinking that's exactly what the docs on CIBW_BUILD / CIBW_SKIP do...?

No, they don't explain why selectors are no not `py3` and `pypy3`.

henryiii:
You can't have a 1:1 map between cibuildwheel selectors and wheel names. The selector defines the platform the wheel is built on. The wheel name defines the platform the wheel is run on. Those are often closely related, but not always. If you have an `abi3` or `none` wheel, then all different CPython versions you build on map to exactly the same final wheel name. You can also have multiple platform_tags - manylinux wheels get both the 1/2010/2014 version of the tag and the 2_* version of the tag. macOS can be multi tagged too - it's based on the _target platform_, not the _build platform_.

The "m" was removed in Python 3.4 or 3.5 (don't remember which), so the last time it was in a build selector was with Python 2, in cibuildwheel 1.x. Yes, you had to build all CPython 2.7 wheels twice. Yes, some wheels (like Pandas) forgot to do this.

Personally, I love the table in https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip and the only thing I'd be tempted to change is to duplicate it onto the main page / readme, as I feel like I have to jump there all the time. Adding explanations to it would extend the docs, and not really add all that much, I think. You can easily see the mapping to Python versions on one axes, and platforms on the other.

PR #876, opened 2021-10-14
chore: TOML 1.0 support: drop toml, use tomli

### LINKED ISSUE #534: Idea: maybe use Tomlkit instead of toml?
TOML doesn't really have a "go to" library; toml is the most popular (maybe due to the name), but rather slowly developed. I noticed that Poetry uses [tomlkit](https://github.com/sdispater/tomlkit), which explicitly supports keeping the same format when reading and writing a file; poetry needs to add requirements to pyproject.toml when you type `poetry add`, so that's important to them. It's also perfect for what we do with our writing, so changing over to use that might be an interesting option?

henryiii:
This is required to get TOML 1.0 support, and lightens our dependencies a bit. I think most other packages have already switched (pip, etc). Manually writing the output file since tomli-w doesn't support formatting, and tomlkit doesn't keep the output structure as promised. This is probably simpler/easier for us.

Closes #534.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,6 +58,7 @@ repos:
       - types-jinja2
       - types-certifi
       - types-toml
+      - tomli
   - id: mypy
     name: mypy 3.7+ on bin/
     files: ^((bin|docs)/.*py|noxfile.py)$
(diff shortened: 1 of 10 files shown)


henryiii:
Yes, that's why. We could move it into bin and just add the testing there too. We don't import from bin, so it would be one big file. But I don't mind that way either.

Issue #840, opened 2021-09-22
Do not use Python's str.format() for expanding placeholders in commands


phoerious:
### Description

Placeholders in command environment variables such as `CIBW_BEFORE_BUILD` and `CIBW_BEFORE_TEST` are expanded using Python's `str.format()`:

https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/util.py#L41

This makes it (nearly) impossible to use curly braces in a command, because a pair of single braces will throw a `KeyError` exception and a pair of double braces gets interpreted by the GitHub Actions YAML parser. The only option that might work is something ridiculous like `{{ '{{' }}`.

You should use a different parameter expansion mechanism that doesn't try to expand unknown parameter names.

Czaki:
Did you try to simply escape curly braces `\{`?

phoerious:
`\{` is not a valid escape mechanism. Braces in Python format strings are escaped as `{{` and `}}`.

Czaki:
YAML allows to escape special characters by putting line in `"`.

phoerious:
I am using a `|` multiline string with Bash commands. Putting everything in quotes is beyond inconvenient.

henryiii:
How are you using `{` in bash commands, then? It expands there too, it's a shortcut. I'd avoid using that shortcut if that's what you are doing.

> You should use a different parameter expansion mechanism that doesn't try to expand unknown parameter names

This would be bug prone, confusing to readers, and wouldn't allow us to add new expansions in the future. `{python}` would expand to our Python, while `{python3}` would get passed through and might not error at all. It's much better to be _explicit_ and learn the correct escaping mechanisms. Users would not be able to tell the difference between `{project}` and something that doesn't get substituted. In the future, if we added a new substitution, we could break everyone's code that was expecting that to magically non get substituted. Debugging would be harder, too. You'd no longer get an error message if you tried an incorrect substitution (like python3).

I'd still recommend placing your config in pyproject.toml. Possibly bundle the logic into a script and avoid complex multiline commands if possible. I don't think it's our fault for having a substitution mechanism. Might be more GitHub Action's fault if you want to blame someone. Honestly, I thought GHA used `${{}}`, not just `{{}}`. I've run into GHA collisions before due to cookiecutter also having a simpler escaping scheme. It's not a bug for either library, you just have to learn how to escape.

phoerious:
I am using Bash variables in the command, so I want the expansion on that level.

[...]

PR #889, opened 2021-10-23
Alternative string-based approach to format string passthrough

joerick:
An alternative approach to, and closes #884 that uses regex instead of format_map. 

While working on #884 I felt that there were still quite a lot of gotchas in that implementation i.e.:

- curly brackets that look like positional replacement fields
- unmatched curly brackets

These would require the user to provide additional escaping a-la `{{` and `}}`.

This is an experiment to see if a non-format_map solution might work better.

I'm not sure which one I prefer! The main differences are that, in this PR,

- this approach doesn't have the gotchas mentioned above. Unrecognised syntax is always passed through.
- the escape sequences are different. This uses hash e.g. `#{wheel}` to passthrough a literal `{wheel}`, rather than `{{wheel}}`
- the escape sequence only works on variables known to the command. So in #884, doing `{{unknown}}` is passed through as `{unknown}`, but here the escape sequences are only removed on known params. e.g. `\{unknown}` passes through as `\{unknown}`, untouched. I think this is less disruptive, as most users won't have to think about the escaping unless they hit a naming conflict directly.
- If there are any `{{` escape sequences in the wild, due to the change in escaping rules, this would break that. However, I can only find [one such instance](https://sourcegraph.com/search?q=context:global+CIBW_%28BEFORE_%28ALL%7CBUILD%7CTEST%29%7C%28REPAIR_WHEEL%7CTEST%29_COMMAND%29.*%5B%5E%24%5D%7B%7B&patternType=regexp)

So I suppose I'm leaning towards this approach [^1]. What do you think @henryiii?

[^1]: Despite what I said in #840, sorry! Sometimes it takes writing the code to know.


joerick:
Managed to fix the backslash issue. 

Having looked at it, `{{wheel}}` is actually quite tricky to ignore in regex, because you need to rely on both a negative lookbehind of `{` and a negative lookahead of `}` to properly match it. Perhaps `#{wheel}` would be okay as an escape sequence, what do you think?

henryiii:
If you fixed the backslash, I would think that's fine. `#{}` looks like Ruby to be, but admittedly, it would probably be pretty rare for shell scripts, I think. No strong opinion here, I expect escaping this would be very uncommon.

joerick:
Cool. I'm only really doing the escaping thing for completeness, I'd be surprised if it's ever used! I added a note about its use to the docs, too, all the way at the bottom of the Options page.

So this is ready for another review.

cibuildwheel/util.py
```diff
@@ -48,14 +48,56 @@
...
+                }}  # literal close curly bracket
+            """,
+            re.VERBOSE,
+        )
+
+        # we use a lambda for repl to prevent re.sub interpreting backslashes
+        # in repl as escape sequences
+        result = re.sub(
+            pattern=find_pattern,
+            repl=lambda _: str(value),
```
henryiii, line 83:
```suggestion
            repl=value.replace('\\', r'\\'),
```

FYI, this is the suggestion in the Python docs for escaping here. Don't have a strong preference, just wanted to point it out. See the final suggestion in https://docs.python.org/3/library/re.html#re.escape.

henryiii (approved):

cibuildwheel/util.py
joerick: Thanks, I missed that. Will stick to the lambda for convenience

PR #899, opened 2021-11-01
Fix broken links incorrectly converted to point to PyPA org

CAM-Gerlach:
Upon browsing the docs, I noticed that in the [Automatic section of the Deliver to PyPI docs](https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#automatic-method), the link to the example repo is broken. It seems 3be2d0c indiscriminately converted all references to joerick/cibuildwheel to pypa/cibuildwheel, including those to other repositories with names starting with `cibuildwheel` that were not being moved. Upon checking that commit, I noticed at least one other link that was broken as well, and given it was a rather trivial, straightforward change, went ahead and fixed both in a PR.

NB, link breakage like this could be fixed by a `linkcheck` tool in your CIs; I'm not really familiar with the mkdocs ecosystem as opposed to Sphinx, but at least at a glance, it seems [this](https://pypi.org/project/mkdocs-linkcheck/) tool exists and is well-maintained.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -470,7 +470,7 @@ _26 May 2019_
 - 🌟 Add support for building on Azure pipelines! This lets you build all
   Linux, Mac and Windows wheels on one service, so it promises to be the
   easiest to set up! Check out the quickstart in the docs, or
-  [cibuildwheel-azure-example](https://github.com/pypa/cibuildwheel-azure-example)
+  [cibuildwheel-azure-example](https://github.com/joerick/cibuildwheel-azure-example)
   for an example project. (#126, #132)
 - 🛠 Internal change - the end-to-end test projects format was updated, so we
   can more precisely assert what should be produced for each one. (#136, #137).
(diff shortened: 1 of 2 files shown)


joerick:
Thanks for this! However, I'm not 100% sure these are still good examples of auto-deployment. I think that probably the `/examples` dir in the repo would be a better place to find examples than this repo. Maybe the `docs/deliver-to-pypi.md` changes could point to there instead?

CAM-Gerlach:
@joerick Indeed; I note one of the repos is archived, and the other has not been updated in several years, with a stale PR to add GH Actions support. However, one instance (Azure) is in a two and a half year old changelog entry, so not sure if its worth more than simply fixing to point to what it is intended to, while the second can either be changed to point to the examples dir generally, or—since two of the three specific ones relevant to deployment are already linked directly earlier in the sentence, and third is a variation of one of the existing with additions (testing) unrelated to deployment, it might make sense to simply elide the reference entirely. As the creator of this terrific project and of the files in question, and far more knowledgeable than I on the matters at hand, I'm happy to implement here whatever you judge to be best.

joerick:
Thanks @CAM-Gerlach ! Let's keep the changelog link, but remove the other ones, as you say. If you're able to make the changes to this PR i'll merge it in.

CAM-Gerlach:
Thanks @joerick ! I removed the mention of the other link, with a few textual tweaks to the sentence so it reads better on its own without the other one.

joerick (approved):
Great, thanks.

CAM-Gerlach:
Thanks @joerick !

PR #903, opened 2021-11-03
Update macos download link

### LINKED ISSUE #902: Broken link to Python 3.10 on macos
### Description

The [configuration file](https://github.com/pypa/cibuildwheel/blob/9ce9574d4e6fce57daeb398479d39e3a388945e1/cibuildwheel/resources/build-platforms.toml) indicates that the 3.10 Python file is at https://www.python.org/ftp/python/3.10.0/python-3.10.0-macos11.pkg, but the URL returns a 404.

I notice on https://www.python.org/ftp/python/3.10.0/ that is called https://www.python.org/ftp/python/3.10.0/python-3.10.0-macos11.pkg_superceded.

The consequence is that my build breaks and I cannot deploy :( 

### Build log

https://github.com/SethMMorton/fastnumbers/runs/4075920362?check_suite_focus=true

### CI config

https://github.com/SethMMorton/fastnumbers/actions/runs/1410787082/workflow

joerick:
Used bin/update_pythons.py.﻿ PyPy update was deselected, it's in #881.

Fixes #902.


henryiii:
The old link has been restored (to the fixed version).

joerick:
Yep, this is no longer needed.

PR #906, opened 2021-11-04
Travis-CI: Fix Container Network

### LINKED ISSUE #904: Travis ppc64le network issue work-around
### Description

Hi,

I currently see network issues with docker containers on Travis CI ppc64le builders.
Essentially, connections seem to time out, e.g., on `yum update`.

The same problem seems to be known in conda-forge and adding `--network=host` to `docker run` seems to solve the issue.

Do you know how I can most simply apply the same work-around in cibuildwheel?

X-ref:
- https://github.com/conda-forge/conda-smithy/pull/1520

### Build log

X-ref:
- https://github.com/openPMD/openPMD-api/pull/1136

### CI config

- https://github.com/openPMD/openPMD-api/tree/wheels

ax3l:
Add `--network=host` to Travis-CI PPC64le runners.

Draft to fix #904

Test in https://github.com/openPMD/openPMD-api/pull/1138

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -31,19 +31,21 @@ class DockerContainer:
     bash_stdout: IO[bytes]
 
     def __init__(
-        self, *, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None
+        self, *, docker_image: str, simulate_32_bit: bool = False, network_host: bool = False, cwd: Optional[PathOrStr] = None
     ):
         if not docker_image:
             raise ValueError("Must have a non-empty docker image to run.")
 
         self.docker_image = docker_image
         self.simulate_32_bit = simulate_32_bit
+        self.network_host = network_host
         self.cwd = cwd
         self.name: Optional[str] = None
 
     def __enter__(self) -> "DockerContainer":
         self.name = f"cibuildwheel-{uuid.uuid4()}"
         cwd_args = ["-w", str(self.cwd)] if self.cwd else []
+        network_args = ["--network=host"] if self.network_host else []
         shell_args = ["linux32", "/bin/bash"] if self.simulate_32_bit else ["/bin/bash"]
         subprocess.run(
             [
@@ -53,6 +55,7 @@ class DockerContainer:
                 f"--name={self.name}",
                 "--interactive",
                 "--volume=/:/host",  # ignored on CircleCI
+                *network_args,        # mainly for TravisCI PPC64le
                 *cwd_args,
                 self.docker_image,
                 *shell_args,
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -31,7 +31,12 @@ class DockerContainer:
     bash_stdout: IO[bytes]
 
     def __init__(
-        self, *, docker_image: str, simulate_32_bit: bool = False, network_host: bool = False, cwd: Optional[PathOrStr] = None
+        self,
+        *,
+        docker_image: str,
+        simulate_32_bit: bool = False,
+        network_host: bool = False,
+        cwd: Optional[PathOrStr] = None,
     ):
         if not docker_image:
             raise ValueError("Must have a non-empty docker image to run.")
@@ -55,7 +60,7 @@ class DockerContainer:
                 f"--name={self.name}",
                 "--interactive",
                 "--volume=/:/host",  # ignored on CircleCI
-                *network_args,        # mainly for TravisCI PPC64le
+                *network_args,  # mainly for TravisCI PPC64le
                 *cwd_args,
                 self.docker_image,
                 *shell_args,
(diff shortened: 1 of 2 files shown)


henryiii:
I don't think this should be added for everyone due to travis issues. I also think we need to think about how to provide a little more flexibility for the docker command (say to support podman). Maybe we could add a configuration option for controlling the command, and then you could use overrides to add it just for ppc64le?

Also, ppc64le is really a mess last I checked. Manylinux can't build using it still (maybe this is related).

ax3l:
Sounds good to me, how would you propose the control for the config option?

As an alternative, I tried again building a my project with emulation on GH actions to avoid travis, and its so slow, that I cannot get a single ppc64le build out in 360min ^^

henryiii:
I'd think some sort of option / set of options. I think it should be flexible enough to allow this and #691 too, if possible.

joerick:
What about wrapping this in a `detect_ci_provider() == CIProvider.travis_ci`? I think I'd prefer that to exposing another option on our API.

henryiii:
I was thinking we need this anyway, #691 was about adding support for other runners (podman). If we allowed configuration here, we could also use this to allow users to work around broken Travis runners, solving both issues with one change.

The problem is Travis's, and I don't know if it's temporary or permanent (it's Travis, so it might be permanent if it it should be temporary). I like that detecting it would automatically fix it (but we should probably limit it to just the broken PPC arch, maybe just the Sao Paulo workers, as I think it's only that cluster that's broken).

henryiii:
Actually https://github.com/pypa/cibuildwheel/issues/340 is a better reference. The idea is that podman (and singularity, if it's close enough) are image runners that don't require sudo access. At least podman is close enough to docker that just configuring the command slightly would enable it to be used. And if we have a way to configure that generally, then we could use that to allow someone to add this flag as a workaround for Travis's issue.

joerick:
My feeling (based on a podman fork I saw) is that a change to a different container system would require more than just changing the invocation command of the runner, it would be more like CIBW_BUILD_BACKEND - choosing between a set of container systems (E.g. `CIBW_CONTAINER_RUNNER`, choices `docker`, `podman` etc) - not choosing the exact command to be run. I suppose that the above could be integrated into such an option (`docker-network-host`), but with this being just a temporary work around, it's probably better just to get it in and worry about the API stuff later.

ax3l:
Re-starting CI

henryiii:
Shouldn't we check for Travis CI, and only apply the fix for that CI system?

cibuildwheel/docker_container.py
```diff
@@ -31,19 +31,26 @@ class DockerContainer:
...
         self.docker_image = docker_image
         self.simulate_32_bit = simulate_32_bit
+        self.network_host = network_host
         self.cwd = cwd
         self.name: Optional[str] = None
 
     def __enter__(self) -> "DockerContainer":
         self.name = f"cibuildwheel-{uuid.uuid4()}"
         cwd_args = ["-w", str(self.cwd)] if self.cwd else []
+        network_args = ["--network=host"] if self.network_host else []
```
henryiii, line 53: I'd make this check for Travis CI + ppc64le, and only apply it there. In fact, you probably could/should check for native ppc64le rather than checking the target (cross compiles are unaffected)

ax3l: Happy to do that. Can you guide me on this please?

- As travis detection, just use `if "TRAVIS" in os.environ`?
- How to check for native ppc64le in cibuildwheel?

Thanks! :)

P.S.: If it's faster to push than answer then feel free to also update as you see fit - both good for me :)

henryiii: `cibuildwheel.util.detect_ci_provider() == cibuildwheel.util.CIProvider.travis_ci`

And `platform.machine()` for the platform?

ax3l: Alright, thanks for the guidance! I moved the work-around directly at the current location and removed the interface control for it.

joerick (approved):
Looks good to me! I think the CircleCI build failure is unrelated.

cibuildwheel/docker_container.py
```diff
@@ -31,7 +34,11 @@ class DockerContainer:
     bash_stdout: IO[bytes]
 
     def __init__(
-        self, *, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None
+        self,
+        *,
+        docker_image: str,
+        simulate_32_bit: bool = False,
+        cwd: Optional[PathOrStr] = None,
```
mayeut, line 41: Can you restore this as it was before ?
I guess it was split over multiple lines when the fix went through an argument to this function.

mayeut (changes requested):

mayeut (approved):
LGTM, thanks @ax3l

henryiii (approved):

mayeut:
The failures are all unrelated so merging this.
Let's see if they also appear on main...

mayeut:
failures probably related to https://github.com/pypa/setuptools/issues/3002

PR #914, opened 2021-11-11
feat: support passenv

henryiii:
This is an initial implementation of what passenv would look like. Some thought into escaping quotes is likely needed?

### DIFF
--- a/cibuildwheel/environment.py
+++ b/cibuildwheel/environment.py
@@ -1,8 +1,9 @@
 import dataclasses
-import shlex
-from typing import Dict, List, Mapping, Optional
+from typing import Any, Dict, List, Mapping, Optional, Sequence
 
 import bashlex
+
+from cibuildwheel.typing import Protocol
 
 from . import bashlex_eval
 
@@ -40,7 +41,44 @@ def split_env_items(env_string: str) -> List[str]:
     return result
 
 
-class EnvironmentAssignment:
+class EnvironmentAssignment(Protocol):
+    name: str
+
+    def evaluated_value(
+        self,
+        *,
+        environment: Dict[str, str],
+        executor: Optional[bashlex_eval.EnvironmentExecutor] = None,
+    ) -> str:
+        """Returns the value of this assignment, as evaluated in the environment"""
+        ...
+
+    def __eq__(self, other: object) -> bool:
+        ...
+
+
+class EnvironmentAssignmentRaw:
+    """
+    An environment variable - a simple name/value pair
+    """
+
+    def __init__(self, name: str, value: str):
+        self.name = name
+        self.value = value
+
+    def __repr__(self) -> str:
+        return f"{self.name}: {self.value}"
+
+    def evaluated_value(self, **kwargs: Any) -> str:
+        return self.value
+
+
+class EnvironmentAssignmentBash:
+    """
+    An environment variable, in bash syntax. The value can use bash constructs
+    like "$OTHER_VAR" and "$(command arg1 arg2)".
+    """
+
     def __init__(self, assignment: str):
         name, equals, value = assignment.partition("=")
         if not equals:
@@ -53,14 +91,13 @@ class EnvironmentAssignment:
         environment: Dict[str, str],
         executor: Optional[bashlex_eval.EnvironmentExecutor] = None,
     ) -> str:
-        """Returns the value of this assignment, as evaluated in the environment"""
         return bashlex_eval.evaluate(self.value, environment=environment, executor=executor)
 
     def __repr__(self) -> str:
         return f"{self.name}={self.value}"
 
     def __eq__(self, other: object) -> bool:
-        if isinstance(other, EnvironmentAssignment):
+        if isinstance(other, EnvironmentAssignmentBash):
             return self.name == other.name and self.value == other.value
         return False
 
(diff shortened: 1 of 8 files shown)


henryiii:
Buggy Travis Windows. Anyway, how does this approach look?

cibuildwheel/environment.py
```diff
@@ -84,6 +84,9 @@ def as_dictionary(
 
         return environment
 
+    def add(self, name: str, value: str) -> None:
+        self.assignments.append(EnvironmentAssignment(f'{name}="{value}"'))
```
joerick, line 88: It would be worth a unit test with a double quote and maybe a trailing backslash to check we're quoting properly. But I _think_ this should be `f'{name}={shlex.quote(value)}'`.

cibuildwheel/options.py
```diff
@@ -408,6 +408,7 @@ def build_options(self, identifier: Optional[str]) -> BuildOptions:
             environment_config = self.reader.get(
                 "environment", table={"item": '{k}="{v}"', "sep": " "}
             )
+            passenv = self.reader.get("passenv", sep=",").split(",")
```
joerick, line 411: Some users might naturally write `CIBW_PASSENV="VAR_A VAR_B"` since the delimiter in the CIBW_ENVIRONMENT option is a space. Perhaps we should change the delimiter to a space, to match it?

joerick (commented):
This approach seems good to me, yes. 

In terms of names, how would you feel about `environment_pass`/`CIBW_ENVIRONMENT_PASS`? I'm just thinking to spiritually nest it within the environment option.

cibuildwheel/options.py
henryiii: My thought was to match extras, but I considered space too; that would be fine. To me it's more of a list, while ENVIRONMENT is a dict, but I see why someone might compare them more than EXTRA (which really is a list).

cibuildwheel/environment.py
henryiii: Ahh! I just found that docker passes environment variables without an equals by taking the value from the environment directly. Nice! I can use that instead of trying to quote things correctly.

henryiii:
I've implemented the Docker passthrough version, feel free to take a look. It can drive a new docker launch if you use an override to set it differently, but that should be rare. I like avoiding trying to escape these, and leaving it up to docker.

It seemed to me passenv might be a bit of convention, possibly? I'm not sure if it is. Otherwise I don't have a strong option on it. I would be more inclined toward nesting if we'd made the nesting real (build.all instead of build-all, etc).

cibuildwheel/options.py
henryiii: I've implemented it with space this time.

joerick:
> I've implemented the Docker passthrough version, feel free to take a look. It can drive a new docker launch if you use an override to set it differently, but that should be rare. I like avoiding trying to escape these, and leaving it up to docker.

Hmm. I'm not so sure this is better - a new Docker launch would be quite a surprise to anyone using this in an override, slowing down the build, and probably quite hard to figure out what was causing it. I think we can get the escaping right, I doubt that's a problem. I'm 99% sure that shlex.quote is the right thing, we already use it for this purpose elsewhere, e.g.

https://github.com/pypa/cibuildwheel/blob/08a694485ded60cad4680897874fcf4a940a29c2/cibuildwheel/docker_container.py#L156-L161

> It seemed to me passenv might be a bit of convention, possibly? I'm not sure if it is. Otherwise I don't have a strong option on it. I would be more inclined toward nesting if we'd made the nesting real (build.all instead of build-all, etc).

I had thought so too, but I can only find two references - the [mod_env module of the Apache webserver](https://httpd.apache.org/docs/2.4/mod/mod_env.html#passenv), and [tox](https://tox.wiki/en/latest/config.html#conf-passenv). PassEnvironment actually seems more common. We haven't used 'env' anywhere yet in our options, so it would be better to be internally consistent, so CIBW_ENVIRONMENT_PASS or CIBW_PASS_ENVIRONMENT seem good. My preference would be slightly towards CIBW_ENVIRONMENT_PASS so it would sit next to CIBW_ENVIRONMENT neatly.

henryiii:
environement_pass reads a little oddly, but is fine, then. I'll use that + the manual addition, but would like some help adding a good test for escaping. :)

henryiii:
Only thing left would be one more test.

joerick:
> I'm 99% sure that shlex.quote is the right thing

Famous last words. Turns out shlex.quote -> bashlex doesn't work for everything. Actually I think the problem is in bashlex, but either way, I've implemented a different EnvironmentAssignment that passes straight through without bashlex for these passthrough variables.

cibuildwheel/options.py
```diff
@@ -438,6 +439,14 @@ def build_options(self, identifier: Optional[str]) -> BuildOptions:
                 traceback.print_exc(None, sys.stderr)
                 sys.exit(2)
 
+            # Pass through environment variables
+            if self.platform == "linux":
+                for passedenv in environment_pass:
+                    try:
+                        environment.add(passedenv, os.environ[passedenv])
```
joerick, line 446:
```suggestion
                for env_var_name in environment_pass:
                    try:
                        environment.add(env_var_name, os.environ[env_var_name])
```

unit_test/options_test.py
```diff
@@ -66,3 +70,46 @@ def test_options_1(tmp_path, monkeypatch):
...
+
+
+@pytest.mark.parametrize(
+    "env_var_value",
+    [
+        "normal value",
+        '"value wrapped in quotes"',
+        "an unclosed single-quote: '",
+        'an unclosed double-quote: "',
+        "string\nwith\ncarriage\nreturns\n" "a trailing backslash \\",
```
joerick, line 102:
```suggestion
        "string\nwith\ncarriage\nreturns\n",
        "a trailing backslash \\",
```

henryiii:
Good to go?

joerick:
Yep!

PR #917, opened 2021-11-12
Add python3.exe symlink on windows builds.

### LINKED ISSUE #915: Windows: python and python3 don't point to the same installation
### Description

Hi guys!

I am building wheels with cibw, but noticed some issues with windows builds, where installed packages during `CIBW_BEFORE_BUILD` step would not show up as installed during the build step.
As it turn out, `python3` is not pointing to the installed python version, but tp the hosts python version (?).

Here is an example from the build log linked to, where `cp37` is used:
| cmd  |  path |
| --- | --- |
| `python3` | `/c/hostedtoolcache/windows/Python/3.10.0/x64/python3`
| `python` | `/c/cibw/python/python.3.7.9/tools/python` |

This differs from the behavior in linux and macos, where both `python` and `python3` point to the _same_ installation, examples below.

| platform | cmd  |  path |
| --- | --- | --- |
| linux | `python3` | `/opt/python/cp310-cp310/bin/python3`
| linux | `python` | `/opt/python/cp310-cp310/bin/python` |
| macos | `python3` | `/Library/Frameworks/Python.framework/Versions/3.7/bin/python3`
| macos | `python` | `/tmp/cibw_bin/python` (which looks like a link to the version above) |

It would be great, if `python` on windows builds can also point to the installed cpython version, what do you think?

### Build log

https://github.com/lukeparser/lukeparser/runs/4182800043?check_suite_focus=true

### CI config

https://github.com/lukeparser/lukeparser/blob/88fcaf2de9599c62fffd26c4f85fee99c17278ec/.github/workflows/cibuildwheel.yml#L100

sbrodehl:
This PR adds a `python3.exe` link pointing to `python.exe` on windows, closes #915.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -97,8 +97,9 @@ def install_cpython(version: str, arch: str, nuget: Path) -> Path:
     nuget_args = get_nuget_args(version, arch)
     installation_path = Path(nuget_args[-1]) / (nuget_args[0] + "." + version) / "tools"
     call([nuget, "install", *nuget_args])
+    # "python3" is not included in the vanilla nuget package,
+    # though it can be present if modified (like on Azure).
     if not (installation_path / "python3.exe").exists():
-        # create python3.exe symlink, if not exists
         (installation_path / "python3.exe").symlink_to(installation_path / "python.exe")
     return installation_path
 
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -109,6 +109,7 @@ def install_cpython(version: str, arch: str, nuget: Path) -> Path:
                 if os.path.isfile(os.path.join(installation_path, f))
             ]
         )
+        raise RuntimeError(f"{(installation_path / 'python3.exe')} already exists!")
     return installation_path
 
 
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -102,7 +102,13 @@ def install_cpython(version: str, arch: str, nuget: Path) -> Path:
         (installation_path / "python3.exe").symlink_to(installation_path / "python.exe")
     else:
         print(f'{(installation_path / "python3.exe")} already exists:')
-        print([f for f in os.listdir(installation_path) if os.path.isfile(os.path.join(installation_path, f))])
+        print(
+            [
+                f
+                for f in os.listdir(installation_path)
+                if os.path.isfile(os.path.join(installation_path, f))
+            ]
+        )
     return installation_path
 
 
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -100,16 +100,6 @@ def install_cpython(version: str, arch: str, nuget: Path) -> Path:
     if not (installation_path / "python3.exe").exists():
         # create python3.exe symlink, if not exists
         (installation_path / "python3.exe").symlink_to(installation_path / "python.exe")
-    else:
-        print(f'{(installation_path / "python3.exe")} already exists:')
-        print(
-            [
-                f
-                for f in os.listdir(installation_path)
-                if os.path.isfile(os.path.join(installation_path, f))
-            ]
-        )
-        raise RuntimeError(f"{(installation_path / 'python3.exe')} already exists!")
     return installation_path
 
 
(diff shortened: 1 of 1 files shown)


henryiii:
Hmm, our Nuget Python should(?) be first in the path, and it looks like it does include python3. Is the bug you are seeing true on all Python versions?

sbrodehl:
I can reproduce the error very well with different cpython versions on different windows images on *Github Actions*.
See [https://github.com/lukeparser/lukeparser/runs/4198732270](https://github.com/lukeparser/lukeparser/runs/4198732270?check_suite_focus=true).

---

Here are some infos for `cp10-*` on `windows-2016`:

#### PATH
- `C:\cibw\python\python.3.10.0\tools`
- `...`
- `C:\hostedtoolcache\windows\Python\3.10.0\x64\Scripts`
- `...`

#### where python
- `C:\cibw\python\python.3.10.0\tools\python.exe`
- `C:\hostedtoolcache\windows\Python\3.10.0\x64\python.exe`
- `C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe`
- `C:\ProgramData\Chocolatey\bin\python.exe`

#### where python3
- `C:\hostedtoolcache\windows\Python\3.10.0\x64\python3.exe`
- `C:\hostedtoolcache\windows\Python\3.7.9\x64\python3.exe`

#### which python
- `/c/cibw/python/python.3.10.0/tools/python`

#### which python3
- `/c/hostedtoolcache/windows/Python/3.10.0/x64/python3`

As far I can tell, `C:\cibw\python\python.3.10.0\tools\python3.exe` is missing and a symlink should be possible.

---

As far as *Azure* builds go (here `cp36` `x86` on `windows-2016`):

```shell
2021-11-13T13:57:57.1464359Z ##[group]Installing Python cp36...
2021-11-13T13:57:57.1465470Z + C:\cibw\nuget.exe install pythonx86 -Version 3.6.8 -FallbackSource https://api.nuget.org/v3/index.json -OutputDirectory C:\cibw\python
2021-11-13T13:57:57.1469502Z Package "pythonx86.3.6.8" is already installed.
Debug output:
2021-11-13T13:57:57.1470172Z C:\cibw\python\pythonx86.3.6.8\tools\python3.exe already exists:
2021-11-13T13:57:57.1470983Z ['python.exe', 'python3.dll', 'python3.exe', 'python36.dll', 'pythonw.exe', 'vcruntime140.dll']
```
The pythonx86.3.6.8 package is **already** installed, I currently don't know where the package is installed from and how python3.exe is created, maybe the _Use Python Version task_ also uses nuget and creates a symlink?
Also I don't understand what the windows builds `windows_36` and `windows_38` are testing, different host python versions? Different cpython build versions?

henryiii:
Oops, sorry, one of us has to push a button to get the GHA builds to go! Sorry! I know GitHub Actions makes a python3 symlink (I requested it, actually), but why is that one overriding the local symlink? The path should be setup correctly so that we override any external environment. Oh, wait a minute, I might know what's happening. If Azure is using nuget to set up Python instead of the Action, I bet they added the python3 symlink manually there too when I asked for it in Actions. Therefor NuGet doesn't come with it by default (though IMO it should), so that's why we need to add it. Okay, conditional adding is fine with me, as long as we know why it's needed!

henryiii (approved):

sbrodehl:
Great to see this resolved!

On a side note: GHA builds will be triggered without your consent by `pre-commit-ci` commits, don't know if this is a bug, or a feature. (See 02c8633 above)

cibuildwheel/windows.py
```diff
@@ -97,6 +97,9 @@ def install_cpython(version: str, arch: str, nuget: Path) -> Path:
     nuget_args = get_nuget_args(version, arch)
     installation_path = Path(nuget_args[-1]) / (nuget_args[0] + "." + version) / "tools"
     call([nuget, "install", *nuget_args])
+    if not (installation_path / "python3.exe").exists():
+        # create python3.exe symlink, if not exists
```
henryiii, line 101:
```suggestion
    # "python3" is not included in the vanilla nuget package,
    # though it can be present if modified (like on Azure).
    if not (installation_path / "python3.exe").exists():
```

joerick (approved):

PR #918, opened 2021-11-13
Enable Pycln since the artificial limit removed for all Python3 versions.

hadialqattan:
Re-enabling [Pycln ](https://github.com/hadialqattan/pycln) after solving the Python3.10 artificial limitation issue.
___
This is reverting [this commit](https://github.com/pypa/cibuildwheel/commit/6db52297745c0c0c212ce4cd9177bf5b1497b00b).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,11 +27,10 @@ repos:
 
 # Autoremoves unused imports
 - repo: https://github.com/hadialqattan/pycln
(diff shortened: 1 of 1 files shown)


henryiii:
The problem has been fixed in the future too - now if 3.11 is not supported, it will be due to an actual problem (which is fine and should be rare unless a warning is being produced on 3.10 or 3.9), rather than an artificial limit. Thanks!

PR #920, opened 2021-11-15
add support for building wheels for windows on arm64

nsait-linaro:
This patch enables building python wheels for windows on arm64 targets.

There is no public CI available yet for win/arm64 targets but expecting Azure support soon. This could help to start releasing binary wheels for python once the CI systems are upgraded.


henryiii (approved):
Seems reasonable enough to me. @joerick?

cibuildwheel/windows.py
```diff
@@ -189,10 +187,28 @@ def setup_python(
...
+                "install",
+                "--force-reinstall",
+                "--upgrade",
+                "pip",
+                *dependency_constraint_flags,
+            ],
+            env=env,
+            cwd=CIBW_INSTALL_PATH,
+        )
+
```
joerick, line 211: This might be overcomplicated. If the pip executable is missing, we should already detect that. Could you try looking for whatever you're missing, like-

```
if python_configuration.arch == "ARM64" and not (installation_path / "Scripts" / "piparm64.exe").exists():
    requires_reinstall = True
```

or just  this would be okay for now I think-

```
if python_configuration.arch == "ARM64":
    requires_reinstall = True
```

probably better to move this up to after the requires_reinstall set line too.

cibuildwheel/windows.py
```diff
@@ -73,10 +74,7 @@ def get_python_configurations(
 
     python_configurations = [PythonConfiguration(**item) for item in full_python_configs]
 
-    map_arch = {
-        "32": Architecture.x86,
-        "64": Architecture.AMD64,
-    }
+    map_arch = {"32": Architecture.x86, "64": Architecture.AMD64, "ARM64": Architecture.ARM64}
```
joerick, line 77: It occurs to me that we should update the `build_platforms.toml` format to use the same arch format as the rest of the codebase, rather than having this map here. But we don't have to solve that in this PR.

joerick (commented):
Cool! Have you tested this on an arm64 machine @nsait-linaro? It looks good so far, just wanted to check. I can't test it locally, and we don't have a CI runner for this so we'd have to take your word for it for now!

Other things we'd need to merge this-

- the build_platforms.toml file is actually generated/modified weekly by `bin/update_pythons.py`, so to make this change stick, we'd need to make sure that script still works and retains arm64. Could you try that and make sure it does the right thing?
- Docs updates. Off the top of my head, we'd need to update the build id table in the CIBW_BUILD section, and the table in the README. For now, in the README, could we note  that win-arm64 support is 'experimental', since we can't test this in CI yet.

cibuildwheel/windows.py
nsait-linaro: That's exactly what I tried at first but unfortunately didn't work.

The issue is that pip bundled with python 3.10 builds executables such as pip.exe for x64 platform. The first re-install of pip updates pip module but builds pip.exe using the old pip which still generates x64 executable. But the second re-install uses updated pip and correctly builds pip.exe for the target.

nsait-linaro:
> Cool! Have you tested this on an arm64 machine @nsait-linaro? It looks good so far, just wanted to check. I can't test it locally, and we don't have a CI runner for this so we'd have to take your word for it for now!

Yes, I've tried it on my surface pro and tried building a few packages like `pyzmq` and it seems to work.

I am working on a patch to get the cross-compilation working but require a setuptools update to help with cross-compilation.

nsait-linaro:
Thanks, @joerick and @henryiii for the review. I've addressed the review comments in the latest commits.

nsait-linaro:
[Failing CI](https://app.circleci.com/pipelines/github/pypa/cibuildwheel/3158/workflows/a530df39-86ac-4893-8a4e-f6ab4de66fa1/jobs/8917) seems like an infrastructure issue. Can someone please rerun the test ?

henryiii:
Yes, CircleCI has been timing out here in the last two days or so. Not sure when it's going to be fixed. I think the latest main build has the timeout failure too.

nsait-linaro:
> Yes, CircleCI has been timing out here in the last two days or so. Not sure when it's going to be fixed. I think the latest main build has the timeout failure too.

Thanks

cibuildwheel/resources/build-platforms.toml
```diff
@@ -87,6 +87,8 @@ python_configurations = [
   { identifier = "cp39-win_amd64", version = "3.9.8", arch = "64" },
   { identifier = "cp310-win32", version = "3.10.0", arch = "32" },
   { identifier = "cp310-win_amd64", version = "3.10.0", arch = "64" },
+  { identifier = "cp310-win_arm64", version = "3.9.9", arch = "ARM64" },
```
joerick, line 90: This identifier appears to be wrong? should be `cp39...` i think.

cibuildwheel/windows.py
joerick: ah, hmm, okay. I think I understand. What is the `Version(get_pip_version(env)) < Version("21.3")` about? And are you assuming that requires_reinstall is always true on arm64?

cibuildwheel/resources/build-platforms.toml
henryiii: Shouldn't we have two, possibly, then? 3.9.9 and 3.10.0? Not sure what's actually available.

henryiii (changes requested):
Python identifier needs to match version. Then I'm happy with it. ;)

cibuildwheel/windows.py
henryiii: I'd expect this is a temporary problem and should be fixed once a newer pip is bundled? That's a guess, though, and it would be nice to verify this really was fixed in 21.3 and is not just a problem with nuget.

cibuildwheel/resources/build-platforms.toml
nsait-linaro: > This identifier appears to be wrong? should be `cp39...` i think.

Ah, copy-paste error. I will fix it.

nsait-linaro: > Shouldn't we have two, possibly, then? 3.9.9 and 3.10.0? Not sure what's actually available.

Yes, both are available. I thought I had it in my patch. Will fix thtat

cibuildwheel/windows.py
nsait-linaro: Yes, the fix is available from pip release 21.3 so we don't have to do the extra install.

nsait-linaro:
> Python identifier needs to match version. Then I'm happy with it. ;)

This has been fixed in the latest commit. Thanks

README.md
```diff
@@ -22,16 +22,17 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
...
-| CPython 3.8   | ✅ | ✅  | ✅  | ✅  | ✅ | ✅  | ✅ | ✅  | ✅  |
-| CPython 3.9   | ✅ | ✅  | ✅  | ✅  | ✅ | ✅  | ✅ | ✅  | ✅  |
-| CPython 3.10  | ✅ | ✅  | ✅  | ✅  | ✅ | ✅  | ✅ | ✅  | ✅  |
-| PyPy 3.7 v7.3 | ✅ | N/A | ✅  | N/A | ✅¹ | ✅¹  | ✅¹ | N/A | N/A |
+| CPython 3.6   | ✅ | N/A | ✅  | ✅  | N/A | ✅  | ✅  | ✅ | ✅  | ✅  |
+| CPython 3.7   | ✅ | N/A | ✅  | ✅  | N/A | ✅ |✅  | ✅ | ✅  | ✅  |
+| CPython 3.8   | ✅ | ✅  | ✅  | ✅   | N/A | ✅ | ✅  | ✅ | ✅  | ✅  |
+| CPython 3.9   | ✅ | ✅  | ✅  | ✅  | ✅² | ✅ | ✅  | ✅ | ✅  | ✅  |
+| CPython 3.10  | ✅ | ✅  | ✅  | ✅  | ✅² | ✅ | ✅  | ✅ | ✅  | ✅  |
+| PyPy 3.7 v7.3 | ✅ | N/A | ✅  | N/A | N/A | ✅¹ | ✅¹  | ✅¹ | N/A | N/A |
```
joerick, line 32: Looks like this table markdown isn't parsing any more. Do the first two Pythons have too many columns?

cibuildwheel/windows.py
```diff
@@ -189,10 +187,28 @@ def setup_python(
     # Install pip
 
     requires_reinstall = not (installation_path / "Scripts" / "pip.exe").exists()
+
     if requires_reinstall:
         # maybe pip isn't installed at all. ensurepip resolves that.
         call(["python", "-m", "ensurepip"], env=env, cwd=CIBW_INSTALL_PATH)
 
+    # pip bundled with python 3.10 for arm64 lacks windows launcher and requires an extra pip force-reinstall to get pip executable
```
joerick, line 195:
```suggestion
    # pip older than 21.3 builds executables such as pip.exe for x64 platform.
    # The first re-install of pip updates pip module but builds pip.exe using
    # the old pip which still generates x64 executable. But the second
    # re-install uses updated pip and correctly builds pip.exe for the target.
    # This can be removed once ARM64 Pythons (currently 3.9 and 3.10) bundle
    # pip versions newer than 21.3.
```

cibuildwheel/windows.py
joerick: Got it. I've suggested adding this explanation as a comment, which will help let us know when it can be removed.

joerick (commented):
Just a couple tweaks needed I think!

README.md
henryiii: The line in-between is wrong.

henryiii:
```
|---------------|----|-----|-----|-----|----|-----|----|-----|-----|
```
 -> 
```
|---------------|----|-----|-----|-----|-----|----|-----|----|-----|-----|
```

nsait-linaro: I've tried in a markdown editor and it didn't spot that. Sorry I will fix that.

joerick (approved):
Looks good!

henryiii (approved):
After this goes in, we should add a third note on the table about musllinux not being supported on Alpine 3.14.

PR #926, opened 2021-11-18
always default to manylinux2014

### LINKED ISSUE #909: Move the default for PyPy to manylinux2014
@mattip suggested moving to manylinux2014, since pypy is looking at dropping manylinux2010 in the future. cibuildwheel 2.3 (which will add pypy3.8) might be a good place to do it.

mattip:
closes #909 

Move the default for all platforms to manylinux2014. This is available for all platforms in `resources/pinned_docker_images.cfg`

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -147,6 +147,10 @@ manylinux-x86_64-image = "manylinux2014"
 [[tool.cibuildwheel.overrides]]
 select = "cp36-*"
 manylinux-x86_64-image = "manylinux1"
+
+[[tool.cibuildwheel.overrides]]
+select = "cp3?-*"
+manylinux-x86_64-image = "manylinux2014"
 ```
 
 This example will build CPython 3.6 wheels on manylinux1, and manylinux2014
--- a/docs/options.md
+++ b/docs/options.md
@@ -888,12 +888,12 @@ Auditwheel detects the version of the manylinux / musllinux standard in the Dock
     CIBW_MANYLINUX_PYPY_X86_64_IMAGE: manylinux2014
     CIBW_MANYLINUX_PYPY_I686_IMAGE: manylinux2014
 
-    # Build using the latest manylinux2010 release, instead of the cibuildwheel
+    # Build using the latest manylinux2014 release, instead of the cibuildwheel
     # pinned version
-    CIBW_MANYLINUX_X86_64_IMAGE: quay.io/pypa/manylinux2010_x86_64:latest
-    CIBW_MANYLINUX_I686_IMAGE: quay.io/pypa/manylinux2010_i686:latest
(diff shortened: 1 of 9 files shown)


mattip:
Probably there are places in the documentation to update as well.

mattip:
Why did the circleci run time out?

henryiii:
Because PyPy 3.8 pushed it over the edge, I believe. It mostly times out now, it used to just barely finish in time.

docs/options.md
```diff
@@ -142,19 +142,15 @@ which means it can be given multiple times.
...
-manylinux-x86_64-image = "manylinux2010"
+manylinux-x86_64-image = "manylinux2014"
 
 [[tool.cibuildwheel.overrides]]
 select = "cp36-*"
 manylinux-x86_64-image = "manylinux1"
-
-[[tool.cibuildwheel.overrides]]
-select = "cp310-*"
-manylinux-x86_64-image = "manylinux2014"
```
henryiii, line 153: This example should be included, but it should be reversed, such as:

```toml
select = "cp3?-*"
manylinux-x86_64-image = "manylinux2010"
```

mattip: fixed, thanks

henryiii: Can you revert the change to the text right below, too? Thanks!

docs/options.md
```diff
@@ -142,19 +142,19 @@ which means it can be given multiple times.
...
 [[tool.cibuildwheel.overrides]]
-select = "cp310-*"
+select = "cp3?-*"
 manylinux-x86_64-image = "manylinux2014"
 ```
 
-This example will build CPython 3.6 wheels on manylinux1, CPython 3.7-3.9
-images on manylinux2010, and CPython 3.10 wheels on manylinux2014.
+This example will build CPython 3.6 wheels on manylinux1, and manylinux2014
+wheels everywhere else.
```
mattip, line 157: I admit I am confused by this example. Which stanza has priority? If the `select = "cp36-*"` overrides the default, then won't the `select = "cp3?-*"` re-override that to reset `cp36-*` to `manylinux2014`?

henryiii: Yes, you are right, that's exactly what it would do. The order needs to be swapped, or better yet, it could be `cp3{7,8,9}-*`. :/ I can do that if you want.

mattip: Yes please

joerick:
@mattip I found a few more places where we need to update options, I've added them, but we'll see what the integration tests make of it...

mayeut:
I've rebased this on the current `main` not to get CI failures with circle CI.

henryiii (approved):

mattip:
Thanks

PR #927, opened 2021-11-20
Workaround CircleCI timeouts by deselecting tests

joerick:
We'll need a better solution to this, but for now, every PR being red is making it hard to work on the project, so this will tide us over.


henryiii (approved):
Works for now. :)

Issue #934, opened 2021-11-23
Tracking issue for broken musllinux wheels


henryiii:
See https://github.com/pypa/manylinux/pull/1225 - all muslllinux wheels are broken; the version of Python used in `musllinux_1_1` makes and uses the wrong binary names inside the wheels (ending in `_gnu` instead of `_musl`) In Aline 3.14, they have a fixed version of Python that makes _and only reads_ the new `_musl` ending. So any wheels produced on `musllinux_1_1` do not work past Alpine 3.14.

Possible solutions:

* Make a new manylinux build on alpine 3.14 `musllinux_1_2` (I don't like this; this is not why new platforms are created)
* Make auditwheel fix the names, and provide compatibility symlinks to the pre-3.14 names
* Get a patch into older versions of Alpine to produce the correct names (3.12 is the basis for musllinux

henryiii:
Looks like wheels are zip files and therefore don't support symlinks due to an old CPython bug. So it is not possible to play nice here. Alpine is applying an unaccepted patch onto CPython to change `-gnu` to `-musl`, so it's clearly in the wrong unless something is decided by CPython (maybe fore 3.11, or maybe it will just be left alone, since not much is to be gained by changing it). We should probably note the problem, and hopefully Alpine will drop their unsupported patch until this is resolved long-term.

henryiii:
@ncopa and I spoke yesterday and came to a potential path forward. Quick summary added to the description. This has already been fixed in Alpine; 3.14 and 3.15 can now load musllinux wheels.

mayeut:
Thanks @henryiii & @ncopa for figuring out a path forward. I will try to do my best to get something in auditwheel (been quite busy at work these days without much time/energy left to work on open-source projects).

joerick:
I thought that the patch in Alpine Linux had fixed the issue? (at least for now? I suppose that suffix will get updated at some point, on a CPython minor version, perhaps)

henryiii:
Upstream resolved, and this is old. ;)

PR #938, opened 2021-11-24
docs: note about musllinux mess

henryiii:
See #934.


joerick (approved):

henryiii:
This has already been fixed, so the correct note would be to upgrade to the latest 3.14 and 3.15 python packages. Should we even mention this? I'm kind of thinking it still is good to mention, because some maintainers removed thier musllinux packages when they got reports of them not working, so a note there that they work again might be useful.

joerick:
I think let's keep the note for a few months, we can remove it early 2022.

PR #940, opened 2021-11-28
Add PyPy 3.8 to supported operating systems / Python versions table

Zuzu-Typ:
With cibuildwheel v2.3.0 came support for PyPy 3.8.
This PR simply adds it to the table under _[What does it do?](https://github.com/Zuzu-Typ/cibuildwheel/tree/patch-1#what-does-it-do)_


henryiii:
Oops, we basically have similar tables in two places (the build identifier one and this one), and only the build identifiers got updated. Last time only this one got updated. 🤣

animalize:
The table in this table is not updated as well:
https://cibuildwheel.readthedocs.io/en/stable/

henryiii:
That only gets updated on releases.

PR #947, opened 2021-12-03
Fix Travis deploy example

caspervdw:
In https://github.com/pygeos/pygeos, I moved the aarch64 builds to Travis. It appeared that the `pypi` deploy provider uploads only wheels from the `dist` directory, not the `wheelhouse` directory.


joerick (approved):

joerick:
Thanksm @caspervdw !

PR #948, opened 2021-12-04
Ignore rmtree errors on Windows during cleanup

joerick:
This [fails our CI occasionally](https://app.travis-ci.com/github/pypa/cibuildwheel/jobs/550840205), but we don't really care about the
occasional file escaping deletion - this is more about hygene than
anything else, so ignore errors on this step.

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -449,7 +449,9 @@ def build(options: Options) -> None:
                 shell(test_command_prepared, cwd="c:\\", env=virtualenv_env)
 
                 # clean up
-                shutil.rmtree(venv_dir)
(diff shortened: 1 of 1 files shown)


joerick:
Another example from today - https://app.travis-ci.com/github/pypa/cibuildwheel/jobs/551321337

henryiii:
I think this is us working around issues in Travis CI's buggy Windows implementation, but reluctantly, okay.

henryiii:
IIRC, they have added a process that monitors your files, and that process might be monitoring our files right when we try to delete them.

joerick:
> I think this is us working around issues in Travis CI's buggy Windows implementation, but reluctantly, okay.

Agreed. But worth it for a less flaky CI!

PR #956, opened 2021-12-09
fix: environment variable affect initial setup macOS/win

### LINKED ISSUE #942: Setting PIP options can break environment setup
### Description

Setting a pip option, like `PIP_USE_DEPRECATED=out-of-tree-build` can break the initial setup, because this option gets applied to the original pip version (21.1 on Windows) instead of our pinned version. We should possibly ignore PIP_* variables or all environment variables during this initial setup for the pinned versions of things.

See https://github.com/pybind/cmake_example/pull/66

Example of problematic line: https://github.com/pypa/cibuildwheel/blob/f717468ff60fbd2ee76b6327596d078ad0ec13d5/cibuildwheel/macos.py#L232

### Build log

https://github.com/pybind/cmake_example/runs/4346290813?check_suite_focus=true

### CI config

https://github.com/pybind/cmake_example/blob/c06fc016af76bd0f6a637705afc5d4cbfa554f5a/.github/workflows/wheels.yml#L49

henryiii:
Closes #942. macOS and Windows currently do not support setting an environment variable that is not supported in the original pip but in the pinned pip we install.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -192,7 +192,6 @@ def setup_python(
     # https://github.com/pypa/virtualenv/issues/620
     # Also see https://github.com/python/cpython/pull/9516
     env.pop("__PYVENV_LAUNCHER__", None)
-    env = environment.as_dictionary(prev_environment=env)
 
     # we version pip ourselves, so we don't care about pip version checking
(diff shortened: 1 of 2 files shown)


joerick (approved):
Looks good to me!

PR #966, opened 2021-12-17
Podman support on Linux

Erotemic:
I've been using an old fork of cibuildwheel for awhile (discussion on this in https://github.com/pypa/cibuildwheel/issues/676) and I've taken an hour to port the changes to the latest version of cibuildwheel.

The basic idea is that docker is now abstracted via a `CIBW_OCI_EXE` variable. By setting CIBW_OCI_EXE="podman". And using that in the appropriate places, it should "just work".

However, podman does have some edge cases that make it different than docker. It can require special additional arguments to make it work correctly. In previous versions of my code I hard coded those options, but in this version, if the user is not using docker, I make them specify exactly what they need. In my use cases the following environment variables work: 

```python
export CIBW_OCI_EXE="podman"
export CIBW_OCI_EXTRA_ARGS_CREATE="--events-backend=file --privileged"
export CIBW_OCI_EXTRA_ARGS_COMMON="--cgroup-manager=cgroupfs --storage-driver=vfs --root=$HOME/.local/share/containers/vfs-storage/"
export CIBW_OCI_EXTRA_ARGS_START="--events-backend=file --cgroup-manager=cgroupfs --storage-driver=vfs"
```

Lastly, there are a few instances where podman worked very differently than docker. The biggest one is `copy_out`, which I had to completely redo. 

Other minor issues are:
* for podman there seems to be some race condition in `__exit__`, but adding a few sleeps fixes it
* podman doesnt have the same cwd semantics as docker, so I couldn't use the `-w` argument, and I just explicitly change working dirs (shouldn't be a big issue).
* There is an exclude hack in `copy_into`. Not sure if that is needed anymore.

I've done some minor testing of this port locally and it seems to work (the general logic I've been using in production for months now). I imagine this PR probably needs some discussion and modifications, but the general idea is here, and ported to the latest version. 

At the very least, I'll try to keep this PR up to date so I can continue using this feature, as I do need it.


Erotemic:
So it looks like most of the tests are passing, but for some reason `unit_test/docker_container_test.py` seems to fail. Any insights on why this is would be helpful.

jshowacre:
Hi @Erotemic, what are your thoughts on using the docker-py package? Podman strives to maintain compatibility with this and I have a slightly modified test suite that seems to work with podman's Compat API. 
<details><summary>docker_container_remote.py</summary>
<p>

  ```python
import io
import json
import os
import shlex
from pathlib import PurePath
import subprocess
import sys
import uuid
from tarfile import TarFile
from docker import DockerClient, from_env
import docker
from docker.models.containers import Container
from pathlib import Path, PurePath
from types import TracebackType
from typing import IO, Dict, List, Optional, Sequence, Type, cast

from .typing import PathOrStr, PopenBytes

class RemoteDockerContainer:
    """
    An object that represents a remote running Docker container.

    Intended for use as a context manager e.g.
    `with DockerContainer(docker_image = 'ubuntu') as docker:`

    A bash shell is running in the remote container. When `call()` is invoked,
    the command is relayed to the remote shell, and the results are streamed
    back to cibuildwheel.
    """

    UTILITY_PYTHON = "/opt/python/cp38-cp38/bin/python"
    client: DockerClient
    cont: Container

    def __init__(
        self, *, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None
    ):
        if not docker_image:
            raise ValueError("Must have a non-empty docker image to run.")

        self.docker_image = docker_image
        self.simulate_32_bit = simulate_32_bit
        self.cwd = cwd
        self.name: Optional[str] = None
        self.client = from_env()

    def __enter__(self) -> "RemoteDockerContainer":
        self.name = f"cibuildwheel-{uuid.uuid4()}"
        cwd_args = ["-w", str(self.cwd)] if self.cwd else []
        shell_args = ["linux32", "/bin/bash"] if self.simulate_32_bit else ["/bin/bash"]
        image = self.client.images.pull(self.docker_image)
        self.cont = self.client.containers.create(image, name=self.name, command=shell_args, working_dir=self.cwd, environment=['CIBUILDWHEEL'], network_mode="host", auto_remove=True, stdin_open=True)
        self.cont.start()
        return self

    def __exit__(
        self,
        exc_type: Optional[Type[BaseException]],
        exc_val: Optional[BaseException],
        exc_tb: Optional[TracebackType],
    ) -> None:
        self.cont.stop()
        # if (self.cont is not None):
            # self.cont.remove()

    def copy_into(self, from_path: Path, to_path: PurePath) -> None:
        with io.BytesIO() as mem, TarFile.open(fileobj=mem, mode='w|gz') as tar:
            tar.add(from_path, arcname=from_path.name)
            # tar.list()
            tar.close()
            mem.seek(0)
            self.cont.put_archive(to_path, mem.getvalue())

    def copy_out(self, from_path: PurePath, to_path: Path) -> None:
        data, stat = self.cont.get_archive(from_path, encode_stream=True)
        with io.BytesIO() as mem:
            for chk in data:
                mem.write(chk)
            mem.seek(0)
            with TarFile.open(fileobj=mem) as tar:
                # tar.list()
                tar.extractall(path=to_path, numeric_owner=True)

    def glob(self, path: PurePath, pattern: str) -> List[PurePath]:
        glob_pattern = os.path.join(str(path), pattern)

        path_strs = json.loads(
            self.call(
                [
                    self.UTILITY_PYTHON,
                    "-c",
                    f"import sys, json, glob; json.dump(glob.glob({glob_pattern!r}), sys.stdout)",
                ],
                capture_output=True,
            )
        )

        return [PurePath(p) for p in path_strs]

    def call(
        self,
        args: Sequence[PathOrStr],
        env: Optional[Dict[str, str]] = None,
        capture_output: bool = False,
        cwd: Optional[PathOrStr] = None,
    ) -> str:
        env = dict() if env is None else env
        env = dict([(shlex.quote(k), shlex.quote(v)) for k, v in env.items()])
        args = shlex.join([(str(p) if isinstance(p, PurePath) else p) for p in args])
        return self.cont.exec_run(args, workdir=cwd, environment=env, demux=False, stream=False).output

    def get_environment(self) -> Dict[str, str]:
        env = json.loads(
            self.call(
                [
                    self.UTILITY_PYTHON,
                    "-c",
                    "import sys, json, os; json.dump(os.environ.copy(), sys.stdout)",
                ],
                capture_output=True,
            )
        )
        return cast(Dict[str, str], env)

    def environment_executor(self, command: List[str], environment: Dict[str, str]) -> str:
        # used as an EnvironmentExecutor to evaluate commands and capture output
        return str(self.call(command, env=environment, capture_output=True), errors="surrogateescape")


def shell_quote(path: PurePath) -> str:
    return shlex.quote(str(path))

  ```
</p>
</details>

<details><summary>docker_container_remote_test.py</summary>
<p>

  ```python
import platform
import random
import shlex
import shutil
import subprocess
import textwrap
from pathlib import Path, PurePath

import pytest

from cibuildwheel.docker_container_remote import RemoteDockerContainer
from cibuildwheel.environment import EnvironmentAssignmentBash

# for these tests we use manylinux2014 images, because they're available on
# multi architectures and include python3.8
DEFAULT_IMAGE = "pypa/manylinux2014_%s:2020-05-17-2f8ac3b" % platform.machine()

@pytest.mark.docker
def test_simple():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        assert container.call(["echo", "hello"], capture_output=True) == str.encode("hello\n")

@pytest.mark.docker
def test_no_lf():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        assert container.call(["printf", "hello"], capture_output=True) == str.encode("hello")

@pytest.mark.docker
def test_environment():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        assert (
            container.call(
                ["sh", "-c", "echo $TEST_VAR"], env={"TEST_VAR": "1"}, capture_output=True
            )
            == str.encode("1\n")
        )

@pytest.mark.docker
def test_cwd():
    with RemoteDockerContainer(
        docker_image=DEFAULT_IMAGE, cwd="/cibuildwheel/working_directory"
    ) as container:
        assert container.call(["pwd"], capture_output=True) == str.encode("/cibuildwheel/working_directory\n")
        assert container.call(["pwd"], capture_output=True, cwd="/opt") == str.encode("/opt\n")

@pytest.mark.docker
def test_container_removed():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        docker_containers_listing = subprocess.run(
            "docker container ls",
            shell=True,
            check=True,
            stdout=subprocess.PIPE,
            universal_newlines=True,
        ).stdout
        assert container.name is not None
        assert container.name in docker_containers_listing
        old_container_name = container.name

    docker_containers_listing = subprocess.run(
        "docker container ls",
        shell=True,
        check=True,
        stdout=subprocess.PIPE,
        universal_newlines=True,
    ).stdout
    assert old_container_name not in docker_containers_listing

@pytest.mark.docker
def test_large_environment():
    # max environment variable size is 128kB
    long_env_var_length = 127 * 1024
    large_environment = {
        "a": "0" * long_env_var_length,
        "b": "0" * long_env_var_length,
        "c": "0" * long_env_var_length,
        "d": "0" * long_env_var_length,
    }

    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        # check the length of d
        assert (
            container.call(["sh", "-c", "echo ${#d}"], env=large_environment, capture_output=True)
            == str.encode(f"{long_env_var_length}\n")
        )

@pytest.mark.docker
def test_binary_output():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        # note: the below embedded snippets are in python2

        # check that we can pass though arbitrary binary data without erroring
        container.call(
            [
                "/usr/bin/python2",
                "-c",
                textwrap.dedent(
                    """
                    import sys
                    sys.stdout.write(''.join(chr(n) for n in range(0, 256)))
                    """
                ),
            ]
        )

        # check that we can capture arbitrary binary data
        output = container.call(
            [
                "/usr/bin/python2",
                "-c",
               textwrap.dedent(
                    """
                    import sys
                    sys.stdout.write(''.join(chr(n % 256) for n in range(0, 512)))
                    """
                ),
            ],
            capture_output=True,
        )

        for i in range(512):
            assert output[i] == i % 256

        # check that environment variables can carry binary data, except null characters
        # (https://www.gnu.org/software/libc/manual/html_node/Environment-Variables.html)
        binary_data = bytes(n for n in range(1, 256))
        binary_data_string = str(binary_data, encoding="raw_unicode_escape", errors="surrogateescape")
        output = container.call(
            ["python2", "-c", 'import os, sys; sys.stdout.write(os.environ["TEST_VAR"])'],
            env={"TEST_VAR": binary_data_string},
            capture_output=True,
        )
        output_string = ''.join(shlex.split(str(output, errors="surrogateescape")))
        assert output_string == binary_data_string
        assert output_string.encode(encoding="raw_unicode_escape", errors="surrogateescape") == binary_data

@pytest.mark.docker
def test_file_operations(tmp_path: Path):
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        # test copying a file in
        binary_data = bytes(random.randrange(256) for _ in range(1000))
        test_file = tmp_path / "test.dat"
        test_file.write_bytes(binary_data)

        dst_folder = PurePath("/tmp/test.dat/")
        container.copy_into(test_file, dst_folder)

        output = container.call(["cat", dst_folder / "test.dat"], capture_output=True)
        assert binary_data == output

@pytest.mark.docker
def test_dir_operations(tmp_path: Path):
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        binary_data = bytes(random.randrange(256) for _ in range(1000))
        original_test_file = tmp_path / "test.dat"
        original_test_file.write_bytes(binary_data)

        # test copying a dir in
        test_dir = tmp_path / "test_dir"
        test_dir.mkdir()
        test_file = test_dir / "test.dat"
        shutil.copyfile(original_test_file, test_file)

        tmp = PurePath("/tmp")
        dst_dir = tmp / "test_dir"
        dst_file = dst_dir / "test.dat"
        container.copy_into(test_dir, tmp)

        output = container.call(["cat", dst_file], capture_output=True)
        assert binary_data == output

        # test glob
        assert container.glob(dst_dir, "*.dat") == [dst_file]

        # test copy dir out
        new_test_dir = tmp_path / "test_dir_new"
        new_test_dir.mkdir()
        container.copy_out(dst_dir, new_test_dir)

        assert binary_data == (new_test_dir / "test_dir" / "test.dat").read_bytes()

@pytest.mark.docker
def test_environment_executor():
    with RemoteDockerContainer(docker_image=DEFAULT_IMAGE) as container:
        assignment = EnvironmentAssignmentBash("TEST=$(echo 42)")
        assert assignment.evaluated_value({}, container.environment_executor) == "42"

  ```
</p>
</details>


<details><summary>Dockerfile.tests.alpine</summary>
<p>

  ```
FROM alpine:3.15
# podman-docker is at 3.4.2 which currently which causes the 'cwd' test to break. Fixed in 3.4.3.
# https://github.com/containers/podman/issues/11842
# Privileged atm, but rootless!
RUN apk add python3 py3-pip py3-setuptools podman shadow shadow-uidmap
RUN adduser worker; usermod -v 10000-42767 -w 10000-42767 worker;
RUN chmod 644 /etc/containers/containers.conf; sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' /etc/containers/storage.conf; \
	sed -i -e 's|^unqualified-search-registries[[:space:]]*=.*$|unqualified-search-registries = ["quay.io", "docker.io"]|g' /etc/containers/registries.conf; \
	# Should we be remote by default?
	sed -i -e 's|^#[[:space:]]remote[[:space:]]*=.*$|remote = true|g' /usr/share/containers/containers.conf /etc/containers/containers.conf; touch /etc/containers/nodocker;
RUN mkdir -p /var/lib/shared/overlay-images /var/lib/shared/overlay-layers /var/lib/shared/vfs-images /var/lib/shared/vfs-layers; touch /var/lib/shared/overlay-images/images.lock; touch /var/lib/shared/overlay-layers/layers.lock; touch /var/lib/shared/vfs-images/images.lock; touch /var/lib/shared/vfs-layers/layers.lock; ln -sv /usr/bin/podman /usr/bin/docker
WORKDIR /home/worker/project
RUN mkdir -p /home/worker/.local/share/containers; chown worker:worker -R /home/worker; ln -sv /home/worker/project /project
USER worker
COPY --chown=worker ./setup* ./
RUN pip install -e '.[test]';
COPY --chown=worker . ./
ENV CONTAINER_HOST="unix:///tmp/podman-run-1000/podman/podman.sock"
ENV DOCKER_HOST="unix:///tmp/podman-run-1000/podman/podman.sock"
  ```
</p>
</details>

I also have a fedora Dockerfile that seems to work ok.


> Not sure what the right way to add podman tests to the test suite is.

Someone's guidance would be greatly appreciated here as well.

If you are interested in my direction I could make a PR to your fork, or make a draft PR instead.

Erotemic:
@jshowacre I think docker-py might make sense, but it is another dependency. However, it would make maintaining this package easier if that logic was outsourced and maintained elsewhere.

I suppose it is up to the maintainers as to which direction to go, but IMO it is a good idea to incorporate docker-py (haven't looked at it in detail, I'm just assuming its well maintained and provides a nice generalized OCI API). But input from the maintainers would be useful.

joerick:
On docker-py, I think that it an implementation based on that would be quite a bit slower than our current implementation. A lot of the stream-based work in the current DockerContainer is because `docker exec` was too slow for cibuildwheel's use. Especially in the test suite we fire commands back-and-forth between Docker and Python very often, so this needs to be pretty fast. For discussion on this see https://github.com/pypa/cibuildwheel/pull/386 .

But I appreciate that you guys have been putting good work into this! So I've had a think this morning, as for the integration of podman into the codebase, here's my take on a design for this feature:

- a `CIBW_CONTAINER_RUNTIME_LINUX` option, choices are `docker`,`podman`, default `docker`. 
- That switch controls if a `DockerContainer` or a `PodmanContainer` are used for the `docker` variable in linux.py (while we're at it, we should probably rename that variable to `container`). 
- Then the PodmanContainer class can be written to fulfil the same interface as DockerContainer - `__enter__`, `__exit__`, `copy_into`, `call` etc. Performance of this might not be quite so critical, so we might be able to get away with `podman exec`, which would probably be nicer as that keeps the implementation simple.
- As for testing, we already have quite good unit tests of the DockerContainer implementation (`unit_test/docker_container_test.py`), so parameterizing this with PodmanContainer would provide pretty good assurance that the implementations are compatible. An option test, to ensure the option is parsed correctly. Then just a simple integration test build using `CIBW_CONTAINER_RUNTIME_LINUX=podman` would ensure that the new object fits into linux.py would be good.
- As a side note, if we can figure out a way to get rid of that pesky `sleep`, that would be good. Those are the fixes that come back to haunt you in my experience!

Erotemic:
@joerick Most of that sounds good. 

Right now I'm calling the `CIBW_CONTAINER_RUNTIME_LINUX` option `CIBW_OCI_EXE`. Where OCI stands for open-container-initiative, as the idea would be that eventually more than docker and podman could be supported. I'm not a big fan of putting `_LINUX` as a suffix, nothing else seems to follow that pattern. Perhaps `CIBW_CONTAINER_ENGINE` would be more a more appropriate and clear name? 

I'm also not sure if there should be a separate class for `DockerContainer` and `PodmanContainer`. The number of instances where there is need to differentiate between them is small, and I imagine updates to podman will reduce those further, as their goal is to eventually have podman become a drop-in replacement for docker. The main issue is that there needs to be a way for the user to specify extra-arguments to whatever the container engine is. This would handle the most cases in the most general way, but it might also make sense for cibuildwheel to supply sensible defaults depending on if the container engine is docker or podman.

I'm not sure how to get rid of the sleep. I agree it is a bad hack, but I'm not sure where to start debugging it. Perhaps it would be a good idea to expose a parameter that lets the user adjust the sleep time? That would provide the user with a workaround if it ever wasn't sufficient for a use-case, and it would allow for easier experimentation with disabling it / finding the root cause.

My question about testing podman on CI was less about how to write the unit tests (I did see `docker_container_test`), it was more about how do I create a CI environment where podman exists ensure podman exists on the CI: which CI do I add that to? github actions? CircleCI? TravisCI? All of them?

So, to summarize, the actionable questions I have are: 

* What name do we give what I'm currently calling `CIBW_OCI_EXE`? @joerick proposed `CIBW_CONTAINER_RUNTIME_LINUX`, but I'm thinking `CIBW_CONTAINER_ENGINE` might be better. 
* What are better names for the other parameters I've added? `CIBW_OCI_EXTRA_ARGS_CREATE`, `CIBW_OCI_EXTRA_ARGS_COMMON`, and 
`CIBW_OCI_EXTRA_ARGS_START`, and where do these get documented?
* If `CIBW_OCI_EXE` is podman, should the above options default to the vfs / cgroup settings I've found to work well? Or if the user is using podman, should they be responsible for setting those?
* What should the `DockerContainer` class be renamed to? `Container`? `OCIContainer`? Should there be a separate `DockerContainer` and `PodmanContainer` subclass (I personally think inheritance is not the right design in this case). Likewise, I imagine the `docker_container.py` module should be renamed similarly?
* Should the sleep time be an option while it is an unresolved issue? Does acceptance of this PR hinge on discovering a proper solution to it?
* Which CI providers should I add podman tests to? One or all of them?

Erotemic:
I added podman parameterization to `docker_container_test.py`, which adds a `pytest.mark.parametrize` for the kwargs used to create the `DockerContainer` instance. This first runs all the tests with docker, then with vanilla podman, and then with podman with cgroupfs/vfs container args (which is what I need to run cibuildwheel inside of docker on my gitlab CI). 

Note: I'm going to hold off on making name changes on any existing code (I'll fixup names for new variables introduced) until things are finalized. This will help keep the diff smaller and easier to review. I might make a second PR on top of this that actually does the name changes, as that should be fairly mechanical.

cibuildwheel/docker_container.py
```diff
@@ -88,12 +199,24 @@ def __exit__(
...
+            time.sleep(self._sleep_time)
+
         assert isinstance(self.name, str)
 
-        subprocess.run(["docker", "rm", "--force", "-v", self.name], stdout=subprocess.DEVNULL)
+        subprocess.run(
+            [self.oci_exe, "rm"] + self._common_args + ["--force", "-v", self.name],
+            stdout=subprocess.DEVNULL,
+        )
         self.name = None
```
jshowacre, line 220: @Erotemic The race condition was happening here you said? I tried:
```python
self.bash_stdin.close()

assert isinstance(self.name, str)
subprocess.run(["docker", "rm", "--force", "-v", self.name], stdout=subprocess.DEVNULL)
self.process.terminate()
self.process.wait()
self.name = None
```

[I copied your `cwd` changes and along with this](https://github.com/jshowacre/cibuildwheel/commit/2e51c6e44b7a429048bb42f06bb9839e4bdbf021), the unit tests and sample project seemed to work! Could you try as well?

unit_test/docker_container_test.py
```diff
@@ -21,23 +28,50 @@
...
+    """
+    HAVE_DOCKER = bool(shutil.which("docker"))
+    if HAVE_DOCKER:
+        yield {"oci_exe": "docker", "docker_image": DEFAULT_IMAGE}
+    HAVE_PODMAN = bool(shutil.which("podman"))
+    if HAVE_PODMAN:
+        # Basic podman usage
+        yield {"oci_exe": "podman", "docker_image": DEFAULT_IMAGE}
+        # VFS Podman usage (for the podman in docker use-case)
+        home = str(Path("~").expanduser())
```
jshowacre, line 47:
```suggestion
        home = str(Path.home())
```

unit_test/docker_container_test.py
```diff
@@ -21,23 +28,50 @@
...
+        yield {
+            "oci_exe": "podman",
+            "oci_extra_args_common": f"--cgroup-manager=cgroupfs --storage-driver=vfs --root={home}/.local/share/containers/vfs-storage",
+            "oci_extra_args_create": "--events-backend=file --privileged",
+            "oci_extra_args_start": "--events-backend=file --cgroup-manager=cgroupfs --storage-driver=vfs",
```
jshowacre, line 52: Could these be set in a `containers.conf` in an appropriate location for you? The CLI arguments take the most precedence and I am a little uncomfortable with vfs being the near default.

cibuildwheel/docker_container.py
Erotemic: It still seems to break for me without the sleeps. The error I get is:

```
Error: read unixpacket @->/proc/self/fd/15/attach: read: connection reset by peer

                                                              ✕ 1.09s
Error: Command podman exec --cgroup-manager=cgroupfs --storage-driver=vfs --root=/root/.local/share/containers/vfs-storage/ -i cibuildwheel-19e7f1f5-ddd7-48f5-a130-7dc62c9776d1 tar -cC /output -f /tmp/output-cibuildwheel-19e7f1f5-ddd7-48f5-a130-7dc62c9776d1.tar . failed with code 255. None
```

Note, this is with the vfs and cgroupfs options:

```python
          export CIBW_SKIP="pp* cp310-* cp39-* cp37-* cp36-* *musllinux*"
          export CIBW_OCI_EXE="podman"
          export CIBW_OCI_EXTRA_ARGS_CREATE="--events-backend=file --privileged"
          export CIBW_OCI_EXTRA_ARGS_COMMON="--cgroup-manager=cgroupfs --storage-driver=vfs --root=$HOME/.local/share/containers/vfs-storage/"
          export CIBW_OCI_EXTRA_ARGS_START="--events-backend=file --cgroup-manager=cgroupfs --storage-driver=vfs"
          cibuildwheel --output-dir wheelhouse --platform linux --archs x86_64
```

When I add them back in it goes back to working.

I further verified this by adding an oci-sleep-time option, so within the same test session I set the variable to `export CIBW_OCI_SLEEP_TIME="0.0"` and I again got an error:

```
ERRO[0000] container not running                        
ERRO[0022] Error forwarding signal 15 to container 56aa55d43bfc75cf39d65ca4a86037ddeb129f3fe7d780c6b3de8604b160d9f1: error sending signal to container 56aa55d43bfc75cf39d65ca4a86037ddeb129f3fe7d780c6b3de8604b160d9f1: `/usr/sbin/runc kill 56aa55d43bfc75cf39d65ca4a86037ddeb129f3fe7d780c6b3de8604b160d9f1 15` failed: exit status 1 
```

and then `export CIBW_OCI_SLEEP_TIME="0.1"`, and again it worked.

unit_test/docker_container_test.py
Erotemic: I'm not familiar with `containers.conf`, it looks like it's a container-level thing and not a cibuildwheel level thing?

What is important to me is that cibuildwheel has an easy way to allow the user to modify _how it_ uses containers to build your wheels. If that involves writing a file to my cwd with options instead of specifying then as environs, that's fine. But I don't want to make any system-level modifications. Everything should work locally.

How would one tell cibuildwheel to use a specific container configuration? Is it something you can pass to each Popen invocation of docker/podman commands? If it was something like:

```
export CIBW_CONTAINER_ENGINE=podman
export CIBW_CONTAINER_CONFIG=path-to-containers.conf
```

that would be fine with me, as long as all of the places where I'm using "_common_args" have an equivalent invocation that respects the given conf file.

jshowacre: If you forward `CIBW_CONTAINER_CONFIG` as `CONTAINERS_CONF` for the podman invocation it should do just that, I believe. If not the precedence(so I don't forget) is: `built-in -> /usr/share/containers/ -> /etc/containers/ -> ~/.config/containers/ (rootless) -> invocation`

Right now I'm putting my 'soft' defaults in `/etc/containers` of the container since in rootless podman the user config has higher precedence. If you want to override the storage driver to be vfs, that's done inside the `storage.conf` and the configuration may be overridden as `CONTAINER_STORAGE_CONF`.

Erotemic: That seems like a better way to handle those special podman options. I'll have to learn about the syntax and options... I've spent a bit of time looking at [this](https://github.com/containers/common/blob/main/pkg/config/containers.conf) and this might take me some time. If anyone feels inclined to create a containers.json that corresponds to how I've setup a extra arguments I pass to docker/podman create, start, and other "common" commands, that would help a lot when I next have time to work on this.

joerick:
> * What name do we give what I'm currently calling `CIBW_OCI_EXE`? @joerick proposed `CIBW_CONTAINER_RUNTIME_LINUX`, but I'm thinking `CIBW_CONTAINER_ENGINE` might be better.

The convention we've used in the past is that options that only affect the linux build are suffixed with `_LINUX`, at least in the documentation. But that's only the env var, in the pyproject.toml, the option would just be `container-runtime = "docker"`. I don't have a strong preference between container-runtime and container-engine, I note that podman describes itself as a 'container engine' so maybe that's a better word, happy to use that instead.

> * What are better names for the other parameters I've added? `CIBW_OCI_EXTRA_ARGS_CREATE`, `CIBW_OCI_EXTRA_ARGS_COMMON`, and
>   `CIBW_OCI_EXTRA_ARGS_START`, and where do these get documented?
> * If `CIBW_OCI_EXE` is podman, should the above options default to the vfs / cgroup settings I've found to work well? Or if the user is using podman, should they be responsible for setting those?

Ah, sorry, I wasn't clear. I was proposing removing these options, and using `CIBW_CONTAINER_ENGINE` as the switch for all of them. I had thought that your particular value of these options is _required_ for podman to work in this scenario? I'd prefer to expose the minimum API surface area that we can get away with.

> * What should the `DockerContainer` class be renamed to? `Container`? `OCIContainer`? Should there be a separate `DockerContainer` and `PodmanContainer` subclass (I personally think inheritance is not the right design in this case). Likewise, I imagine the `docker_container.py` module should be renamed similarly?

This is the kind of thing that might be easier to work through in code. But I don't mind a bit of duplication if the code reads clearer.

> * Should the sleep time be an option while it is an unresolved issue? Does acceptance of this PR hinge on discovering a proper solution to it?

It probably doesn't _hinge_ on it, but at least I'd like to understand why it's there a bit better.

> * Which CI providers should I add podman tests to? One or all of them?

Perhaps another job on GitHub Actions would do the trick.

unit_test/docker_container_test.py
joerick: > What is important to me is that cibuildwheel has an easy way to allow the user to modify how it uses containers to build your wheels.

Let's talk about this, because this isn't how cibuildwheel has approached problems like this before. The docker 'backend' has no options exposed to customise (beyond the existing env vars that docker itself reads). Why would users need to customise these options for podman?

henryiii:
Just to note, the problem with #906 would be fixable by users if we exposed more control over the API. Also, haven't looked into this lately, but I don't know if these should be normal options - they are backend only, maybe they would be environment only options. As just a single backend (podman/docker) selector, that would probably be okay to allow in toml, though I'd probably usually recommend it be CI specific unless you really only can build in podman. It's pretty easy to override a single option the other direction, too, so that's not that bad. That's why the "advanced" controls would be really bad in TOML, since you'd have to override all of them. Maybe that's a good argument against the "advanced" controls (assuming they are fixed).

cibuildwheel/docker_container.py
```diff
@@ -88,12 +201,24 @@ def __exit__(
     ) -> None:
 
         self.bash_stdin.close()
+
+        if self._sleep_time:
+            time.sleep(self._sleep_time)
+
         self.process.terminate()
```
jshowacre, line 208:
```suggestion
```

I believe the error is this call itself, since we exhausted the last handle(stdin) and we did not specify detach? Hopefully it is reproducible with your conf.

Erotemic:
>  I had thought that your particular value of these options is required for podman to work in this scenario?

It is not required if you are just running cibuildwheel on your host machine with podman as the only container engine/runtime.  But, it is required for me to use podman in docker on my gitlab-ci instance. I don't quite understand the exact reason, I think it has to due with stricter host machine security settings. My expertise wrt to details of containers is spotty at best.

If I don't include the VFS / cgroup on my gitlab-ci instance I get:

```
time="2021-12-21T16:52:39Z" level=warning msg="Couldn't run auplink before unmount /var/lib/containers/storage/aufs/mnt/f98c1aa0bcc73fedb9642ba1b825491be8847e49da01ffca457958c8684d7730: exit status 22"
Error: unable to start container 91bec90b5d9777c8b4e9871aae3b0fb1314aece559e6e72ca26d747b98571f37: error mounting storage for container 91bec90b5d9777c8b4e9871aae3b0fb1314aece559e6e72ca26d747b98571f37: error creating aufs mount to /var/lib/containers/storage/aufs/mnt/f98c1aa0bcc73fedb9642ba1b825491be8847e49da01ffca457958c8684d7730: invalid argument
```

If I try to run in podman-in-docker on my local machine it does seem to work. I think the gitlab-CI has some sort of security mechanism on the host, which is forcing me to use the [VFS](https://docs.docker.com/storage/storagedriver/vfs-driver/) / [cgroup](https://unix.stackexchange.com/questions/246978/what-is-cgroupfs-and-how-is-it-used-with-docker) options (which for the record, I am still learning about, and I do not have a great understanding of).

> I don't have a strong preference between container-runtime and container-engine, I note that podman describes itself as a 'container engine' so maybe that's a better word, happy to use that instead.

Either `CIBW_CONTAINER_RUNTIME` or `CIBW_CONTAINER_ENGINE` makes sense to me. 

> Ah, sorry, I wasn't clear. I was proposing removing these options,

Learning more about container configuration, I agree with removing those options.
Based on [containers.conf.5.md](https://github.com/containers/common/blob/main/docs/containers.conf.5.md) I think it might make sense to have a `CIBW_CONTAINERS_CONF` and `CIBW_CONTAINERS_STORAGE_CONF`. Or I suppose we could rely on implicit behavior and just use the standard `CONTAINERS_CONF` and `CONTAINERS_STORAGE_CONF` environment variables, but I have a preference for having some way to pass that information around explicitly and coupled with CIBW itself. I can imagine this being very helpful to a newcomer who (like my past self) encounters problems with using cibuildwheel and does not realize that containers are configurable, or that the container configuration itself is the problem.

> Just to note, the problem with #906 would be fixable by users if we exposed more control over the API.

If I understand correctly, setting `--network=host` should be possible in a configuration file. The question is, what is the best way to point cibuildwheel at it?

cibuildwheel/docker_container.py
Erotemic: That does seem to work on my local tests. I'm going to remove the sleep and give it a try on the gitlab instance.

unit_test/docker_container_test.py
Erotemic: As mentioned in another reply, running podman inside docker, which is running on a host with stricter permission settings (I'll have to ask to figure out what these are, as I don't manage the gitlab-ci servers I use), need to run podman in "rootless" mode and I think there is some other issue with the overlay file driver, which is why the vfs driver is needed. But again, my understanding of containers is limited, 

If I can determine how to get the CI working with only using existing environment variables that docker reads themselves, I'm perfectly happy to use those instead of relying on cibuildwheel (probably a better design decision anyway, although it would be nice if cibuildwheel had a note about how to configuring containers and then point at the docs to CONTAINERS_CONF and CONTAINERS_STORAGE_CONF). What I'm currently trying to figure out is what those config settings are. They unfortunately, do not correspond exactly to command line options, which is a design decision I always find to be frustrating.

Erotemic:
I've verified that I can use config files on my gitlab instance to get the behavior I'm looking for. That is good because it reduces the scope of this PR.

However, I'd still like to add a test which ensures cibuildwheel generally works with setting needed on more secured hosts. This involved expanding the docker_container_test.py such that it can write these config files and the pass environment variables to `DockerContainer`.

To obtain something like the effect of podman create with `--privileged`, I used the config files to specify extra default_capabilities and I changed the aufs mount options to be "rw". Still not 100% sure I have the config files in a 1-to-1 correspondence with the CLI params I passed earlier, but it does seem to work for me.

The main issue I'm running into here is that the temporary directory doesn't get cleaned up neatly on exit. The files written to the storage dir don't have user write permissions for whatever reason, and there are symlinks to /lib64 that seem to be causing issues. Hopefully these can be resolved.

mayeut:
@Erotemic, does terminating the container with something like that still exhibits a race condition ?
```python
self.bash_stdin.write(b"exit 0\n")
self.bash_stdin.flush()
self.process.wait()
self.process.stdin.close()
self.process.stdout.close()
```

Erotemic:
I don't remember if I tried the workaround from @mayeut or not. I probably didn't. Looks like this at least needs a rebase.

Honestly, I'm a bit fuzzy on all of the details of podman. It's been awhile since I've used it actively. My motivation for working on this was to get my CI working and I've been using my fork (which is effectively this branch) successfully for over a year now. I'm forgetting the details on the race condition, and I can try the workaround, but even if that doesn't work I think this is probably good enough to merge given a rebase and re-review. I imagine this feature would be helpful to others and it would generalize cibuildwheel in a good direction (not locking into docker).

I may not have time to get to the rebase in the near future, so if anyone does need this urgently, I wouldn't mind some help getting the last issues fixed.

Erotemic:
I've gotten into another situation where I need podman support in cibuildwheel. I've rebased this branch on the lastest main. I'll work on fixing any issues that crop up.

I'll see if I can lookup and describe the race condition. It may just be something we have to live with as a known issue. It can be exclusively locked into podman execution though so docker users will be unaffected.

joerick:
Another look into the race condition would be appreciated. And a general tidy up of this would be good too, if you're still looking to get it merged. This PR has been on a bit of a journey, I see some debug code still in there, and the tests are quite confusing to me. It would probably benefit from asking 'what are the minimum changes that get this working'. this is probably quite a niche feature, so it's more likely to be worth it to merge if it's not a lot of code. 🙏 thanks!

Erotemic:
@joerick I've tidied up and removed the debug code. 

The new test code is less daunting than it appears. Let me explain.

### Changes in tests

In the tests, there is only one major change: adding the `basis_container_kwargs`, which we can rename or add extra documentation to if needed. The idea is it just generates different keyword arguments that can be used to construct an instance of the existing `DockerContainer` object. It yields a dictionary that correspond to a container engine and its configuration. I've added some more comments and removed commented debugging statements. The first two dictionaries yielded are the basic podman and docker configurations. I also added a VFS podman configuration, which is my primary use case. 

The other added function is `_cleanup_tempdir`, which removes the test configurations. I put a comment in there indicating that I thought there might be a "better way" to do this with pytest itself, but I couldn't figure it out so I fell back on using atexit, which works fine.

The only other test changes are parameterizing the existing tests so they run with both podman and docker.

To support the tests, CI rules to install podman were added in `test.yml` and `install_latest_podman.sh`. 

### Changes in module code

Most other changes generally stem from adding the new `container_engine` option, which just holds the name of the executable using to run the containers: docker or podman. The default is "docker".

The majority of these changes are in the `DockerContainer` class. I've also noted the name `DockerContainer` is no longer appropriate, but I didn't want to introduce that change in this PR. I think it makes more sense to merge in the functionality, ensure it doesn't need to be reverted, and then gradually move terminology over to the generalized docker-agnostic setting. 

As previously mentioned, there are a few incompatibilities between docker and podman, which is where all other changes stem from. 

* I've added the `env` parameter which is used in the tests.

* Podman doesn't let you specify a `cwd` in the container create command, so we just add a command after the create to make the directory. We also add the ":Z" flag to the volume to support SELinux. I don't think this has any impact on exiting invocations (or at least the tests don't flag it).

* The copy out procedure works differently in podman. I think docker might have some extra magic going on to make the concise command work, whereas podman needs the steps spelled out more explicitly. Suffice to say, the logic for docker is unchanged, so only  podman is using the new codepath. I added some more comments explaining this different in the code.

* I added a debug_info function, which could be removed, but I think it's helpful and worth keeping in. 

* The extra `self.process.wait`. This is the elephant in the PR. I have no idea why this is necessary, and it would be a lot of investigative work to drill down into it. Frankly I don't have the time to dedicate to this problem. It seems to avoid a race condition. Perhaps that is a problem with what we are doing, perhaps its a problem with podman itself. I think a good compromise is to leave it in, document why it is there, and only execute it in the podman case. My hope is that this issue doesn't derail the PR.

unit_test/docker_container_test.py
```diff
@@ -21,23 +25,178 @@
...
+    (virtual file system) enabled as the storage driver.
+    """
+
+    global temp_test_dir
+    if temp_test_dir is None:
+        # Only setup the temp directory once for all tests
+        temp_test_dir = tempfile.TemporaryDirectory(prefix="cibw_test_")
+
+    HAVE_DOCKER = bool(shutil.which("docker"))
+    HAVE_PODMAN = bool(shutil.which("podman"))
```
joerick, line 97: These tests rely on the runtime detection of something, meaning that they might stop being run one day if podman/docker isn't found, and we wouldn't know, the tests would still be green.

I'd propose moving to a test command line flag, like the existing `--run-docker` command, so adding a `--run-podman` flag that would enable the running of these tests with podman. I set up the existing `--run-docker` flag, I'd be happy to help with this if you like?

unit_test/docker_container_test.py
```diff
@@ -21,23 +25,178 @@
...
     DEFAULT_IMAGE = "quay.io/pypa/manylinux2014_s390x:2020-05-17-2f8ac3b"
+else:
+    DEFAULT_IMAGE = ""
+
+
+temp_test_dir = None
+
+
+@atexit.register
+def _cleanup_tempdir():
```
joerick, line 36: I wonder if this could be resolved more cleanly by using the pytest [tmp_path fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html). Did you try using that?

unit_test/docker_container_test.py
```diff
@@ -21,23 +25,178 @@
...
+                "CONTAINERS_CONF": str(vfs_containers_conf_fpath),
+                "CONTAINERS_STORAGE_CONF": str(vfs_containers_storage_conf_fpath),
+            }
+        )
+
+        yield {
+            "container_engine": "podman",
+            "docker_image": DEFAULT_IMAGE,
+            "env": oci_environ,
+        }
```
joerick, line 170: I don't understand all the complexities here! But do we need to run all the tests with this configuration? Or would just one be fine?

unit_test/docker_container_test.py
```diff
@@ -1,11 +1,15 @@
...
 import platform
 import random
 import shutil
 import subprocess
+import tempfile
 import textwrap
 from pathlib import Path, PurePath, PurePosixPath
 
 import pytest
+import toml  # type: ignore[import]
```
joerick, line 12: You might not need the `ignore[import]`, if you install all the dev-dependencies on your development machine - we have `types-toml` listed there.

cibuildwheel/docker_container.py
```diff
@@ -155,12 +200,47 @@ def copy_out(self, from_path: PurePath, to_path: Path) -> None:
...
+            )
+            command = f"tar -xvf output-{self.name}.tar"
+            subprocess.run(
+                command,
+                shell=True,
+                check=True,
+                cwd=to_path,
+                env=self.env,
+            )
+            os.unlink(to_path / f"output-{self.name}.tar")
```
joerick, line 232: I wonder if you tried a simple `podman cp` without the tar step?  In the Docker implementation, the tar step is to work around [a bug in Docker](https://github.com/moby/moby/issues/38995), but perhaps that bug doesn't exist in podman?

cibuildwheel/docker_container.py
```diff
@@ -57,39 +80,49 @@ def __enter__(self) -> "DockerContainer":
...
 
         # run a noop command to block until the container is responding
-        self.call(["/bin/true"])
+        self.call(["/bin/true"], cwd="")
+
+        if self.cwd:
+            # Although `docker create -w` does create the working dir if it
+            # does not exist, podman does not. There does not seem to be a way
+            # to setup a workdir for a container running in podman.
+            self.call(["mkdir", "-p", str(self.cwd)], cwd="")
```
joerick, line 125: Thinking about alternative approaches here... is the problem that the directory doesn't exist or that podman doesn't support `-w`? Perhaps we could do a `podman exec mkdir -p {self.cwd}` before attaching to the container?

cibuildwheel/docker_container.py
```diff
@@ -57,39 +80,49 @@ def __enter__(self) -> "DockerContainer":
...
+            env=self.env,
         )
 
         assert self.process.stdin and self.process.stdout
         self.bash_stdin = self.process.stdin
         self.bash_stdout = self.process.stdout
 
         # run a noop command to block until the container is responding
-        self.call(["/bin/true"])
+        self.call(["/bin/true"], cwd="")
```
joerick, line 119: The use of empty string as semantically different from None is a little too tricky. I think there's probably a better way to do this, that handles the working dir stuff at the start, rather than setting for each command.

joerick (commented):
Thanks for the detailed response @Erotemic.

I think we can probably get this in, with a little work. I've added my comments inline.

Actually, I don't have a huge problem with the `process.wait()` call. I seem to recall that it used to be a `time.sleep()`, which I was objecting to. But `process.wait()` does make a lot of sense in this context - the podman process has to exit before you can rm the container.

cibuildwheel/docker_container.py
```diff
@@ -57,39 +80,49 @@ def __enter__(self) -> "DockerContainer":
...
+                self.container_engine,
                 "create",
                 "--env=CIBUILDWHEEL",
                 f"--name={self.name}",
                 "--interactive",
-                "--volume=/:/host",  # ignored on CircleCI
                 *network_args,
-                *cwd_args,
+                # Z-flags is for SELinux
+                "--volume=/:/host:Z",  # ignored on CircleCI
```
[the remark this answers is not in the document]
Erotemic: Hmm, this is a good catch. I'll need to run more tests and look further into this. Given these docs, blindly using :Z seems like a bad idea. I'm not sure if this will need an additional option.

cibuildwheel/docker_container.py
```diff
@@ -34,7 +49,13 @@ class DockerContainer:
...
 
     def __init__(
-        self, *, docker_image: str, simulate_32_bit: bool = False, cwd: Optional[PathOrStr] = None
+        self,
+        *,
+        docker_image: str,
+        simulate_32_bit: bool = False,
+        cwd: Optional[PathOrStr] = None,
+        container_engine: str = "docker",
+        env: Optional[Dict[str, str]] = None,
```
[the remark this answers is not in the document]
Erotemic: I think it is a logical extension of the object that may prove useful in the future. I feel like an object that is easy to test without framework specific test-time dependencies has advantages. I would be in inclined to leave this as-is, I try to avoid using pytest features when possible, but I'm not going to die on that hill.

unit_test/docker_container_test.py
Erotemic: This is a good point; an oversight on my part. I'd be glad to have the help setting up `--run-podman`.

unit_test/docker_container_test.py
Erotemic: I'll have to run tests. I the need to chown was an issue, but I don't remember the details.

unit_test/docker_container_test.py
Erotemic: I recall encountering instances where the logic broke one but not the other. Is there some mechanism to mark a test as "extra" or "slow?" It might be reasonable to move that into a group that isn't tested regularly on the CI, if runtime is the concern. If I find an instance where it is necessary I'll describe it in comments, otherwise I'm not opposed to removing it as long as my use case isn't breaking (which is what it's testing). I'll send in an issue if it does :)

unit_test/docker_container_test.py
Erotemic: I'll try removing it, but I think the CI will error. I added it in because mypy told me to.

cibuildwheel/docker_container.py
Erotemic: It would make sense to at least document the alternative "simpler" route in case the bug is fixed in docker. I'll try it.

cibuildwheel/docker_container.py
Erotemic: I think it is the second one. It does look like it has `-w`, but I think the create step itself fails if the "image" doesn't have the directory already in it, which we can't do anything about (AFAIK). I can try it if you think it is cleaner, but the difference between the methods seems like a wash to me.

cibuildwheel/docker_container.py
Erotemic: I'm fine with it as long as it works.

cibuildwheel/docker_container.py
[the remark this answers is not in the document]
joerick: My perspective is more along the lines of YAGNI (you ain't gonna need it), and we use setenv quite a lot already. I prefer setenv because it's closer to how the code would perform if the user set the environment before calling cibuildwheel.

cibuildwheel/docker_container.py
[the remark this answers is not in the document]
joerick: One option could be to disable the host mount on SELinux? It's already conditional on where the docker is running.

unit_test/docker_container_test.py
joerick: Sure, I'll take a look.

cibuildwheel/docker_container.py
joerick: Ah, interesting, so it's the create step that fails. I guess that makes sense given that's where the `-w` is applied. So the `exec` approach wouldn't work then.

cibuildwheel/docker_container.py
joerick: Maybe we can just do `cwd="/"` for these first couple commands instead. That would be preferable to me.

cibuildwheel/docker_container.py
Erotemic: To close the loop here, I have a MWE. It is not the create step that fails, but it seems that the start step fails. I'm not sure if you can apply an exec before the container is started, it gave me an error when I tried. However, it worked fine under docker.

```bash
# Create a podman container with "-w"
podman create --name=foobar-podman-test --interactive -w "/createme-dir" python:3.7 /bin/bash

# Running the exec here will fail because the container is not started
podman exec -i foobar-podman-test mkdir -p "/createme-dir"

# But starting the container raises an error when create was used with "-w"
podman start foobar-podman-test

# Cleanup
podman rm --force -v foobar-podman-test
```

Here is similar logic working with docker: 

```bash
# Demo where this sequence works with docker
docker create --name=foobar-docker-test --interactive -w "/createme-dir"  python:3.7 /bin/bash
docker start foobar-docker-test
docker exec -i foobar-docker-test mkdir -p "/createme-dir"
docker rm --force -v foobar-docker-test
```

I've documented this MWE in my misc repo: https://github.com/Erotemic/misc/blob/main/tests/containers/test_container_create_with_cwd.sh so I don't lose it.

cibuildwheel/docker_container.py
Erotemic: I've added a stub for an alternate method: 

```python
        TRY_SIMPLE_CP = (self.container_engine == "podman")
        if TRY_SIMPLE_CP:
            # There is a bug in docker that prevents this simple implementation
            # from working https://github.com/moby/moby/issues/38995
            command = f"{self.container_engine} cp {self.name}:{shell_quote(from_path)} {shell_quote(to_path)}"
            subprocess.run(
                command,
                shell=True,
                check=True,
                cwd=to_path,
                env=self.env,
            )
```

But this causes the `test_dir_operations` test to fail. 

```
>           assert test_binary_data == (new_test_dir / "test.dat").read_bytes()

container  = <cibuildwheel.docker_container.DockerContainer object at 0x7f822933a640>
container_kwargs = {'container_engine': 'podman', 'docker_image': 'quay.io/pypa/manylinux2014_x86_64:2020-05-17-2f8ac3b', 'env': {'BASH_F...'/home/joncrall/.local/cuda::/home/joncrall/.local', 'CODE_DIR': '/home/joncrall/code', 'COLORTERM': 'truecolor', ...}}
dst_dir    = PurePosixPath('/tmp/test_dir')
dst_file   = PurePosixPath('/tmp/test_dir/test.dat')
new_test_dir = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new')
original_test_file = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test.dat')
output     = '8\udcea\udc91\x1f\udcfd\udcf9X\udcb8H\udcb4\udce3\x06\udca7?\udc82\udcfd\x16\udcbd\x068\udcb6\x02\udca5\udcea\x0b{\x0...dc88ya\udcc9\x04\udc9a\udcb6\udc8d@ot\x08s\udc91\udcf7\udc88\udcb4,\udc82\x03\x17*6\udcba\x12[4w=\udcc4Bn)\udc9f\udce0'
test_binary_data = b'8\xea\x91\x1f\xfd\xf9X\xb8H\xb4\xe3\x06\xa7?\x82\xfd\x16\xbd\x068\xb6\x02\xa5\xea\x0b{\x03G\xda\xb0\xea\xa0\xdf\x93\...\x91h\xd2,\xbd\x1bd\xb3\r"\x88ya\xc9\x04\x9a\xb6\x8d@ot\x08s\x91\xf7\x88\xb4,\x82\x03\x17*6\xba\x12[4w=\xc4Bn)\x9f\xe0'
test_dir   = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir')
test_file  = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir/test.dat')
tmp_path   = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1')

unit_test/docker_container_test.py:353: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../.pyenv/versions/3.9.9/lib/python3.9/pathlib.py:1259: in read_bytes
    with self.open(mode='rb') as f:
        self       = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat')
../../.pyenv/versions/3.9.9/lib/python3.9/pathlib.py:1252: in open
    return io.open(self, mode, buffering, encoding, errors, newline,
        buffering  = -1
        encoding   = None
        errors     = None
        mode       = 'rb'
        newline    = None
        self       = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat')
name = '/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat', flags = 524288, mode = 438

    def _opener(self, name, flags, mode=0o666):
        # A stub for the opener argument to built-in open()
>       return self._accessor.open(self, flags, mode)
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat'

flags      = 524288
mode       = 438
name       = '/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat'
self       = PosixPath('/tmp/pytest-of-joncrall/pytest-18/test_dir_operations_container_1/test_dir_new/test.dat')

../../.pyenv/versions/3.9.9/lib/python3.9/pathlib.py:1120: FileNotFoundError
============================================================================= short test summary info ==============================================================================
FAILED unit_test/docker_container_test.py::test_dir_operations[container_kwargs0] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-joncrall/pytest-18/te...
FAILED unit_test/docker_container_test.py::test_dir_operations[container_kwargs1] - FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-joncrall/pytest-18/te...
```

unit_test/docker_container_test.py
Erotemic: I think it's unlikely that the temp path fixture will work here. But I've been able to simplify the logic of the cleanup hook. 

The issue is that when you create a podman filesytem, it will create it with a fake UID of 0, which makes the system think you need to be root to remove it. This causes permission errors. I found a blog that documents the case: https://podman.io/blogs/2018/10/03/podman-remove-content-homedir.html

I was able to rewrite the logic using the "podman unshare" executable: 

```python
    global temp_test_dir
    if temp_test_dir is not None:
        print(f"CLEANUP temp_test_dir = {temp_test_dir!r}")  # type: ignore[unreachable]
        # When podman creates special directories, they cant be cleaned up
        # unless you fake a UID of 0. The package rootlesskit helps with that.
        import subprocess
        subprocess.call(['podman', 'unshare', 'rm', '-rf', temp_test_dir.name])
        try:
            temp_test_dir.cleanup()
        except Exception as ex:
            print(f"Issue cleaning up ex = {ex!r}")
    temp_test_dir = None
```

This deletes the temp directory without any permission issues. 

Is there a way to nicely turn this into a pytest hook? We also should only need to do this if the podman tests run (implying the podman executable is available), otherwise the existing tempdir cleanup mechanism should work.

unit_test/docker_container_test.py
Erotemic: I see the logic in conftest.py to add the "--run-docker" flag. If there is a corresponding podman mark, is there a way to apply specific marks to different instances of the `pytest.mark.parametrize`? The current way I have it setup is that I parameterize the same test over different configuration of the OCI backend. I'm not sure how to get some of them to trigger and others not to.

cibuildwheel/docker_container.py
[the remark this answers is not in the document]
Erotemic: I've been testing modifications I'm making here on the project which is my use case for working on this PR: https://gitlab.kitware.com/computer-vision/kwimage_ext/-/merge_requests/4

It seems that not having the `:Z` is no longer breaking me (or maybe it never was to begin with...), either way I just reverted it to the state without this patch.

cibuildwheel/docker_container.py
Erotemic: I just removed the cwd argument, which reverts this to its original state.

cibuildwheel/docker_container.py
Erotemic: I have this in as a note in the code itself. I can leave it in as a node, (it is effectively commented out with an `if 0`) or just remove it. It seems like a simple podman cp won't work. Is there any major objection to the way I originally wrote it?

unit_test/docker_container_test.py
Erotemic: I was able to get around this by adding a line to `.pre-commit-config.yaml` indicating that ` - types-toml` is a mypy dependency.

unit_test/docker_container_test.py
Erotemic: I moved the logic for setting up the environs and paths for the VFS test into it's own function. This should make it easier to read what configurations `basis_container_kwargs` generates.

unit_test/docker_container_test.py
Erotemic: I modified the logic here. I've added another global called "using_podman", such that this only runs if the podman tests are being executed. I also had to move the atexit registration to be after the TemporaryDirectory was created, otherwise it would try to clean up the tempdir first.

I might be able to rewrite this as a pytest fixture with a module level scope. But I think it might depend on how `--run-podman` is implemented here, so I'm going to wait until those decisions are made to try messing with this. But I feel like I have a much better understanding of the issue I was working around now, and using `podman unshare` seems much cleaner than what I was doing.

cibuildwheel/docker_container.py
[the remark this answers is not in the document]
Erotemic: I removed the `.env` attribute and modified the tests to use `monkeypatch.setenv`. Personally, I think the previous method was a lot cleaner, and the monkeypatch fixture adds unnecessary obfuscation (there is no in-code indication of where it comes from!), but I understand the value of limiting the complexity of the main class, so I'll compromise.

cibuildwheel/docker_container.py
joerick: Got a local podman setup working! I managed to get the simple cp to work, there's just a trick in how the src/dest paths are supplied on the command line (a bit like rsync). I've pushed that change.

cibuildwheel/docker_container.py
joerick: Oh cool, that works too :)

cibuildwheel/docker_container.py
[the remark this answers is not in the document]
joerick: thanks! I've refactored the monkeypatch.setenv calls a bit, I think they're a bit neater now.

unit_test/docker_container_test.py
joerick: This is done, actually it ended up quite a lot neater using pytest 'parametrized fixtures'

unit_test/docker_container_test.py
Erotemic: This looks much cleaner! The renames look very nice. I guess, my initial stab at using "cp" in `copy_out` had a mistake? Your refactor seems to use it, and the tests aren't breaking, so that's good.

The only thing I worry about is that the VFS configuration won't run through all of the container operations, but I've tested this branch on my use case and it doesn't break, so that's good enough for me. In general I'm happy with the state of the PR. I think it's ready to go

joerick:
Thank you for your responses to feedback on this @Erotemic!

I've done a pass at this PR, I think it should be nearly ready to go. 

Apologies for the diff explosion. I've ended up changing lots of references of 'docker' to 'container', which is language that appears throughout the project. It's probably a good thing anyway, since there were lots of occurrences of things like "copy files into the docker", which doesn't really make sense :P

The main files to look at are "oci_container.py", "oci_container_test.py", and "test_podman.py".

The only thing missing that I'm aware of currently is documentation for this new "CIBW_CONTAINER_ENGINE" option.

cibuildwheel/docker_container.py
joerick: I just noticed that leaving the cwd argument out means that there's output like this in the terminal:

```text
    + /bin/true
/bin/bash: line 2: cd: /project: No such file or directory
    + mkdir -p /project
/bin/bash: line 8: cd: /project: No such file or directory
```

Which are inconsequential errors, but the user wouldn't know that.

So maybe `cwd="/"` would be better after all.

.github/workflows/test.yml
```diff
@@ -45,6 +45,13 @@ jobs:
       with:
         python-version: ${{ matrix.python_version }}
 
+    # Install podman on this CI instance for podman tests on linux
+    # Snippet from: https://github.com/redhat-actions/podman-login/blob/main/.github/workflows/example.yml
+    - name: Install latest podman
+      if: runner.os == 'Linux'
+      run: |
+        bash .github/install_latest_podman.sh
```
henryiii, line 53: Personally I prefer placing CI specific code directly in CI, rather than in a helper file, especially things with sudo's in them. Especially one sitting directly in the `.github` directory, which is mostly for GitHub's use.

cibuildwheel/options.py
```diff
@@ -392,12 +395,22 @@ def globals(self) -> GlobalOptions:
         archs_config_str = args.archs or self.reader.get("archs", sep=" ")
         architectures = Architecture.parse_config(archs_config_str, platform=self.platform)
 
+        container_engine_str = self.reader.get("container-engine")
```
henryiii, line 398: This is technically general or linux only, but we don't limit manylinux either, so that's fine. If we ever to limit win/macOS to not allow manylinux, we could also enforce that for container-engine too.

setup.py
```diff
@@ -14,6 +14,7 @@
         "pytest-timeout",
         "pytest-xdist",
         "build",
+        "toml",
```
henryiii, line 17:
```suggestion
```

Why is a dead package listed here? `toml` was replaced by `tomli` (which is replaced by built-in `tomllib` in Python 3.11)?

test/conftest.py
```diff
@@ -7,20 +7,7 @@ def pytest_addoption(parser) -> None:
...
-
-
-def pytest_collection_modifyitems(config, items) -> None:
-    if config.getoption("--run-emulation"):
-        # --run-emulation given in cli: do not skip emulation tests
-        return
-    skip_emulation = pytest.mark.skip(reason="need --run-emulation option to run")
-    for item in items:
-        if "emulation" in item.keywords:
-            item.add_marker(skip_emulation)
```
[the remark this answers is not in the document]
henryiii: I'm not sure it's better - the markers would allow a user to statically see all the tests marked with emulation, and read in the normal pytest style, where a test is decorated with it's properties. This is now just procedural manual hacks in each test rather than declarative markers, and you have to look inside for skips, and pytest doesn't know beforehand it's skippable.

The marker should have been in pyproject.toml and not in contest.py, but handing markers in this way was fine in pytest, IMO. There's actually at least one plugin for "opt-in" markers (https://github.com/reece/pytest-optional-tests)  that IMO would have been even better - it's not the best maintained, though; I've manually implemented it with fixes inside plumbum at https://github.com/tomerfiliba/plumbum/blob/master/tests/conftest.py, for example. I'm not sure if @reece would be interested in having some help maintaining the plugin.

.github/workflows/test.yml
Erotemic: This was because I followed https://github.com/redhat-actions/podman-login as an example for how to install podman in the CI. We can replace that with:

```
. /etc/os-release
echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
curl -sSfL "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key" | sudo apt-key add -
sudo apt-get update
sudo apt-get -y upgrade
sudo apt-get -y install podman
```

and remove the install file if you'd like.

setup.py
henryiii: Please use the same structure we use elsewhere, which is tomli if <3.11 and tomllib (identical) otherwise.

Erotemic: err, I guess we have to use `tomli_w` here instead?

cibuildwheel/options.py
Erotemic: There are podman builds for windows, so I think container-engine should be considered a general setting. This could extend to other engines like lima, lxd, or something else that doesn't exist yet.

cibuildwheel/docker_container.py
Erotemic: Seems reasonable. I pushed up that change.

cibuildwheel/options.py
henryiii: We don't support using a container engine for a Windows target, and I don't really see what that would be useful (you could just run cibuildwheel _in_ the Windows container, after all), and it would require major changes. But this doesn't need any changes and we might not add that level of checking (it's not an easy mistake to make).

henryiii (approved):

PR #977, opened 2021-12-29
open setup.py as utf-8 in get_requires_python_str

### LINKED ISSUE #976: on windows, setup_py_python_requires attempts to open utf-8 setup.py as Windows-1252 and fails
### Description

This [setup.py file](https://github.com/fgregg/fastcluster/blob/master/setup.py) is valid utf-8, and has a few non-ascii characters. In a windows build, `setup_py_python_requires` appears to be opening this file as if it was encoded like Windows-1252 and thus fails on some non-ascii characters.

### Build log

https://github.com/fgregg/fastcluster/runs/4660766954?check_suite_focus=true#step:5:40

### CI config

https://github.com/fgregg/fastcluster/blob/master/.github/workflows/pythonpackage.yml#L41-L47

fgregg:
will close #976

without this, windows can attempt to open this file as Windows-1252

### DIFF
--- a/cibuildwheel/projectfiles.py
+++ b/cibuildwheel/projectfiles.py
@@ -70,7 +70,7 @@ def get_requires_python_str(package_dir: Path) -> Optional[str]:
         pass
 
     try:
-        with (package_dir / "setup.py").open(encoding='utf8') as f2:
+        with (package_dir / "setup.py").open(encoding="utf8") as f2:
             return setup_py_python_requires(f2.read())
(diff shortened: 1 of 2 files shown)


henryiii (approved):

mayeut (approved):
I missed the character reproducing the issue in the first test I added.
The test now fails without the fix on Windows.

mayeut:
Thanks for  the contribution @fgregg

PR #978, opened 2021-12-30
refactor: Use utils.call & utils.shell in macos.py & windows.py

mayeut:
Closes #672, uses an implementation like the one in https://github.com/pypa/cibuildwheel/pull/672#issuecomment-878323359
Extracted from #974

It might be easier to review this one on its own. I'll rebase #974 on this one.

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -3,6 +3,7 @@ import fnmatch
 import itertools
 import os
 import re
+import shlex
 import ssl
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


cibuildwheel/util.py
```diff
@@ -47,6 +48,60 @@
...
+    text: Literal[True],
+) -> str:
+    ...
+
+
+def call(
+    *args: PathOrStr,
+    env: Optional[Dict[str, str]] = None,
+    cwd: Optional[PathOrStr] = None,
+    text: bool = False,
```
joerick, line 78: How about renaming this `text` parameter to `capture_stdout`?

cibuildwheel/util.py
```diff
@@ -47,6 +48,60 @@
...
+) -> Optional[str]:
+    """
+    Run subprocess.run, but print the commands first. Takes the commands as
+    *args. Should use shell=True on Windows due to a bug. Also converts to
```
joerick, line 82:
```suggestion
    *args. Uses shell=True on Windows due to a bug. Also converts to
```

joerick (commented):
This seems like a good approach to me.

joerick (approved):

PR #979, opened 2022-01-02
fix: remove warnings on macOS when installing certificates

mayeut:
It seems `sudo` is not required anymore (at least on GHA), let's see for other providers.


joerick (approved):

joerick:
Always nice to remove a bit of log noise :)

Issue #882, opened 2021-10-18
Stalling emulated CI runs.


joerick:
### Description

I keep seeing the emulated tests stalling on our CI. This is an example of the output:

```
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/runner/work/cibuildwheel/cibuildwheel, configfile: pyproject.toml
plugins: forked-1.3.0, xdist-2.4.0, timeout-2.0.1
collected 2 items

Error: The operation was canceled.
```

They are cancelled after the job times out - 3 hours. Restarting normally fixes it.

It seems that our process is not running, as we have pytest-timeout that should cancel a test that's taking too long.

Is this only an issue in our CI? or is it an issue with GHA, or the emulation step?

### Build log

https://github.com/pypa/cibuildwheel/pull/854/checks?check_run_id=3920499554

Czaki:
> I don't think anything that the test process is doing would be installing a signal handler, it's just waiting on a subprocess.run call AFAIK.

I do not know what happens here. When I play with Qt and calculation stuck in C code then SIGINT was ignored. It has handled first-time calculation returns from C to python.

henryiii:
Not that it fixes anything, but https://twitter.com/hugovk/status/1450088663304581120?s=20 -> https://github.com/github/roadmap/issues/271 will be nice. :)

joerick:
This seems to have stopped happening :)

PR #983, opened 2022-01-06
Debug Appveyor failures

joerick:


mayeut:
I think it might be related to https://github.com/pypa/virtualenv/issues/2203 / https://github.com/pypa/virtualenv/pull/2256 as I can locally reproduce the issue / fix it / reproduce it by messing with virtualenv cache.
It been solved in virtualenv 20.11.2, bump the test to this one to see if it works.

mayeut:
Seems good: https://ci.appveyor.com/project/joerick/cibuildwheel/builds/42113260/job/1nhf54nmlvnuon88

henryiii:
One of them timed out, otherwise fine? (wasn't really following what was wrong, though)

joerick:
Awesome, thank you @mayeut ! That would have taken me a lot longer to figure out! Yes @henryiii I think the timeout is unrelated.

joerick:
The linux timeout is also problematic, I'll see if I can request an extension with Appveyor.

joerick:
Appveyor increased the build time, so we should be good to go here!

mayeut:
cleaned up history & rebased on main.

mayeut:
there's the same issue with pip on travis-ci windows, will fix that also

mayeut:
only the emulation test remains and this one is a bit flaky as mentioned in #882 so merging this.

PR #971, opened 2021-12-22
Remote container build

jshowacre:

Issue #676, opened 2021-05-14
Allow running cibuildwheel inside of a previously configured docker image.


Erotemic:
On my internal gitlab-ci runners docker-inside-docker is disabled. This seems to be causing cibuildwheel to fail.

As a workaround, I would like to be able to use cibuildwheel inside of a base manylinux image. That is, I want to give it the information that it is already inside of a manylinux image like `quay.io/pypa/manylinux2010_x86_64`, and then I want it to do its thing.

I do something similar in this script, if you are inside of docker image (you have to give it this information), it executes the script, but if you tell it you want to run in docker, then it executes itself inside inside of the docker image. 
https://gitlab.kitware.com/computer-vision/kwimage/-/blob/master/run_manylinux_build.sh

And my gitlab yaml explicitly has to call out when I use the base quay.io/pypa/manylinux2010_x86_64 image: 
https://gitlab.kitware.com/computer-vision/kwimage/-/blob/master/.gitlab-ci.yml

I was poking around in cibuildwheel.linux, and I see that it loops over several configurations and then executes a block of code in a DockerContainer context manager. If I was to write a PR that refactored that inner part into a function the CLI could invoke directly (where the user likely has to provide some information that the looped settings are currently providing), would that be of interest to the maintainers?

(A lot of the inner loop actually looks like a more sophisticated version of what I'm doing in that shell script, I think it would be nice to have that as a callable function)

henryiii:
Do you really need to refactor it in this way? It seems like we could just write a new context manager that simply wraps subprocess.run and otherwise does pretty much nothing, and then using that instead would perform the work in process instead of in a docker container. Then you could have some escape hatch that uses that instead if cibuildwheel was running inside the manylinux image - say maybe even with pipx! :) (See https://github.com/pypa/manylinux/pull/1055#issuecomment-841253348).

I actually would really like to have a `run_all` and `run` function inside each linux/macOS/windows module, see #560. But there are challenges to doing it properly, I think, and the interface might not be that "nice" for outside users.  Longer term, having an API usage might be of interest?

Erotemic:
I started hacking on this on https://github.com/Erotemic/cibuildwheel/tree/dev/flow 

I'm not sure what the best way to refactor my stuff to use cibuildwheel is yet. I'm not tied to any one solution.

[...]

Issue #78, opened 2018-06-30
Add support for Py_LIMITED_API ?


jlaine:
This is a more a "should we do this" question rather than an actual issue.

Background : Python 3.2 and up offer a subset of the C API which is guaranteed to be stable at the ABI level. For packages that only make use of this limited API, it is possible to build wheels which can be used across multiple Python versions.

https://docs.python.org/3/c-api/stable.html

For instance `cryptography` switched to Py_LIMITED_API a few months ago:

https://github.com/pyca/cryptography/issues/3307

Adding support for this would for instance have avoid the need to rebuild wheels for Python 3.7.

YannickJadoul:
Hmmm, I hadn't come across this before. So how do you reckon would this then be supported by cibuildwheel? What's the difference to the build process?

And, trying to assess the idea further: how relevant is `cibuildwheel` to those users, if they don't have to run the build for all different Python versions anyway?

jlaine:
I think the only change to the build process is passing something like `--py-limited-api=cp34` to setup.py

As for for "does it make sense to use this with cibuildwheel" I can think of a couple of benefits:

- some APIs may only have been declared "ABI stable" from a certain version of Python onwards. This is the case for cryptography: pythons 3.4 and up can use a common wheel, previous ones need their own wheels

- Python 2.x has no equivalent for this feature -> you still need to build a separate wheel

- cibuildwheel's ability to run a test suite on the resulting wheels remains a great bonus

- cibuildwheel remains a big help for the "cross platform" aspect

As stated initially, I mainly wanted to get a discussion started about whether this makes sense or not, I don't really have the answer at this stage.

joerick:
Interesting. I didn't know about this! It would be a great improvement if we could support this.

A few notes off the top of my head:

- It would be nice to be able to build limited-api wheels just by specifying the lowest python version's limited API e.g. `CIBW_LIMITED_API=cp34`. Builds below that python version (that haven't been skipped) would still build individual wheels. Builds at or above would build a limited-api wheel if there isn't one already.
- We'd still like to _test_ the limited-api wheels on many versions of Python. So those builds would run as normal, but if there's already a suitable limited-api wheel in the output directory (from a previous limited-api build), skip the build and just run the tests using that.
- Can support for the limited-api be detected from setup.py? It would be lovely to just do the right thing without adding a new option!
[...]

PR #996, opened 2022-01-22
refactor: split generic workflow from OS specific workflow

### LINKED ISSUE #560: Factor our run_on_each
I'd like to make the following change. The current structure of the build function looks like this:

```python
    temp_dir = Path(tempfile.mkdtemp(prefix='cibuildwheel'))
    built_wheel_dir = temp_dir / 'built_wheel'
    repaired_wheel_dir = temp_dir / 'repaired_wheel'

    try:
        if options.before_all:
            log.step('Running before_all...')
            env = options.environment.as_dictionary(prev_environment=os.environ)
            before_all_prepared = prepare_command(options.before_all, project='.', package=options.package_dir)
            call([before_all_prepared], shell=True, env=env)

        python_configurations = get_python_configurations(options.build_selector)

            log.build_start(config.identifier)
            ...
```

I'd like to refactor it like this:

```python
    try:
        if options.before_all:
            log.step('Running before_all...')
            env = options.environment.as_dictionary(prev_environment=os.environ)
            before_all_prepared = prepare_command(options.before_all, project='.', package=options.package_dir)
            call([before_all_prepared], shell=True, env=env)

        python_configurations = get_python_configurations(options.build_selector)

            with TemporaryDirectory(prefix='cibuildwheel') as tmp_dir:
                build_on_each(config, env, Path(tmp_dir))

    except subprocess.CalledProcessError as error:
        log.error(f'Command {error.cmd} failed with code {error.returncode}. {error.stdout}')
        sys.exit(1)


def build_on_each(config, env: Dict[str, str], tmp_dir: Path) -> None:
    built_wheel_dir = temp_dir / 'built_wheel'
    repaired_wheel_dir = temp_dir / 'repaired_wheel'

    log.build_start(config.identifier)
    ...
```

This provides a new temporary directory that is always cleaned up for each build configuration, is much better for unit testing and mocking, and removes the ridiculous levels of indentation we have for 90% of the code in build and that is keeping us from applying good programming practices, like context managers (and auto formatting? ;) ). I also think it's much easier to read the try/except, and "build_on_each" is a clearly defined concept, and still reads nicely linearly. See #558.

mayeut:
This PR started as an attempt to close #560 in the backlog.
As mentioned in the issue, there's probably little benefit in doing just this so I went further than that just to get some ideas where we could go. This is just a draft to see if any of those ideas are worth looking into, I don't expect this PR to be merged ever.

The end game idea was to split generic stuff from OS specific stuff which could help for #78, #676, #966, #971
To help reviewing/checking if anything is of interest:
- Virtual Environments are now handled as context managers for either build or tests
- There's now a PlatformBackend abstract class which should be implemented for all that's related to call/shell/file operations. This allows to have Windows/macOS to implement a "native/direct" one, linux the DockerContainer one.
- There's now a BuilderBackend class which handles the default build workflow thus avoiding code duplication for generic stuff. It can be customized to handle OS specific workarounds, specific implementations.

As a whole:
Pros:
- Better workflow consistency between all OS (and we could enforce the same constraints on linux as on windows/macOS easily for even more consistency) with less code duplication.
- Allows easier extendability through abstract class (one might even suggest a plug-in system as an end goal)

Cons:
- Higher complexity
- Harder to see the order of specific patch/workarounds/specifics at OS level


cibuildwheel/backend.py
```diff
@@ -0,0 +1,296 @@
...
+from .typing import PathOrStr, assert_never
+from .util import (
+    NonPlatformWheelError,
+    get_build_verbosity_extra_flags,
+    prepare_command,
+)
+from .virtualenv import VirtualEnv, VirtualEnvBase
+
+
+class BuilderBackend:
```
henryiii, line 18: I think this would be a lot simpler as a dataclass.

cibuildwheel/backend.py
```diff
@@ -0,0 +1,296 @@
...
+            self._before_build()
+            with self.platform.tmp_dir(self.identifier + "-built-wheel") as built_wheel_dir:
+                built_wheel = self._build_wheel(built_wheel_dir)
+                wheels = self._repair_wheel(built_wheel, repaired_wheel_dir)
+            # save build environment
+            self._constraints_dict = self.venv.constraints_dict
+        return wheels
+
+
+def test_one(
```
henryiii, line 244: Starting with `test_` makes this look like a pytest test.

cibuildwheel/platform_backend.py
```diff
@@ -0,0 +1,213 @@
...
+from contextlib import contextmanager
+from pathlib import Path, PurePath
+from typing import Dict, Iterator, List, Optional
+
+from cibuildwheel.bashlex_eval import local_environment_executor
+from cibuildwheel.typing import Final, Literal, PathOrStr, PlatformName
+from cibuildwheel.util import call, ensure_virtualenv, shell
+
+
+class PlatformBackend:
```
henryiii, line 14: Shouldn't this be an ABC? (either through the ABCMeta metaclass or inheriting from abc.ABC)?

henryiii (commented):
I did a quick first pass. In general, I like this direction as long as it's easy to unit test (mocking and such should be easy?). I'll need to do a second pass looking not at the diff, but at the full code, to see if it is still reasonable to track what is happening as a reader. Also to see why there's a backend and a platform backend.

I'm not sure if @joerick will like it, given https://cibuildwheel.readthedocs.io/en/stable/contributing/#design-goals, but since it's this far, I think a "final version" should be compared with the current version (probably not through a diff) to see if it looks better, to see if it looks easy to make large specific changes without breaking the other platforms, etc. (PS: to be clear, I'm not claiming to like it yet either, just positive about the direction but will need further review later.)

henryiii:
PS: Also, regardless of the final comparison, there are some good changes here we should pull out, like more `Final` usage.

henryiii:
By the way, I've thought about this too, and what I likely would have attempted would have been to implement smaller methods and a Protocol that would verify that a backend had the required components. Protocols are generally nicer than ABC's, though it's easier to share code with an ABC (which is not always great for reading the code, which is part of why Protocols are nice...).

joerick:
just a quick note - i'm very curious to review this but I won't have time for a few days!

joerick:
Thanks for this @mayeut, it's interesting! Reflections on a quick 30-min skim through...

- Reading the code itself (outside of the diff), I get a bit lost. For example, it's not very clear to me where the _actual building_ takes place. The existing codebase, you can go to the `build` function of the platform you're interested in and read it from top-to-bottom to understand what happens. I like that - especially in the context of building things, a 'build script' style of code is more suitable I think. In this PR, we lose this straight-forwardness. Of course, the existing codebase pays for that in some code duplication, but I think readability is important enough to warrant this.
- I'm surprised how 'big' BuilderBackend is. It's got a lot of functionality in there, so perhaps suffers from the age-old problems of inheritance - where execution flow gets pretty confusing. Perhaps there would be a different version of this which tried to make the platform-specific bits as a Protocol, with as small a surface-area as possible. This might be nicer. But I'm not sure how small that surface area could get...
- Maybe it's possible to retain the build-script style, but with state-less (or minimally stated) utilities for the platform-specific bits via a Protocol? I've thought about this a little... possibly yes, but I guess we'd have to have loads of `if platform == 'windows':` and `if platform == 'macos':` for the various workarounds. I'm not sure if this would be more or less readable... potentially it could be more of a mess than the existing code. 
- I know that Lines of Code is not a great benchmark for a number of reasons, but I am surprised that this increases the count, given the duplication that was eliminated. I suppose we pay for it in genericness.

mayeut:
@joerick, @henryiii, thanks for taking the time to review this.

I'll try to address the comments but not sure when I'll be able to address all of them.
In the meantime, if there are bits that you think should be extracted from this on their own, please let me know (as @henryiii did for `Final` usage).

mayeut:
I rebased on main. Only the last 3 commits are addressing some of the concerns raised during review.

cibuildwheel/platform_backend.py
mayeut: done.

cibuildwheel/backend.py
mayeut: changed to `do_test_wheel`

cibuildwheel/backend.py
mayeut: I will need some time to think about what it would look like going this way.

mayeut:
>  The existing codebase, you can go to the build function of the platform you're interested in and read it from top-to-bottom to understand what happens. I like that - especially in the context of building things, a 'build script' style of code is more suitable I think. In this PR, we lose this straight-forwardness. Of course, the existing codebase pays for that in some code duplication, but I think readability is important enough to warrant this.

@joerick, I probably went too far in `BuilderBackend`: as I was doing `_setup_build_env`, I thought it might be good for other steps to be functions as well. I removed functions for those other steps and expanded them in `BuilderBackend::run`. It still does not answer your concern but is probably better now than before in this regard.

joerick:
There are some good ideas here, I think it was worth attempting for sure, but ultimately I don't think that the extra layers of abstraction give us enough upside in maintainability to outweigh the downsides of indirectness.

I'm cleaning up PRs, so I'll close this one for now. Ofc feel free to reopen if you'd like to continue discussion.

PR #999, opened 2022-01-25
chore: add PyLint and fix most issues

henryiii:
This adds PyLint. I've disabled quite a few checks - PyLint is well known for producing a lot of false positives by default, but some of the checks were really useful. I'll try to outline the major ones below. This only runs in nox / ci, not in pre-commit, because pylint expects your app to be installed with all dependencies, so is more like a pytest check than a normal linter check. (In fact, there's probably a pytest plugin for it, that seems familiar).

* Unused parameters removed. In the one place it was really there because we have a makeshift ABC/Protocol (`linux.build`), marked with a pylint comment.
* `check=` was missing on subprocess.run in once instance - best to always specify it.
* Too-wide catching for errors. I've specified the proper error, and improved handling a bit.
* Cleaned up manual exception handling on `read_python_configs`, still not as nice as something like `urllib3`, but simpler than before.
* Cell capture in loop - this causes a value to get captured, but _later_ - to avoid confusion or later mistakes, I've either used default value capture (which is at define site, not call site), or used `functools.partial`.
* Fixed a f-string missing an f.
* Replaced ugly globals and out-of-init attribute definitions with `@functools.lru_cache`.
* Use matching argument names on subclass (on ast.AST)
* Added a missing encoding when opening a file (constraints).
* Used guard-style everywhere. This is the largest change, and possibly the most controversial, but I've decided I like it. Guard-style is clearly less mental effort when you have nested structures, and is at least consistent when you have shallow structures. It's also a bit easier to read, as you are never combining `elif` or `else` with control flow. It also encouraged some nice cleanup on some structures like the return Color/Symbol code!

I've added this to GHA, but I'd also understand it if you just want to run it manually once in a while from nox - pylint's dynamic nature is rather finicky about things like what version of Python it's running on, etc.


mayeut:
This looks good but I'm not sure it shall be run & enforced in GHA as it might discourage contributors if they can't reproduce the issue easily locally for the reason you mention.
I'd be fine with manual run or even a weekly run in GHA (just to get a heads up at least).

henryiii:
Only a significant number of new issues (probably 10+ or so) will trigger a "failure" now, though a severe error still will. But GHA will annotate any warnings it sees!

Example from a mistake elsewhere using this matcher:

Screen Shot 2022-01-29 at 12 07 50 PM

---

PS: I really don't think linter red x's are that discouraging; personally I'd rather have quick automated feedback for the simple stuff over having to wait for a maintainer to comment with picky suggestions. We also don't have droves of people contributing, and I'd rather jump in and help someone fix red x's vs. have crummy code merged. I also feel safer _using_ a package with strict checks, because it means they care about code quality, making it likely to be reliable. But my general "minimal" recommendation is to run pre-commit, and pylint is more like pytest, and not very good as a pre-commit check, so I'm also completely fine with not having it in the required "red x" set.

mayeut:
> PS: I really don't think linter red x's are that discouraging; personally I'd rather have quick automated feedback for the simple stuff over having to wait for a maintainer to comment with picky suggestions. We also don't have droves of people contributing, and I'd rather jump in and help someone fix red x's vs. have crummy code merged. I also feel safer using a package with strict checks, because it means they care about code quality, making it likely to be reliable. But my general "minimal" recommendation is to run pre-commit, and pylint is more like pytest, and not very good as a pre-commit check, so I'm also completely fine with not having it in the required "red x" set. It's also picky about what version of Python it runs on, unlike the other checks.

I do agree with what you're saying overall. If it's only picky about what version of Python it runs on (& not OS for example), then pinning the python version in the noxfile should allow anyone to reproduce issues locally right ?

henryiii:
It's only lightly picky from what I can tell, not bad, just not as independent as other static checks. "Any recent Python version" doesn't seem too bad, and nox isn't great about finding my brewed Python 3.10 (and it doesn't like pyenv), so I prefer letting letting checks pick the "current" version until we find a difference. I think @joerick will also want to keep this from being too picky for contributors, so the current model (doesn't put a red X unless many issues are found, adds annotations) is likely the best (for now, possibly changing after we see how this works in the future?)

cibuildwheel/macos.py
```diff
@@ -460,15 +461,19 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                            msg = f"don't know how to emulate {testing_arch} on {machine_arch}"
+                            raise RuntimeError(msg)
 
                     # define a custom 'call' function that adds the arch prefix each time
-                    def call_with_arch(*args: PathOrStr, **kwargs: Any) -> None:
-                        call(*arch_prefix, *args, **kwargs)
-
-                    def shell_with_arch(command: str, **kwargs: Any) -> None:
+                    call_with_arch = functools.partial(call, *arch_prefix)
+
```
joerick, line 469: Personally, I'm not a massive fan of functools.partial, it's a bit high-order for my taste. I do understand what it's doing, but I really have to think about it, whereas I can just read the explicit version. What's the warning that pylint produces here?

cibuildwheel/macos.py
```diff
@@ -460,15 +461,19 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                    def shell_with_arch(
+                        command: str,
+                        arch_prefix: Tuple[str, ...] = tuple(
+                            arch_prefix
+                        ),  # required to capture the value of arch_prefix now, rather than later (less confusion)
+                        **kwargs: Any,
+                    ) -> None:
```
joerick, line 476: I have no idea what's going on here....! Is this refactor necessary?

cibuildwheel/options.py
```diff
@@ -139,7 +140,7 @@ def _dig_first(*pairs: Tuple[Mapping[str, Setting], str], ignore_empty: bool = F
 
             return value
 
-    raise KeyError(key)
+    raise KeyError(key)  # pylint: disable=undefined-loop-variable
```
joerick, line 143: This line did trip me up on first reading it too, maybe this would avoid the comment?

```suggestion
    last_key = pairs[-1][1]
    raise KeyError(last_key)
```

cibuildwheel/options.py
```diff
@@ -310,17 +311,22 @@ def get(
...
+        if isinstance(result, int):
             return str(result)
-        else:
-            return result
+
+        return result
 
 
 class Options:
+    # pre-decleration to avoid property decorator mypy clash.
```
joerick, line 327:
```suggestion
    # pre-declaration to avoid property decorator mypy clash.
```

cibuildwheel/options.py
```diff
@@ -345,12 +351,11 @@ def config_file_path(self) -> Optional[Path]:
...
-    @property
+    @property  # type: ignore[no-redef, misc]
+    @functools.lru_cache(maxsize=None)
     def package_requires_python_str(self) -> Optional[str]:
-        if not hasattr(self, "_package_requires_python_str"):
-            args = self.command_line_arguments
-            self._package_requires_python_str = get_requires_python_str(Path(args.package_dir))
-        return self._package_requires_python_str
+        args = self.command_line_arguments
+        return get_requires_python_str(Path(args.package_dir))
```
joerick, line 358: I'm kinda surprised we haven't got a `@cached_property` decorator in this project. I wonder if mypy would like that better...?

cibuildwheel/util.py
```diff
@@ -150,11 +150,14 @@ def format_safe(template: str, **kwargs: Any) -> str:
...
+        # we use a function for repl to prevent re.sub interpreting backslashes
+        # in repl as escape sequences. Capturing the current value of value.
+        def repl_func(_: Any, value: Any = value) -> str:
+            return str(value)
+
         result = re.sub(
             pattern=find_pattern,
-            repl=lambda _: str(value),
+            repl=repl_func,
             string=result,
```
joerick, line 161: This is a little eyebrow-raising for me. What's wrong with the original code?

cibuildwheel/bashlex_eval.py
```diff
@@ -45,14 +45,14 @@ def evaluate(
...
         node_result = evaluate_command_node(node.command, context=context)
         # bash removes training newlines in command substitution
         return node_result.rstrip()
-    elif node.kind == "parameter":
+    if node.kind == "parameter":
         return evaluate_parameter_node(node, context=context)
-    else:
-        raise ValueError(f'Unsupported bash construct: "{node.kind}"')
+
+    raise ValueError(f'Unsupported bash construct: "{node.kind}"')
```
joerick, line 55: Yeah, 90% of these if..elif refactors make it harder to read IMO.

joerick (commented):
There are some nice improvements here. Also, I really appreciate how this is set up to only provide warnings, rather than fail the build. Let's go ahead with that setup and see how it fares. I really like the lint errors screenshots you showed.

Style-wise, my biggest complaint is with the if...elif...else refactors, as noted inline.

cibuildwheel/macos.py
henryiii: `functools.partial` does exactly what you expect it to do. It uses the value of arch_prefix. While cell capture (which I absolutely hate, it makes things much, much harder to read - I should not have to read the entire body of a function to see what variables it references the are not in the arguments) does _not_ do that, instead it takes the variable and captures it _later_. So this code:

```python
y = 3
def f():
    return y
y = 4
print(f"{f()=}")
```
Does not print `f()=3`, it prints `f()=4`.

This code is much more readable, doesn't require reading the body of the function, and has no surprises:

```python
y = 3
def f_x(x):
    return x
f = functools.partial(f_x, y)
y = 4
print(f"{f()=}")
```

It prints `f()=3`.

cibuildwheel/options.py
henryiii: `cached_property` is 3.8+ only.

cibuildwheel/bashlex_eval.py
henryiii: The final expression is "special", it's the "nothing matched" case, so this highlights it as such. It also actually makes the code easier to lint; see https://github.com/tomerfiliba/plumbum/pull/584/files#diff-f8094a3c5928e29d7188c84488925d2728abc59163ba387451cca5a8cdcd54e8 - besides removing a lot of white space, "guard" style fixed an issue parsing the code that make liners think there was a way to return without a return statement.

Some of them are neither better nor worse, IMO, but this is one that's a tiny bit better. If we could use match statements, that would be even better, but we can't, yet.

cibuildwheel/architecture.py
```diff
@@ -81,12 +81,14 @@ def all_archs(platform: PlatformName) -> "Set[Architecture]":
...
+        if platform == "macos":
             return {Architecture.x86_64, Architecture.arm64, Architecture.universal2}
-        elif platform == "windows":
+
+        if platform == "windows":
             return {Architecture.x86, Architecture.AMD64, Architecture.ARM64}
-        else:
-            assert_never(platform)
+
+        assert_never(platform)
```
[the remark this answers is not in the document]
henryiii: > Could we disable this lint rule and restore these refactors?

We don't _have_ to have this lint rule enabled 😆. However, on reading large amounts of code and performing this refactor (cibuildwheel, plumbum, boost-histogram, and partially awkward), I find it to be very helpful for nested structures, and a tradeoff for shallow ones; and even then, it sometimes prompts real improvements, like moving the "return" clauses to the top. It reduces unnessiary whitespace, and it avoids dangling "else"'s that that are hard to match up with if's.

The key thing for me is you _can't_ ignore the body of the if - they have return's in them, and returns _are_ control flow, so the "else" is completely redundant and can be removed. Since it _can_ be removed, you have three choices: alway include it (terrible, would add tons of whitespace), always disallow it (this check), or mix and match and force the reader to have to digest both.

While when you write it, you might have an intent you are trying to convey, but there is no code impact at all; it is transparent to Python, so it really doesn't help the reader. It also creates long elif chains that are mentally taxing, while you can look at each guard clause in isolation; they are just "narrowing" (and this is exactly how MyPy works, actually).

cibuildwheel/macos.py
henryiii: Same issue with capture. If you put `=thing` in your defaults, it is both explicitly captures instead of implicitly captured (good!), and it also uses the _current_ value of variable, rather than the value of the variable at the call site, which is _very_ bug prone.

henryiii: But I forgot to format this correctly in Black. 🤦

cibuildwheel/options.py
henryiii: Oh, nice. :)

cibuildwheel/options.py
henryiii: I need to think about this and make sure I didn't leak here. Not that it would really hurt us, but would be a bad example.

cibuildwheel/macos.py
henryiii: I think these sorts of changes were:

> Cell capture in loop - this causes a value to get captured, but later - to avoid confusion or later mistakes, I've either used default value capture (which is at define site, not call site), or used functools.partial.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
henryiii: I really wish there was some sort of setting that would allow one layer, but not more. Many of the other checks have this sort of a knob, but I don't think these three do.

cibuildwheel/environment.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, name: str, value: str):
     def __repr__(self) -> str:
         return f"{self.name}: {self.value}"
 
-    def evaluated_value(self, **kwargs: Any) -> str:
+    def evaluated_value(self, **_: Any) -> str:
```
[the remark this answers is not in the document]
henryiii: `**_kwargs` is fine if you like that better, I think it was complaining about it being unused. `**kwargs` means you are using the extra arguments, usually - the convention is arguments that start with an underscore might be unused, but normal arguments always will be used. It could also be an ignore comment, but I think `_` or `_kwargs` conveys the "we won't use it" meaning better than a comment and an unused `kwargs`. But not strongly opinionated on it; it's so short it's hard to miss that it's unused.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
joerick: > While when you write it, you might have an intent you are trying to convey, but there is no code impact at all; it is transparent to Python, so it really doesn't help the reader. 

Ah, see, well, I completely disagree with this statement, philosophically. We carefully name all our functions, variables, organise our code into classes and modules - this has no bearing on the ultimate working of the script, it could all be minified into one file and Python would execute it exactly the same. The code, the structure is for humans, to make it easier to understand!

> The key thing for me is you can't ignore the body of the if - they have return's in them, and returns are control flow, so the "else" is completely redundant and can be removed. 

Hmm, not sure I understand. If you're reading the above code while debugging a Windows issue, you don't need to read the mac section.

> Since it can be removed, you have three choices: alway include it (terrible, would add tons of whitespace), always disallow it (this check), or mix and match and force the reader to have to digest both.

Yeah, the third one. Except I'd phrase it as "stop generic rules from interfering with humans expressing ideas" :P

cibuildwheel/environment.py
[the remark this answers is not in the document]
henryiii: Pylance (which is running in vscode) likes `**_` much better than `**kwargs` or `**_kwargs`. The later seems buggy to me, since leading underscore for unused args is a common convention.

cibuildwheel/macos.py
joerick: "Cell capture in loop" is a valid criticism, but it doesn't apply in this case as the function isn't used outside of that iteration. I'd class this as a false positive.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
henryiii: > We carefully name all our functions

We have to choose a name, so we choose a name that conveys meaning. Here are not doing that, we are adding extra control structures that have no meaning and are completely ignorable.

Just curious, what about:

```python
for f in funcs:
    if f():
        return True
else:
    return False

```

? This is not hypothetical - this is an actual structure I'm clearing out of a different package.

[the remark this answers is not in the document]
henryiii: (note the alignment of the else - that was intentional)

cibuildwheel/logger.py
```diff
@@ -262,8 +257,6 @@ def file_supports_unicode(file_obj: IO[AnyStr]) -> bool:
     return "utf" in codec_info.name
 
 
-"""
-Global instance of the Logger.
-"""
+# Global instance of the Logger.
 # (there's only one stdout per-process, so a global instance is justified)
 log = Logger()
```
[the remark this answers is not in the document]
henryiii: Instances don't have have docstrings, only modules, classes, and functions (and in slots in Python 3.8+). I think there might be a bit of syntax somewhere kind of like this, like for class instances or something, because I've written things like that before, but wasn't able to get it just now with a little trying.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
joerick: > Here are not doing that, we are adding extra control structures that have no meaning and are completely ignorable.

I'd argue that they do have meaning. They serve to express that this statement is meant to be read like a 'switch'.



> Just curious, what about:
> 
> ```python
> for f in funcs:
>     if f():
>         return True
> else:
>     return False
> ```

for...else has never made much sense to me. The analogy to if...else is weak, because we did execute the `for` already. My opinion is that the above is more readable without the `else`.

cibuildwheel/environment.py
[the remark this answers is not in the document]
joerick: Yeah, I think `_kwargs` seems okay as a compromise. `**_` looks like a magic incantation of punctuation, like Perl or something :P

cibuildwheel/options.py
joerick: we can keep this, though, if you prefer.

cibuildwheel/util.py
joerick: ahh, it's the cell-capture in a loop again. Still feels like a false-positive to me. That lambda is going nowhere outside of that line of code.

cibuildwheel/macos.py
henryiii: I really, really don't like capture. It's misused and makes code really hard to read.  I've even included it in a problem to students before, https://github.com/henryiii/compclass/blob/master/problems/Week10.ipynb :)

It only works in this case because arch_prefix doesn't change. However, if someone were to refactor this one day to overwrite arch_prefix later, it would suddenly change the behavior of these to functions - the dependence on `arch_prefix` is completely hidden after the fact and requires studying the contents of the functions to see.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
henryiii: So is:
```Python
if g():
    if f():
        return True
else:
    return False
```
better than:
```Python
if g():
    if f():
        return True
return False
```
? Imagine this much longer, of course, where aligning the else's is harder.

cibuildwheel/options.py
henryiii: I'm still working though the code, but I think I might have created a leak here with the current code - https://youtu.be/sVjtp6tGo0g?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY. It's probably better to vendor in a cached_property if you are fine with that.

cibuildwheel/util.py
henryiii: It's a little ugly, but I'm okay to just suppress this one, since you can't accidentally break it later by assigning `value`. I'll restore it and verify that this was the error it triggered.

cibuildwheel/options.py
henryiii: Vendored a copy from the stdlib + typeshed.

cibuildwheel/util.py
henryiii: Added a suppression and went with the original code.

cibuildwheel/options.py
henryiii: Not needed at all now that mypy knows what a `cached_property` is. :)

henryiii:
Also pushed a fix for https://github.com/theacodes/nox/issues/340 - nox happy report a 0 error code if no Pythons are found. Let's just make sure we don't get hit by that.

cibuildwheel/architecture.py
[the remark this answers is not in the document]
mayeut: Shouldn't we get rid of `if`/`elif`/`else` altogether for this one ? IMHO, the real intent is a map and maybe doing just that would be better for readability ?

```python
    @staticmethod
    def all_archs(platform: PlatformName) -> "Set[Architecture]":
        all_archs_map = {
            "linux": {
                Architecture.x86_64,
                Architecture.i686,
                Architecture.aarch64,
                Architecture.ppc64le,
                Architecture.s390x,
            },
            "macos": {Architecture.x86_64, Architecture.arm64, Architecture.universal2},
            "windows": {Architecture.x86, Architecture.AMD64, Architecture.ARM64},
        }
        return all_archs_map[platform]
```

cibuildwheel/bashlex_eval.py
mayeut: > Some of them are neither better nor worse, IMO, but this is one that's a tiny bit better. If we could use match statements, that would be even better, but we can't, yet.

agreed

cibuildwheel/macos.py
joerick: That's fine, but I still maintain that it's a false positive. Lint rules are friction. But they're good if they catch issues that save us as maintainers time make code review easier. This is not something that I would object to if I was reviewing the code, because it's not a bug!

henryiii: A really, really easy bug to introduce is to assign the `arch_prefix` later. The capture version is securely holding a reference to that variable, rather than capturing the current value, which is naively what most users would expect.

This makes the code much easier to read, since you don't have to inspect the contents of the function to notice it is actually capturing something in the outer scope. I'd maintain that any capture inside a function is harder to read than placing it into the signature - that's what the signature is for. I shouldn't be forced to read bodies of functions to know what variables are being used!

And yes, to the best of my knowledge, I always object to use of variable capture in reviews. (Actually, I object to nested functions in reviews, too, they reduce the readability and just tend to be used to capture variables, which is bad. I'm sure you can see such reviews in, for example, DecayLanguage & Particle where I've done exactly that. ;) )

henryiii: What about moving these nested functions out, actually! I think that would probably help quite a bit here. Yes, it did! New version is just:

```python
call_with_arch = functools.partial(call, *arch_prefix)
shell_with_arch = functools.partial(shell, *arch_prefix)
```

And no tested functions at all. :)

cibuildwheel/macos.py
henryiii: This is now gone with the refactor.

henryiii:
What's the plan on this? I think there are some nice cleanups here. Let me know what direction to go.

mayeut:
> What's the plan on this? I think there are some nice cleanups here. Let me know what direction to go.

On my end, I also think there are some nice cleanups here. All my comments have been addressed so, providing a rebase enables CI to give the green light, I'm all in.

joerick:
Apologies, I have been meaning to come back to this. I find this kind of thing quite tiring. Thank you for the responses to feedback so far. Aside from the if..elif refactors, it looks good. I can't agree with the if..elif rule however, and I fundamentally disagree that something so subjective should be decided by a rule. If you're okay with it, I can take a pass and add a commit to undo some of those refactors, and then we can get it merged.

joerick:
I've pushed a commit to disable some checks and revert some of the changes that I disagree with. Also have resolved conflicts. Feel free to merge when you're ready @henryiii.

PR #1010, opened 2022-02-05
chore: use `Final` type annotation on every top-level constant

mayeut:
per https://github.com/pypa/cibuildwheel/pull/996#issuecomment-1021446632, this annotates top-level constants as such.

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -4,9 +4,9 @@ import re
 from enum import Enum
 from typing import Set
 
-from .typing import Literal, PlatformName, assert_never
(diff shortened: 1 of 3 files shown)


henryiii (approved):

joerick (approved):
Why this type is called `Final` and not `Const` baffles me. But either way, looks good!

henryiii:
I think because it only refers to the reference assignment, not methods.

```const
x: Final = [3]
x.append(4)
```

Is sadly completely valid. It protects against `=`, including `+=`, but does not understand "constness" of methods (which it can't, there's no `const` for methods in Python).

PR #1011, opened 2022-02-06
chore: cleanup and harden pytest config

henryiii:
Minor cleanup, and make sure pytest is strict.

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -102,7 +102,7 @@ class DockerContainer:
 
         self.bash_stdin.write(b"exit 0\n")
         self.bash_stdin.flush()
-        self.process.wait(timeout=5)
+        self.process.wait(timeout=30)
         self.bash_stdin.close()
         self.bash_stdout.close()
 
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -100,9 +100,11 @@ class DockerContainer:
         exc_tb: Optional[TracebackType],
     ) -> None:
 
+        self.bash_stdin.write(b"exit 0\n")
+        self.bash_stdin.flush()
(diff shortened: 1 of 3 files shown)


mayeut:
It seems the warnings (turned in error) comes from https://github.com/pypa/cibuildwheel/blob/b52698dad6af15f292cef3e9d2f148969ac730fa/cibuildwheel/docker_container.py#L96-L107

adding the following gets rid of the error.
```python
self.bash_stdout.close()
```

I wonder if terminating differently might help with #966 race conditions. Something like:
```python
self.bash_stdin.write(b"exit 0\n")
self.bash_stdin.flush()
self.process.wait()
self.process.stdin.close()
self.process.stdout.close()
```

henryiii:
I've implemented something close to that, thoughts?

cibuildwheel/docker_container.py
```diff
@@ -100,9 +100,11 @@ def __exit__(
         exc_tb: Optional[TracebackType],
     ) -> None:
 
+        self.bash_stdin.write(b"exit 0\n")
+        self.bash_stdin.flush()
+        self.process.wait(timeout=5)
```
mayeut, line 105: Should we fail on timeout or just `docker kill` the container and try to wait again (and go on) ?

joerick: Failing here would be a non-critical error, so it would be acceptable to just log a warning and move on. But until we see such an error reported to us, I don't think it's necessary to code around it - this seems okay to me.

But, maybe it would be good to increase this timeout to decrease the chance of this happening, on a remote container for example. It would be frustrating to fail a build because of this. And if we do ever get an error report, we'd like to be sure that it really had stallled.

```suggestion
        self.process.wait(timeout=30)
```

joerick (approved):

PR #1014, opened 2022-02-08
Add Arbor to working-examples.md

brenthuisman:
Might be interesting to mention that we have a [small script](https://github.com/arbor-sim/arbor/blob/master/scripts/patchwheel.py) to patch `rpath` in included libraries, potentially useful to others. The comment seemed to indicate an interest in such things, but there is no obvious place to add it. Maybe just append to the description?


henryiii:
This needs to be added into https://github.com/pypa/cibuildwheel/blob/main/docs/data/projects.yml instead. And you can add a `notes:` entry with that info, otherwise it draws from the GH description.

brenthuisman:
Whoops! I'll make the correct PR.

PR #1020, opened 2022-02-11
Support building from an sdist archive

pombredanne:
It is now possible to call the cibuildwheel command line and point it
to a package source distribution archive tarball (sdist) instead of
only a directory.

Reference: https://github.com/pypa/cibuildwheel/issues/597
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


cibuildwheel/__main__.py
```diff
@@ -147,13 +149,20 @@ def main() -> None:
...
-    package_files = {"setup.py", "setup.cfg", "pyproject.toml"}
-
-    if not any(package_dir.joinpath(name).exists() for name in package_files):
-        names = ", ".join(sorted(package_files, reverse=True))
-        msg = f"cibuildwheel: Could not find any of {{{names}}} at root of package"
-        print(msg, file=sys.stderr)
-        sys.exit(2)
+    from_sdist = package_dir.is_file()
+    from_dir = not from_sdist
+    if from_dir:
```
nsoranzo, line 154:
```suggestion
    if not from_sdist:
```

cibuildwheel/__main__.py
```diff
@@ -147,13 +149,20 @@ def main() -> None:
...
+        package_files = {"setup.py", "setup.cfg", "pyproject.toml"}
+        if not any(package_dir.joinpath(name).exists() for name in package_files):
+            names = ", ".join(sorted(package_files, reverse=True))
+            msg = f"cibuildwheel: Could not find any of {{{names}}} at root of package"
+            print(msg, file=sys.stderr)
+            sys.exit(2)
+    elif from_sdist:
+        if not str(package_dir).endswith(".tar.gz"):
+            msg = f"cibuildwheel: Package is not an sdist'.tar.gz' tarball"
```
nsoranzo, line 163:
```suggestion
            msg = f"cibuildwheel: Package {package_dir} is not an sdist '.tar.gz' tarball"
```

nsoranzo:
@pombredanne Thanks for working on this, it's a feature I'd need for a wheel-building project and I was hoping to have time to implement it at some point.
I haven't looked into it much, but I was wondering if unpacking the sdist in a temporary directory would simplify things. You then could re-assign `package_dir` to the temporary directory, so wouldn't need to skip the search for `pyproject.toml` and maybe also the `build` backend works?

pombredanne:
@nsoranzo Thanks for the review!

> I haven't looked into it much, but I was wondering if unpacking the sdist in a temporary directory would simplify things. You then could re-assign `package_dir` to the temporary directory, so wouldn't need to skip the search for `pyproject.toml` and maybe also the `build` backend works?

yeah, I have thought about it and decided to go simpler, but it would be fairly trivial to extract the tarball and rejoin the "package_dir" groove after this. This would make "build" work... though there are subtle things about an sdist layout that the "build" frontend may not recognize, chiefly the `PKG-INFO` METADATA-like file? 

I am fine either way.

cibuildwheel/options.py
```diff
@@ -339,9 +339,11 @@ def config_file_path(self) -> Optional[Path]:
             return Path(args.config_file.format(package=args.package_dir))
 
         # return pyproject.toml, if it's available
-        pyproject_toml_path = Path(args.package_dir) / "pyproject.toml"
-        if pyproject_toml_path.exists():
-            return pyproject_toml_path
+        package_dir = Path(args.package_dir)
```
henryiii, line 342: Can't we pull this out of the tar.gz? Or even zip, but I'm fine in following build and only allowing .tar.gz's.

(There's also the temporary dir idea - the only benefit of this I can immediately think of is just validation, but if the build tools support it (and build should at some point), this seems cleaner)

pombredanne:
@henryiii Thanks for your review.... I just pushed a vastly simplified version that extracts a tarball in place... much less churn and code involved with this one. This is extracting the tarball as suggested by @nsoranzo and I think as you suggested too.

test/test_sdist_package.py
```diff
@@ -0,0 +1,29 @@
...
+    package_dir = project_dir / sdist_filename
+    shutil.copy2(sdist, package_dir)
+
+    # build the wheels from an sdist
+    actual_wheels = utils.cibuildwheel_run(
+        project_dir,
+        package_dir=package_dir,
+        add_env={
+            # this shouldn't depend on the version of python, so build only CPython 3.6
+            "CIBW_BUILD": "cp36-*",
```
henryiii, line 23: I'd probably go with something a bit newer, just so we don't have to mess with this when 3.6 goes away (mid next year, I think, but still). 3.9 or something.

I'd also really like to see that `--config-file=cibuildwheel.toml` still works with this (I _think_ so?), maybe this could be dropped into a toml file instead, and the equivalent of this passed in?

cibuildwheel/__main__.py
```diff
@@ -147,8 +150,19 @@ def main() -> None:
...
+    from_sdist = package_dir.is_file()
+    if from_sdist:
+        if not str(package_dir).endswith(".tar.gz"):
+            msg = f"cibuildwheel: Package {package_dir} is not an sdist '.tar.gz' tarball"
+            print(msg, file=sys.stderr)
+            sys.exit(2)
 
+        extract_dir, _, _ = str(package_dir).rpartition(".tar.gz")
+        with tarfile.open(package_dir) as tar:
+            tar.extractall()
```
nsoranzo, line 162: Should with pass a `path=` argument here to extract in a temporary directory instead of the current working directory?

pombredanne: The deal for a proper sdist is that the root directory inside the tarball is the base of name of the archive (e.g., minus the tar.gz extension). Exacting in place means that an extract dir will be created alright.

henryiii:
```
NotADirectoryError: [Errno 20] Not a directory: '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-1/popen-gw1/test2/project/spam-0.1.0.tar.gz/pyproject.toml'
```

test/test_sdist_package.py
pombredanne: > I'd probably go with something a bit newer, just so we don't have to mess with this when 3.6 goes away (mid next year, I think, but still). 3.9 or something.

I modeled this after the other tests for the pure wheel and it is using a 3.6 version. Therefore I thought that was the right way to process this and there must have been a reason for sticking to 3.6

henryiii: That was probably shiny and new at the time. 🤣 We can update them all at the same time, then, fine as is.

pombredanne: > That was probably shiny and new at the time. rofl We can update them all at the same time, then, fine as is.

This was the intent in adopting the existing ways

pombredanne: > I'd also really like to see that --config-file=cibuildwheel.toml still works with this (I think so?), maybe this could be dropped into a toml file instead, and the equivalent of this passed in?

can you elaborate what you mean a bit?

pombredanne: Note: there are no changes that relates to configuration. The process is simply extract and carry on with the regular flow

cibuildwheel/__main__.py
pombredanne: @nsoranzo I am fine to create a temp directory but this will then need to be moved back after under the project dir which is the contract for the "package dir" argument so far. I am not sure trying to keep things elsewhere would help and this would require extensive surgery

cibuildwheel/options.py
pombredanne: This is extracted now.

pombredanne:
@henryiii re:

> NotADirectoryError: [Errno 20] Not a directory: '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-1/popen-gw1/test2/project/spam-0.1.0.tar.gz/pyproject.toml'


This has been fixed.

cibuildwheel/__main__.py
```diff
@@ -144,11 +147,26 @@ def main() -> None:
...
+    # check if we have an sdist tarball as "package_dir" and extract first
+    package_dir = args.package_dir and Path(args.package_dir)
+    from_sdist = package_dir and package_dir.is_file()
+    if from_sdist:
+        if not str(package_dir).endswith(".tar.gz"):
+            msg = f"cibuildwheel: Package {package_dir} is not an sdist '.tar.gz' tarball"
+            print(msg, file=sys.stderr)
+            sys.exit(2)
+
+        extract_dir, _, _ = str(package_dir).rpartition(".tar.gz")
```
nsoranzo, line 159: We can't use [str.removesuffix()](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) yet, but this still looks simpler to me:

```suggestion
        extract_dir = str(package_dir)[:-len(".tar.gz")]
```

nsoranzo: Actually the code here and below seems to make some assumptions that are not guaranteed, namely:
- that the tarball is located in the current directory (can be easily fixed by a `os.path.basename()`)
- that the tarball contains a directory with the same name as the archive (minus the `.tar.gz`)
- that the tarball doesn't have files that will be created outside of the c.w.d. (see the warning in https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall )

I think we could at least verify that the tarball is "well-behaved" by calculating the common prefix directory of the members of the archive: if there is one, it can replace `extract_dir `here; otherwise, we just bail out.

Some inspiration can be taken e.g. from https://github.com/pypa/pip/blob/main/src/pip/_internal/utils/unpacking.py

cibuildwheel/__main__.py
joerick: Yes, I think a temp dir is better than extracting inplace. That way cibuildwheel isn't leaking its implementation into the filesystem. But you're right that it's a more substantial change. I'll have a think about the best way to go about this.

cibuildwheel/__main__.py
```diff
@@ -82,10 +83,12 @@ def main() -> None:
         default=".",
         nargs="?",
         help="""
-            Path to the package that you want wheels for. Must be a subdirectory of
-            the working directory. When set, the working directory is still
-            considered the 'project' and is copied into the Docker container on
-            Linux. Default: the working directory.
+            Path to the package that you want wheels for. Must be either a
+            a subdirectory of the working directory or a source distribution
+            (sdist) tar.gz archive file that will be extracted in-place.
```
joerick, line 88: > extracted in-place

Seems surprising for a user, I think. I wonder if we can improve this...

test/test_sdist_package.py
joerick: I'd also like to see `--config-file=cibuildwheel.toml` in a test. We should check that it's resolved relative to the cwd of the invocation, not to the sdist or a temp dir (if that's what's used).

test/test_sdist_package.py
```diff
@@ -0,0 +1,29 @@
+import shutil
+from pathlib import Path
+
+from . import utils
+
+
+def test(capfd, tmp_path):
+    sdist_filename = "spam-0.1.0.tar.gz"
+    sdist = Path(__file__).parent / sdist_filename
```
joerick, line 9: Rather than including a tar.gz in the repo, I'd rather build this within the test, so we can all see what's actually being tested by reading the code.

joerick:
Ah, this is actually more tricky to build in than I initially thought! In my opinion, the best way to do this is to build in a temporary directory, with the CWD set as the sdist root, so pyproject.toml is at `./pyproject.toml`. Having CWD at the sdist root is best, because 1) we can expand the tarball to a tmpdir, and 2) pyproject.toml might contain commands like `before_build = "sh scripts/prepare.sh"`, and these should be executed relative to that location. 

But, this change to the CWD makes me a bit uncomfortable, because it's changing the behaviour of all options that contain paths, because their cwd is not resolved relative to the cwd of the invocation. And, I'm also not super keen on making `__main__` more complicated for the sake of this feature.

I'm leaning towards a different solution, which doesn't actually modify cibuildwheel, but gives us another, separate utility. So instead of calling `cibuildwheel my-package.tar.gz`, you'd call `cibuildwheel-from-sdist my-package.tar.gz`. 

```
usage: cibuildwheel-from-sdist [-h] [--output-dir OUTPUT_DIR] 
                               [--config-file CONFIG_FILE]
                               [cibuildwheel-args...] sdist

Build wheels from an sdist archive.

Extracts the sdist and calls cibuildwheel on the resulting package directory.
Note that cibuildwheel will be invoked with its working directory as the
package directory, so options aside from --output-dir and --config-file are
relative to the package directory.

positional arguments:
  package               Path to the sdist archive that you want wheels for.
                        Must be a tar.gz archive file.
  cibuildwheel-args...
                        Any further arguments will be passed on to cibuildwheel.

optional arguments:
  -h, --help            show this help message and exit
  --output-dir OUTPUT_DIR
                        Destination folder for the wheels. Default: wheelhouse.
  --config-file CONFIG_FILE
                        TOML config file. Default: "", meaning
                        the pyproject.toml inside the sdist, if it exists.
```

I think this would be a lot less complexity for cibuildwheel itself to manage, and clearer API for users too. 

Would this fit your use case @pombredanne  and @nsoranzo ?

henryiii:
My longer-term hope would be that `pypa/build` would pick up the ability to build from SDist too (`pip wheel` already supports it), and I'd hope that then cibulidwheel could just support it naturally - the only thing we'd need to add is the ability to look inside for the pyproject.toml, which doesn't even need to be unpacked on disk, just memory.

Having a new command will not work well with either the GitHub Action, or pipx, where you'll have to run something like `pipx run cibuildwheel-from-sdist --spec cibuildwheel` to make it work. I'm a strong -1 on that design. That's basically already how https://github.com/astrofrog/autowheel worked.

In the interim, is there really that much of a problem just manually unpackaing an sdist into the current directory and then calling cibuildwheel? Simpler, composable commands vs. complicated integrated ones, and all that? It's just an extra command you'd need to stick before the GitHub action or pipx call, however you are running cibuildwheel. Might be a hair harder to make work on windows, but probably not terrible.

joerick:
> My longer-term hope would be that pypa/build would pick up the ability to build from SDist too (pip wheel already supports it), and I'd hope that then cibulidwheel could just support it naturally

Still, the issue of broken paths in options persists here. Pulling options from pyproject.toml and executing them from outside the sdist makes very little sense to me.

> Having a new command will not work well with either the GitHub Action, or pipx, where you'll have to run something like `pipx run cibuildwheel-from-sdist --spec cibuildwheel` to make it work.

This seems like a pretty minor concern to me, this wouldn't be a mainstream feature, so I don't mind if users need to have a slightly longer command line invocation to do it.

henryiii:
If you are running from the SDist, it's quite unlikely you have commands in your pyproject.toml for cibuildwheel. If you support cibuildwheel, you'd probably not build separately, and if you did, you'd want your configuration separate too - otherwise, you would not be able to edit it while you are working on making a release. The main thing to pull from it would be the python requires settings (and even then we could just not support any discovery of settings when running on a SDist, requiring everything to be in a local config file)

joerick:
> If you are running from the SDist, it's quite unlikely you have commands in your pyproject.toml for cibuildwheel.

I'm not so sure. I think that building from sdist could open up some interesting opportunities longer-term. I'm not sure what exactly, but when I think about a process like-

```
┌─────┐         ┌─────┐
│sdist│────┬───▶│wheel│
└─────┘    │    └─────┘
           │    ┌─────┐
           ├───▶│wheel│
           │    └─────┘
           │    ┌─────┐
           ├───▶│wheel│
           │    └─────┘
           │    ┌─────┐
           ├───▶│wheel│
           │    └─────┘
           │    ┌─────┐
           └───▶│wheel│
                └─────┘
```
it opens up

Issue #1021, opened 2022-02-11
Support building pure Python wheels


pombredanne:
As a follow up to https://github.com/pypa/cibuildwheel/discussions/1007 I would like to also allow building a pure wheel.
Why? When I build arbitrary wheels, I cannot know ahead of time if a wheel will be pure Python or native. And if I build many wheels including sdist tarballs as in https://github.com/pypa/cibuildwheel/issues/597#issuecomment-1035937707 there is no way I can know ahead of time if a built wheel will be pure python or not.

pombredanne:
> short inspection of setup.py will inform if there is any compiled code. Or simply run a local build of the wheel and check the output file.

that's easy for one, hard for 100, impossible for a 1000.

pombredanne:
I pushed a minimal implementation with a test. It works with a new --pure-wheel option and keeps on trucking rather than failing in this case. See https://github.com/pypa/cibuildwheel/pull/1022

Czaki:
> > short inspection of setup.py will inform if there is any compiled code. Or simply run a local build of the wheel and check the output file.
> 
> that's easy for one, hard for 100, impossible for a 1000.

Writle simple shell script which will call `tar -xvf sdist` (or equivalent for zip). Then `python -m build package_dir` and check if there is wheele in dist (if not then there is need of manual intervention) and if there is system name in wheel (linux, win, macosx depending on host system) and you will got information if you need to use cibuildwheel. 

And if You have such script then you could even process 10 000 packages

> I pushed a minimal implementation with a test. It works with a new --pure-wheel option and keeps on trucking rather than failing in this case

To use such flag you need first determine if package is pure python. What is advantage of using cibuildwheel?

pombredanne:
@Czaki I appreciate the feedback and I know how to drive this alright ... See https://github.com/nexB/skeleton/tree/main/etc/scripts if you are interested
That said and FWIW you cannot determine if a wheel will be pure before you build it. Just building it and discaring it as it was done until now in cibuildwheel does not make sense. I am just keeping the built wheel optionally.

> To use such flag you need first determine if package is pure python. What is advantage of using cibuildwheel?

Not at all. cibuildwheel today ALWAYS build a wheel, pure or not. And then raise an Exception if the wheel was pure.

At scale you cannot know if a wheel will be pure unless you build it first as a given build  context could lead to build a plain wheel for some package that have optional native speedup extensions. 

Now,  are suggesting something like:
[...]

PR #1022, opened 2022-02-11
Optionally build a pure wheel

pombredanne:
Add a new --pure-wheel command line option to also build pure Python
wheel. If not selected, build will fail with a wheel that does not
contain platform-specific code once built.
Also add a new CIBW_PURE_WHEEL environment variable for this option.

Reference: https://github.com/pypa/cibuildwheel/issues/1021
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -107,6 +107,15 @@ def main() -> None:
         help="Enable pre-release Python versions if available.",
     )
 
+    parser.add_argument(
+        "--pure-wheel",
+        action="store_true",
+        help="""
+            Build pure Python wheel. If not selected, build will fail with a
+            wheel that does not contain platform-specific code.
+        """,
+    )
+
     args = parser.parse_args(namespace=CommandLineArguments())
 
     if args.platform != "auto":
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -229,10 +229,11 @@ def build_on_docker(
         docker.call(["rm", "-rf", repaired_wheel_dir])
         docker.call(["mkdir", "-p", repaired_wheel_dir])
 
-        if built_wheel.name.endswith("none-any.whl"):
+        is_pure_python = built_wheel.name.endswith("none-any.whl")
+        if not build_options.globals.pure_wheel and is_pure_python:
             raise NonPlatformWheelError()
 
-        if build_options.repair_command:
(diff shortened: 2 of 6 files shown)


henryiii:
Could we possibly detect the pure wheel immediately (before the repair!), and then error out with a unique error code? Then you could continue if you receive that specific error code?

henryiii:
So basically `raise NonPlatformWheelError()` would have a unique "error" code.

pombredanne:
> So basically `raise NonPlatformWheelError()` would have a unique "error" code.

@henryiii if you think this is not a good idea in the first place, do you still want to have something in place that would be guarded only by an environment variable then?

henryiii:
No, if there was a unique error, then you could detect it and just continue on if that's actually what you want to do. No need for a secret or hidden setting, or special logic on our end.

pombredanne:
> No, if there was a unique error, then you could detect it and just continue on if that's actually what you want to do. No need for a secret or hidden setting, or special logic on our end.

I am not entirely clear about what you want. Let me reformulate:
1. remove the --pure-wheel option. There will be no CLI option at all?
2. in the main build loop ("for config in platform_configs:"  as in https://github.com/pypa/cibuildwheel/blob/51b50db510b941bbbaf2ee1eb7d263f852095f36/cibuildwheel/linux.py#L135 ) we would accumulate some list of "errors" to  tracks when pure wheels are created (likely which ones),  
3. if there were pure wheels a exception would be raised at the end?

I am not sure I get how this design would work, at least without some CLI option or env var unless you are saying that building pure wheels would never go through the tests?

joerick:
Closing this for the reasons stated in #1021.

joerick:
My apologies @pombredanne. But we can't support every user request! There is a small chance that such functionality might find its way into the project as a side-effect of limited-api support, but my advice would be to find a workaround external to cibuildwheel. We could add a custom error code, to address your question above, one way to do this would be to catch NonPlatformWheelError at this [location](https://github.com/pypa/cibuildwheel/blob/ab13f7a9de7c78d5dc0b80120f5c3071470f7a50/cibuildwheel/__main__.py#L214) and handle it by printing the message and exiting the program with `sys.exit(3)`.

pombredanne:
@joerick You do not have to be sorry! I was about to close this anyway as it truly does not make sense to have it here.

PR #1024, opened 2022-02-11
fix: A type alias cannot be final

henryiii:
Addition of Python 3.10 to the checks exposed this.


mayeut:
I will assume the failing tests have nothing to do with the fix and merge this.

mayeut:
the failure for Circle CI indeed disappeared on main.
I opened #1025 for GitLab CI which is still there (and unrelated to this PR)

PR #1047, opened 2022-03-08
Use tomllib on Python 3.11

hauntsaninja:


henryiii:
Nice, this went in 5 hours ago! https://github.com/python/cpython/tree/main/Lib/tomllib

joerick (approved):
Great!

henryiii:
FYI, I'm thinking that merging these as soon as the next Python 3.11 alpha is released would be a good idea. For https://github.com/pypa/build/pull/443, it matters, since that is testing against the latest alpha. Here, it doesn't really matter too much (using 3.11 alpha for host seems a little overkill), but might as well do them at the same time, and support the latest alpha, I think.

henryiii:
Waiting on https://github.com/actions/python-versions/pull/155 to merge and deploy.

henryiii:
It's in and up, so don't see a reason to delay this. Let's merge!

PR #1060, opened 2022-03-22
fix: minor touchups based on stricter mypy settings

henryiii:
Using some stricter (and at least one new in 0.940) setting for mypy.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
   hooks:
   - id: pycln
     args: [--all]
-    additional_dependencies: [click<8.1]  # temporary workaround until typer updates
+    stages: [manual]
 
 - repo: https://github.com/PyCQA/isort
   rev: 5.10.1
@@ -54,7 +54,7 @@ repos:
   - id: mypy
     name: mypy 3.6 on cibuildwheel/
     exclude: ^(bin|cibuildwheel/resources|docs)/.*py$
-    args: ["--python-version=3.6", "--show-error-codes"]
+    args: ["--python-version=3.6"]
     additional_dependencies: &mypy-dependencies
       - nox
       - packaging>=21.0
@@ -72,11 +72,11 @@ repos:
   - id: mypy
     name: mypy 3.7+ on bin/
     files: ^((bin|docs)/.*py)$
-    args: ["--python-version=3.7", "--show-error-codes"]
+    args: ["--python-version=3.7"]
     additional_dependencies: *mypy-dependencies
   - id: mypy
     name: mypy 3.10
-    args: ["--python-version=3.10", "--show-error-codes"]
+    args: ["--python-version=3.10"]
     additional_dependencies: *mypy-dependencies
 
 - repo: https://github.com/asottile/yesqa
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -30,7 +30,8 @@ repos:
   rev: v1.2.5
   hooks:
   - id: pycln
-    args: [--config=pyproject.toml]
+    args: [--all]
+    stages: [manual]
```
henryiii, line 34: Moving this to manual, since it's still very handy, but can get in the way.

.pre-commit-config.yaml
```diff
@@ -97,8 +98,6 @@ repos:
   hooks:
   - id: python-check-blanket-noqa
     stages: [manual]
-  - id: python-check-blanket-type-ignore
-    stages: [manual]
```
henryiii, line 101: In MyPy 0.940, this is now an option, and using MyPy, it not only tells you that you need an error code, it suggests the correct one! :)

cibuildwheel/options.py
```diff
@@ -96,7 +96,7 @@ def architectures(self) -> Set[Architecture]:
         return self.globals.architectures
 
 
-Setting = Union[Dict[str, str], List[str], str]
+Setting = Union[Dict[str, str], List[str], str, int]
```
henryiii, line 99: Discovered by the unreachable check.

cibuildwheel/options.py
```diff
@@ -496,7 +496,7 @@ def build_options(self, identifier: Optional[str]) -> BuildOptions:
 
                     config_value = self.reader.get(f"musllinux-{build_platform}-image")
 
-                    if config_value is None:
+                    if not config_value:
```
henryiii, line 499: @mayeut The original version should be impossible to reach - `self.reader.get` should always return strings, not `None`, right? If it can return None, then we need some significant fixes to the typing in other places.

Discovered by unreachability.

henryiii: (Going to mark this draft until verified)

mayeut: > The original version should be impossible to reach

Seems right. Not sure why it wasn't checked like manylinux here.

mayeut: There's another difference between musllinux & manylinux.
manylinux has:
```
if not config_value:
    # default to manylinux2014
    image = pinned_images.get("manylinux2014")
elif ....
...
assert image is not None
```

while musllinux has (after the change):
```
if not config_value:
    image = pinned_images["musllinux_1_1"]
elif ....
....
```

Not sure why `get` is used for manylinux (which is the only reason for `assert image is not None`). It would be nicer to use `image = pinned_images["manylinux2014"]` no ?

I've not been through the entire history so there might be some historic reasons for this being written this way but might not make sense anymore.

joerick: I'd guess it's a historical thing, perhaps from the days that manylinux2010 was the default on some platforms and manylinux2014 on others.

> It would be nicer to use `image = pinned_images["manylinux2014"]` no ?

Agreed.

cibuildwheel/options.py
joerick: Nice.

joerick (approved):

henryiii:
Actually, I'll leave auto merge off, so we can merge all three ready PRs in a row to reduce GHA usage.

PR #1061, opened 2022-03-24
Change github.event.ref to github.ref

kjelljorner:
According to the [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables), the variable `GITHUB_REF` which is accessed here via `github.ref` has the value `refs/tags/<tag_name>` for tags. `github.event.ref` just contains `<tag_name>` (as described [here](https://stackoverflow.com/a/59349906)).

An alternative edit is of course 
```yaml
if: github.event_name == 'push' && startsWith(github.event.ref, 'v')
```

### DIFF
--- a/examples/github-deploy.yml
+++ b/examples/github-deploy.yml
@@ -45,7 +45,7 @@ jobs:
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
     # upload to PyPI on every tag starting with 'v'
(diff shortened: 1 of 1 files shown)


joerick (approved):

joerick:
thanks!

kjelljorner:
No problem, keep up the good work with cibuildwheel, it helps me a lot 😄

PR #1065, opened 2022-03-26
docs: add note about opting into PEP 621

### LINKED ISSUE #1064: [Docs] Example may accidentally be encouraging users to write invalid `pyproject.toml` files
### Description

Hi guys, first of all thank you for the amazing project (always good to remember).

While I was working on adding support for PEP 621 to setuptools, I noticed a series of users having problems with invalid `pyproject.toml` files. The common pattern seem to be a almost empty `[project]` table with only a `requires-python` field set, which is invalid according to PEP 621.

It took me some time to find the reason for this behaviour but I think it comes from:
https://cibuildwheel.readthedocs.io/en/stable/options/#requires-python

I suspect that the example and notes about the preferred way of setting the config has been encouraging users that are unaware of PEP 621 to write technically invalid `pyproject.toml` files.

Please note that this issue is not necessarily related to setuptools itself.
The existence of the `[project]` table in the `pyproject.toml` allows (according to the standard) the installer/builder/consumer program to treat the package differently (specially regarding `dynamic`).

I think it would be nice to at least add a note about this so users became aware of the implications of adding a `[project]` table.

### Build log

_No response_

### CI config

_No response_

abravalheri:
Closes #1064

This is just a quick suggestion covering the bare minimum without touching the other parts of the doc.

You guys might want to do it in a completely different way. In that case, please feel free to ignore the PR.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -402,7 +402,7 @@ This option can also be set using the [command-line option](#command-line)
 > Manually set the Python compatibility of your project
 
 By default, cibuildwheel reads your package's Python compatibility from
-`pyproject.toml` following [PEP621](https://www.python.org/dev/peps/pep-0621/)
+`pyproject.toml` following the [project metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/)
 or from `setup.cfg`; finally it will try to inspect the AST of `setup.py` for a
 simple keyword assignment in a top level function call. If you need to override
 this behaviour for some reason, you can use this option.
@@ -416,43 +416,26 @@ Default: reads your package's Python compatibility from `pyproject.toml`
 `setup.py` `setup(python_requires="...")`. If not found, cibuildwheel assumes
 the package is compatible with all versions of Python that it can build.
 
-
 !!! note
-    Rather than using this option, it's recommended you set
-    `project.requires-python` in `pyproject.toml` instead:
+    Rather than using this option, it's recommended you set this value
+    statically in a way that your build backend can use it, too. This ensures
+    that your package's metadata is correct when published on PyPI.
 
-    ```toml
-    [project]
-    requires-python = ">=3.6"
-    ```
-
-    A few things to note:
-
-      - If your project didn't already have a `pyproject.toml`, be sure to
-        set the build system to your build backend.
+      - If you have a `pyproject.toml` containing a `[project]` table, you can
+        specify `requires-python` there.
 
         ```toml
-        [build-system]
-        requires = ["setuptools>=42", "wheel"]
-        build-backend = "setuptools.build_meta"
+        [project]
+        requires-python = ">=3.6"
         ```
 
-      - If you didn't already have a `pyproject.toml` that had a `[project]`
-        table, you should migrate values from `setup.py` or `setup.cfg`, or
-        list them in the [the `dynamic` field](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic).
-        The values to include and the format is listed in the
-        [project source metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic)
-        (as originally defined in PEP 621).
+        Note that build backend support for the `[project]` table is still patchy, and
+        adding `[project]` to `pyproject.toml` can change the behaviour of your build
+        (e.g. setuptools may ignore `install_requires` specified via `setup.py` or
+        `setup.cfg`). Make sure to double-check the build after adding.
 
-      - Adding `[project]` to `pyproject.toml` can
-        change the behaviour of your build (e.g. `setuptools` may ignore
-        `install_requires` specified via `setup.py` or `setup.cfg`). Make sure
-        to double-check the build after adding.
-
-This option is not available in `pyproject.toml` under
-`tool.cibuildwheel.project-requires-python`, since it should be set with the
-[PEP621](https://www.python.org/dev/peps/pep-0621/) location instead,
-`project.requires-python`.
+      - If you're using setuptools, [you can set this value in `setup.cfg` or `setup.py`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement)
+        and cibuildwheel will read it from there.
 
 #### Examples
 
--- a/docs/options.md
+++ b/docs/options.md
@@ -420,39 +420,33 @@ the package is compatible with all versions of Python that it can build.
 !!! note
     Rather than using this option, it's recommended you set
     `project.requires-python` in `pyproject.toml` instead:
-    Example `pyproject.toml`:
 
-        [project]
-        requires-python = ">=3.6"
+    ```toml
+    [project]
+    requires-python = ">=3.6"
+    ```
 
-        # Aside - in pyproject.toml you should always specify minimal build
-        # system options, like this:
+    A few things to note:
 
+      - If your project didn't already have a `pyproject.toml`, be sure to
+        set the build system to your build backend.
+
+        ```toml
         [build-system]
         requires = ["setuptools>=42", "wheel"]
         build-backend = "setuptools.build_meta"
+        ```
 
-    Note however, that adding a `[project]` table to your `pyproject.toml` may
-    trigger a series of validations as specified in
-    [PEP621](https://www.python.org/dev/peps/pep-0621/). Also consider that
-    adding `[project]`, if not done carefully, will change the behaviour of your
-    build (e.g. `setuptools` may ignore `install_requires` specified via
-    `setup.py` or `setup.cfg`).
+      - If you didn't already have a `pyproject.toml` that had a `[project]`
+        table, you should migrate values from `setup.py` or `setup.cfg`, or
+        [list them in the the `dynamic` field](https://peps.python.org/pep-0621/#dynamic).
+        The values to include and the format is listed in
+        [PEP 621](https://peps.python.org/pep-0621/#details)
 
-    Currently, setuptools has not yet added support for reading this value from
-    pyproject.toml yet, and so does not copy it to Requires-Python in the wheel
-    metadata. This mechanism is used by pip to scan through older versions of
-    your package until it finds a release compatible with the current version
-    of Python compatible when installing, so it is an important value to set if
-    you plan to drop support for a version of Python in the future.
-
-    If you don't want to list this value twice, you can also use the setuptools
-    specific location in `setup.cfg` and cibuildwheel will detect it from
-    there. Example `setup.cfg`:
-
-        [options]
-        python_requires = ">=3.6"
-
+      - Adding `[project]` to `pyproject.toml` can
+        change the behaviour of your build (e.g. `setuptools` may ignore
+        `install_requires` specified via `setup.py` or `setup.cfg`). Make sure
+        to double-check the build after adding.
 
 This option is not available in `pyproject.toml` under
 `tool.cibuildwheel.project-requires-python`, since it should be set with the
--- a/docs/options.md
+++ b/docs/options.md
@@ -439,9 +439,10 @@ the package is compatible with all versions of Python that it can build.
 
       - If you didn't already have a `pyproject.toml` that had a `[project]`
         table, you should migrate values from `setup.py` or `setup.cfg`, or
-        [list them in the the `dynamic` field](https://peps.python.org/pep-0621/#dynamic).
-        The values to include and the format is listed in
-        [PEP 621](https://peps.python.org/pep-0621/#details)
+        list them in the [the `dynamic` field](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic).
+        The values to include and the format is listed in the
+        [project source metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic)
+        (as originally defined in PEP 621).
 
       - Adding `[project]` to `pyproject.toml` can
         change the behaviour of your build (e.g. `setuptools` may ignore
--- a/docs/options.md
+++ b/docs/options.md
@@ -432,6 +432,12 @@ the package is compatible with all versions of Python that it can build.
         requires = ["setuptools>=42", "wheel"]
         build-backend = "setuptools.build_meta"
 
+    Note however, that adding a `[project]` table to your `pyproject.toml` may
+    trigger a series of validations as specified in
+    [PEP621](https://www.python.org/dev/peps/pep-0621/). Also consider that
+    adding `[project]`, if not done carefully, will change the behaviour of your
+    build (e.g. `setuptools` may ignore `install_requires` specified via
+    `setup.py` or `setup.cfg`).
 
     Currently, setuptools has not yet added support for reading this value from
     pyproject.toml yet, and so does not copy it to Requires-Python in the wheel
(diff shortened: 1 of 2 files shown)


joerick:
Thanks again for the PR. On this change, what would the minimum fields be that we could show in this example? It would be nice to just include the minimum, that's kinda what the example was trying to do before (hence why it includes the `[build-system]` section. 

I had a read of the PEP, is the minimum just `name = "..."`? or does it also need `version = "..."`? Of course we'd also link to the PEP with the full list, but a minimal example is always good to give users a starting point.

abravalheri:
The PEP requires all the projects to specify `name` statically and `version` either statically or dynamically.

However the PEP also forbids setuptools from backfilling using external sources (e.g. `setup.py`) the values of any attribute that is not listed under `dynamic`. This may be difficult to deal with.

abravalheri:
Version `v61.2.1` of setuptools will use the values of `setup.py` anyway when nothing is found in `pyproject.toml`. However there will be a bunch of warnings. I added this provisional behaviour just now, so we don't break the builds while users are getting used to he consequences of PEP 621.

joerick:
I had a go at making a bigger improvement to this section. If you have time, it would be great if you could let me know what you think. Probably easiest to read the rendered output - https://cibuildwheel--1065.org.readthedocs.build/en/1065/options/#requires-python

docs/options.md
```diff
@@ -420,33 +420,33 @@ the package is compatible with all versions of Python that it can build.
...
-    specific location in `setup.cfg` and cibuildwheel will detect it from
-    there. Example `setup.cfg`:
-
-        [options]
-        python_requires = ">=3.6"
-
+        ```
+
+      - If you didn't already have a `pyproject.toml` that had a `[project]`
+        table, you should migrate values from `setup.py` or `setup.cfg`, or
```
abravalheri, line 441: > you should migrate

My only remark is that support for PEP 621 was just added last Friday, so there might be sharp edges left.

---

The following is just maintainer babbling 😅:

We can actively start encouraging people to migrate to it, however in setuptools we decided to keep the `experimental` status for a while to guarantee that we have the freedom to modify how we implement it before committing to an specific design.

Changes in interface will not occur in the `[project]` table (after all it is standardized), but if the users reach out for `[tool.setuptools]` there is no guarantee the fields there will stay unchanged between versions.

If in the following days we are reported that the releases seriously break important packages for the community we might also consider yanking it and backing off a bit while we re-implement any problematic bits.

abravalheri (commented):
Hi @joerick thank you very much for the changes.
In general it does look good to me.

docs/options.md
joerick: Thanks for reviewing !

> My only remark is that support for PEP 621 was just added last Friday, so there might be sharp edges left.

Hm. Yes okay I see. So what's the conservative route for users with a `[project]` table? Would listing options in `dynamic` be the safest way to go?

abravalheri: If the users specify as dynamic the fields that they are providing via `setup.py`, that would be safe.

I think the parts of PEP 621 important to keep in mind are:

> Build back-ends MUST raise an error if the metadata specifies a field in dynamic but the build back-end was unable to provide the data for it.

> Build back-ends MUST raise an error if the metadata specifies a field statically as well as being listed in dynamic.

> If the metadata does not list a field in dynamic, then a build back-end CANNOT fill in

deepika094:
@abravalheri  so if i had pyproject.toml file with project tag where  i was just defining name and python-requires and for version i had a dynamic logic in setup.py , now setuptools will complain -
ERROR:setuptools.config.pyprojecttoml:configuration error: `project` must contain ['version'] properties

[06:26:51]	[Step 3/5]     ############################
[06:26:51]	[Step 3/5]     # Invalid `pyproject.toml` #
[06:26:51]	[Step 3/5]     ############################
[06:26:51]	[Step 3/5] 
[06:26:51]	[Step 3/5]     Any configurations in `pyproject.toml` will be ignored.
[06:26:51]	[Step 3/5]     Please note that future releases of setuptools will halt the build process
[06:26:51]	[Step 3/5]     if an invalid file is given.
[06:26:51]	[Step 3/5] 
[06:26:51]	[Step 3/5]     To prevent setuptools from considering `pyproject.toml` please
[06:26:51]	[Step 3/5]     DO NOT include the `[project]` or `[tool.setuptools]` tables in your file.


To fix this i should change it to below?-

[project] 
name = "***"
**dynamic = ["version"]**
requires-python = ">=3.7" #Python version

abravalheri:
Hi @deepika094, yes that change makes sense.

If you are using any other field in `setup.py` that has a correspondence in `pyproject.toml [project]`, you should include them in `dynamic` as well[^1].

For example let's say you define `version`, `install_requires` and `extras_require` in `setup.py`.
Then you should have `dynamic = ["version", "dependencies", "optional-dependencies"]`.
Please note that if you define something in `dynamic` that cannot be found by setuptools, there might also been an error.

These principles are defined in the PEP 621:

> Build back-ends MUST raise an error if the metadata specifies a field in dynamic but the build back-end was unable to provide the data for it.

> If the metadata does not list a field in dynamic, then a build back-end CANNOT fill in

Please note that this maintenance overhead is a direct consequence of opting into PEP 621.
If that is too for the project, the alternative is to remove `[project]` from `pyproject.toml` and rely on the other mechanisms listed in the `cibuildwheel` documentation to specify which versions of Python are supported.

[^1]: [PEP 621](https://peps.python.org/pep-0621/) lists the corresponding fields between setuptools and `pyproject.toml`.

deepika094:
@abravalheri  thank you. that's very  helpful. Indeed we define lots of fields in setup.py like install_requires, extra_requires, entry-points so I guess we will have to make change in all the projects and make these fields dynamic in pyproject.toml.  
@abravalheri  @joerick  If we look for alternative and we remove [project] from pyproject.toml  and for name , version and python-requires we keep all these just in **setup.py** 
eg.
**Before-**
setup(
    name=pyproject_toml['project']['name'],
    version=_determine_version(),
    python_requires=pyproject_toml['project']['requires-python'],
....

**After-**
setup(
    name='myproject_name,
    version=_determine_version(),
    python_requires= ">=3.7",

will that be ok and compliant with pep621? in that case we can just keep pyproject.toml to have below and not [project]-
[build-system]
requires = [
    "setuptools >= 46.1.0",
    "wheel",
    "toml"
]

Or the recommended way is to use [project] and for fields we define in setup.py we should mark them as dynamic in pyproject.toml?

CAM-Gerlach:
Personally, while I'm a big fan of PEP 621 (and am thrilled by @abravalheri 's work there) and look forward to all tools supporting it, IMO `cibuildwheel` strongly pushing users to use it without a full, in-depth explanation of the consequences, caveats and complexities. seems a bit premature—experimental Setuptools support _just_ landed, and Poetry still doesn't support it with no clear timeline for when it will be added—whereas there are serious implications users need to understand before they switch to it or they are in for a fair amount of trouble (as was the case for Black when they required `pyproject.toml` for configuration with no alternative, which triggered PEP 517 builds on unprepared projects).

Furthermore, it means that `cibuildwheel` has to independently keep up with the rapidly changing real-world situation thereof as tools implement and iterate on it, or otherwise risk giving users bad advice, which seems to be a non-trivial mantience burden for a package not specifically focused on this part of the ecosystem (project source metadata).

Since this is just about adding a single, simple metadata key as a way for `cibuildwheel` to better infer the Python versions supported by a project, I'd suggest not bringing in the whole big issue of PEP 621 migration into it, at least for now, and instead just list the different mechanisms in order of preference, mentioning that users should use the highest one they are currently using, and linking to the authoritative documentation and specifications for more detail on each.

Also, on another note, now that it has been accepted, PEP 621 is *not* the canonical specification or its documentation and thus should *not* be linked to in that capacity, as opposed to the [official specification on the packaging guide](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/).

docs/options.md
```diff
@@ -420,33 +420,33 @@ the package is compatible with all versions of Python that it can build.
...
+        [list them in the the `dynamic` field](https://peps.python.org/pep-0621/#dynamic).
+        The values to include and the format is listed in
+        [PEP 621](https://peps.python.org/pep-0621/#details)
```
CAM-Gerlach, line 444:
```suggestion
        list them in the [the `dynamic` field](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic).
        The values to include and the format is listed in the
        [project source metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dynamic)
        (as originally defined in PEP 621).
```

henryiii:
This is heavily my fault - I added this (about a year ago, maybe more) with the understanding that the PEP 621 location (`project.requires-python`) was the canonical location for this information (as stated in the PEP), and I was hoping to avoid a proliferation of locations for this information. I hadn't really thought about someone only adding this one value.

When we later added `tool.cibuildwheel` support, I explicitly avoided adding this field, as it was supposed to come from the one of the other sources. It's really optional for cibuildwheel - it just helps cibuildwheel discover the correct minimum - it was very important when we were still supporting Python 2 to give Python 3 users a much nicer experience. (And still helps for Python 3.7+ or 3.8+ users who don't want to be bothered with our following of Manylinux's ubuntu-18.04-inspired support for 3.6+). But I think that's when the emphasis causing just this value to be filled came. I certainly was not thinking about consequences of just adding that one value. We never were pushing for full PEP 621 because there were no backends for it yet.

Something similar happened with metadata.license and setup.cfg, with projects filling in just that one field in setup.cfg. PEP 621 is not the same, though, in that it's all-or-nothing; as a benefit to tools wanting to statically infer information from pyproject.toml, it requires everything to either be specified or listed explicitly in `dynamic`. 

My thought would be (I think closely following @CAM-Gerlach's suggestion above):

* Try to remove the emphasis on the PEP 621 location, and just state that this is supposed to be discovered, and the environment variable is available if the discovery doesn't work.
* Point out PEP 621 backends that support building (currently setuptools (experimental) and masonpy, hopefully scikit-build in the near future).
* Fix the links to the PEP 621 docs.
* _Maybe_ support this being specified in `tool.cibuildwheel`. I'm still biased toward not allowing it, though.

docs/options.md
joerick: thanks!

CAM-Gerlach:
> Thanks for the input, all. I still think that adopting the project.requires-python field for this option was the right thing to do - it's an interoperability spec after all - but I totally get that for good, practical reasons, this has ended up a more complicated picture than we originally thought.

Just to be clear, supporting the `project.requires-python` key and giving it precedence over the others was definitely the right thing to do, and in fact a bit part of the intention of the standardized project table was to allow other tools to statically read the project source metadata from a common location without having to parse and extract it from each tool's bespoke configuration format. My comments were just about how the documentation was framed and worded around that in a way that eased the effort and issues surrounding it for both you and your users, which is sounds like we all pretty much agree on :+1:

docs/options.md
```diff
@@ -416,43 +416,26 @@ Default: reads your package's Python compatibility from `pyproject.toml`
...
-        set the build system to your build backend.
+      - If you have a `pyproject.toml` containing a `[project]` table, you can
+        specify `requires-python` there.
 
         ```toml
-        [build-system]
-        requires = ["setuptools>=42", "wheel"]
-        build-backend = "setuptools.build_meta"
+        [project]
+        requires-python = ">=3.6"
```
henryiii, line 429:
```suggestion
        ...
        requires-python = ">=3.6"
```

docs/options.md
```diff
@@ -416,43 +416,26 @@ Default: reads your package's Python compatibility from `pyproject.toml`
...
-        to double-check the build after adding.
-
-This option is not available in `pyproject.toml` under
-`tool.cibuildwheel.project-requires-python`, since it should be set with the
-[PEP621](https://www.python.org/dev/peps/pep-0621/) location instead,
-`project.requires-python`.
+        Note that build backend support for the `[project]` table is still patchy, and
+        adding `[project]` to `pyproject.toml` can change the behaviour of your build
+        (e.g. setuptools may ignore `install_requires` specified via `setup.py` or
+        `setup.cfg`). Make sure to double-check the build after adding.
```
henryiii, line 435:
```suggestion
        Note that not all build backends fully support using a `[project]` table yet;
        specifically setuptools just added experimental support in version 61.
        Adding `[project]` to `pyproject.toml` requires all the other supported
        values to be specified there, or to be listed in `dynamic`.
```

docs/options.md
```diff
@@ -416,43 +416,26 @@ Default: reads your package's Python compatibility from `pyproject.toml`
...
+
+      - If you're using setuptools, [you can set this value in `setup.cfg` or `setup.py`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement)
+        and cibuildwheel will read it from there.
```
henryiii, line 438:
```suggestion
      - If you're using setuptools, [you can set this value in `setup.cfg` (preferred) or `setup.py`](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement)
        and cibuildwheel will read it from there.
```

docs/options.md
```diff
@@ -416,43 +416,26 @@ Default: reads your package's Python compatibility from `pyproject.toml`
...
-    Rather than using this option, it's recommended you set
-    `project.requires-python` in `pyproject.toml` instead:
-
-    ```toml
-    [project]
-    requires-python = ">=3.6"
-    ```
+    Rather than using this option, it's recommended you set this value
+    statically in a way that your build backend can use it, too. This ensures
+    that your package's metadata is correct when published on PyPI.
```
henryiii, line 422:
```suggestion
    Rather than using this option, it's recommended you set this value
    statically in a way that your build backend can use it, too. This ensures
    that your package's metadata is correct when published on PyPI. This option
    is provided by cibuildwheel as an emergency override, and therefore is only
    available in environment variable form.
```

CAM-Gerlach:
```suggestion
    Rather than using this environment variable, it's recommended you set this value
    statically in a way that your build backend can use it, too. This ensures
    that your package's metadata is correct when published on PyPI. This override
    is provided by cibuildwheel only for emergencies, and therefore is only
    available in environment variable form.
```

I suggest a modest rewrite of the above suggestion to be more clear, avoid repetition and read more cleanly. (In particular, it confused me as to what exactly "this option" was referring to, since it was several paragraphs distant from the original reference, and the intervening ones had talked about the static methods of setting this value)

joerick: Thanks both. I'll take @CAM-Gerlach's version, but just tweak 'emergencies', as it seems a bit strong. 

```suggestion
    Rather than using this environment variable, it's recommended you set this value
    statically in a way that your build backend can use it, too. This ensures
    that your package's metadata is correct when published on PyPI. This 
    cibuildwheel-specific option is provided as an override, and therefore is only
    available in environment variable form.
```

docs/options.md
joerick: I'll take @henryiii's version, poetry isn't very relevant to wheel building, and it's the existence of the `[project]` table that invokes the change in behaviour.

joerick (approved):

docs/options.md
CAM-Gerlach: Sounds good; I wasn't too fond of that wording either to be honest.

PR #1072, opened 2022-04-03
Build wheels from sdist (separate entrypoint)

### LINKED ISSUE #597: Idea: build from SDists
Just an idea, but how about allowing the "source dir" to be (a possibly remote?) SDist? If you separate the build into a different repository, having a way to put a SDist path, or simply require a user to download an SDist then put it in as a path would negate most of the need for "astrofrog/autowheel", which doesn't seem to be maintained. Just a thought.

joerick:
Alternative to #1020. Fixes #597.

This provides a separate entrypoint for building wheels from sdist, rather than from a source tree.

I'm using a different entrypoint because the way that command line arguments are parsed are different. This is because cibuildwheel is run with working dir as the root of the sdist, not the current working dir. That is because the docker 'project' is the contents of the sdist, and paths in options (e.g. options within pyproject.toml) should be relative to the sdist root.

The command line options `--output-dir` and `--config-file` are the exceptions to this - they are options of cibuildwheel-from-sdist and thus resolved relative to the working dir.

Docs are still to come, but help output is:

```
$ cibuildwheel-from-sdist --help
usage: cibuildwheel-from-sdist [-h] [--output-dir OUTPUT_DIR]
                               [--config-file CONFIG_FILE]
                               package

Build wheels from an sdist archive.

Extracts the sdist to a temp dir and calls cibuildwheel on the
resulting package directory. Note that cibuildwheel will be
invoked with its working directory as the package directory, so
options aside from --output-dir and --config-file are relative to
the package directory.

positional arguments:
  package               Path to the sdist archive that you want wheels for.
                        Must be a tar.gz archive file.

optional arguments:
  -h, --help            show this help message and exit
  --output-dir OUTPUT_DIR
                        Destination folder for the wheels. Default:
                        wheelhouse.
  --config-file CONFIG_FILE
                        TOML config file. To refer to a file inside the sdist,
                        use the `{project}` or `{package}` placeholder. e.g.
                        `--config-file {project}/config/cibuildwheel.toml`
                        Default: "", meaning the pyproject.toml inside the
                        sdist, if it exists.

Any further arguments will be passed on to cibuildwheel.
```

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -28,6 +28,7 @@ def main() -> None:
     platform: PlatformName
 
     parser = argparse.ArgumentParser(
+        prog="cibuildwheel",
         description="Build wheels for all the platforms.",
         epilog="""
             Most options are supplied via environment variables or in
--- a/cibuildwheel/from_sdist.py
+++ b/cibuildwheel/from_sdist.py
@@ -22,7 +22,7 @@ def main() -> None:
             the package directory.
             """,
         ),
-        epilog="""Any further arguments will be passed on to cibuildwheel.""",
+        epilog="Any further arguments will be passed on to cibuildwheel.",
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
 
@@ -65,7 +65,7 @@ def main() -> None:
 
         temp_dir_contents = list(temp_dir.iterdir())
 
-        if len(temp_dir_contents) != 1 or not temp_dir_contents[0].is_dir:
+        if len(temp_dir_contents) != 1 or not temp_dir_contents[0].is_dir():
             exit("invalid sdist: didn't contain a single dir")
 
         project_dir = temp_dir_contents[0]
--- a/cibuildwheel/from_sdist.py
+++ b/cibuildwheel/from_sdist.py
@@ -11,6 +11,7 @@ from cibuildwheel.util import format_safe
 
 def main() -> None:
     parser = argparse.ArgumentParser(
+        prog="cibuildwheel-from-sdist",
         description=textwrap.dedent(
             """
             Build wheels from an sdist archive.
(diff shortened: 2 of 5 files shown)


henryiii:
I'm still not very fond of this at all, especially after seeing it. It is unavailable from the GitHub action, much harder to get from pipx run, a new command to learn, all for exactly no user benefit. It has exactly the same commands, and is exactly orthogonal in primary argument - you can only run this on SDists, you can only run the original command on folders. From what I understand, this is entirely to keep the original command simple? Adding a brand new command to maintain and test likely negates that benefit. And document - since Pip already takes folder or SDist, there's already a mental model that you can pass a folder or an SDist to a builder. If pypa/build adds SDist support, it will work the same way.

From a user perspective, they are passing in the exact same arguments, the API is identical. The only difference is some relative paths - but that's already a little murky, so I think just a little documentation would cover this just fine. Something like "if you pass an SDist, all paths except `--output-dir` and `--config-file` are relative to the temporary build folder".

Since there's no rush, can you leave this around for a week or two so I can try doing it without a new command in an alternative PR to see how bad it actually looks?

joerick:
> Since there's no rush, can you leave this around for a week or two so I can try doing it without a new command in an alternative PR to see how bad it actually looks?  

Sure. I did try that when reviewing #1020 which was my motivation for this approach. The problem I hit was changing the cwd to the expanded tar directory in-process meant that all the paths in options were wrong. At that point, not wanting to further complicate things, a separate process seemed a much more pragmatic solution. But maybe I missed something.

Another option to avoid a new command (though it might appear slightly perverse at first blush) would be for cibuildwheel to call /itself/ as a subprocess, once it's done the expansion of the sdist. That would avoid the extra entrypoint, and keep complexity down. But the complications around relative paths remain of course.

PR #569, opened 2021-01-30
Error on creation of wheel with a name that already exists, and add Py_LIMITED_API to the docs' FAQ

YannickJadoul:
Closes #78﻿, by providing clear error messages and documentation on how to use `cibuildwheel` to create wheels with Python's limited API/stable ABI.

TODO:
- [x] Show error on copying an already existing file to the output directory
- [ ] Update the FAQ with some information about Python's limited API.
- [x] Add tests (?)

PR #1091, opened 2022-04-18
feature: add support for ABI3 wheels

### LINKED ISSUE #78: Add support for Py_LIMITED_API ?

mayeut:
Closes #78

While #569 errors out on duplicate abi3 wheel with an updated error message, this PR skips compatible abi3 builds and do proper testing with all python versions as specified by the build selector.

real-world usage can be seen there: https://github.com/mayeut/psutil/actions/runs/2184123889


henryiii (approved):
This is much easier to review after checking "ignore whitespace" in GH. :) The extra indentation is a minor downside - I think we should see if we can pull out each step (build, test, etc) into functions, then call them - that will keep the whitespace from growing here, and keep the logic easier to read (if this, build this, if that, test that) instead of having a massive inline function, too. Not for this PR though.

cibuildwheel/util.py
```diff
@@ -562,6 +565,46 @@ def virtualenv(
...
+    ...
+
+
+@overload
+def find_abi3_wheel(wheels: Sequence[PurePath], identifier: str) -> Optional[PurePath]:
+    ...
+
+
+def find_abi3_wheel(
+    wheels: Sequence[Union[Path, PurePath]], identifier: str
```
henryiii, line 579: Couldn't you use `TypeVar("P", bound=PurePath)` instead of the overload? Maybe even `Traversable`, though I forget if you can bind to something structurally subclassing, I think you can.

mayeut: I was not very keen on what I came up with but couldn't find easily other ways to do that. I'm not very comfortable with  typing yet except for basic stuff but I'll try to look into your proposals over the week-end. If you have something you want to commit in the meantime, please do.

test/test_limited_api.py
```diff
@@ -0,0 +1,50 @@
...
+                    return python, "abi3", plat
+
+            cmdclass["bdist_wheel"] = bdist_wheel_abi3
+            extension_kwargs["define_macros"] = [("Py_LIMITED_API", "0x03080000")]
+            extension_kwargs["py_limited_api"] = True
+    """
+    ),
+    setup_py_extension_args_add="**extension_kwargs",
+    setup_py_setup_args_add="cmdclass=cmdclass",
+)
```
joerick, line 29: I suppose this is the way to get abi3 wheel from `pip wheel`/`python -m build`? Because we can't pass `--py-limited-api`? I wonder if there's a cleaner way to do it. I suppose we'd want to document this too.

joerick: A follow-up on this question - should a reference version of this bdist_wheel specify in the tag that it's a "3.8-ABI" wheel, rather than using whatever it's being built by?

cibuildwheel/util.py
```diff
@@ -562,6 +565,46 @@ def virtualenv(
...
+def find_abi3_wheel(wheels: Sequence[Path], identifier: str) -> Optional[Path]:
+    ...
+
+
+@overload
+def find_abi3_wheel(wheels: Sequence[PurePath], identifier: str) -> Optional[PurePath]:
+    ...
+
+
+def find_abi3_wheel(
```
joerick, line 578: Minor suggestion: `find_compatible_abi3_wheel(identifier, wheels)` might be clearer?

cibuildwheel/util.py
```diff
@@ -562,6 +565,46 @@ def virtualenv(
...
+    wheels: Sequence[Union[Path, PurePath]], identifier: str
+) -> Optional[Union[Path, PurePath]]:
```
joerick, line 580: It would be good to add a docstring here detailing what it's doing. e.g. 

```
'''
Finds an ABI3 wheel in `wheels` compatible with the Python
interpreter specified by `identifier`. 
'''
```

cibuildwheel/linux.py
```diff
@@ -174,74 +177,79 @@ def build_on_docker(
...
-                    container_package_dir,
-                    "--wheel",
-                    f"--outdir={built_wheel_dir}",
-                    f"--config-setting={config_setting}",
-                ],
-                env=env,
-            )
+        abi3_wheel = find_abi3_wheel(built_wheels, config.identifier)
+        if abi3_wheel:
+            repaired_wheels = [abi3_wheel]
```
joerick, line 182: It would be nice to add a log statement here, "Using previously built wheel filename.whl". You might have to close the previous log step first, otherwise it'll end up in the _Installing build tools_ step.

(same for the other platforms)

joerick (commented):
Great stuff @mayeut. Thank you for this, it's been on my todo list for a while!

I enjoy how pragmatic it is, simpler than I'd have thought to do it.

- The normal operation of cibuildwheel is to build from oldest Python to newest. Is that what we want in this case? I.e. the abi wheel will be built on the oldest Python that supports limited api.
- Documentation - I suppose we could add a FAQ entry about limited API, and include how to do that within setup.py, since we don't allow passing the `--py_limited_api` option.

test/test_limited_api.py
henryiii: Most packages will hard code this into setup.py. A package either produces abi3 wheels or regular wheels, usually they don't select between the two, I think. So you'd hardcode in the define in the source code, and `py_limited_api=True` is passed to `setup()`.

cibuildwheel/util.py
henryiii: I tried what I was thinking in ba9a91e.

mayeut: Thanks @henryiii, looks much better this way and I learned something.

mayeut: Thanks @henryiii, looks much better this way and I learned something.

mayeut:
> The normal operation of cibuildwheel is to build from oldest Python to newest. Is that what we want in this case? I.e. the abi wheel will be built on the oldest Python that supports limited api.

I think that's what we want in this case. This way, when a new version of CPython is available, it'll be tested but it won't change the build so I'd say it allows more reproducibility in the long run.

>  Documentation - I suppose we could add a FAQ entry about limited API, and include how to do that within setup.py, since we don't allow passing the --py_limited_api option.

I really don't think it belongs in cibuildwheel. IMHO, it belongs in https://packaging.python.org/en/latest/ and cibuildwheel should reference an entry there but it does not exist yet.

cibuildwheel/util.py
mayeut: will do

test/test_limited_api.py
mayeut: I tested something a bit more complicated than what most packages will do. As @henryiii mentioned, and I think so to, "a package either produces abi3 wheels or regular wheels, usually they don't select between the two" and it'll be hardcoded in the source code.
Even for some packages that still support cp27 or some that depends on an API being made stable in a version more recent than the minimum supported, the logic will probably hardcoded.
I really don't think it's cibuildwheel's place to document how to build an `abi3` wheel. Only if workarounds were to arise, specific to cibuildwheel, those should be documented then.

cibuildwheel/util.py
mayeut: done

cibuildwheel/linux.py
mayeut: done

cibuildwheel/linux.py
```diff
@@ -174,74 +177,81 @@ def build_on_docker(
...
-                    "--wheel",
-                    f"--outdir={built_wheel_dir}",
-                    f"--config-setting={config_setting}",
-                ],
-                env=env,
-            )
+        abi3_wheel = find_compatible_abi3_wheel(built_wheels, config.identifier)
+        if abi3_wheel:
+            log.step_end()
+            print(f"SKIP building wheel. Using previously built wheel {abi3_wheel.name}")
```
joerick, line 183: I'm thinking there might be a way to rephrase that doesn't have the potential to confuse with CIBW_SKIP here...

```suggestion
            print(f"Found previously built wheel {abi3_wheel.name}, that's compatible with {config.identifier}. Skipping build step...")
```

cibuildwheel/macos.py
```diff
@@ -318,84 +321,92 @@ def build(options: Options, tmp_path: Path) -> None:
...
-                    build_options.package_dir,
-                    "--wheel",
-                    f"--outdir={built_wheel_dir}",
-                    f"--config-setting={config_setting}",
-                    env=build_env,
-                )
+            abi3_wheel = find_compatible_abi3_wheel(built_wheels, config.identifier)
+            if abi3_wheel:
+                log.step_end()
+                print(f"SKIP building wheel. Using previously built wheel {abi3_wheel.name}")
```
joerick, line 327:
```suggestion
            print(f"Found previously built wheel {abi3_wheel.name}, that's compatible with {config.identifier}. Skipping build step...")
```

cibuildwheel/windows.py
```diff
@@ -274,83 +277,91 @@ def build(options: Options, tmp_path: Path) -> None:
...
-            if build_options.before_build:
-                log.step("Running before_build...")
-                before_build_prepared = prepare_command(
-                    build_options.before_build, project=".", package=options.globals.package_dir
-                )
-                shell(before_build_prepared, env=env)
+            abi3_wheel = find_compatible_abi3_wheel(built_wheels, config.identifier)
+            if abi3_wheel:
+                log.step_end()
+                print(f"SKIP building wheel. Using previously built wheel {abi3_wheel.name}")
```
joerick, line 283:
```suggestion
            print(f"Found previously built wheel {abi3_wheel.name}, that's compatible with {config.identifier}. Skipping build step...")
```

test/test_limited_api.py
joerick: > I tested something a bit more complicated than what most packages will do.

Understood.

> I really don't think it's cibuildwheel's place to document how to build an abi3 wheel. Only if workarounds were to arise, specific to cibuildwheel, those should be documented then.

The only thing that might be considered specific is the inability to pass `--py-limited-api` to bdist_wheel - most docs online still reference this. But since that calling method is deprecated, it does indeed make sense to document a more modern way across the ecosystem. Ideally, indeed, in packaging.python.org. Then we can link it from our docs.

cibuildwheel/macos.py
henryiii: Was the indent change intentional?

cibuildwheel/linux.py
mayeut: suggestion applied

cibuildwheel/macos.py
mayeut: suggestion applied

cibuildwheel/windows.py
mayeut: suggestion applied

henryiii:
LGTM

joerick:
I'm currently thinking how best to document this...

PR #1095, opened 2022-04-24
fix: extract project as current user on linux

mayeut:
This allows git to work properly, e.g., `versioneer`/`setuptools_scm` with latest git versions.
c.f. https://github.com/pypa/manylinux/issues/1309

e.g. for ninja which uses `versioneer` when built on `quay.io/pypa/manylinux2014_x86_64:latest`:
- `ninja-0+unknown-py2.py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` without the fix
- `ninja-1.10.2.3.post17.dev0+g3a8b005-py2.py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` with the fix

### DIFF
--- a/cibuildwheel/docker_container.py
+++ b/cibuildwheel/docker_container.py
@@ -122,7 +122,7 @@ class DockerContainer:
         if from_path.is_dir():
             self.call(["mkdir", "-p", to_path])
             subprocess.run(
-                f"tar cf - . | docker exec -i {self.name} tar -xC {shell_quote(to_path)} -f -",
+                f"tar cf - . | docker exec -i {self.name} tar --no-same-owner -xC {shell_quote(to_path)} -f -",
                 shell=True,
                 check=True,
                 cwd=from_path,


henryiii (approved):

henryiii:
The downstream projects can fix this too by specifying the git root explicitly, such as in https://github.com/pypa/setuptools_scm/pull/708. But this seems like a good fix, the ownership should not be based on the contents of the tar file.

henryiii:
This is a fix for main, or for using the latest manylinux manually? I’m assuming older pins in released versions have the older git and are fine?

mayeut:
> This is a fix for main, or for using the latest manylinux manually? I’m assuming older pins in released versions have the older git and are fine?

Pinned images in main are still using older git and are not affected (will be affected once updates kick-in).
Users that are specifying latest manylinux images or custom images are affected (the report in manylinux repo is a cibuildwheel user using custom image based on manylinux, or https://github.com/mayeut/patchelf-pypi/pull/16 with a custom image which is now failing).

joerick (approved):
For anyone curious:

```
     --no-same-owner
             (x mode only) Do not extract owner and group IDs.  This is the
             reverse of --same-owner and the default behavior if tar is run as
             non-root.
```

It seems to me very reasonable ignore ownership here 👍  I guess the UIDs don't even match up between the host and container.

PR #1096, opened 2022-04-27
feat: SDist builds (attempt 3)

henryiii:
This is a followup to both #1072 and #1020 to attempt to build from SDists. This avoids a new entry point, works better with pipx and GitHub Actions, and should be the same as #1072 conceptually.

- Add from_sdist script
- Add entrypoint
- Add test for cibuildwheel.from-sdist
- Add test for argument passthrough
- Tidy ups, bug fixes
- Add prog arguments to ArgumentParser to prevent the wrong inferred name
- refactor: use single entry for SDist builds


cibuildwheel/options.py
```diff
@@ -46,9 +48,9 @@
 class CommandLineArguments:
     platform: Literal["auto", "linux", "macos", "windows"]
     archs: Optional[str]
-    output_dir: Optional[str]
+    output_dir: Optional[Path]
```
joerick, line 51: I suspect we can remove the Optional, if we use the argparse default param.

cibuildwheel/__main__.py
```diff
@@ -109,6 +116,42 @@ def main() -> None:
 
     args = parser.parse_args(namespace=CommandLineArguments())
 
+    args.package_dir = args.package_dir.resolve()
+
+    # This are always relative to the base directory, even in SDist builds
+    args.output_dir = Path(
+        args.output_dir
+        if args.output_dir is not None
+        else os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse")
```
joerick, line 125: Might be easier to add this envvar fallback to the `parser.add_argument` call i.e. `default=os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse")`

unit_test/utils.py
```diff
@@ -8,8 +10,8 @@ def get_default_command_line_arguments() -> CommandLineArguments:
     defaults.allow_empty = False
     defaults.archs = None
     defaults.config_file = ""
-    defaults.output_dir = None
-    defaults.package_dir = "."
+    defaults.output_dir = Path("wheelhouse")  # This must be resolved from "None" before passing
```
joerick, line 13: Perhaps we won't need the comment if it's not really optional after all.

cibuildwheel/options.py
```diff
@@ -361,12 +363,9 @@ def package_requires_python_str(self) -> Optional[str]:
     @property
     def globals(self) -> GlobalOptions:
         args = self.command_line_arguments
-        package_dir = Path(args.package_dir)
-        output_dir = Path(
-            args.output_dir
-            if args.output_dir is not None
-            else os.environ.get("CIBW_OUTPUT_DIR", "wheelhouse")
-        )
+        assert args.output_dir is not None, "Must be resolved"
```
joerick, line 366: Possibly this can also be removed if it's not Optional.

test/test_from_sdist.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+
+    # fmt: off
+    assert process.stdout == textwrap.dedent(
+        """
+        cp38-manylinux_ppc64le
+        cp38-musllinux_ppc64le
+        """
+    ).lstrip()
+    # fmt: on
```
joerick, line 214: I think we can remove this test, now? It doesn't test anything now that the calling method is the same

```suggestion
```

cibuildwheel/__main__.py
```diff
@@ -65,27 +66,33 @@ def main() -> None:
...
-            TOML config file. Default: "", meaning {package}/pyproject.toml,
-            if it exists.
+            TOML config file. Default: "", meaning {package}/pyproject.toml, if
+            it exists. To refer to a project inside your project, use {package};
+            this matters if you build from an SDist.
         """,
     )
 
     parser.add_argument(
         "package_dir",
```
joerick, line 84: Could this argument be renamed to `package`? since it's not a dir when it's an sdist archive.

cibuildwheel/__main__.py
```diff
@@ -65,27 +66,33 @@ def main() -> None:
...
-            the working directory. When set, the working directory is still
-            considered the 'project' and is copied into the Docker container on
-            Linux. Default: the working directory.
+            Path to the package that you want wheels for. Must be a subdirectory
+            of the working directory. When set to a directory, the working
+            directory is still considered the 'project' and is copied into the
+            Docker container on Linux. Default: the working directory. This can
+            also be a tar.gz file - if it is, then --config-file and
+            --output-dir are relative to the current directory, and other paths
+            are relative to the expanded SDist directory.
```
joerick, line 95:
```suggestion
            Path to the package that you want wheels for. Default: the working
            directory. Can be a directory inside the working directory, or an
            sdist. When set to a directory, the working directory is still
            considered the 'project' and is copied into the Docker container
            on Linux.  When set to a tar.gz sdist file, --config-file
            and --output-dir are relative to the current directory, and other
            paths are relative to the expanded SDist directory.
```

joerick (commented):
You're right! This approach is much cleaner than the extra entrypoint. Let's hope it passes the tests. I've a few minor comments, too.

cibuildwheel/__main__.py
henryiii: I renamed it in the CLI, left it in the code because it quickly is resolved to a dir in both cases.

cibuildwheel/__main__.py
```diff
@@ -109,6 +118,38 @@ def main() -> None:
 
     args = parser.parse_args(namespace=CommandLineArguments())
 
+    args.package_dir = args.package_dir.resolve()
+
+    # This are always relative to the base directory, even in SDist builds
+    args.output_dir = args.output_dir.resolve()
+
+    # Standard builds if a directory or non-existent path is given
```
henryiii, line 126: Avoiding making a new directory if this ends with .tar.gz

joerick (approved):

nsoranzo:
Thanks @henryiii , at a first test this seems to work great!

nsoranzo:
I couldn't resist and created https://github.com/nsoranzo/wheelforge to use this new feature to replace https://github.com/galaxyproject/starforge / https://github.com/galaxyproject/starforge-recipes/

PR #1097, opened 2022-04-28
docs: add some projects

henryiii:
Updating docs with some new projects out of the 600+ projects using cibuildwheel. Based on the show and tell discussion and a sourcegraph search.


joerick:
Some big names in here!

PR #1098, opened 2022-04-28
feat: add --only

### LINKED ISSUE #1087: How do I build one specific wheel from the command line?
I'm running cibuildwheel from the command line on my local machine. I want to debug a build failure for the `cp39-musllinux_i686` wheel. But I can't figure out how to say "only build `cp39-musllinux_i686` not anything else; I don't have all day here".

I was expecting this to work:

`cibuildwheel --build-identifiers cp39-musllinux_i686`

but there is no such flag.

henryiii:
Closes #1087 by adding a `--only` arg. Closes #1262. By providing an alternative that only uses the wheel identifier.


joerick:
Hmm. I'm thinking more about this and I wonder if perhaps `--build` should override both CIBW_BUILD and CIBW_SKIP from environment/config. There seems something weird about the user doing explicitly `cibuildwheel --build 'pp38-manylinux_x86_64'` but that selects zero identifiers because there's a `CIBW_SKIP=pp*` in the environment. Perhaps `--build` should override both build & skip from environment and config. What do you think @henryiii ?

henryiii:
`--target=<...>` or `--local=<...>` or `--only=<...>` might be options? Given it is merging several things, not just BUILD, and selector expressions are not valid. Personally I'd often do `CIBW_BUILD=cp310-*` since I'm lazy and don't want to look up the spelling of the rest of the os/arch tag, but being able to skip `--platform` (and maybe `--arch`) would likely make it more worth while.

joerick:
`--only` makes sense to me!

cibuildwheel/options.py
```diff
@@ -388,6 +389,15 @@ def globals(self) -> GlobalOptions:
...
         requires_python = None if requires_python_str is None else SpecifierSet(requires_python_str)
 
+        archs_config_str = args.archs or self.reader.get("archs", sep=" ")
+        architectures = Architecture.parse_config(archs_config_str, platform=self.platform)
+
+        # Process `--only`
+        if args.only is not None:
+            build_config = args.only
+            skip_config = ""
+            architectures = Architecture.all_archs(self.platform)
```
joerick, line 399: i was wondering how this would work. neat! :)

docs/options.md
```diff
@@ -907,7 +911,7 @@ If this option is blank, it will fall though to the next available definition (e
...
 
-Auditwheel detects the version of the manylinux / musllinux standard in the image through the `AUDITWHEEL_PLAT` environment variable, as cibuildwheel has no way of detecting the correct `--plat` command line argument to pass to auditwheel for a custom image. If a custom image does not correctly set this `AUDITWHEEL_PLAT` environment variable, the `CIBW_ENVIRONMENT` option can be used to do so (e.g., `CIBW_ENVIRONMENT='AUDITWHEEL_PLAT="manylinux2010_$(uname -m)"'`).
+Auditwheel detects the version of the manylinux / musllinux standard in the image through the `AUDITWHEEL_PLAT` environment variable, as cibuildwheel has no way of detecting the correct `--platform` command line argument to pass to auditwheel for a custom image. If a custom image does not correctly set this `AUDITWHEEL_PLAT` environment variable, the `CIBW_ENVIRONMENT` option can be used to do so (e.g., `CIBW_ENVIRONMENT='AUDITWHEEL_PLAT="manylinux2010_$(uname -m)"'`).
```
joerick, line 914: Auditwheel's argument is indeed called `--plat`, unlike ours

https://github.com/pypa/auditwheel/blob/560509888bdb9cdfa98b41cc1641ec749b64b949/src/auditwheel/main_repair.py#L47

docs/options.md
```diff
@@ -101,7 +101,7 @@ The complete set of defaults for the current version of cibuildwheel are shown b
 
 !!! tip
     Static configuration works across all CI systems, and can be used locally if
-    you run `cibuildwheel --plat linux`. This is preferred, but environment
+    you run `cibuildwheel --platform linux`. This is preferred, but environment
```
joerick, line 104: hm. good spot.

unit_test/main_tests/main_platform_test.py
```diff
@@ -191,3 +197,64 @@ def test_archs_platform_all(platform, intercepted_build_args, monkeypatch):
...
+    assert options.globals.build_selector.build_config == only
+    assert options.globals.build_selector.skip_config == ""
+    assert options.platform == plat
+    assert options.globals.architectures == Architecture.all_archs(plat)
+
+
+@pytest.mark.parametrize("only", ("cp311-manylxinux_x86_64", "some_linux_thing"))
+def test_only_failed(intercepted_build_args, monkeypatch, only):
+    monkeypatch.setattr(sys, "argv", sys.argv + ["--only", only])
+    monkeypatch.delenv("CIBW_PLATFORM")
```
joerick, line 237: I'm curious why these delenvs are needed? Also on line 244, and 254.

joerick (commented):
Overall structure looks good!

unit_test/main_tests/main_platform_test.py
henryiii: Because `intercepted_build_args` uses `platform` which injects it. The point of this is to see if adding something like `--platform` or `--archs` fails or some broken `--only` is used, but instead it will fail because `CIBW_PLATFORM` is set & `--only` is present.

docs/options.md
henryiii: Oops, overdid it. :)

unit_test/main_tests/main_platform_test.py
joerick: I've removed the coupling of intercepted_build_args from platform - it now monkeypatches all three. I'll push that to this PR so you can have a look.

joerick:
I've just pushed a few commits. After removing the `intercepted_build_args` to `platform` coupling, I realised that CIBW_PLATFORM was causing an error when used with the `--only` command line option, which doesn't seem right to me (command line options should override env vars). So I wrote a test for that and got it passing.

Let me know if that works for you!

unit_test/main_tests/main_platform_test.py
```diff
@@ -10,6 +10,12 @@
 from ..conftest import MOCK_PACKAGE_DIR
 
 
+class ArgsInterceptor:
+    def __call__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+
```
joerick, line 18: Actually I think we can remove this now too

```suggestion
```

joerick:
The CI failures are the
> WARNING: Skipping page https://pypi.org/simple/xxx/ because the GET request got Content-Type: Unknown
ones.

Having seen more of this recently I investigated a little https://github.com/pypa/cibuildwheel/issues/1254

henryiii:
That's fine with me - IMO it should be really rare to have `CIBW_PLATFORM` & `--only` together, and it's probably a mistake on the user's part if they are both there, so I'd be fine with the error, but don't have a strong feeling on it.

cibuildwheel/__main__.py
```diff
@@ -149,10 +160,40 @@ def main() -> None:
...
+                "--arch cannot be specified with --only, it is computed from --only",
+                file=sys.stderr,
+            )
+            sys.exit(2)
+    elif platform_option_value != "auto":
+        if platform_option_value not in PLATFORMS:
+            print(f"cibuildwheel: Unsupported platform: {platform_option_value}", file=sys.stderr)
+            sys.exit(2)
+
+        platform = cast(PlatformName, platform_option_value)
```
henryiii, line 196: Personally I like `typing.cast` better than `cast` (and, in general, only importing classes from typing and using `typing.<thing>` for functions and constants). `cast()` especially is really general.

docs/options.md
henryiii: Technically, unless we turn off argparse's option shortcut feature, this would work. But it's better to use the actual arg name (and [personally not fond](https://github.com/CLIUtils/CLI11#features-not-supported-by-this-library) of that feature).

henryiii:
Added `only:` to the Action, since I think it would solve https://github.com/pypa/cibuildwheel/pull/1262.

PR #1099, opened 2022-04-28
refactor: use dataclasses instead of NamedTuples

henryiii:
Yesterday's MyPy 0.950 fixed a bug with dataclasses holding Callables (the one in the bashlex file), so this is now supported (via static typing, it worked before but failed a typecheck). NamedTuples should not be used for new API, they are intended to be a backwards compatibility thing (they add all of tuple syntax too). 

See https://twitter.com/brettsky/status/1516493479282503687?s=20&t=2i7N_WHyCCDZ7GeiCqmgOA and other places.


cibuildwheel/__main__.py
```diff
@@ -107,7 +107,7 @@ def main() -> None:
         help="Enable pre-release Python versions if available.",
     )
 
-    args = parser.parse_args(namespace=CommandLineArguments())
```
joerick, line 110: I guess this'll give us a runtime failure if the args don't match the object, which would help us catch mismatches.

joerick (approved):
Looks good!

joerick:
Hm. CI failed for some reason. Oh well, it's not critical to get this in before the release.

henryiii:
That's okay, no rush. I can't actually see why it failed ATM.

henryiii:
Seems to have been an CI infrastructure error. Rebuilding the failed job worked.

PR #1100, opened 2022-04-30
Add interactive diagram to the docs

### LINKED ISSUE #468: Ideas for improving the first-run experience

joerick:
Fix #468.

This adds a diagram to the docs that summarises what steps cibuildwheel takes when building a wheel. It's interactive, so mousing over a part will show more info. 

One thing I'm not sure about is where to put it in the docs. Currently it's on the 'home' page, it could have its own page. If it had its own page, I could increase the page width a bit too, which might let me increase the font size a bit.


joerick:
Here it is: https://cibuildwheel--1100.org.readthedocs.build/en/1100/#how-it-works

henryiii:
Would it make sense to have the png image as a no script fallback, since you already have it? (Or does everything fall apart without JS? Do people still browse without JS sometimes?)

henryiii:
This looks fantastic, by the way!

henryiii:
Was going ask about mobile! Noticed it didn’t handle mobile very well when showing it off at PyCon.

henryiii:
The place you press seems small for going to the link on the second press?

mayeut:
this looks very nice !

joerick:
> The place you press seems small for going to the link on the second press?

Yes, good point. I think I'll change the tap behaviour to be: 1st tap = open tooltip. 2nd tap = follow link to docs.

joerick:
> Would it make sense to have the png image as a no script fallback, since you already have it? (Or does everything fall apart without JS? Do people still browse without JS sometimes?)

Indeed our docs do use JS for a few things already. But they still basically work so the noscript fallback is a good idea 👍

henryiii (approved):
Thought I already did this.

henryiii:
Should we merge before release so it gets in the stable docs?

abitrolly:
Wow. That appeared to be very helpful. How it was made?

abitrolly:
@joerick I wanna reuse that. What are the conditions? :D

joerick:
Same as the rest of the project, BSD license! https://github.com/pypa/cibuildwheel/blob/main/LICENSE

Go for it :)

PR #1104, opened 2022-05-08
fix: update workflow

mayeut:
Use `wntrblm/nox` instead of `excitedleigh/setup-nox` to set-up `nox`.
There's no release yet so use the sha256 of the latest commit on main for now.


mayeut:
I cancelled travis CI jobs on this one.

.github/workflows/update-dependencies.yml
```diff
@@ -12,7 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: excitedleigh/setup-nox@v2.0.0
```
henryiii, line 15: FYI, the latest commit of `excitedleigh/setup-nox`, it's just that @excitedleigh has not made a release.

IMO using the nox action is fine, though.

mayeut: > FYI, the latest commit of `excitedleigh/setup-nox`, it's just that @excitedleigh has not made a release.

I've been using the latest commit of `excitedleigh/setup-nox` in the manylinux repo which was available before your PR was merged in `wntrblm/nox`

> IMO using the nox action is fine, though.

Given `wntrblm/nox` is now available as an action, maybe going with this is safer in the long run ? (given it's a composite action based on `setup-python@v3` it might be less likely to break)

.github/workflows/update-dependencies.yml
```diff
@@ -12,7 +12,10 @@ jobs:
...
+    - uses: actions/setup-python@v3
+      with:
+        python-version: "3.6"
+    - uses: wntrblm/nox@f12fb342cc72bbe6df00b14077b42f1e1d55d69a
```
henryiii, line 18:
```suggestion
    - uses: excitedleigh/setup-nox@v2.1.0
```

mayeut: yes, we'll see then.
For now I've merged #1110

PR #1113, opened 2022-05-22
action: use python 3.10 to run the GitHub Action

mayeut:
We can use setup-python in the composite GitHub Action & use that python to run cibuildwheel with pipx.
This allows to drop python 3.6 without losing support for ubuntu-18.04 runners c.f. #508.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,19 @@ jobs:
         name: sample_wheels
         path: wheelhouse
 
+    - name: Get some sample wheels (GitHub Action)
+      uses: ./
+      with:
+        package-dir: sample_proj
+        output-dir: wheelhouse_action
+      env:
+        CIBW_ARCHS_MACOS: x86_64 universal2 arm64
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+        package-dir: sample_proj
+        output-dir: wheelhouse_action
+      env:
+        CIBW_ARCHS_MACOS: x86_64 universal2 arm64
+
+    - uses: actions/upload-artifact@v3
+      with:
+        name: sample_wheels_action
+        path: wheelhouse_action
+
```
mayeut, line 81: There was no test for the GitHub Action so I duplicated the sample wheels steps to test the action.

mayeut:
On second thought, this might have too many side-effects (mainly, PATH is updated but never reverted by setup-python).

joerick:
> On second thought, this might have too many side-effects (mainly, PATH is updated but never reverted by setup-python).

thanks for the attention to detail here!

PR #1114, opened 2022-05-22
action: use the most recent python from tool cache to run the GitHub Action

mayeut:
This allows to drop python 3.6 without losing support for ubuntu-18.04 runners c.f. https://github.com/pypa/cibuildwheel/issues/508.

~~Blocked on https://github.com/actions/setup-python/pull/411~~
~~Blocked on a new release of [actions/setup-python](https://github.com/actions/setup-python)~~
v4.1 has been released.


.github/workflows/test.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+        package-dir: sample_proj
+        output-dir: wheelhouse_action
+      env:
+        CIBW_ARCHS_MACOS: x86_64 universal2 arm64
+
+    - uses: actions/upload-artifact@v3
+      with:
+        name: sample_wheels_action
+        path: wheelhouse_action
+
```
mayeut, line 81: There was no test for the GitHub Action so I duplicated the sample wheels steps to test the action.

action.yml
```diff
@@ -20,10 +20,52 @@ branding:
 runs:
   using: composite
   steps:
+    # Set up a non-EOL, cibuildwheel & pipx supported Python version
```
mayeut, line 23: This could be reworked as a helper script with its own tests.
Syntax is Python 3.4+
Unfortunately, setup-python messes with the environment for other steps & I'd rather avoid side-effects for the rest of users workflow when using the cibuildwheel GitHub Action.
It should allow self-hosted runners to continue working out-of-the-box.

action.yml
```diff
@@ -20,10 +20,52 @@ branding:
...
+            print("::set-output name=interpreter-path::{}".format(sys.executable))
+            exit(0)
+        print(
+            "::error ::The current interpreter is Python {}, cibuildwheel requires one of {}.".format(
+                current_version, ", ".join(SUPPORTED_VERSIONS)
+            ),
+            "You can use actions/setup-python to install a supported version."
+        )
+        exit(1)
+      shell: "${{ runner.os == 'Windows' && 'python' || 'python3' }} {0}"
```
mayeut, line 63: This could be reworked as a helper script with its own tests.
Syntax is Python 3.4+
Unfortunately, setup-python messes with the environment for other steps & I'd rather avoid side-effects for the rest of users workflow when using the cibuildwheel GitHub Action.
It should allow self-hosted runners to continue working out-of-the-box.

.github/workflows/test.yml
mayeut: It might be worth isolating those tests a bit more (and maybe reduce them a bit as well in this case to avoid duplicate docker pulling / python installation).

action.yml
henryiii: Can't this just be `python`? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-running-a-python-script

henryiii: This could be an action, we could put it in scikit-build/latest-python or something like that.

mayeut: > or does that get Python 2 somewhere (like on macOS)?

Indeed... https://github.com/mayeut/cibuildwheel/actions/runs/2373174130

| runner | python |
|---|---|
|ubuntu-18.04|2.7.17|
|ubuntu-20.04|3.8.10|
|ubuntu-22.04|3.10.4|
|windows-2019|3.7.9|
|windows-2022|3.9.12|
|macos-10.15|2.7.18|
|macos-11|2.7.18|
|macos-12|2.7.18|

> If not, why not python3 {0}?

It works https://github.com/mayeut/cibuildwheel/actions/runs/2373224962. Good to know they've added python3 on windows !

| runner | python |
|---|---|
|ubuntu-18.04|3.6.9|
|ubuntu-20.04|3.8.10|
|ubuntu-22.04|3.10.4|
|windows-2019|3.7.9|
|windows-2022|3.9.12|
|macos-10.15|3.9.12|
|macos-11|3.9.12|
|macos-12|3.9.12|

> This could be an action, we could put it in scikit-build/latest-python or something like that.

Yes, we could. I'll try to get this into [actions/setup-python](https://github.com/actions/setup-python). (option not to mess with the environment, together with an `python-path` output - already got a PR for that last one)

mayeut: > I'll try to get this into [actions/setup-python](https://github.com/actions/setup-python).

PR created at https://github.com/actions/setup-python/pull/411

I'll wait for feedback there to see which direction this PR will take (marking as draft for now)

mayeut: It seems https://github.com/actions/setup-python/pull/411 will be merged in some form or another

action.yml
```diff
@@ -20,10 +20,17 @@ branding:
 runs:
   using: composite
   steps:
+    # Set up a non-EOL, cibuildwheel & pipx supported Python version
+    - uses: mayeut/setup-python@no-envvar
+      id: python
+      with:
+        python-version: "3.7 - 3.10"
```
mayeut, line 27: I went with a range to allow self-hosted runners not to download python 3.10 if they have a supported version in their tool cache but no python 3.10 yet.
We could however just say `3.10` if we prefer, self-hosted runners would download & install it just fine if missing from their tool cache.

joerick: The range seems great to me. No download time, and a supported Python!

action.yml
```diff
@@ -20,10 +20,17 @@ branding:
 runs:
   using: composite
   steps:
+    # Set up a non-EOL, cibuildwheel & pipx supported Python version
+    - uses: actions/setup-python@766e8c60885ef280a3e447209d1f1083b3c7039d
```
henryiii, line 24:
```suggestion
    - uses: actions/setup-python@v4
```

(4.1 supports this)

.github/workflows/test.yml
joerick: Its probably worth consolidating these Sample Build steps. I'm happy to use the Github Action one as the version we keep, given that the tests are checking the command-line invocation.

joerick: I've pushed a change to that effect.

joerick (approved):
LGTM!

henryiii:
Ahh, `v4` doesn't point at `v4.1.0` yet. Why is this passing, actually?

mayeut:
> Why is this passing, actually?

Seems that invalid inputs are just warnings... Since Python gets setup properly, there's no reason for the workflow to fail.

> Warning: Unexpected input(s) 'update-environment', valid inputs are ['python-version', 'python-version-file', 'cache', 'architecture', 'token', 'cache-dependency-path']

Let's wait for v4 being tagged again.
I'm not sure it's worth adding another workflow to test the action does not update the environment.

henryiii:
This also should avoid getting the CMake environment variables set pointing at the global Python instance (also added in 4.1.0).

joerick:
Shall we go back to `ubuntu-latest` runner before merging?

henryiii:
Leaving this until after we drop 3.6 might be helpful. Okay either way.

mayeut:
Let's get this in then.

PR #1138, opened 2022-06-13
ensure tests run on prerelease Pythons

### LINKED ISSUE #1137: Tests are skipped on pre-release Pythons
### Description

On cibuildwheel==2.6.1, when running with `CIBW_PRERELEASE_PYTHONS=1`, any configured test command is always skipped when targeting a pre-release Python version. A quick glance at the code makes me think it's an unintended consequence of having `TestSelector` inherit the prerelease exclusion behavior from `IdentifierSelector`.

### Build log

https://github.com/rolpdog/cffi-mirror/runs/6867587654?check_suite_focus=true

### CI config

https://github.com/rolpdog/cffi-mirror/commit/b7a01156bafba52bd2bfedc1222c9eb8d36491c7#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd

nitzmahone:
fixes #1137 

This is a pretty naive approach to the fix- not sure if maybe the Python prerelease stuff should just move to another layer in the class hierarchy instead, but this makes things work as expected at least...

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -387,7 +387,7 @@ class Options:
             requires_python=requires_python,
             prerelease_pythons=prerelease_pythons,
         )
-        test_selector = TestSelector(skip_config=test_skip)
+        test_selector = TestSelector(skip_config=test_skip, prerelease_pythons=prerelease_pythons)
 
         archs_config_str = args.archs or self.reader.get("archs", sep=" ")
         architectures = Architecture.parse_config(archs_config_str, platform=self.platform)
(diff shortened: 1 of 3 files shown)


joerick:
I've pushed a change here to refactor the class hierarchy as there was some unnecessary coupling before.

henryiii (approved):

mayeut (approved):

joerick:
Thanks you @nitzmahone and reviewers :)

PR #1143, opened 2022-06-17
Fail on errors inside multiline commands

abitrolly:


abitrolly:
Looks like tests caught some unhandled error.

For Windows it might also by possible https://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error.

henryiii:
> Looks like tests caught some unhandled error.

This is causing this:

https://github.com/pypa/cibuildwheel/blob/cc56a140d2e514e2a46bdf6640b035917ee53b77/test/test_troubleshooting.py#L50-L51

to break with `err Error: Command ['sh', '-e', '-c', 'false'] failed with code 1.` instead of throwing an error here as expected. I'm not sure why, but it's not an unhandled error, it's just breaking a unit test by changing the error somehow.

abitrolly:
@henryiii it throws error as expected. The failure happens when comparing `stderr` here.

https://github.com/pypa/cibuildwheel/blob/cc56a140d2e514e2a46bdf6640b035917ee53b77/test/test_troubleshooting.py#L32

The new `stderr` is this.
```
Error: Command ['sh', '-e', '-c', 'false'] failed with code 1.
```
And the expected one is something with this line.
```
"NOTE: Shared object (.so) files found in this project."
```
Probably this line was written later and then an error was thrown. Now the error is caught early based on exit code.

Here is where the NOTE is printed.
https://github.com/pypa/cibuildwheel/blob/c00b59fe8b76e16c47fa98046f80dbcf5b620903/cibuildwheel/linux.py#L400

abitrolly:
This last commit should fix that. I'd prefer to pass the prepared command name to `troubleshoot` handler, but I can not immediately see how to do that. Perhaps `docker.call` could return this info in error.

joerick:
If we can figure out a way to make this consistent cross-platform, I'd be all for it. I did a little search and I can find a [few](https://sourcegraph.com/github.com/vispy/vispy/-/blob/.github/workflows/wheels.yml) [examples](https://sourcegraph.com/github.com/openai/triton/-/blob/.github/workflows/wheels.yml) of people configuring it wrong (assuming that the first error will fail it).

The trouble with it not being cross platform is that if somebody configures their tests like 

```yaml
CIBW_TEST_COMMAND: |
    pytest unit_tests
    pytest tests
```

That would be a correct config on Linux and macOS, but `pytest unit_tests` could fail on windows but report success.

The current situation is at least consistent.

p.s. we use `cmd` to run these commands on Windows, currently. I'm not sure if that matters.

abitrolly:
Well I did that mistake too, because other CI systems manage to do it somehow. Maybe on Windows it is possible to feed the line one by one and check error status there? How GitHub manages to do this? Maybe this part is open?

abitrolly:
`: >` is too error-prone for `bash` scripts. Another 15 minutes spent catching this.

```
        CIBW_BEFORE_ALL: >
          # dnf install -y $(grep "^[^#]" dependencies.txt) &&
          pip install Cython &&
          cd python && cmake .
```

joerick:
> Maybe on Windows it is possible to feed the line one by one and check error status there?

Maybe... but that's probably getting deeper into the execution than I'd like. Shells are tricky things. Commands that contain characters like `;`, `'`, `"` can really mess things up.

> How GitHub manages to do this?

According to [these (rather nice) docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference), it's not possible with `cmd` (which is what cibuildwheel uses). But they actually use `pwsh` instead. I wonder if that's broadly compatible with pwsh? That could be an option 🤔

joerick:
Sadly pwsh isn't part of the system, it's installed separately. Which would mean adding a non-python dependency to cibuildwheel. Probably not worth it...

henryiii:
I think staying simple and close to the metal here is fine and less likely to cause issues - Windows should continue to use the user-level trickery we already support - user can use TOML arrays or they can add `&&` on every line (which is supported cross platform).

I'm not sure it would be that bad to have differing behavior here on Windows (there's a limit to how much you can write cross platform anyway for a command). The downside is that a working linux/macOS build could hide Windows problems, discovering that a multiline string doesn't work out of the box for errors on Windows might be a bit harder if it works on UNIX. But not sure it's that bad since cmd doesn't support it.

One nice feature of the manual `&&` (which the TOML array syntax method adds for you) is that it displays nicely (and is copy-pasteable) in our pre-run printout.

So if you ask for `cmd` GHA still runs powershell? I know they have the option to use cmd, but powershell is the default.

joerick:
> So if you ask for `cmd` GHA still runs powershell? I know they have the option to use cmd, but powershell is the default.

No, sorry, I meant that `pwsh` is their default. You can ask for `cmd`, but they note that it doesn't fail-fast like the others.

> I think staying simple and close to the metal here is fine and less likely to cause issues - Windows should continue to use the user-level trickery we already support - user can use TOML arrays or they can add `&&` on every line (which is supported cross platform).

I tend to agree. And the current situation is easily explained - "it won't fail-fast unless you do it like this" - whereas if we add options like this then we have to start saying "on linux and mac, it does work like this, but on Windows, it's different, because cmd...". So it's just more to think about. Just having one best-practice that always works is much preferable to me.

abitrolly:
> Just having one best-practice that always works is much preferable to me.

One best practice doesn't work, because commands for Linux rarely match commands for Windows. If there is a budget, I would be actually interested to see stats about matching and non-matching lines in join Windows and Linux builds.

For me the proper resolution of this ticket is to leave the && comment just for Windows users instead of demanding that Linux users should equally suffer too.

abitrolly:
The best practice as I see it is to move all build instructions into separate script and call just this one script without embedding shell instructions into `pyproject.toml` or CI YAMLs.

henryiii:
> The best practice as I see it is to move all build instructions into separate script and call just this one script without embedding shell instructions into `pyproject.toml` or CI `YAML`s.

I disagree with this for many cases. If the code can't be reused, that is, if it's CI specific, it should _not_ sit in a `.sh` file, littering the repo, making the CI config harder to follow, and begging a user to try to run it locally. Especially if it has sudo commands that could mess with the user's system! It's much harder to track down where things are running if you have a bunch of `.ci/build_doxygen.sh` and such calls that then have to be looked up. This is an especially sore spot because I used to do this a lot with Travis, and moving to a new CI has been much _more_ painful than if I'd had all the code inside the travis.yml file, not less! Stuff like downloading dependencies really is CI specific, and should sit in CI files. Putting it in the TOML file means you can run it locally in the environment it was designed for without locking yourself to a specific CI system, which is ideal, I think. You technically still need to look in two places (pyproject.toml & CI), but the CI config is generally very simple, making the TOML file the only required reading to understand the build process.

If it is reusable and not dangerous to run locally, then it can sit in a file. And both TOML and environment files are perfectly capable of running files. And you then can put your `set -euf -o pipefail` line on top and now you don't care about errors in multiline commands here at all, so you are not affected by the multiline setting here at all.

abitrolly:
I do not support your decision not to fail on multiline commands. And would encourage every user to add +1 comment to this issue if they had to deal with builds that should have been failed, but didn't because of this issue.

PR #1145, opened 2022-06-18
Use preinstalled Python 3 for `cibuildwheel` action

abitrolly:
There is at least 3.7 in every runner
https://github.com/actions/virtual-environments

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,19 +80,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-10.15]
+        os: [ubuntu-20.04, windows-2019, macos-11]
 
     steps:
       - uses: actions/checkout@v2
 
-      # Used to host cibuildwheel
-      - uses: actions/setup-python@v2
-
       - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==2.7.0
+        run: python3 -m pip install cibuildwheel==2.7.0
 
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
+        run: python3 -m cibuildwheel --output-dir wheelhouse
         # to supply options, put them in 'env', like:
         # env:
         #   CIBW_SOME_OPTION: value
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -191,7 +191,7 @@ To build Linux, Mac, and Windows wheels using GitHub Actions, create a `.github/
     ```
 
 !!! tab "Generic"
-    This is the most generic form using setup-python and pip; it looks the most
+    This is the most generic form using python3 and pip; it looks the most
     like the other CI examples. If you want to avoid having setup that takes
     advantage of GitHub Actions features or pipx being preinstalled, this might
     appeal to you.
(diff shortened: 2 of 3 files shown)


henryiii:
[18.04 has 3.6](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md). And using setup-python is recommended on GHA, and this example is more about showing generally how to use cibuildwheel, not the specifics of what's installed in a runner. You really should use pipx, or the action itself, etc. if this was about GHA. GHA recommends using setup-python if you need Python.

abitrolly:
> [18.04 has 3.6](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md).

And `cibuildwheel` is compatible with 3.6 https://github.com/pypa/cibuildwheel/blob/cc56a140d2e514e2a46bdf6640b035917ee53b77/setup.cfg#L43

> And using setup-python is recommended on GHA

"[because it ensures consistent behavior across different runners](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#specifying-a-python-version)". For me `setup-python` is probably needed to setup `python-dev` and all dependencies for binary extensions properly, but it looks like `cibuildwheel` uses its own environment control. Having `setup-python` makes scenarios less clear as it may give a false impression that it is how Python version for wheel is selected.

abitrolly:
> `python3` to `setup-python` on Windows, but I'm not sure it's available without setup-python on Windows

`windows-latest` has it https://github.com/yakshaveinc/python/runs/6945448591?check_suite_focus=true

image

henryiii:
> And cibuildwheel is compatible with 3.6

It would be nice to change that, though! Hopefully if setup-python gains better composite actions support, we can drop 3.6 from the host.

Don't we want consistent behavior across runners? And if we tell people to use `python3`, we'll be getting an older and less consistent Python version. And if people really are using cibuildwheel on GHA, they should ideally be using the action. Which uses pipx. Which currently uses the (inconsistent) built-in Python 3 (albeit with venvs, etc). So technically it's not that different (and part of why we still support 3.6 on the host). We are actively trying to fix this, though, by adding a feature to setup-python for composite actions.

I'm not terribly against it but I think the current method is a bit closer to best-practices for GHA examples. Put me down as -0.1.

henryiii:
> less clear as it may give a false impression that it is how Python version for wheel is selected

It's something you do have to be at least a little aware of, and I've been in favor of making it more obvious (by printing the host Python version in the initial printout), but I'm not in the majority there. It was much better before v4 started forcing you to specify a version (discussion in https://github.com/actions/setup-python/issues/421), as leaving it verisonless was clearer.

abitrolly:
> Don't we want consistent behavior across runners? And if we tell people to use python3, we'll be getting an older and less consistent Python version.

Cover all runners with a test that builds simple wheel will signal when this version becomes a problem. Speaking of this test, why there are no tests running with GHA?

> So technically it's not that different (and part of why we still support 3.6 on the host). We are actively trying to fix this, though, by adding a feature to setup-python for composite actions.

And switching to 3.6 would leave out builds without GHA on Ubuntu 18.04 (https://packages.ubuntu.com/bionic/python3).

The project already can bootstrap needed Python versions to run code inside it, so if coding with 3.6 is that painful, it might be possible to keep the bootstrap part minimal (and probably reused from outside) and run the rest of `cibuildwheel` with that Python. That would come with a small performance penalty though.

joerick:
Calling `setup-python` would seem to be a best-practise kind of thing. I'd prefer to keep it. The github example workflows all seem to use it.

e.g. https://github.com/actions/starter-workflows/blob/main/ci/python-app.yml

PR #1147, opened 2022-06-19
Clear up that BEFORE_ALL runs outside of Python env

abitrolly:


joerick:
I'm not totally sure I understand the problem here. What was your misunderstanding with the previous way it was worded?

henryiii:
How about keeping the old wording but adding after the parenthetical remark "before creating Python environments"?

Czaki:
I think that @abitrolly forget to link #1146 discussion.

docs/options.md
```diff
@@ -635,7 +635,7 @@ To specify more than one environment variable, separate the variable names by sp
 ### `CIBW_BEFORE_ALL` {: #before-all}
 > Execute a shell command on the build system before any wheels are built.
 
-Shell command to prepare a common part of the project (e.g. build or install libraries which does not depend on the specific version of Python).
+Shell command that runs before creating Python environment, to build or install parts that do not depend on the specific version of Python.
```
henryiii, line 638:
```suggestion
Shell command to prepare a common part of the project (e.g. build or install libraries which does not depend on the specific version of Python). This runs before creating the Python environments for each build.
```

abitrolly:
> I'm not totally sure I understand the problem here.

The current doesn't contain necessary technical details to prevent people from shooting themselves in the foot as I did in #1146.

docs/options.md
abitrolly: If  a "common part of the project" doesn't include pieces that are shared between operating system environments, it is better to leave this scope escape off to avoid bloating the description beyond that is mere necessary.

joerick:
I've added a little more detail to this. I just wanted to make sure the point was made - 'before the Python environment is created' seemed a little hand-wavy to me (which Python environment - cibuildwheel creates many).

Look okay?

image

https://cibuildwheel--1147.org.readthedocs.build/en/1147/options/#before-all

Issue #1142, opened 2022-06-15
cibuildwheel should error out if the same build produces multiple wheels with the same filename


joerick:
As discussed in #1129, we shouldn't silently overwrite wheels that are produced in the same run that have the same filename, as this implies a misconfiguration.

See https://github.com/pypa/cibuildwheel/pull/1129#issuecomment-1152905408 and https://github.com/pypa/cibuildwheel/pull/1129#issuecomment-1153611787 for further detail.

(alternatively, we could try to cover more wheel overlap cases with the same technique used in our ABI3 wheel support. See https://github.com/pypa/cibuildwheel/pull/1129#issuecomment-1152392491 for an idea on this)

PR #1151, opened 2022-06-19
feature: add support for `py3-none-{platform}` wheels

mayeut:
This extends the mechanism introduced in #1091 for `abi3` wheels.
Most of the mentions to `abi3` have been removed and replaced by a more generic `compatible_wheel`.
This allows to build a wheel `foo-0.1-py3-none-win_amd64.whl` only once and still test with every configured python.

relates to #1142


henryiii (approved):

cibuildwheel/util.py
```diff
@@ -574,23 +574,25 @@ def virtualenv(
...
         _, _, _, tags = parse_wheel_filename(wheel.name)
         for tag in tags:
-            if tag.abi != "abi3":
-                continue
-            if not tag.interpreter.startswith("cp3"):
+            if tag.abi == "abi3":
+                if not (interpreter.startswith("cp3") and tag.interpreter.startswith("cp3")):
+                    continue
+            elif tag.abi == "none":
+                if tag.interpreter[:3] != "py3" and tag.interpreter[:3] != interpreter[:3]:
```
joerick, line 591: This line is confusing me... if the wheel tag doesn't start with 'py3', then we check that the part after that is different?

Also, (genuine question), why do we care about python interpreter version if abi is 'none'? To me, 'none' would mean that we don't actually use the Python C interface (e.g. a ctypes/cffi library) I suppose the obvious answer is "that's what `packaging` does", but I'm curious about the semantics of 'none'.

mayeut: I was sure cp35-none would be compatible with cp36, 37, ....
`pip debug` proves me wrong (or there's a bug in packaging).
I will restrict this to `py3`

joerick:
I added an integration test for py3-none wheels, using ctypes. I was curious anyway about them, so it was a good chance to check them out. They're pretty cool actually - one wheel that works for all Pythons, including PyPy. 

The test seems to pass, at least it did locally on linux. I'll check the execution time in case we want to disable some Python versions for performance. Setting up the test virtualenvs does slow the tests down that use CIBW_TEST_COMMAND. It might also be worth counting the number of times the test says "Building wheel..." versus "Found previously built wheel".

joerick (approved):

PR #1152, opened 2022-06-19
fix: error out if multiple wheels with the same name are produced

### LINKED ISSUE #1142: cibuildwheel should error out if the same build produces multiple wheels with the same filename

mayeut:
We shouldn't silently overwrite wheels that are produced in the same run that have the same filename, as this implies a misconfiguration.

fix #1142 
builds on top of #1151


mayeut:
I'll have to debug the Windows failure later, probably not before next week-end.

joerick (commented):
This looks good, thank you @mayeut. I've just reviewed the commit ef18186c455b3d4225b9de89e623cb2a77c6091c, since the rest is in a different PR.

henryiii (approved):

PR #1155, opened 2022-06-19
Add `data/images.yml` with info about OS inside

abitrolly:
There is a missing info about which package manager should be used in which manylinux image. I am not an expert in MkDocs, but I've seen another file here, so if this YAML can be turned into some table in documentation, that would be helpful I guess. Which is missing is to how to search package names in AlmaLinux online. https://repology.org/ works somewhat, but it still not clear where the package sources and issue tracker for requests.

### DIFF
--- a/docs/data/images.yml
+++ b/docs/data/images.yml
@@ -1,0 +1,23 @@
+# name: image name
+# os: operating system in the image 
+# packman: package management tool
+
+- name: manylinux_2_28
+  os: AlmaLinux 8
+  packman: dnf
+
+- name: manylinux_2_24
+  os: Debian 9
+  packman: apt-get
+
+- name: manylinux2014
+  os: CentOS 7
+  packman: yum
+
+- name: manylinux2010
+  os: CentOS 6
+  packman: yum
+
+- name: manylinux1
+  os: CentOS 5
+  packman: yum


joerick:
This information could be useful as a Markdown table in the [option docs](https://cibuildwheel.readthedocs.io/en/stable/options/#linux-image). We probably don't need the YAML file though. Could you reformat the information as a Markdown table and put it in the relevant place in [options.md](https://github.com/pypa/cibuildwheel/blob/cc56a140d2e514e2a46bdf6640b035917ee53b77/docs/options.md)?

henryiii:
As a markdown table you could also easily add a footnote about `manylinux_2_24` being likely to be discontinued (with a link to https://github.com/pypa/manylinux/issues/1332). Or simply not include it.

FYI, `yum` still works on RHEL 8, though if it's present, `dnf` is "dandified yum" and better (and faster from what I can tell). There are small differences for redistributions like Alma; one of the main ones is likely activating EPEL, which had a "centos" in the package name for CentOS. Could be worth a mention too.

We should include musllinux which uses `apk`. Also `apt` is the package manager for Debian.

abitrolly:
I would actually prefer to fetch the YAML from https://github.com/pypa/manylinux#manylinux to extract the relevant info from it, but as it also ships with markdown table, so decided to start from here.

As a proponent of open data and datacentric approach to everything, I don't feel like formatting markdown tables by hand. However, if somebody points me how to render data files with MkDocs, I may be able to contribute the formatter.

abitrolly:
> for redistributions like Alma; one of the main ones is likely activating EPEL

`Alma` doesn't activate EPEL. I even got downvoted for asking why.

https://www.reddit.com/r/AlmaLinux/comments/vg1kj2/comment/iczn7cw/?utm_source=share&utm_medium=web2x&context=3

abitrolly:
I hope we are still speaking in the context of `cibuildwheel` users who just need to build a wheel for a new version of Linux without getting into the intricate details of RHEL Alma EPEL relationships (#1157).

joerick:
> I don't feel like formatting markdown tables by hand. However, if somebody points me how to render data files with MkDocs, I may be able to contribute the formatter.

Sounds overengineered to me! This data doesn't change that often, a hard-coded markdown table would be fine.

joerick:
Feel free to reopen if you want to format as a markdown table. Actually I think it might be more useful in the Tips and Tricks section of the docs, [here](https://cibuildwheel--1155.org.readthedocs.build/en/1155/faq/#linux-builds-on-docker).

PR #1159, opened 2022-06-21
Name downloadable artifacts by OS name

abitrolly:
They will be available as separate downloads from GitHub interface.

### DIFF
--- a/examples/github-minimal.yml
+++ b/examples/github-minimal.yml
@@ -25,4 +25,5 @@ jobs:
 
       - uses: actions/upload-artifact@v3
         with:
+          name: wheels-${{ matrix.os }}
           path: ./wheelhouse/*.whl


abitrolly:
@Czaki for minimal example I think it is good to have distinguished names as those are not upload anywhere. For deploy example I would name it `{projectname}-wheels` and maybe adding a date to it, so that when those files end on disk, it is clear where they from.

abitrolly:
Actually `${{ github.event.repository.name }}-wheels` will produce `cibuildwheel-wheels.zip` for this repo.

henryiii:
This makes it much harder to use in an upload step. You'd need a separate download artifacts for each one. The main use for uploading artifacts is downloading artifacts, which this makes harder. I think we should stick to the default and let users rename them if they want to. Something like `${{ github.event.repository.name }}-wheels` adds complexity that is not related to cibuildwheel. It's also wrong - this is where you put the SDist too. It's not just wheels. It's a collection of artifacts and in the UI, it shows up under the "artifacts" heading.

henryiii:
(To be clear, I'm not against someone giving them a custom name. I probably wouldn't for consistency across repos, but I'm not against it. I just don't think it's helpful to _teach_ it here as a requirement for using cibuildwheel. The simpler we keep the config, the better.)

abitrolly:
> The main use for uploading artifacts is downloading artifacts, which this makes harder.

Right. And this case point users to web UI downloads, where they most likely want wheels for their platform.

> Something like `${{ github.event.repository.name }}-wheels` adds complexity that is not related to cibuildwheel. It's also wrong[1](#user-content-fn-1-f63186754dc561f9cb2e92789847f647) - this is where you put the SDist too.

This specific example doesn't upload SDists.

abitrolly:
The main complain is that wheels in artifacts are not discoverable by users, who are not aware that these should be wheels.

image

henryiii:
Still don't see the problem downloading `artifact` under `Artifacts`, instead of `wheels` under `Artifacts`. You can't rename the `Artifacts` part. It's not perfect, but IMO not worth adding complexity to the simple "minimal" example just to make this name a little prettier. It then adds complexity to followup too, since you now have to know match this new name you set to download the wheels and usually SDist. This _is_ the artifact of the CI run.

henryiii:
By the same logic, why not add job cancellation, names to all the steps, a markdown summary, etc. to the _minimal_ example? It might make the output nicer and the job better. I think the point is to try to keep it as simple as possible to highlight exactly what you _need_ to get cibuildwheel to work. You don't need to customize this name to get cibuildwheel to work. This is not a general tutorial on how to use GitHub Actions.

abitrolly:
For a minimal example then upload-artifacts step is not needed at all. It just shows that wheels are built.

henryiii:
Building wheels then not uploading them anywhere is not helpful. Someone following this example could make functional wheels and manually upload them to PyPI. Removing the artifact step would break that. Adding a slightly nicer name to the wheels would not change that. Breaking them up into three separate artifacts as is done here (which also introduces the concept of expanding a GHA variable into an existing string) would make it even harder to download - you'd need three clicks instead of one.

abitrolly:
> Someone following this example could make functional wheels and manually upload them to PyPI. Removing the artifact step would break that.

There is no need use artifacts in GitHub Actions to upload wheels to PyPI https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

abitrolly:
That was just an idea. Anyway, it probably doesn't worth the time.

abitrolly:
Artifact are good, but their `name` handling logic is bugging me, so I opened https://github.com/actions/download-artifact/issues/166

PR #1160, opened 2022-06-22
fix CalledProcessError message output

kemingy:
According to:

https://github.com/pypa/cibuildwheel/blob/0bf4f7abc36bd33bed60eac98c0ee2faefcc4ae6/cibuildwheel/docker_container.py#L252-L253

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -367,7 +367,7 @@ def build(options: Options, tmp_path: Path) -> None:  # pylint: disable=unused-a
 
         except subprocess.CalledProcessError as error:
             log.step_end_with_error(
-                f"Command {error.cmd} failed with code {error.returncode}. {error.stdout}"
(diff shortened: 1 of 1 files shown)


henryiii:
https://github.com/python/cpython/blob/ad2e9f97c2cf270a33705e32427738e56e809bea/Lib/subprocess.py#L130-L132

Though, is it correct to not show stderr too? I am currently dealing with a problem and I think it might be that stderr is being swallowed (locally).

henryiii:
Ahh, it's because stderr is being printed directly, but stdout is being moved down to the end. Weird, but at least it's not missing.

Anyway, output and stdout are interchangeable, so don't think it's necessary to change from one to the other.

kemingy:
> https://github.com/python/cpython/blob/ad2e9f97c2cf270a33705e32427738e56e809bea/Lib/subprocess.py#L130-L132

I see. Thanks for pointing it out. Will close this one.

PR #1175, opened 2022-07-14
chore: drop python 3.6 for cibuildwheel driver

### LINKED ISSUE #508: [question] Raise minimum required Python to 3.7

mayeut:
This drops python 3.6 for cibuildwheel.
Building python 3.6 packages is still supported.

closes #508

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,12 +17,6 @@ repos:
   rev: v2.37.1
   hooks:
   - id: pyupgrade
-    name: PyUpgrade 3.6+
-    args: ["--py36-plus"]
-    exclude: ^bin/
-  - id: pyupgrade
-    name: PyUpgrade 3.7+ on bin
-    exclude: ^(cibuildwheel|unit_test|test)/
     args: ["--py37-plus"]
 
 # Autoremoves unused imports
@@ -52,9 +46,9 @@ repos:
   rev: v0.961
   hooks:
   - id: mypy
-    name: mypy 3.6 on cibuildwheel/
-    exclude: ^(bin|cibuildwheel/resources|docs)/.*py$
-    args: ["--python-version=3.6"]
+    name: mypy 3.7 on cibuildwheel/
+    exclude: ^cibuildwheel/resources/.*py$
+    args: ["--python-version=3.7"]
     additional_dependencies: &mypy-dependencies
       - nox
       - packaging>=21.0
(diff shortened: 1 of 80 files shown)


unit_test/conftest.py
```diff
@@ -27,4 +28,11 @@ def mock_path_exists(path):
     args = ["cibuildwheel", str(MOCK_PACKAGE_DIR)]
     monkeypatch.setattr(Path, "exists", mock_path_exists)
     monkeypatch.setattr(sys, "argv", args)
-    return args
+
+    tmp_path.joinpath(MOCK_PACKAGE_DIR).mkdir()
+    cwd_backup = os.getcwd()
+    os.chdir(tmp_path)
```
joerick, line 34: `monkeypatch.chdir(path)` should work here, which would save a little cleanup code.

I'm curious - is this related to the 3.7 changes or incidental?

.pre-commit-config.yaml
```diff
@@ -17,12 +17,7 @@ repos:
   rev: v2.37.1
   hooks:
   - id: pyupgrade
-    name: PyUpgrade 3.6+
-    args: ["--py36-plus"]
-    exclude: ^bin/
-  - id: pyupgrade
-    name: PyUpgrade 3.7+ on bin
-    exclude: ^(cibuildwheel|unit_test|test)/
```
henryiii, line 25: We don’t need the name anymore.

mayeut: name removed

unit_test/conftest.py
mayeut: I applied your suggestion & made a different commit for the explanation on this one.

.pre-commit-config.yaml
```diff
@@ -37,6 +32,8 @@ repos:
   rev: 5.10.1
   hooks:
   - id: isort
+    args: ["-a", "from __future__ import annotations"]
```
henryiii, line 35: You added this before I could suggest it. :D

joerick:
Let's hold this until after the next patch release (2.8.1)

PR #1178, opened 2022-07-14
DOC: Typo manylinux_2_24

larsoner:
Just a little typo `manylinux2_24` to `manylinux_2_24`


larsoner:
... and another one I just hit in https://github.com/openmeeg/openmeeg/pull/466 where you need to quote a string that starts with a brace in YAML

henryiii (approved):

PR #1181, opened 2022-07-16
fix(docs): pin things due to the mkdocs pin

henryiii:
We really need to upgrade mkdocs! The old version we are stuck on doesn't support Python 3.10 or the latest Markdown.

Should patch the error seen in #1180 with docs.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -92,12 +92,13 @@ def update_proj(session: nox.Session) -> None:
     )
 
 
-@nox.session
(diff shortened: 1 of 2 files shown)


joerick:
Thanks. I originally pinned because of theme incompatibilities. I'll take a look at the upgrade soon.

joerick (approved):

henryiii:
The “failure” is that the “allowed to fail” job reported a status of cancelled. :/

PR #1182, opened 2022-07-17
DOC: Bump to non-deprecated image

larsoner:
Bumps docs and Azure to refer to macos-11 since macos-10.15 is deprecated.

Feel free to close if this isn't the right fix and more is needed, I figured I'd get the ball rolling at least in case it isn't.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-11.0]
+        os: [ubuntu-20.04, windows-2019, macOS-11]
 
     steps:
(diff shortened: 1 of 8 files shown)


henryiii (approved):
I don’t think there are any remaining downsides to 11? IIRC there was a limitation originally? Something with PyPy? Hoping that’s not the case anymore?

larsoner:
I can let you know once I get CIs working on my project since I plan to try to build x86_64 and arm64 for both CPython and PyPy variants. TBD how long that will actually take me to accomplish, though!

henryiii:
Yep, that’s what I was thinking of. 👍 

Also I think PyPy now defines a reasonable macOS min version now 10.10, IIRC?) so we might be able to remove the macOS min version default.

PR #1188, opened 2022-07-22
Provide alternative to actions/setup-python for self-hosted runners

jbarlow83:
Since commit 25c1bf53d079d1fa52a5144b63bca3e66d8e21d4 (v2.8.1),
cibuildwheel uses actions/setup-python to ensure that a reasonable
Python is installed on the GitHub Actions runner. This is a reason
solution for standard runners, but unfortunately, actions/setup-python
is not capable of furnishing some self-hosted runners with a working
Python. In particular, it only provides x64 and x86 Pythons, so breaks
anyone using an ARM self-hosted runner.

This change fixes the issue by allowing self-hosted runners by
introducing a new environment variable, CIBW_GITHUB_ACTIONS_PYTHON,
allowing one to specify the host Python without calling
actions/setup-python. The user is responsible for ensuring that
Python and pipx are installed on the self-hosted runner.

Here is the list of Pythons that actions/setup-python can install:
https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -21,16 +21,28 @@ runs:
   using: composite
   steps:
     # Set up a non-EOL, cibuildwheel & pipx supported Python version
+    # setup-python@v4 does not support ARM runners (e.g., self-hosted runners).
+    # These runners must ensure cibuildwheel and provided on their own before
+    # cibuildwheel.
+    # See https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
     - uses: actions/setup-python@v4
       id: python
       with:
         python-version: "3.7 - 3.10"
         update-environment: false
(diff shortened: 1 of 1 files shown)


jbarlow83:
I realize the environment variable may not be ideal option here since it's a GHA specific problem and solution....

Czaki:
maybe just use `pip install cibuildwheel` and direct call in your workflow? like for other CI providers?

henryiii:
Yes, this is just a shortcut for `pipx run cibuildwheel==2.8.1`. If you are self hosted and you can’t use setup-python too, why not use pipx directly?

I’m not strongly against this either, but I would think it should be an input, not an envvar.

jbarlow83:
Yes on second thought, it's probably to do it manually. I'll withdraw.

PR #1191, opened 2022-07-23
Add support for Cirrus CI

### LINKED ISSUE #145: Support for Cirrus CI
Yet another good workhorse.
https://cirrus-ci.org/features/

> **Free for Open Source**
> To support Open Source community Cirrus CI provides Linux, Windows, macOS and FreeBSD services free of charge.

### LINKED ISSUE #1240: Add support for native Linux arm64 builds on Cirrus CI
Support for native Linux arm64 on Cirrus CI wasn't added in #1191 due to some blockers on Cirrus CI's side.

It is likely going to be easy to solve this issue once at least one of these gets resolved:
- https://github.com/cirruslabs/cirrus-ci-docs/pull/1027
    - this would allow running a custom **arm64** VM rather than just a container which will enable us to use Cirrus CI's Docker Builder VM image for the build (this could maybe even support Windows arm64 if and when Microsoft starts supporting generic arm64 rather than just what they can support while they have an exclusivity deal with Qualcomm)
- https://github.com/cirruslabs/cirrus-ci-docs/issues/1034
    - this would allow running Docker in Docker on Cirrus CI which would enable us to run Docker daemon in an additional container and use that to run manylinux images

Until then, it might be possible to resolve this in some different way (potential contenders including spinning up a docker daemon in some different way or using cibuildwheel's Podman support instead) *but* it would not be *that* easy so I don't think it's worth pursuing before then.

jack1142:
Resolves #145 and resolves #1240

For macOS arm64, there was a need for some changes to the test suite:
- some of the tests were failing because they tried to build for CPython 3.6 (`cp36-*`) and on macOS arm64, you can only use CPython 3.8 and above ([`4bc6bd3`](https://github.com/pypa/cibuildwheel/commit/4bc6bd3c5fcc2d3b4822464fab4d32f030b0ff91), [`b39dc5b`](https://github.com/pypa/cibuildwheel/commit/b39dc5bb5e1db77da817c198fbd0ca4dd4b409a3))
- due to an issue with `wheel` 0.36.1 and earlier, generated wheel filenames had a wrong minimum macOS version ([`3c246c7`](https://github.com/pypa/cibuildwheel/commit/2bf4c14249bdbff865b8825d1d1842a64dc34639))
- due to #1169, cibuildwheel generated `cp38-macosx_arm64` wheels without testing them which caused it to not delete the wheel when the test expected it to ([`2b35f3c`](https://github.com/pypa/cibuildwheel/commit/2b35f3c658f3d7506362be90d0fa0509dcfca19f))

Link to a Cirrus CI test suite build: https://cirrus-ci.com/build/6463822541094912
Link to a Cirrus CI minimal example build: https://cirrus-ci.com/build/6073962235953152
Link to a Cirrus CI macOS Intel example build: https://cirrus-ci.com/build/6736767930859520


.cirrus.yml
```diff
@@ -0,0 +1,67 @@
...
+    DOCKER_HOST: tcp://localhost:2375
+
+  install_pre_requirements_script:
+    - curl -sSL https://get.docker.com/ | sh
+  <<: *RUN_TESTS
+
+windows_x86_task:
+  # The task takes ~55 minutes while the timeout happens
+  # after 60 minutes by default, let's allow some wiggle room.
+  timeout_in: 70m
```
joerick, line 46: Experience tells me a bit extra leeway would be good

```suggestion
  timeout_in: 90m
```

test/test_dependency_versions.py
```diff
@@ -58,6 +59,8 @@ def test_pinned_versions(tmp_path, python_version, build_frontend_env):
     build_environment = {}
 
     if python_version == "3.6":
+        if utils.platform == "macos" and platform.machine() == "arm64":
+            pytest.skip("macOS arm64 does not support Python 3.6")
```
joerick, line 63: This is confusing me... 3.6 should run under Rosetta I think? It does on my M1:

```
$ which python3.6
/Library/Frameworks/Python.framework/Versions/3.6/bin/python3.6
$ python3.6 --version
Python 3.6.8
$ python3.6 -m platform
Darwin-21.5.0-x86_64-i386-64bit
```

test/test_testing.py
```diff
@@ -140,6 +140,8 @@ def test_failing_test(tmp_path):
                 # problems with this, so let's check that.
                 "CIBW_MANYLINUX_I686_IMAGE": "manylinux1",
                 "CIBW_MANYLINUX_X86_64_IMAGE": "manylinux1",
+                # tests don't run on CPython 3.8 when running on macOS arm64
+                "CIBW_SKIP": "cp38-macosx_arm64",
```
joerick, line 144: Your solution is right, I think. Just trying to add a bit of detail to this (rather unintuitive) situation...

```suggestion
                # CPython 3.8 when running on macOS arm64 is unusual. The build
                # always runs in x86_64, so the arm64 tests are not run. See 
                # #1169 for reasons why. That means the build succeeds, which 
                # we don't want. So we skip that build.
                "CIBW_SKIP": "cp38-macosx_arm64",
```

unit_test/environment_test.py
```diff
@@ -41,9 +41,15 @@ def test_shell_eval():
...
+    env = {"PREV_VAR": "1 2 3"}
+    # The above uses `echo` binary which, on Windows, is only shipped by msys2
+    # that ships with Git. This binary just hangs when USERPROFILE is missing.
+    if "USERPROFILE" in os.environ:
+        env["USERPROFILE"] = os.environ["USERPROFILE"]
 
-    assert environment_dict == {"PREV_VAR": "1 2 3", "VAR": "1 2 3 string"}
+    environment_dict = environment_recipe.as_dictionary(prev_environment=env)
+
+    assert environment_dict == {**env, "VAR": "1 2 3 string"}
```
joerick, line 52: Ugh. So yeah, to me it feels like we should just fix the dependency on msys2. Since `echo` is built-in to `cmd`, can we use that instead? I tried this on my Windows PC:

```
joeri@JOERICK-PC C:\Users\joeri>python -c "import subprocess; subprocess.run(['echo', 'something'])"              
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 1438, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified

joeri@JOERICK-PC C:\Users\joeri>python -c "import subprocess; subprocess.run(['cmd', '/c', 'echo', 'something'])"
something
```

So `echo` isn't installed, but `cmd /c echo` does work.

I think I'd slightly prefer that to a `python -c "import os;..."` because invoking Python here makes things more complicated again and might not properly test the shell-like behaviours.

joerick (commented):
Exciting! I'd love to have native macos-arm64 CI, and another native linux-aarch64 is no bad thing either!

I've tried to add the service to this repo, but it seems that it requires a pypa admin. @pradyunsg would you mind adding the free tier of CirrusCI to this repo? https://github.com/marketplace/cirrus-ci

test/test_dependency_versions.py
jack1142: Sure but x86_64 target is not enabled by default when running on arm64. I suppose I should try enabling that, at least for the test suite.

unit_test/environment_test.py
jack1142: > Ugh. So yeah, to me it feels like we should just fix the dependency on msys2. Since `echo` is built-in to `cmd`, can we use that instead?

Probably but then we would need to run different command depending on platform which seems to somewhat defeat the point of the cross-platform evaluator of these expressions? `python -c` with os.environ and print would at least work the same on all platforms. There's also some case to be made here that maybe `echo` should be special-cased by cibuildwheel's evaluator and treat it more similarly to how shells treat it - as a builtin command. But that would add to the evaluator's complexity and require a separate PR so I don't know...

test/test_dependency_versions.py
jack1142: I thought about this a bit more and I think that the default (`auto`) architectures built on Apple Silicon should include `x86_64` since Rosetta is a thing and current default is `arm64` + `universal2` which is not ideal for older Pythons (which simply don't support Apple Silicon natively) as well as older pip versions.

Is there any benefit to building `x86_64` + `universal2` wheels over building `x86_64` + `arm64`? I feel like a reasonable default (`auto`) would be to build `x86_64` + `arm64`.
However, I suppose `x86_64` + `universal2` is fine too, just makes it a kind of an untypical case where we wouldn't be building wheels for the native architecture as we do in all the other cases. There's also an option of building `x86_64` + `universal2` + `arm64` but cibuildwheel's documentation seems to somewhat discourage that (+ it would increase build time somewhat, though building wheels on Apple Silicon takes the least time anyway :smile:).

jack1142: So I've tried changing `auto` to include all of the arches - `x86_64`, `arm64`, and `universal2`. It required some changes to the generated wheel file names in the `expected_wheels()` utility in the test suite and temporarily skipping (only for the sake of testing this) two tests that don't work properly because of #1193.
Once all of that was done, it worked flawlessly without the need for *this* adjustment in the test suite as well as another one that I had to make in this PR.
So if the decision on which architectures `auto` should build on macOS arm64 would be made, it shouldn't be a big issue to adjust this PR appropriately (presumably no changes other than reverts of unnecessary test suite changes would be required if the change to `auto` was done in a separate PR).

Diff: https://github.com/jack1142/cibuildwheel/compare/add_cirrus_ci_support...jack1142:cibuildwheel:try_x86
Build log: https://cirrus-ci.com/task/5300735129681920?logs=run_cibuildwheel_tests

unit_test/environment_test.py
joerick: So to summarise, the options are:

1. patch the test so USERPROFILE is included in the env
1. patch the test so that it uses `cmd /c echo` on windows
1. modify the bash evaluator to add 'echo' as a built-in command

My slight preference is still (2) I think. While we've found on CIrrusCI that the `echo` used is from msys2, so the USERPROFILE fix works there, we haven't needed that fix elsewhere. So I'm guessing that the other platforms have different `echo`s available on path. However, all platforms should have the same `cmd`, since that's part of windows, so that would bring uniformity to this test.

Yes, (3) is also an option, but there are lots of shell builtins, but generally they're backed by a system utility -  I'm not aware of this being a problem in practise. So let's keep that in the back pocket - if there's a bug report about missing `echo`, we can implement a builtin.

test/test_dependency_versions.py
joerick: Sorry, I forgot that the default `archs` would mean that Python 3.6/3.7 build nothing on arm64. 

The question of the default is a tricky one. My view when this decision was made is that we're designing defaults for a world where arm64 and x86_64 runners coexist. The best situation is to build+test x86_64 natively and arm64 natively, rather than using emulation - if we can help it. 

I'm still assuming that we're going to get arm64 runners on the major CI providers at some point. At that point, I think that the recommended configuration will have both a x86_64 runner and a arm64 runner. So I think we should design defaults that 'just work' for that.

That's where the current defaults come from. Since CirrusCI only has M1 runners, I can see that we might include an option in our example config: `CIBW_ARCHS: x86_64, arm64, universal2`.

The question of whether to default building both `universal2` and `arm64` on the arm runner is still open, in my mind. I'm still mostly of the opinion that building both is good, because universal2 packages are preferred in some contexts (e.g. distributable GUI apps), and that most packages are small and the extra build time isn't much of a hit. Of course, people building big libraries would be wise to change this to just build arm64.

jack1142: I imagine we would want the test suite to be altered so that on macOS arm64 it runs x86_64 builds in majority of the tests as well to allow detection of issues such as #1193. I can probably look into that.

unit_test/environment_test.py
jack1142: > While we've found on CIrrusCI that the `echo` used is from msys2, so the USERPROFILE fix works there, we haven't needed that fix elsewhere.

I'm reasonably sure that all Windows runners use `echo` from Git/msys2 but for some reason the bug is only reproducible on Cirrus CI (maybe it's specific to the version of Git/msys2?). I was not able to reproduce the hang locally in a Windows VM (without Git it does fail due to `FileNotFoundError` as `echo` is indeed missing) which would explain why it wouldn't necessarily be reproducible in all other CIs, even if they use `echo` from msys2 (which I would imagine they have to unless there's some other external program that provides `echo` on Windows).

jack1142: Either way, patching the test to use `cmd /c echo` on windows is fine I guess, I can make that change.

jack1142: Resolved by #1196.

test/test_dependency_versions.py
jack1142: Okay, so what do you think should be done about this?

henryiii: Personally, I think the default should be native only. Eventually no one will care about Intel / universal and will only produce native binaries again. Also, there are still workarounds related to cross compiling for universal (like requiring all dependencies to also be universal). And we don’t support it yet, but there are two ways to make a universal binary, the default way and merging two native binaries. And finally universal vs. native is a choice, and I’d you pick universal, you don’t need native.

So I’d recommend native being default, and only producing native binaries with it.

test/test_subdir_package.py
```diff
@@ -51,13 +51,13 @@ def test(capfd, tmp_path):
         add_env={
             "CIBW_BEFORE_BUILD": "python {project}/bin/before_build.py",
             "CIBW_TEST_COMMAND": "python {package}/test/run_tests.py",
-            # this shouldn't depend on the version of python, so build only CPython 3.6
-            "CIBW_BUILD": "cp36-*",
+            # this shouldn't depend on the version of python, so build only CPython 3.8
+            "CIBW_BUILD": "cp38-*",
```
mayeut, line 55:
```suggestion
            # this shouldn't depend on the version of python, so build only CPython 3.10
            "CIBW_BUILD": "cp310-*",
```
If bumping this, maybe it's worth bumping to 3.10 to further delay the time when it needs bumping again ?

test/test_subdir_package.py
```diff
@@ -51,13 +51,13 @@ def test(capfd, tmp_path):
...
         },
     )
 
     # check that the expected wheels are produced
-    expected_wheels = [w for w in utils.expected_wheels("spam", "0.1.0") if "cp36" in w]
+    expected_wheels = [w for w in utils.expected_wheels("spam", "0.1.0") if "cp38" in w]
```
mayeut, line 60:
```suggestion
    expected_wheels = [w for w in utils.expected_wheels("spam", "0.1.0") if "cp310" in w]
```

test/test_dependency_versions.py
joerick: > Also, there are still workarounds related to cross compiling for universal (like requiring all dependencies to also be universal)

Certainly the idea that we're forcing _more_ work on packagers to support universal2 cross-compilation alongside the native arm64 seems like a strong argument against universal2 by-default.

Yes, that seems like a well-reasoned position @henryiii. I'd be happy to go down the native-only route, too.

joerick: > Okay, so what do you think should be done about this?

Even if we do end up changing the default archs on arm64, this PR isn't the place to do it. I've created #1204 for that. Lets keep the archs defaults the same here.

henryiii:
Is this ready to review/go? Might be nice to get it in for the upcoming release?

jack1142:
> Is this ready to review/go? Might be nice to get it in for the upcoming release?

Sadly as I mentioned in the Discord server, my account on Cirrus CI got suspended, most likely by an automated system, so I can't really do anything until their support responds. I emailed them about this 2 days ago, but no response yet. It's possible that if Pradyun added the Cirrus CI app to the PyPA org, it would be possible to see the build output but it's also equally if not more likely that it just wouldn't run the build when it sees me as the one triggering it.

joerick:
@pradyunsg has added cirrus ci to this repo, so let's try again... perhaps try pushing an empty commit and see if cirrus picks it up?

jack1142:
My account has actually been unsuspended earlier today after I re-asked on their GitHub rather than their support email so their CI should just work now for me again. I'm going to merge `main` into this branch first and I'll update this PR to not make changes that are no longer necessary.

Note that Linux arm64 build still doesn't work due to cirruslabs/cirrus-ci-docs#1034, I think it might make sense to leave that work for a future PR.

test/test_dependency_versions.py
jack1142: We ended up with only building arm64 wheels by default (#1204) so this skip needs to remain here as Python 3.6 can only run on macOS arm64 through Rosetta which uses x86_64 wheels.

jack1142:
Alright, all review comments should be addresses and the tests pass. I dropped Linux arm64 support - I don't think it's worth blocking over this and once this PR is merged, a new issue should be created for that with explanation what is a potential blocker for it.

I updated the PR description appropriately and included links to successful build for example files. This PR should now be ready for review.

mayeut:
> Note that Linux arm64 build still doesn't work due to https://github.com/cirruslabs/cirrus-ci-docs/issues/1034, I think it might make sense to leave that work for a future PR.

Linux arm64 might also be solved by https://github.com/cirruslabs/cirrus-ci-docs/pull/1027, whichever comes first. It makes to leave that work for a future PR.

jack1142:
I was hoping to re-trigger CircleCI check due to https://status.circleci.com/incidents/3d31fnpthj7y but looks like that didn't work... Gonna try an empty commit.

jack1142:
Well, that definitely made the failing check disappear, I guess that's solved then :thinking:

joerick (approved):
This all looks good to me!

jack1142:
I went ahead and made #1240 to track Linux arm64 support on Cirrus CI.

Czaki (approved):

henryiii (approved):
I guess the intel Mac example is separate from the rest of the example because the arm64 one cross-compiles Intel wheels? At least for now, I'd guess the ideal workflow for most projects on Cirrus would be to have native wheels & run both runners - cross compiling is harder than native compiling. Eventually, as Intel runners become hard to find, then this workflow would be better.

Just a comment, looks great!

jack1142:
> I guess the intel Mac example is separate from the rest of the example because the arm64 one cross-compiles Intel wheels?

The cirrus-ci-minimal example file does not do any cross-compilation since the default arch on Apple Silicon is only arm64 (which was done in cibuildwheel 2.9.0).

> I'd guess the ideal workflow for most projects on Cirrus would be to have native wheels & run both runners - cross compiling is harder than native compiling.

Cirrus CI does not have an Intel runner. And in case of Apple Silicon, it's not really cross-compilation process per se since it uses Rosetta2 emulation and so it doesn't involve a typical cross-compilation toolchain.

README.md
```diff
@@ -50,16 +51,19 @@ Usage
...
+| Azure Pipelines | ✅    | ✅    | ✅      |           | ✅²       |
+| Travis CI       | ✅    |       | ✅      | ✅        | ✅²       |
+| AppVeyor        | ✅    | ✅    | ✅      |           | ✅²       |
+| CircleCI        | ✅    | ✅    |         |           | ✅²       |
+| Gitlab CI       | ✅    |       |         |           | ✅²       |
+| Cirrus CI       | ✅    | ✅³   | ✅      | ✅        | ✅        |
 
 <sup>¹ [Requires emulation](https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation), distributed separately. Other services may also support Linux ARM through emulation or third-party build hosts, but these are not tested in our CI.</sup><br>
+<sup>² [Uses cross-compilation](https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2). It is not possible to test `arm64` and the `arm64` part of a `universal2` wheel on this CI platform.</sup><br>
+<sup>³ [Uses cross-compilation](https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2). Thanks to Rosetta 2 emulation, it is possible to test `x86_64` and both parts of a `universal2` wheel on this CI platform.</sup><br>
```
jack1142, line 66: As I said in my previous comment, I'm not sure cross-compilation is the right term here, could anyone confirm whether "emulation" makes more sense here?
```suggestion
<sup>³ [Uses Rosetta 2 emulation](https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2). Thanks to the emulation, it is possible to test `x86_64` and both parts of a `universal2` wheel on this CI platform.</sup><br>
```

jack1142:
> Linux arm64 might also be solved by [cirruslabs/cirrus-ci-docs#1027](https://github.com/cirruslabs/cirrus-ci-docs/pull/1027), whichever comes first. It makes to leave that work for a future PR.

Good news, the linked issue got resolved and I've already made a configuration that so far appears to just work, I'm waiting for the tests on my fork to finish before pushing it here.

jack1142:
All tests ran in Cirrus CI succeeded, I've also updated the PR description with the new link to a build from `examples/cirrus-ci-minimal.yml` which also works fine.

It seems that one of the GitHub workflows failed due to an intermittent pip/PyPI issue but would have worked otherwise. I can't re-run it myself.

README.md
joerick: Well, technically it's both! Cross-compilation is right, the wheel is built using cross-compilation, but we can still test the wheel using Rosetta 2 emulation.

I think the existing copy is fine as-is.

examples/cirrus-ci-minimal.yml
```diff
@@ -20,6 +20,20 @@ linux_x86_task:
     - apt install -y python3-venv python-is-python3
   <<: *BUILD_AND_STORE_WHEELS
 
+linux_arm64_task:
+  name: Build Linux arm64 wheels.
```
joerick, line 24: Python calls this architecture `aarch64`, as does cibuildwheel, so it's probably best to stick to the Python convention here.

```suggestion
linux_aarch64_task:
  name: Build Linux aarch64 wheels.
```

.cirrus.yml
```diff
@@ -17,6 +17,19 @@ linux_x86_task:
     - apt install -y python3-venv python-is-python3
   <<: *RUN_TESTS
 
+linux_arm64_task:
```
joerick, line 20:
```suggestion
linux_aarch64_task:
```

joerick (commented):
Great! The Cirrus aarch64 builder is is 1.5x faster than the Travis one, and 5x faster than emulation on GHA!

test/test_dependency_versions.py
henryiii: Are you sure this needs to be here? The exact same situation occurs for Python 3.7, there is no (official) AS support, but a skip wasn’t added there.

henryiii (approved):

test/test_dependency_versions.py
jack1142: It's a confusingly written test - `3.7` is not part of `pytest.mark.parametrize` and so it isn't actually being run while `3.6` is.

This is the only part of the test that mentions 3.7, it's not mentioned in the `if` at the end of the test nor in `parametrize`. Do you want me to add or remove 3.7 from the test?

jack1142: Just to add onto this, here's how it fails on 3.6:
https://cirrus-ci.com/task/5307329800306688?logs=run_cibuildwheel_tests#L868

mayeut (approved):

joerick:
PR merged. 🎉

Stellar work on this @jack1142! Thank you.

PR #1196, opened 2022-07-25
Remove dependency on Unix tools in Windows test suite

jack1142:
See https://github.com/pypa/cibuildwheel/pull/1191#discussion_r928228837

### DIFF
--- a/unit_test/environment_test.py
+++ b/unit_test/environment_test.py
@@ -3,6 +3,7 @@ import sys
 
 from cibuildwheel.environment import parse_environment
 
+# this command is equivalent to Unix 'echo', but works on Windows too
 PYTHON_ECHO = f"'{sys.executable}' -c \"import sys; print(*sys.argv[1:])\""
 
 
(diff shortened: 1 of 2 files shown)


unit_test/environment_test.py
```diff
@@ -28,7 +29,11 @@ def test_inheritance():
 
 
 def test_shell_eval():
-    environment_recipe = parse_environment('VAR="$(echo "a   test" string)"')
+    if sys.platform == "win32":
+        env_string = 'VAR="$(cmd /C "echo a   test string")"'
```
joerick, line 33: I'm wondering why the change in quoting here? The original was deliberately quoted weird to check that the command is invoked with the right arguments

```suggestion
        env_string = 'VAR="$(cmd /C echo "a   test" string)"'
```

unit_test/environment_test.py
```diff
@@ -39,7 +44,11 @@ def test_shell_eval():
 
 
 def test_shell_eval_and_env():
-    environment_recipe = parse_environment('VAR="$(echo "$PREV_VAR" string)"')
+    if sys.platform == "win32":
+        env_string = 'VAR="$(cmd /C "echo $PREV_VAR string")"'
```
joerick, line 48: This one should also match the quoting of the other, if possible.

unit_test/environment_test.py
joerick: just chatted with @jack1142 in the discord. `cmd /c echo` actually _adds_ double-quotes into arguments with spaces.

```
Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)] on win32
>>> import subprocess
>>> subprocess.run(['cmd', '/C', 'echo', 'a string with spaces'])
"a string with spaces"
CompletedProcess(args=['cmd', '/C', 'echo', 'a string with spaces'], returncode=0)
>>> subprocess.run(['cmd', '/C', 'echo', 'a string with spaces', 'word', 'another string with spaces']) 
"a string with spaces" word "another string with spaces"
```

joerick: we decided to use `python -c 'import sys; print(*sys.argv[1:])'` instead.

henryiii (approved):

PR #1209, opened 2022-08-01
fix: update workflow

mayeut:
move the `pip-compile` command from `bin/update_dependencies.py` directly in `noxfile.py`

Alternative to #1207 
I prefer this one as it's probably less likely to break and does not require specific exclusions in pre-commit config.


henryiii (approved):

henryiii:
IMO this looks better.

PR #1210, opened 2022-08-01
chore: run update workflow on pull request

mayeut:
Allows to catch issues with the update workflow with related PRs.

### DIFF
--- a/.github/workflows/update-dependencies.yml
+++ b/.github/workflows/update-dependencies.yml
@@ -1,6 +1,15 @@
 name: Update dependencies
 
 on:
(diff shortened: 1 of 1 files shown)


joerick (approved):
I'm a little confused... so this would run the update workflow and ~push the results back to the PR branch?~

Ah, I see[ the run](https://github.com/pypa/cibuildwheel/runs/7640323475?check_suite_focus=true) now. So on PRs it runs the workflow, but doesn't push the results. Neat!

henryiii (approved):

PR #1217, opened 2022-08-06
fix: don't build `universal2` as a default on macOS arm64 runners

### LINKED ISSUE #1204: Should arm64 runners build both universal2 and arm64 wheels by default?
Continuing the discussion from https://github.com/pypa/cibuildwheel/pull/1191#discussion_r928749039

The current situation is that arm64 runners builds universal2 and arm64 by default. This discussion is whether we change that to just arm64.

> @joerick: The question of whether to default building both universal2 and arm64 on the arm runner is still open, in my mind. I'm still mostly of the opinion that building both is good, because universal2 packages are preferred in some contexts (e.g. distributable GUI apps), and that most packages are small and the extra build time isn't much of a hit. Of course, people building big libraries would be wise to change this to just build arm64.

> @henryiii: Personally, I think the default should be native only. Eventually no one will care about Intel / universal and will only produce native binaries again. Also, there are still workarounds related to cross compiling for universal (like requiring all dependencies to also be universal). And we don’t support it yet, but there are two ways to make a universal binary, the default way and merging two native binaries. And finally universal vs. native is a choice, and I’d you pick universal, you don’t need native. You can always add tags for the native arch’s to a universal binary. And pip won’t even ever download universal if both native wheels exist.
> 
> So I’d recommend native being default, and only producing native binaries with it.

Let's get a decision on this before we release CirrusCI support, which would be our first officially supported arm64 CI.

mayeut:
Per discussion in #1204, only build `arm64` on arm64 macOS runners.

Fixes #1204
Relates to #1191

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -66,10 +66,6 @@ class Architecture(Enum):
         if platform == "windows" and native_architecture == Architecture.AMD64:
             result.add(Architecture.x86)
 
-        if platform == "macos" and native_architecture == Architecture.arm64:
-            # arm64 can build and test both archs of a universal2 wheel.
-            result.add(Architecture.universal2)
-
         return result
 
     @staticmethod
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -53,11 +53,11 @@ to load on Apple Silicon.
     available.
 
 Generally speaking, because Pip 20.3 is required for the `universal2` wheel,
-most packages should provide both `x86_64` and `universal2` wheels for now.
-Once Pip 20.3+ is common on macOS, then it should be possible to ship only the
-`universal2` wheel.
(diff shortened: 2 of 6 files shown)


joerick:
Just pushed a few docs updates. The only other thing I was wondering about was removing this line: 

https://github.com/pypa/cibuildwheel/blob/03c8f5b741ca25a1ece914117bfe31b2eaa59c6f/test/utils.py#L203

...because we no longer expect it to be built by default. Or perhaps we need another parameter on `expected_wheels()` to enable it when we're specifying ARCH=universal2 in the tests.

joerick:
I pushed a commit that should fix the expected_wheels thing. Of course, we'll only know once we try it on an arm64 CI, which should be coming in #1191 !

henryiii (approved):

PR #1220, opened 2022-08-09
ci: use official nox action

henryiii:
The workflow should be tested in the PR due to #1210. :)


joerick (approved):

joerick:
Aside: CI is flaky at the moment. Seems to be an intermittent issue with PyPI, I've been seeing lots of errors like

> WARNING: Skipping page https://pypi.org/simple/filelock/ because the GET request got Content-Type: Unknown. The only supported Content-Types are application/vnd.pypi.simple.v1+json, application/vnd.pypi.simple.v1+html, and text/html

PR #1233, opened 2022-08-16
Add ril to projects.yml

Cryptex-github:
A great example on how to use cibuildwheel for Pyo3 Rust bindings that uses maturin and how to utilize Github Action caches for performance and auto release.


joerick (approved):
Great, thanks!

PR #1235, opened 2022-08-17
Update docs toolchain

joerick:
This PR updates our docs to run on the latest version of mkdocs. 

There are quite a few extra changes to support this, because I wanted to retain the sidebar experience from the 
previous version. However, in the process of reworking that, I think I happened upon some good style 
improvements.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,8 @@
 title: Changelog
 ---
 
+# Changelog
+
 ### v2.9.0
 
 _11 August 2022_
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -2,6 +2,8 @@
 title: Contributing
 ---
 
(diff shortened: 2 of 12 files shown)


henryiii (approved):
One problem I noticed is that the top bar on mobile doesn't stick to the screen - you have to go to the very top of the page to access the hamburger menu. It seems like it might be worth looking into a different theme (material for MkDocs, for example) that is closer to working for us instead of continuing to tweak this one? Anyway, this seems like a positive move.

joerick:
The toolbar on mobile doesn't stick to the top before this PR. Though we could consider it in the future.

Personally I find the top, left, and right navigation split in Material for mkdocs pretty confusing. Maybe that's just me. RtD is a little dated, but it's not worth changing for now IMO.

PR #1244, opened 2022-08-23
feat: support config-settings

henryiii:
Addresses https://github.com/pypa/cibuildwheel/issues/1227. Don't have much for tests yet, open to ideas.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -20,6 +20,7 @@ from .util import (
     get_build_verbosity_extra_flags,
     prepare_command,
     read_python_configs,
+    split_config_settings,
     unwrap,
 )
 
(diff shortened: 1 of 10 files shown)


joerick:
This looks good to me! Only thing I'm wondering about is, since the values use quotes and shlex.split, what happens with a config like this:

```toml
[tool.cibuildwheel.config-settings]
example = "a string with \"some quotes\" in it"
```

I don't know if that works, but... maybe that's an evil/unrealistic input, so if it's awkward to support we could just raise an error when a config-settings value has a double-quote in it (and request the user raise an issue on the tracker).

henryiii:
Would it be better to use `shlex.quote`? `_inner_fmt` could provide the quoted value too, as `quoted_v` or something.

joerick:
Yeah I think shlex.quote would work better. I think that shlex.split and shlex.quote should roundtrip perfectly.

joerick (approved):
Oh nice. I hadn't realised the shlex.quote fix would apply to environment too.

PR #1248, opened 2022-08-24
Add cibuildwheel config in issue template

CaselIT:
I think it would be useful to provide a place to link to the configuration of cibuildwheel in the issue template


joerick (commented):
Thanks. I'm kinda -0.5 on this... it's pretty easy to find a pyproject.toml in a project once we have a URL to the CI config, and the existence of the field in the issue template means the header is printed in all issues, even when it's not applicable because projects don't configure things with a config file. But if another maintainer like the idea we can try it.

PR #1258, opened 2022-09-06
chore: nicer tracebacks and dev improvements

henryiii:
- chore: drop [all] extra

This could be a hindrance to moving to static configuration, and isn't that useful or that great of an idea (the more deps in one environment, the more fragile it is). Let's just remove it.

- chore(bin): remove ghapi requirement

We already require PyGitHub, and that's fully typed, so let's just use that. Also cleaned up this script (not sure if we really need it, but let's keep it nice as long as we have it).

- chore: cleaner exception tracebacks

I wrote a flake8 check and description for this here: https://github.com/henryiii/flake8-errmsg. I'm not adding the check at the moment, just cleaning up the existing exceptions. I improved the text of a few exceptions, too. I think we shouldn't throw raw `Exception`s, but should at least do `RuntimeError`, but I didn't change any exception types. Edit: it now supports setting a length if you'd rather not do this for very short messages. But I think it's fine to be consistent).

### DIFF
--- a/cibuildwheel/functools_cached_property_38.py
+++ b/cibuildwheel/functools_cached_property_38.py
@@ -21,7 +21,7 @@ class cached_property(Generic[_T]):
         if self.attrname is None:
             self.attrname = name
         elif name != self.attrname:
-            msg = "Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
+            msg = f"Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
             raise TypeError(msg)
 
     @overload
--- a/cibuildwheel/functools_cached_property_38.py
+++ b/cibuildwheel/functools_cached_property_38.py
@@ -21,10 +21,8 @@ class cached_property(Generic[_T]):
         if self.attrname is None:
             self.attrname = name
         elif name != self.attrname:
-            raise TypeError(
-                "Cannot assign the same cached_property to two different names "
-                f"({self.attrname!r} and {name!r})."
-            )
+            msg = "Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
+            raise TypeError(msg)
 
     @overload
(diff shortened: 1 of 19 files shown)


cibuildwheel/functools_cached_property_38.py
```diff
@@ -21,10 +21,8 @@ def __set_name__(self, owner: type[Any], name: str) -> None:
         if self.attrname is None:
             self.attrname = name
         elif name != self.attrname:
-            raise TypeError(
-                "Cannot assign the same cached_property to two different names "
-                f"({self.attrname!r} and {name!r})."
-            )
+            msg = "Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
```
joerick, line 24: This one is missing a `f` on the string.

joerick:
```suggestion
            msg = f"Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
```

cibuildwheel/logger.py
```diff
@@ -197,14 +197,16 @@ def build_description_from_identifier(identifier: str) -> str:
     elif python_interpreter == "pp":
         build_description += "PyPy"
     else:
-        raise Exception("unknown python")
+        msg = "unknown python {python_interpreter!r}"
```
joerick, line 200:
```suggestion
        msg = f"unknown python {python_interpreter!r}"
```

joerick (approved):
Looks good, modulo a couple missing f-strings - suggestions below.

cibuildwheel/functools_cached_property_38.py
henryiii, line 24:
```suggestion
            msg = f"Cannot assign the same cached_property to two different names ({self.attrname!r} and {name!r})."
```

Not sure why GitHub won't let me see @joerick's commit in the files tab so I can batch it...

PR #1262, opened 2022-09-08
feat: add formatted indentifier printout

henryiii:
See #1261 - this would enable easier production of dynamic matrix outputs from `cibuildwheel`.

Example:

```console
$ cibuildwheel --print-build-identifiers '{{"CIBW_BUILD": "{identifier}", "CIBW_ARCHS":"{arch}", "os": "ubuntu-latest"}}' --platform linux | jq -s
[
  {
    "CIBW_BUILD": "cp36-manylinux_x86_64",
    "CIBW_ARCHS": "x86_64",
    "os": "ubuntu-latest"
  },
  {
    "CIBW_BUILD": "cp37-manylinux_x86_64",
    "CIBW_ARCHS": "x86_64",
    "os": "ubuntu-latest"
  },
  ...
```

You should be able to concat multiple platforms as well.

Todo:

- [ ] Verify this is a good idea / makes using this easier in #1261.
- [ ] Tests
- [ ] Any other values useful in the format string?
- [ ] Is there any way to use this from the action, so the version stays locked with the cibuildwheel invocation that runs the builds?
- [ ] This does not have a way to group smaller chunks of runs, sadly, but maybe that could be done via jq? I did include `version` if that's useful - so we have `version`, `identifier`, and `arch`. You'd have to combine the contents of the build field. But you could write a processor to do this at least much easier than now.

I could have done this by just processing the build identifier, and not touching `get_build_identifiers` - version, pp/cp, os, and arch are all obtainable by processing the identifier. Maybe that would be better. Edit: trying that in a new commit. Also added `impl` and `os`.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -100,8 +100,14 @@ def main() -> None:
 
     parser.add_argument(
         "--print-build-identifiers",
-        action="store_true",
-        help="Print the build identifiers matched by the current invocation and exit.",
+        action="store",
+        const="{identifier}",
+        nargs="?",
+        help="""
+            Print the build identifiers matched by the current invocation and
+            exit.  Optionally, specify a format string to print the identifiers.
+            Available replacements are {identifier}, {arch}, {version}. Default: {identifier}.
+        """,
     )
 
     parser.add_argument(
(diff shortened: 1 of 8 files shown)


cibuildwheel/__main__.py
```diff
@@ -208,9 +216,21 @@ def build_in_directory(args: CommandLineArguments) -> None:
         architectures=options.globals.architectures,
     )
 
-    if args.print_build_identifiers:
+    if args.print_build_identifiers is not None:
```
henryiii, line 219: Actually, I think this is safer:

```suggestion
    if args.print_build_identifiers:
```

That would make it easier to pass this though in the action as an empty string.

henryiii:
While we'd lose a little extra info, and it would not work with grouping (not that this does, but at least its closer), #1098 might allow this to be used directly without having to be able to change the formatting. You'd still have to format JSON to build the entry, but maybe it would be doable.

henryiii:
```
cibuildwheel --print-build-identifiers --platform linux | sed 's/\(^.*$\)/{"only": "\1", "os": "ubuntu-latest"}/g'
```

This could be fed into `--only`. Maybe we should add `--only` to the GHA?

PR #1263, opened 2022-09-08
chore: cleanup types

henryiii:
PyLance doesn't like raw Final (and Tuples hard-code the length if you don't specify it). 

Moved to #1270 to unblock this:

Also moved to raising SystemExit over sys.exit. Doesn't require an import & is just the implementation of `sys.exit` anyway. See [python: raise SystemExit (beginner - intermediate)](https://youtu.be/ZbeSPc5wL0g?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY).

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -63,13 +63,13 @@ __all__ = [
     "split_config_settings",
 ]
 
-resources_dir: Final = Path(__file__).parent / "resources"
+resources_dir: Final[Path] = Path(__file__).parent / "resources"
 
-install_certifi_script: Final = resources_dir / "install_certifi.py"
+install_certifi_script: Final[Path] = resources_dir / "install_certifi.py"
 
 BuildFrontend = Literal["pip", "build"]
 
-MANYLINUX_ARCHS: Final = (
+MANYLINUX_ARCHS: Final[tuple[str, ...]] = (
     "x86_64",
     "i686",
     "pypy_x86_64",
@@ -80,7 +80,7 @@ MANYLINUX_ARCHS: Final = (
     "pypy_i686",
 )
 
-MUSLLINUX_ARCHS: Final = (
+MUSLLINUX_ARCHS: Final[tuple[str, ...]] = (
     "x86_64",
     "i686",
     "aarch64",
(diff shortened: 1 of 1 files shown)


bin/sample_build.py
```diff
@@ -24,4 +24,6 @@
         check=True,
     )
 
-    sys.exit(subprocess.run([sys.executable, "-m", "cibuildwheel"], cwd=project_dir).returncode)
+    raise SystemExit(
+        subprocess.run([sys.executable, "-m", "cibuildwheel"], cwd=project_dir).returncode
+    )
```
joerick, line 29: This is poor form, IMO. I watched the video and follow the logic, but sys.exit is more idiomatic to me. The `raise` implies an error. I realise that the exception is how Python works internally, but I believe that it does that to conveniently reuse the try/finally semantics, and not because it should be viewed as an error. It comes across strangely in the case that this is `raise SystemExit(0)` - there, the program is exiting successfully but our code is raising an exception. 

More familiar is a function that never returns, like C's `exit()`. That's also how most (all?) other mainstream languages do it, e.g. [Rust](https://doc.rust-lang.org/std/process/fn.exit.html), [Ruby](https://stackoverflow.com/questions/30337972/how-to-tell-a-ruby-program-to-close-itself), [Node.js](https://nodejs.org/api/process.html#processexitcode), [Java](https://www.geeksforgeeks.org/system-exit-in-java/), [Go](https://gobyexample.com/exit)...

cibuildwheel/__main__.py
```diff
@@ -211,7 +211,7 @@ def build_in_directory(args: CommandLineArguments) -> None:
     if args.print_build_identifiers:
         for identifier in identifiers:
             print(identifier)
-        sys.exit(0)
+        raise SystemExit(0)
```
joerick, line 214: Also not super fond of this...

joerick (commented):
I like the stronger `Final` types - I have also been enduring PyLance errors for these! 

But I'm kinda -0 on most of the `raise SystemExit` changes. I've watched the video and I get it, but it just seems a little 'clever' to me. Importing `sys` isn't bad, and it can be useful information - the absence of a `sys` import is actually quite a useful signal to me that a Python file doesn't interact with the system and is more likely to be pure.

I'm -1 on uses of `raise SystemExit(0)`, as these look like errors and that isn't how they're intended to be read.

PR #1264, opened 2022-09-09
docs: only macOS platforms support macOS ARM

henryiii:
Pretty sure this was a mistake - we don't support cross compile ARM macOS on platforms where we don't support macOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,10 +55,10 @@ Usage
 |-----------------|-------|-------|---------|-----------|-----------|
 | GitHub Actions  | ✅    | ✅    | ✅      | ✅¹       | ✅²       |
 | Azure Pipelines | ✅    | ✅    | ✅      |           | ✅²       |
-| Travis CI       | ✅    |       | ✅      | ✅        | ✅²       |
+| Travis CI       | ✅    |       | ✅      | ✅        |           |
 | AppVeyor        | ✅    | ✅    | ✅      |           | ✅²       |
 | CircleCI        | ✅    | ✅    |         |           | ✅²       |
-| Gitlab CI       | ✅    |       |         |           | ✅²       |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,10 +55,10 @@ Usage
 |-----------------|-------|-------|---------|-----------|-----------|
 | GitHub Actions  | ✅    | ✅    | ✅      | ✅¹       | ✅²       |
 | Azure Pipelines | ✅    | ✅    | ✅      |           | ✅²       |
```
Czaki, line 57:
```suggestion
| GitHub Actions  | ✅    | ✅    | ✅      | ✅¹       | ✅³       |
| Azure Pipelines | ✅    | ✅    | ✅      |           | ✅³       |
```

README.md
```diff
@@ -55,10 +55,10 @@ Usage
...
-| Travis CI       | ✅    |       | ✅      | ✅        | ✅²       |
+| Travis CI       | ✅    |       | ✅      | ✅        |           |
 | AppVeyor        | ✅    | ✅    | ✅      |           | ✅²       |
 | CircleCI        | ✅    | ✅    |         |           | ✅²       |
```
Czaki, line 60:
```suggestion
| AppVeyor        | ✅    | ✅    | ✅      |           | ✅³       |
| CircleCI        | ✅    | ✅    |         |           | ✅³       |
```

Czaki (approved):
Maybe also fix superscript for ram host runners? it should be 3, not 2

henryiii:
@Czaki "2" is for cross-complies from Intel, and "3" is for cross-compiles from ARM. The far right column is not ARM host runners except for the case of Cirrus CI, it's Intel cross-compile.

Czaki:
I see. I was a little too optimistic when considering the availability of macos ARM hosts.

PR #1266, opened 2022-09-13
fix: better cross-platform auto archs

henryiii:
This improves the default arch selection when targeting a different platform.

If you run `cibuildwheel --platform linux --print-build-identifiers` from 64-bit Windows, I'd expect to see the identifiers for 64/32-bit Linux. I noticed this because I ran `cibuildwheel --platform windows --print-build-identifiers` from macOS, and got nothing - I expected the 64/32-bit Windows identifiers. However, this _does_ work if you run macos from linux, or vs. versa, as long as you are on Intel, due to a quirk of the names.

I've just fixed the auto arch discovery to be based on the target platforms's matching architecture. `auto` now uses the target OS instead of the current OS to compute the auto architectures.

(This also happens to be a bit nicer for #1261. It also should be _much_ nicer when using `--platform linux` to do builds with Docker from Windows!)

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -10,17 +10,11 @@ from .typing import Final, Literal, PlatformName, assert_never
 
 PRETTY_NAMES: Final = {"linux": "Linux", "macos": "macOS", "windows": "Windows"}
 
-ARCH_CATEGORIES: Final[dict[str, set[str]]] = {
-    "32": {"i686", "x86"},
-    "64": {"x86_64", "AMD64"},
-    "arm": {"ARM64", "aarch64", "arm64"},
-}
-
-PLATFORM_CATEGORIES: Final[dict[PlatformName, set[str]]] = {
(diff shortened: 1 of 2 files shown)


ddelange:
@joerick I wanted to give the 'advanced parallelization' setup from #1261 a try when I ran into this limitation. Any chance this PR could get a merge&release? Many thanks!

ddelange:
fwiw https://github.com/ddelange/asyncpg/actions/runs/3092263953/jobs/5003341864#step:4:43

in https://github.com/ddelange/asyncpg/pull/2

joerick:
@ddelange that looks to be an issue with the --only flag. I've raised a separate issue.

joerick:
@henryiii I refactored this a little, I was a bit confused with the names of variables. Hopefully this will make sense to both of us!

henryiii:
Okay, LGTM, miming changes is fine.

PR #1273, opened 2022-09-17
Revert to previous quoting behaviour for environment values in TOML

### LINKED ISSUE #1271: `sh: No such file or directory` in `before_all` for linux builds with cibuildwheel v2.10.0
### Description

After updating from `2.9.0` to `v2.10.0`, I get the following error for linux builds using a simple `before-all` command like `before-all = "ls"`:
```       
Running before_all...
    + sh -c ls
env: sh: No such file or directory
```
This seems to be related to the fact that we're updating the `PATH` environment variable:
```
[tool.cibuildwheel]
before-all = "ls"
environment = { PATH="$PATH:/usr/local/bin" }
```

joerick:
Fix #1271.

A (perhaps) temporary fix for #1271, reverting to the previous behaviour of wrapping environment variables with double-quotes, which seems to work in 95% of use cases.

There might be a better approach that handles the last 5% of strange/evil inputs. But for now, this should restore the ability to use single-quotes in values and to reference $PARAMs.

### DIFF
--- a/cibuildwheel/environment.py
+++ b/cibuildwheel/environment.py
@@ -4,6 +4,7 @@ import dataclasses
 from typing import Any, Mapping, Sequence
 
 import bashlex
+import bashlex.errors
 
 from cibuildwheel.typing import Protocol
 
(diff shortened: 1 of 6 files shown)


henryiii (approved):
Ahh, interesting.  You kept the new behavior for the new feature. Looks good.

Issue #1277, opened 2022-09-20
Unable to verify SSL certificate when trying to build `cp311-macosx_x86_64`


andyfaff:
### Description

I'm trying to build a wheel for `cp311-macosx_x86_64`. Part of the `CIBW_BEFORE_BUILD` retrieves a pre-built OpenBLAS library using the `urllib` module.

We're experiencing issues retrieving that library with the Python install that cibuildwheel provides for that target, SSL verification is failing. We don't have the issue on any other target.

```
++ python tools/openblas_support.py
  Traceback (most recent call last):
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/urllib/request.py", line 1348, in do_open
      h.request(req.get_method(), req.selector, req.data, headers,
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line 1282, in request
      self._send_request(method, url, body, headers, encode_chunked)
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line 1328, in _send_request
      self.endheaders(body, encode_chunked=encode_chunked)
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line 1277, in endheaders
      self._send_output(message_body, encode_chunked=encode_chunked)
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line 1037, in _send_output
      self.send(msg)
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line 975, in send
      self.connect()
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/http/client.py", line [145](https://github.com/scipy/scipy/pull/17046/checks#step:6:151)4, in connect
      self.sock = self._context.wrap_socket(self.sock,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 517, in wrap_socket
      return self.sslsocket_class._create(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 1075, in _create
      self.do_handshake()
[...]

PR #1279, opened 2022-09-24
Add a cloudflare URL to the SSL tests

joerick:
Testing out #1277.


joerick:
It's failing, but with 403 forbidden instead, not an SSL error. Need to find a cloudflare-hosted URL that doesn't mind being poked by urllib.request.

joerick:
This ended up being an issue with actions/setup-python: https://github.com/actions/setup-python/issues/512 - now fixed.

PR #1283, opened 2022-09-25
feature: allow testing cp38-macosx_arm64 wheel

mayeut:
Per discussion in https://github.com/pypa/cibuildwheel/pull/1169, the default installer used for cp38 is an Intel installer. It makes sense to skip testing arm64 wheels in this case.

However, if the user choose to manually install the universal2 CPython version (c.f. https://github.com/pypa/cibuildwheel/issues/1278#issuecomment-1256943685), then, tests shall be run on arm64.

This allows users that either target 11.0+ on intel, universal2 or only build for arm64 to get the arm64 wheel tested on AppleSilicon.

I'm not a fan of the way I added the test here but there's a test. Maybe the test suite should be updated to run fully with both installers ?

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -56,3 +56,19 @@ macos_arm64_task:
     - brew install python@3.10
     - ln -s python3 /opt/homebrew/opt/python@3.10/bin/python
   <<: *RUN_TESTS
+
+macos_arm64_cp38_task:
+  macos_instance:
+    image: ghcr.io/cirruslabs/macos-monterey-xcode
+
+  env:
+    PATH: /opt/homebrew/opt/python@3.10/bin:$PATH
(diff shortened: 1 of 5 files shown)


henryiii:
I've got an idea here. I'll try to get to it in a day or so.

joerick:
Thanks @mayeut. This provides a workaround for users with compiled dependencies that need native arm64 builds. I'm curious what @henryiii has in mind also...

> I'm not a fan of the way I added the test here but there's a test. Maybe the test suite should be updated to run fully with both installers ?

Yeah, maybe. I'll save a full review until I've seen the approach henryiii has in mind. We could also limit the cp38-arm64 test job to test_0_basic, test_testing, test_macos_archs to save some CI time.

henryiii (approved):
Playing with it a bit, I think we'll need to discuss some more, and I don't think this gets in the way & it might help.

joerick (approved):
Agreed. This is a good step, but maybe we'll do more here too.

joerick:
Thank you @mayeut!

PR #1291, opened 2022-10-03
chore: clean up a little typing

henryiii:
I was working on my lectures for this week (which are [over static typing](https://henryiii.github.io/se-for-sci/content/week5/typing)), and remembered we can clean this up. It doesn't even need a typing extensions bump; 4.1 is sufficient.

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -114,6 +114,7 @@ class Architecture(Enum):
         return all_archs_map[platform]
 
     @staticmethod
+    # pylint: disable-next=inconsistent-return-statements
     def bitness_archs(platform: PlatformName, bitness: Literal["64", "32"]) -> set[Architecture]:
         archs_32 = {Architecture.i686, Architecture.x86}
         auto_archs = Architecture.auto_archs(platform)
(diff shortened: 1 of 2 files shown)


joerick:
My goodness our tests are flaky at the moment. It's the pip error again.

henryiii:
Haha, yes. :) Currently running into a bug with pylint that it understands `-> NoReturn`, but not `-> Never`. IMO, the signature should have been `Never) -> NoReturn`, but it's not, and PyLint should understand that they are the same. It does in a few places, but not here. Not a failing check, but would be nice to fix.

cibuildwheel/architecture.py
```diff
@@ -114,6 +114,7 @@ def all_archs(platform: PlatformName) -> set[Architecture]:
         return all_archs_map[platform]
 
     @staticmethod
+    # pylint: disable-next=inconsistent-return-statements
     def bitness_archs(platform: PlatformName, bitness: Literal["64", "32"]) -> set[Architecture]:
```
henryiii, line 118: This is a really ugly place to put this lint disable comment. :/

henryiii: Should (hopefully) be temporary.

PR #1293, opened 2022-10-05
chore: minor cleanups from reverb

henryiii:
Some small cleanups inspired by `refurb`. It also doesn't like return after else, and I don't either, but left that off. :'(


joerick (approved):
These look good to me!

joerick:
> It also doesn't like return after else, and I don't either, but left that off. :'(

Haha! Well, I appreciate it :)

PR #1297, opened 2022-10-07
chore: move to hatchling

henryiii:
Moving over to using pypa/hatch's hatching.

Removed the `[mypy]` extra as it wasn't maintained and wasn't correct.


henryiii:
Sorry! I tried to add that last commit via the web editor (the WebAssembly one you get by pressing the period key) and it pushed it to master instead. Reverted. :flushed:

henryiii:
Does Cirrus have isolated builds disabled? Not sure why it's not picking up hatchling. It could be added - but why doesn't it pick it up?

.cirrus.yml
```diff
@@ -1,6 +1,7 @@
 run_tests: &RUN_TESTS
   install_cibuildwheel_script:
-    - python -m pip install -e ".[dev]" pytest-custom-exit-code
+    - pip install hatchling
```
henryiii, line 3:
```suggestion
    - python -m pip --version
    - python -m pip install hatchling
```

henryiii:
Something seems broken with homebrew * isolated installs on M1. Vaguely thought I remember hearing about that somewhere else.

joerick:
Yeah I can recreate that locally, too. Seems it only happens on homebrew python.

https://gist.github.com/joerick/42e9e31ceb5d1dfe43c69fb6a59436e2#file-gistfile1-txt-L5541-L5551

ofek:
There are conflicts btw

ofek:
Maybe one more try? 😅

ofek:
Strange job timeout

joerick:
Checks looks okay now, perhaps a retry fixed it. The other thing I think we'd need is an update to `bin/bump_version.py`, as it has references to setup.cfg. I'll push that now.

mayeut:
I cancelled the travis ci job (which would have failed on windows anyway)

henryiii:
> I cancelled the travis ci job (which would have failed on windows anyway)

Is that an issue with this?

mayeut:
> Is that an issue with this?

No, Windows has been failing the last couple builds.

henryiii:
Any objections to merging?

PR #1300, opened 2022-10-09
chore: test with Python 3.11

mayeut:
With 3.11.0 available in roughly a couple weeks, let's start testing as draft.
I will update this once 3.11 reaches GA.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, windows-latest, macos-11]
-        python_version: ['3.10']
+        python_version: ['~3.11.0-0']
     timeout-minutes: 180
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, windows-latest, macos-11]
-        python_version: ['~3.11.0-0']
+        python_version: ['3.11']
     timeout-minutes: 180
     steps:
     - uses: actions/checkout@v3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,7 +44,7 @@ repos:
   rev: v2.0.0
   hooks:
   - id: setup-cfg-fmt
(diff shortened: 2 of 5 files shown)


henryiii:
3.11 as max needs to be specified in pre-commit setup-cfg-fmt.

mayeut:
> 3.11 as max needs to be specified in pre-commit setup-cfg-fmt.

just read the docs of setup-cfg-fmt and saw that.

I won't bother to change this & will wait for 3.11 GA at which point, setup-cfg-fmt should be updated (or I'll do it then).
The important part is to check everything is working with 3.11 (no obvious reason why not but at least we'll know for sure).

henryiii:
I bumped it - it won't automatically update, the automatic classifier feature was removed from setup-cfg-fmt (I do prefer the explicit limit, though IMO this was one of the best features of setup-cfg-fmt).

henryiii:
IMO, this is ready to go in any time - I'd really like the trove classifier, since we support producing 3.11 wheels. Ideally, we should also be able to run on 3.11, and we can.

FYI, are we checking to make sure `MACOSX_DEPLOYMENT_TARGET is set to a lower value (10.9) than the version on which the Python interpreter was compiled` never shows up in our integration tests? Looks like it's showing up in https://github.com/mypyc/mypy_mypyc-wheels/actions/runs/3245210848/jobs/5322448904 (on all Pythons, actually).

mayeut:
> IMO, this is ready to go in any time - I'd really like the trove classifier, since we support producing 3.11 wheels. Ideally, we should also be able to run on 3.11, and we can.

I would have liked for the action to be tested on 3.11 as well before getting this in (& remove the `~3.11.0-0` specifier) but fair enough, it can probably be merged in as-is.

> FYI, are we checking to make sure...

I'm almost sure we don't & I think this was discussed in the past.

joerick (approved):

joerick:
Any last changes on this now that 3.11 is released? It would be neat to get it in before the next cibuildwheel release with 3.11-final.

.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, windows-latest, macos-11]
-        python_version: ['3.10']
+        python_version: ['~3.11.0-0']
```
henryiii, line 41:
```suggestion
        python_version: ['3.11']
```

PR #1301, opened 2022-10-09
fix(tests): use a specific `PIP_CACHE_DIR` per worker

### LINKED ISSUE #1254: CI failures from pip `WARNING: Skipping page https://pypi.org/simple/xxx/ because the GET request got Content-Type: Unknown`
Having seen more of this recently I investigated a little

> WARNING: Skipping page https://pypi.org/simple/xxx/ because the GET request got Content-Type: Unknown
ones.

My assumption is that it's this bug https://github.com/pypa/pip/issues/5345, related to pip concurrently accessing its cache - we do run tests in parallel.

This bug appears to be fixed in pip. However, we are still seeing it in cibuildwheel's CI - I'm not sure why that is, but ~it might be because we use an old version of pip to upgrade itself, and the old version of pip has this bug~ this doesn't appear to be the case!

Let's gather information about these failures here. Specifically, if we can figure out the version of pip that's outputting the error , that would be good.

mayeut:
This should most likely resolve random failures seen with pip when running tests in parallel.

fix #1254


test/utils.py
```diff
@@ -47,6 +49,18 @@ def cibuildwheel_get_build_identifiers(project_path, env=None, *, prerelease_pyt
     return cmd_output.strip().split("\n")
 
 
+def _update_pip_cache_dir(env: dict[str, str]) -> None:
+    if platform == "linux":
```
joerick, line 53:
```suggestion
    # Fix for pip concurrency bug https://github.com/pypa/pip/issues/11340
    # See https://github.com/pypa/cibuildwheel/issues/1254 for discussion.
    # We should be able to remove this once we drop Python 3.6.

    if platform == "linux":
```

joerick (approved):
I like this solution. Relying on the concurrency aspect of the bug rather than the pip versions makes it easier to implement.

Czaki (approved):

henryiii (approved):

test/utils.py
mayeut: If we add `# We should be able to remove this once we drop Python 3.6.` then we need to modify `test_dependency_versions.py` now (we will forget later).
I'm not even sure it's a good idea to drop this ever. We might be in a similar situation every time pip cache breaks between version (& they'll most likely drop python 3.7 soon after EOL in 8 months so we'll be in the same situation).

joerick: > I'm not even sure it's a good idea to drop this ever. We might be in a similar situation every time pip cache breaks between version (& they'll most likely drop python 3.7 soon after EOL in 8 months so we'll be in the same situation).

Yeah, maybe. I'm not sure what the root cause is yet. I'll keep an eye on the pip issue and if we get some assurances maybe it can be reverted at some point. Generally I'd like to keep the test environment as close as possible to user environments. 

Anyway I'm happy, let's get this merged and maybe our CI will mean something again :)

PR #1311, opened 2022-10-14
fix: hide root pip warning and upgrade check

### LINKED ISSUE #1310: Minor warning: still showing 'root' user warning on Linux
Still seeing this in 2.11.1, even after #1304:

```
  Successfully installed mypy_extensions-0.4.3 setuptools-65.4.1 tomli-2.0.1 types-psutil-5.9.5.1 types-setuptools-65.4.0.0 types-typed-ast-1.5.8 typing_extensions-4.4.0 wheel-0.37.1
  WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
  Running command Getting requirements to build wheel
```

Looks like it's missing from an env still?

henryiii:
Fixing this, try 2. Closes #1310.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -189,6 +189,8 @@ def build_in_container(
         log.step("Setting up build environment...")
 
         env = container.get_environment()
+        env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
+        env["PIP_ROOT_USER_ACTION"] = "ignore"
 
         # put this config's python top of the list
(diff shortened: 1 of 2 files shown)


mayeut:
We might want to add a small test for this ? (or like #1312, modify an existing one ?)

henryiii:
I'll rebase and add it after #1312 goes in.

PR #1312, opened 2022-10-14
fix: warning when building macOS arm64 wheels

mayeut:
When `MACOSX_DEPLOYMENT_TARGET` is not set by the user, cibuildwheel always defaults to `10.9`. This leads to a warning when building arm64 wheels.
This commit removes the warning by setting the default to `11.0` for arm64 builds.

Reported in https://github.com/pypa/cibuildwheel/pull/1300#issuecomment-1278396244

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -207,12 +207,14 @@ def setup_python(
         )
         sys.exit(1)
 
-    # Set MACOSX_DEPLOYMENT_TARGET to 10.9, if the user didn't set it.
-    # PyPy defaults to 10.7, causing inconsistencies if it's left unset.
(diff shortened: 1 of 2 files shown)


mayeut:
I chose to add the test in an existing one not increase test time.
I chose to test both stdout & stderr streams on purpose even if, today, it's on stderr.
I was a bit hesitant to check for a shorter string. If the string changes, the test won't be doing anything useful & maybe just checking for warning is better even if not targeted (thus could report unrelated warnings in the future).

joerick:
@mayeut i hope it's okay I added another test to this, just checking that the warning is detected by forcing it to be produced with a `MACOSX_DEPLOYMENT_TARGET=10.8`. It's only a 5s test.

henryiii (approved):

PR #1314, opened 2022-10-16
Suppress a pylint warning in Github Checks

joerick:


henryiii (approved):
I've been meaning to do this, thanks!

PR #1319, opened 2022-10-19
Fix docs on which dirs are mounted in container

### LINKED ISSUE #1318: Docs seem wrong about which folders are mounted in container
### Description

The FAQ ("Tips and tricks") section of the docs claims:

> The project directory is mounted inside the container as `/project`, the output directory for the wheels as `/output`.

It seems to me like neither of these folders are actually mounted: The only thing that is mounted according to the [docker/podman CLI args](https://github.com/pypa/cibuildwheel/blob/6c361f0193ca23e7355ad157a38e99903bc3c921/cibuildwheel/oci_container.py#L91) is `/host`, while `/project` is [copied in at the start](https://github.com/pypa/cibuildwheel/blob/6c361f0193ca23e7355ad157a38e99903bc3c921/cibuildwheel/linux.py#L149) and `/output` is [copied out at the end](https://github.com/pypa/cibuildwheel/blob/6c361f0193ca23e7355ad157a38e99903bc3c921/cibuildwheel/linux.py#L363).

For `/output` this is fairly inconsequential because it starts out empty anyway and the only purpose is to copy files out at the end, which they will. But claiming that `/project` is mounted when it isn't is confusing in that people will expect to be able to make changes there during the build and have them reflected in the project directory outside of the container (e.g. for caching), which won't happen as it's only copied in. So IMO this should be fixed.

A minimal reproducible example demonstrating this issue can be found [here](https://github.com/smheidrich/bug-cibuildwheel-proj-mount-minimal-example).

### Build log

https://github.com/smheidrich/bug-cibuildwheel-proj-mount-minimal-example/actions/runs/3285447450/jobs/5412547178

### CI config

https://github.com/smheidrich/bug-cibuildwheel-proj-mount-minimal-example/blob/main/.github/workflows/build.yml

### Related issues

- https://github.com/pypa/cibuildwheel/issues/466 mentions the same thing
- https://github.com/pypa/cibuildwheel/issues/1030

smheidrich:
Fixes #1318

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,7 +14,7 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
 
     `cibuildwheel` supports this by providing the [`CIBW_ENVIRONMENT`](options.md#environment) and [`CIBW_BEFORE_ALL`](options.md#before-all) options to setup the build environment inside the running container.
 
--   The project directory is mounted inside the container as `/project`, the output directory for the wheels as `/output`. In general, this is handled transparently by `cibuildwheel`. For a more finegrained level of control however, the root of the host file system is mounted as `/host`, allowing for example to access shared files, caches, etc. on the host file system.  Note that `/host` is not available on CircleCI due to their Docker policies.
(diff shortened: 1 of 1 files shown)


joerick (approved):
Thanks! It used to be a mount, but was changed to a copy because mounts don't always work.

PR #1321, opened 2022-10-20
Correct macOS ARM64 CPython 3.8 handling

altendky:


cibuildwheel/macos.py
```diff
@@ -207,10 +207,6 @@ def setup_python(
         )
         sys.exit(1)
 
-    # Set MACOSX_DEPLOYMENT_TARGET to 10.9, if the user didn't set it.
-    # PyPy defaults to 10.7, causing inconsistencies if it's left unset.
```
henryiii, line 211: Actually, pypy is now 10.10, so this is no longer required. We could probably drop the whole thing if we wanted. Though not all builders know to check the version Python was built with and take a maximum of the two.

cibuildwheel/resources/build-platforms.toml
```diff
@@ -76,8 +76,8 @@ python_configurations = [
   { identifier = "cp36-macosx_x86_64", version = "3.6", url = "https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.9.pkg" },
   { identifier = "cp37-macosx_x86_64", version = "3.7", url = "https://www.python.org/ftp/python/3.7.9/python-3.7.9-macosx10.9.pkg" },
   { identifier = "cp38-macosx_x86_64", version = "3.8", url = "https://www.python.org/ftp/python/3.8.10/python-3.8.10-macosx10.9.pkg" },
-  { identifier = "cp38-macosx_arm64", version = "3.8", url = "https://www.python.org/ftp/python/3.8.10/python-3.8.10-macosx10.9.pkg" },
-  { identifier = "cp38-macosx_universal2", version = "3.8", url = "https://www.python.org/ftp/python/3.8.10/python-3.8.10-macosx10.9.pkg" },
+  { identifier = "cp38-macosx_arm64", version = "3.8", url = "https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg" },
+  { identifier = "cp38-macosx_universal2", version = "3.8", url = "https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg" },
```
henryiii, line 80: This can't be changed. It's quite intentional. You can't build Universal 10.9 from the experimental ARM installers.

altendky: Just the `universal2` can't change?  Or the `arm64` as well.

altendky:
Since it seems I may or may not be going down a bad path, let me do what I should have done first and describe my scenario.  I was trying to expand my build matrix to include macOS ARM64 CPython 3.8.  I got a weird error where CMake was seemingly unable to find the GMP library for 3.8 while it worked fine in 3.9.  Eventually I realized that the Python in use was built for x86_64 only, or at least that's how I interpreted [`CONFIGURE_CFLAGS = "-arch x86_64 -g"`](https://github.com/Chia-Network/chiavdf/actions/runs/3289712847/jobs/5424696839#step:6:183).  Digging around I ended up back here noticing that the ARM64 and Universal2 options are both using the Intel installer.

Maybe I'm jumping too far ahead of what you wrote, but if you can't build a Universal2 for 3.8, should there be an error for that instead of just using the Intel Python build?  Or is the expectation that the Intel Python can 'cross' build the universal?

Though my actual need is the arm64 as we already have x86_64 wheels for macOS CPython 3.8.  So, am I onto anything here with this PR?  Or does it show I don't know what's going on and need to go back to my own builds to understand the issue.

joerick:
@altendky See https://github.com/pypa/cibuildwheel/issues/1278 for discussion of this.

The TLDR is this- on the different versions of CPython 3.8, you can only have one installed at once. So (at least currently), all the 3.8 URLs have to match, so we install the most compatible one, that goes back to macOS 10.9, but this installer is available for x86_64 only. It can cross-compile arm64 wheels. This is fine for simple C extensions, but for more complicated builds on arm64 it becomes an issue.

The current workaround works like this- if you're building arm64 wheels on a arm64 runner, before invoking cibuildwheel, install the universal2 version of Python on your runner, something like-

```
curl -o /tmp/Python38.pkg https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg
sudo installer -pkg /tmp/Python38.pkg -target /
```

Then cibuildwheel will detect that it's installed and use it instead. However, you probably don't want to build x86_64 wheels on this Python, unless you're happy with them being macOS 11+.

altendky:
Thanks so much for the explanation and sorry for the detour with this PR.  I think it will be fine for us to do Intel on the Intel runners and arm on the arm, so the preinstall suggestion should work fine.

PR #1322, opened 2022-10-21
Tweak default mac repair cmd: use delocate-wheel verbose flag instead of extra listdeps command

ankith26:
Minor PR to make a small tweak to cibuildwheels default mac wheel repair command. Instead of having to call `delocate-listdeps` separately, `delocate-wheel` can be invoked with a verbose flag, and this should print very similar information to the previous extra command.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -323,18 +323,17 @@ To work around this, use a different environment variable such as `REPAIR_LIBRAR
 
     ```yaml
     CIBW_REPAIR_WHEEL_COMMAND_MACOS: >
-        DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-listdeps {wheel} &&
-        DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -323,18 +323,17 @@ To work around this, use a different environment variable such as `REPAIR_LIBRAR
...
     ```toml
     [tool.cibuildwheel.macos]
-    repair-wheel-command = [
-        "DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-listdeps {wheel}",
-        "DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}"
-    ]
+    repair-wheel-command = """\
+    DYLD_LIBRARY_PATH=$REPAIR_LIBRARY_PATH delocate-wheel \
+    --require-archs {delocate_archs} -w {dest_dir} -v {wheel}\
+    """
```
joerick, line 336: The line split here makes sense as it's a much longer line i think

joerick (approved):
Thanks for this, it looks good to me. Nice to simplify the command a bit.

PR #1324, opened 2022-10-21
Fix #1161 - stdout is line-buffered during command execution

### LINKED ISSUE #1161: Stderr on docker run at the top
Currently, stdout is captured and placed in the error. But stderr is not, so it's displayed inline. `CalledProcessError` does support stderr too, so should we do the same thing with stderr? (Not sure I'm a big fan of separating stderr and stdout, sometimes they might be best in the original order).

Thoughts?

joerick:
Fix #1161 - making stdout line-buffered.

I can't think how to add a unit test for this because capfd keeps stdout and stderr separately. But on my machine, this-

```
python -m test.test_projects test.test_0_basic.basic_project /tmp/basic_project
(cd /tmp/basic_project && CIBW_BEFORE_BUILD="python -uc 'import sys, time; print(1, file=sys.stdout); time.sleep(0.1); print(2, file=sys.stderr)'" cibuildwheel --platform linux)
```

Prints 

```
2
1
```

before this PR and ...
```
1
2
```
after.

### DIFF
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -303,9 +303,11 @@ class OCIContainer:
                 return_code = int(return_code_str)
                 # add the last line to output, without the footer
                 output_io.write(line[0:footer_offset])
+                output_io.flush()
                 break
             else:
                 output_io.write(line)
+                output_io.flush()
 
         if isinstance(output_io, io.BytesIO):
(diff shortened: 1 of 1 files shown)


henryiii (approved):
There should be no harm, so maybe let's get this in before release?

PR #1328, opened 2022-10-29
chore: remove comment block in `docs/working-examples.md`

mayeut:
Per comment in https://github.com/pypa/cibuildwheel/pull/1302#issuecomment-1273856532

This produces a lot of line diffs on every update and was mostly used to debug star-based sorting.


henryiii (approved):
This should still get printed if you do a dry run, so :+1:.

joerick (approved):

mayeut:
travis-ci has trouble starting s390x jobs at the moment. Merging this without s390x travis-ci status (well, a red one).

PR #1346, opened 2022-11-22
fix: action using powershell on Windows

### LINKED ISSUE #1338: Add support to configure cibuildwheel shell (Github Actions).
### Description

Sometimes you don't want to run the build step under the bash shell, since on the bash shell things such as cl.exe are not part of the PATH.

This is relevant for projects building with meson, since meson is not smart enough to find your VS installation if it's not on the PATH.

### Build log

_No response_

### CI config

_No response_

henryiii:
Fixes #1338 by using the native GHA shell (powershell) on Windows. Using bash changes the path unexpectedly.

You can see it working for existing projects at https://github.com/pybind/scikit_build_example/actions/runs/3524040183/jobs/5908931400. "pipx run cibuildwheel" was verified to work with meson-python independently.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -52,8 +52,8 @@ runs:
         --spec "${{ github.action_path }}"
         cibuildwheel
         "${{ inputs.package-dir }}"
-        --output-dir "${{ inputs.output-dir }}"
-        --config-file "${{ inputs.config-file }}"
-        --only "${{ inputs.only }}"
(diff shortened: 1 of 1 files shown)


joerick (approved):
That quoting is mad. But it does look like it works! 

Let's keep an eye out for stdout/stderr mismatches on Windows. I'm not sure that we can get by without the `2>&1` logic, but maybe we can on Windows.

Issue #1342, opened 2022-11-15
Consider running abi3audit on abi3 wheels


hauntsaninja:
I saw this post earlier today: https://blog.trailofbits.com/2022/11/15/python-wheels-abi-abi3audit/

Seems like a thing that could be useful for cibuildwheel to run automatically, like we do for auditwheel

hugovk:
Note the https://github.com/trailofbits/abi3audit repo says:

> ⚠️ This project is not ready for general-purpose use! ⚠️

woodruffw:
Whoops, I meant to remove that 😅 -- it's still definitely a beta and there will be bugs, but I'd consider it ready for general use. I'll remove that note in the README.

henryiii:
I think we should at least mention this in docs, that would be a good first step.

woodruffw:
JFYI: We've dropped `abi3audit`'s minimum Python version down to 3.8, just in case that helps with any future integration plans! I haven't made a release for those changes yet, but I'll do so in the near future.

joerick:
Seeing as cbuildwheel is a collection of best-practise around wheel building, I'd like to integrate this. I don't think it actually fits the model of `repair-wheel-command`, because (correct me if I'm wrong):

- it doesn't ever change the wheel, just produces a report on stdout
- it doesn't need to be run in the build environment (e.g. the docker container), it can run wherever. 
- it doesn't support the build pythons we do (we're still 3.6+), and it's 3.8+.

So I'm thinking that it doesn't need to be installed in each build environment, rather, it can be run at the end of the build once all wheels have been built. 

1. We could have a command CIBW_AFTER_ALL analogous to CIBW_BEFORE_ALL, that runs in the container. The default would be a noop, unless there are abi3 wheels present, where it would be `pip install abi3audit && abi3audit {abi3wheels}`
2. We could run abit3audit in the host python instead. Once all containers are finished, check for abi3 wheels, if there are any, install abi3audit into a venv and check them there. This doesn't feel like an CIBW_AFTER_ALL-style command, there's not really any point for a user to run anything else at this point, since they could just run a command after cibuildwheel instead. But the user might want to disable the check (or change options?) so it might have an option like `CIBW_ABI3AUDIT=enabled|disabled|custom-command --with --flags {abi3wheels}`

Curious if anyone has any strong opinions/preferences on the above.

[...]

PR #1347, opened 2022-11-22
Suggest running abi3audit in docs

hauntsaninja:
Suggested in #1342

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -118,7 +118,7 @@ The CPython Limited API is a subset of the Python C Extension API that's declare
 
 To create a package that builds ABI3 wheels, you'll need to configure your build backend to compile libraries correctly create wheels with the right tags. [Check this repo](https://github.com/joerick/python-abi3-package-sample) for an example of how to do this with setuptools.
 
-You could also consider running [abi3audit](https://github.com/trailofbits/abi3audit) against the produced wheels in order to check for abi3 violations or inconsistencies.
+You could also consider running [abi3audit](https://github.com/trailofbits/abi3audit) against the produced wheels in order to check for abi3 violations or inconsistencies. You can run it alongside the default in your [CIBW_REPAIR_WHEEL_COMMAND](options.md#repair-wheel-command).
 
 ### Packages with optional C extensions {: #optional-extensions}
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -118,6 +118,8 @@ The CPython Limited API is a subset of the Python C Extension API that's declare
 
 To create a package that builds ABI3 wheels, you'll need to configure your build backend to compile libraries correctly create wheels with the right tags. [Check this repo](https://github.com/joerick/python-abi3-package-sample) for an example of how to do this with setuptools.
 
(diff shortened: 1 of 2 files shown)


Czaki:
could this be used in `repair-wheel-command`? If Yes, then maybe suggest this in docs?

hauntsaninja:
I think specifying it in repair-wheel-command would override the default of running auditwheel / delocate-wheel instead of running abi3audit in addition, which would be undesirable?

henryiii:
You'd want to `&&` them (or make a TOML list).

henryiii:
It would be another example here: https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command

(we used to have a mutli-line option there before, it would be nice to have an example like that again)

hauntsaninja:
Unless I'm messing up how I'm running abi3audit, I think this is blocked on https://github.com/trailofbits/abi3audit/issues/23

joerick:
Looks like that issue was fixed - are you able to incorporate the above @hauntsaninja ?

docs/faq.md
```diff
@@ -118,6 +118,8 @@ The CPython Limited API is a subset of the Python C Extension API that's declare
 
 To create a package that builds ABI3 wheels, you'll need to configure your build backend to compile libraries correctly create wheels with the right tags. [Check this repo](https://github.com/joerick/python-abi3-package-sample) for an example of how to do this with setuptools.
 
+You could also consider running [abi3audit](https://github.com/trailofbits/abi3audit) against the produced wheels in order to check for abi3 violations or inconsistencies.
```
joerick, line 121: It would be nice to link to the repair-command option here, so user can find the example.
```suggestion
You could also consider running [abi3audit](https://github.com/trailofbits/abi3audit) against the produced wheels in order to check for abi3 violations or inconsistencies. You can run it alongside the default in your [CIBW_REPAIR_WHEEL_COMMAND](options.md#repair-wheel-command).
```

docs/options.md
```diff
@@ -915,10 +917,12 @@ Platform-specific environment variables are also available:<br/>
     repair-wheel-command = "auditwheel repair --lib-sdir . -w {dest_dir} {wheel}"
 
     # Multi-line example
+    # Use abi3audit to catch issues with Limited API wheels
     [tool.cibuildwheel]
     repair-wheel-command = [
       'python scripts/repair_wheel.py -w {dest_dir} {wheel}',
       'python scripts/check_repaired_wheel.py -w {dest_dir} {wheel}',
+      'pipx run abi3audit --strict --report {wheel}',
     ]
```
joerick, line 926: Adding to this example is perhaps a little confusing - I think we should show an other multiline example that extends the default, something like

```
    # Use abi3audit to catch issues with Limited API wheels
    [tool.cibuildwheel.linux]
    repair-wheel-command = [
      "pipx run abi3audit --strict --report {wheel}",
      "auditwheel repair -w {dest_dir} {wheel}",
    ]
    [tool.cibuildwheel.macos]
    repair-wheel-command = [
      "pipx run abi3audit --strict --report {wheel}",
      "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}",
    ]
    [tool.cibuildwheel.windows]
    repair-wheel-command = "pipx run abi3audit --strict --report {wheel}"
```

PR #1351, opened 2022-11-25
Fix gitlab windows CI config

joerick:
Gitlab Windows CI setup is broken. Let's try to fix it...


joerick:
This is pretty obviously just some debug code left in by mistake, so I'm gonna merge this.

PR #1352, opened 2022-11-26
Improve the formatting of the preamble

joerick:
The preamble has been getting a bit scruffy lately, this tidies it up. 

- fixed the issue with environment thinking it was overridden when it wasn't
- objects like BuildSelector provide a custom, nested output, which is more friendly than the `repr`
- overrides are grouped when they have the same value
- using `str` rather than `repr` for other objects, makes e.g. Paths easier to read
- options that are set to their defaults are dimmed using an ANSI code

## Simple example

Here I'm setting `CIBW_ENVIRONMENT="a=1 b=2" CIBW_PLATFORM=linux`

### Before
image

### After
image

## more complicated example 

This example has a pyproject.toml:

```
[tool.cibuildwheel]
build = ["cp38*", "cp37*"]
skip = ["*musllinux*"]
environment = {FOO="BAR"}

test-command = "pyproject"

manylinux-x86_64-image = "manylinux1"

environment-pass = ["EXAMPLE_ENV"]

[tool.cibuildwheel.macos]
test-requires = "else"

select = "cp37*"
test-command = "pyproject-override"
manylinux-x86_64-image = "manylinux2014"
```

### Before
image

### After
image

---

At a code level, worth noting is that the Options object is passed in an `environ` object, rather than reading it from `os.environ`. This allows us to construct a 'default' Options object to compare with. Also, it requires less monkey-patching to test, which is nice.


joerick:
image

ANSI 'gray' (`\033[90m` aka 'bright black') doesn't look good on github mobile for some reason, i guess it's just interpreted as 'black'. I'll try a mid-gray from the [256-color palette](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) instead.

joerick:
Github actions on web appears to forget the current ANSI color on newline. I'll need to refresh the color on each newline.

cibuildwheel/options.py
```diff
@@ -362,13 +380,21 @@ def _inner_fmt(k: str, v: Any, table: TableFmt) -> Iterator[str]:
 
 
 class Options:
-    def __init__(self, platform: PlatformName, command_line_arguments: CommandLineArguments):
+    def __init__(
+        self,
+        platform: PlatformName,
+        command_line_arguments: CommandLineArguments,
+        environ: Mapping[str, str],
```
henryiii, line 387: Nitpick, but this is almost always called `env` rather than `environ` when passed and carried around (like in subprocess, for example), so I've usually called it `env` too. `environ` to me indicates the current environment, while `env` is a passed or modified environment. I am a fan of passing this around, but I like the `env` name a hair better, due to usage elsewhere.

cibuildwheel/options.py
```diff
@@ -594,37 +620,122 @@ def check_for_deprecated_options(self) -> None:
...
+            k: self.option_summary_value(v) for k, v in (overrides or {}).items()
+        }
+        # if the override value is the same as the non-overridden value, don't print it
+        overrides_value_strs = {k: v for k, v in overrides_value_strs.items() if v != value_str}
+
+        has_been_set = (value_str != default_value_str) or overrides_value_strs
+        c = log.colors
+
+        result = ""
+        result += c.gray if not has_been_set else ""
```
henryiii, line 688:
```suggestion
        result = c.gray if not has_been_set else ""
```

? Otherwise, I'm not sure this is a great use of a ternary, I'd probably just waste one more line and use a normal if.

cibuildwheel/util.py
```diff
@@ -655,3 +672,31 @@ def chdir(new_path: Path | str) -> Generator[None, None, None]:
         yield
     finally:
         os.chdir(cwd)
+
+
+def fix_ansi_codes_for_github_actions(text: str) -> str:
+    """
+    Github Actions forgets the current ANSI style on every new line. This
+    function repeats the current ANSI style on every new line.
```
henryiii, line 680: This is weird - though I see how it could happen, GH does a lot of optimization for logs. Is there a GH ticket or commend anywhere? I've not generally noticed GH breaking elsewhere with color, but maybe most things like compilers and pytest do per-line coloring too.

henryiii (approved):
Looks nice, a few minor nit picks / comments that you can ignore or address at your discretion. :)

cibuildwheel/options.py
joerick: Ah, interesting. I don't usually use `environ` either, but I felt it was clearer when there are options called `environment` kicking around in the Options object. `environ` felt a lot closer to 'host environment' rather than 'environment option'. `env` is somewhere in the middle IMO.

cibuildwheel/options.py
joerick: Yep 👍  this might just be a remnant from some refactoring

cibuildwheel/util.py
joerick: There is not. I should raise an issue somewhere. Not sure where though. Maybe I'll just raise it at actions/runner and see what they say.

joerick: Raised an issue here: https://github.com/orgs/community/discussions/40864

cibuildwheel/options.py
joerick: I've just had another look through the code and I can't see anywhere that the `environment` option and `env` would get confused. So I'll go with the more common convention.

PR #1372, opened 2022-12-11
feat: add PyPy macOS arm64

### LINKED ISSUE #1369: Update PyPy versions, add macos arm64 PyPy
### Description

PyPy has released version 7.3.10 which includes macos arm64 versions. This release only has python 3.8 and 3.9.

### Build log

_No response_

### CI config

_No response_

mayeut:
Add PyPy macOS arm64
Cross compilation is not supported.

fixes #1369


mayeut:
Let's see what fails with a simple update.

mattip:
It seems the appveyor 10.14 image [falls afoul](https://ci.appveyor.com/project/joerick/cibuildwheel/builds/45814406/job/dchdwjn3rr28jy10#L945) of [PyPy issue 3880](https://foss.heptapod.net/pypy/pypy/-/issues/3880). There is a 10.15 Catalina [appveyor image](https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds). Would it be acceptable to update or are there reasons to stay with 10.14 Mojave? Github Actions [has deprecated all 10.x images](https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/). Note this is just for the platform to build the wheels, `delocate` still makes wheels that are compatible with the `macosx_10_9_x86_64` tag.

henryiii:
That would just be for our tests; downstream projects will stop working unless they also update to a 10.15 image too if we make this change. IMO, that's fine, but I don't feel strongly about it if someone else does. (Also IMO, PyPy should have a way to release patch releases that are separate from minor releases for these sorts of fixes ;) )

henryiii:
I'd also be okay to update to (at least) MSVC 2017+ for all examples - we've been dropping 2015 support elsewhere (pybind11, scikit-build) as GHA dropped support a while ago (And Python 3.7+ really should use 2017+ anyway, and it works for older ones too)

mattip:
CI is passing

joerick (dismissed):

.github/workflows/test.yml
```diff
@@ -73,6 +73,7 @@ jobs:
         output-dir: wheelhouse
       env:
         CIBW_ARCHS_MACOS: x86_64 universal2 arm64
+        CIBW_SKIP: pp*-macosx_arm64
```
joerick, line 76: Hmmm. on further consideration this does seem a little like a footgun - many people will have cross-compilation written into their configs, which will fail when this is added. Can we perhaps provide a softer failure on macOS when pp*_arm64 is erroneously selected on x86_64?

joerick: e.g. when we attempt to build pp-arm64 on x86_64, we print-

```
warning: pp38-macosx_arm64 was selected but can't be built on this architecture (x86_64). Deselect these builds using CIBW_SKIP=pp*-macosx_arm64 to remove this warning.
```

I realise this is a special case, but I expect many people to hit a failed build otherwise.

On this other hand, we could say that we'd rather error out for consistency. But since our example configs have included this since arm64 support was added, it's gonna hit a lot of people.

joerick: We could even downgrade that from a warning to a notice, since there's no real problem with not having skipped pp*_arm64.

henryiii: Actually, since we don't support pypy cross compilation, I don't even know if we need a notice - it should simply not be a selectable option on a non-native platfrom, IMO, unless we add cross compilation.

mayeut: I went with @henryiii proposal to filter out this configuration. Tests were updated to check this, let's see CI results.

joerick: I'd still be in favour of a notice, since it's the only build identifier that's not selectable in this way.

cibuildwheel/macos.py
```diff
@@ -83,6 +83,23 @@ def get_python_configurations(
...
+            "arm64": {"arm64", "x86_64"},
+            "x86_64": {"x86_64"},
+        }[platform.machine()]
+    )
+    python_configurations = [
+        c
+        for c in python_configurations
+        if (not c.identifier.startswith("pp"))
+        or any(c.identifier.endswith(a.value) for a in pypy_architectures)
+    ]
```
joerick, line 101: maybe this would work and be simpler?

```suggestion
    python_configurations = [
        c
        for c in python_configurations
        if not (c.identifier.startswith("pp") and c.identifier.endswith('arm64') and platform.machine() == 'x86_64')
    ]
```

mayeut: yes, I'm going with a simpler version.

joerick:
I added a quiet notice about the auto-skipping of these builds - hope that works for you @mayeut and @henryiii

PR #1377, opened 2022-12-15
fix: not skipping the tests stops the copy (Windows ARM)

henryiii:
If you don't skip the tests, the final copy of the wheels is skipped after printing the warning message when cross-compiling for Win ARM!

IMO, these should be refactored to one function per action (prepare, build, repair, test, finalize, something like that) instead of massive, all-in-one function that's hard to track; it would make bugs like this harder to make. Blocks should not be more than a few lines.


joerick (approved):
Yikes! Yes that's a dodgy copy/paste indeed. 

On the refactor note, hard to say without seeing it, but you might be right. Consistency across the three platforms (e.g. function names) would also be nice.

PR #1388, opened 2023-01-03
ci: use normal AppVeyor macOS image

mayeut:
use normal AppVeyor macOS image i.e. catalina (10.15) per https://github.com/pypa/cibuildwheel/pull/1372#issuecomment-1369773672 comment.

towards #1372 & #1381 

Created a PR cherry-picking @henryiii commits (I think this might be cleaner looking at commit history depending how it would be merged).

Order for merging would be this PR then #1381 and then #1372


henryiii (approved):

joerick (approved):

henryiii:
The AppVeyor failure is unrelated (in the linux run).

PR #1402, opened 2023-01-19
Examples: Updated info for pillow_heif

bigcat88:
Some people will be interested in additional examples for Cirrus CI and how to include third party libraries using MSYS2 for Windows.
For Cirrus, use a slightly different technique than scipy to show another way to get wheels out of it.


henryiii:
This is a generated file, changes to it would get wiped. The change needs to go into `docs/data/projects.yml`.

PR #1405, opened 2023-01-20
chore: use Ruff

henryiii:
- chore: use ruff
- chore: use ruff (manual changes)

Is 10-100x faster (written in Rust), combines flake8 and lots of plugins, pyupgrade, isort, pygrep, yesqa, and some of flake8 (so pre-commit setup is faster - also just a simple binary, I think). And pyproject.toml configured! Also can do auto-fixes for quite a few errors.Ran into a few minor bugs that have already been fixed upstream. Used by Pandas, FastAPI, SciPy, Build, etc. More pre-commit friendly, since we don't have breakage due to plugins not being pinned (because they are all built-in, so there's only the one pre-commit ref).

Suggested in https://github.com/pypa/cibuildwheel/pull/1399#issuecomment-1397340132.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,32 +13,16 @@ repos:
   - id: mixed-line-ending
   - id: trailing-whitespace
 
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.3.1
-  hooks:
-  - id: pyupgrade
-    args: ["--py37-plus"]
-    exclude: ^cibuildwheel/resources/.*py$
-
-# Autoremoves unused imports
-- repo: https://github.com/hadialqattan/pycln
-  rev: v2.1.2
-  hooks:
-  - id: pycln
-    args: [--all]
-    stages: [manual]
-
-- repo: https://github.com/PyCQA/isort
-  rev: 5.11.4
-  hooks:
-  - id: isort
-    args: ["-a", "from __future__ import annotations"]
-    exclude: ^cibuildwheel/resources/.*py$
-
 - repo: https://github.com/psf/black
   rev: 22.12.0
   hooks:
   - id: black
+
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: v0.0.237
+  hooks:
+  - id: ruff
+    args: ["--fix"]
 
 - repo: https://github.com/asottile/setup-cfg-fmt
   rev: v2.2.0
@@ -73,29 +57,6 @@ repos:
     exclude: ^cibuildwheel/resources/.*py$
     args: ["--python-version=3.10"]
     additional_dependencies: *mypy-dependencies
-
-- repo: https://github.com/asottile/yesqa
-  rev: v1.4.0
-  hooks:
-  - id: yesqa
-    additional_dependencies: &flake8-dependencies
-    - flake8-bugbear
-
-- repo: https://github.com/PyCQA/flake8
-  rev: 6.0.0
-  hooks:
-  - id: flake8
-    exclude: ^cibuildwheel/resources/
-    additional_dependencies: *flake8-dependencies
-
-- repo: https://github.com/pre-commit/pygrep-hooks
-  rev: v1.9.0
-  hooks:
-  - id: python-check-blanket-noqa
-    stages: [manual]
-  - id: python-no-log-warn
-  - id: python-no-eval
-  - id: python-use-type-annotations
 
 - repo: https://github.com/shellcheck-py/shellcheck-py
   rev: v0.9.0.2
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -460,7 +460,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     env=env,
                     capture_stdout=True,
                 ).strip()
-                testing_archs: list[Literal["x86_64", "arm64"]]
+                testing_archs: list[Literal["x86_64", "arm64"]]  # noqa: F821
 
                 if config_is_arm64:
                     testing_archs = ["arm64"]
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -44,7 +44,7 @@ from .util import (
 
 @dataclasses.dataclass
 class CommandLineArguments:
-    platform: Literal["auto", "linux", "macos", "windows"] | None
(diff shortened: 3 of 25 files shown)


test/test_pure_wheel.py
```diff
@@ -35,8 +35,7 @@ def test(tmp_path, capfd):
     pure_python_project.generate(project_dir)
 
     with pytest.raises(subprocess.CalledProcessError):
-        actual_wheels = utils.cibuildwheel_run(project_dir)
-        print("produced wheels:", actual_wheels)
+        print(utils.cibuildwheel_run(project_dir))
```
henryiii, line 38:
```suggestion
        print("produced wheels:", utils.cibuildwheel_run(project_dir))
```

This is only a debug printout if it fails, but missed this.

unit_test/main_tests/main_options_test.py
```diff
@@ -29,14 +30,16 @@ def test_output_dir(platform, intercepted_build_args, monkeypatch):
     assert intercepted_build_args.args[0].globals.output_dir == OUTPUT_DIR.resolve()
 
 
-def test_output_dir_default(platform, intercepted_build_args, monkeypatch):
+@pytest.mark.usefixtures("platform")
+def test_output_dir_default(intercepted_build_args):
```
henryiii, line 34: The recommendation was to name the fixtures with an underscore if they don't return anything. I instead used this method.

Czaki: I like this

Czaki (approved):

.pre-commit-config.yaml
```diff
@@ -13,27 +13,11 @@ repos:
...
-    stages: [manual]
-
-- repo: https://github.com/PyCQA/isort
-  rev: 5.11.4
-  hooks:
-  - id: isort
-    args: ["-a", "from __future__ import annotations"]
-    exclude: ^cibuildwheel/resources/.*py$
+  - id: ruff
+    args: ["--fix"]
```
Czaki, line 20: Maybe move this after black. 
Then there will be a lower probability that some formatting rule would be reported.

henryiii: Okay. Though Ruff has auto-fixers too.

.pre-commit-config.yaml
```diff
@@ -13,33 +13,17 @@ repos:
...
 - repo: https://github.com/psf/black
   rev: 22.12.0
   hooks:
   - id: black
 
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: v0.0.226
+  hooks:
+  - id: ruff
+    args: ["--force-exclude", "--fix"]
```
henryiii, line 25: The `--force-exclude` flag can be dropped in the next version of the action, https://github.com/charliermarsh/ruff-pre-commit/pull/20

Czaki: Wait 1-2 days and the next release will land.

henryiii: I'd expect so. :)

henryiii: Actually, it did, an hour ago. :)

.pre-commit-config.yaml
henryiii: Ruff after black caused a double fix: https://results.pre-commit.ci/run/github/525643569/1674247644.l9WEjn23TbWUFrKVbXv-2w (The other way might too, but this way certainly did)

henryiii: It's because the fix rewrite created unformatted code. I think Ruff -> Black might be a better order.

joerick (commented):
Looks cool, fewer tools is better. And indeed it is crazy fast. Makes it much more suitable for a pre-commit hook.

cibuildwheel/options.py
```diff
@@ -555,13 +555,11 @@ def build_options(self, identifier: str | None) -> BuildOptions:
...
-                        image = pinned_images["manylinux2014"]
-                    elif config_value in pinned_images:
-                        image = pinned_images[config_value]
-                    else:
-                        image = config_value
+                    image = (
+                        pinned_images.get(config_value, config_value)
+                        if config_value
+                        else pinned_images["manylinux2014"]
+                    )
```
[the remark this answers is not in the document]
Czaki: Maybe old ruff was used. SIM108 was changed to not trigger if the new code will not fit in a single line. https://github.com/charliermarsh/ruff/pull/1802

cibuildwheel/macos.py
```diff
@@ -460,7 +460,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     env=env,
                     capture_stdout=True,
                 ).strip()
-                testing_archs: list[Literal["x86_64", "arm64"]]
+                testing_archs: list[Literal["x86_64", "arm64"]]  # noqa: F821
```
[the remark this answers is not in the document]
charliermarsh: (Please feel free to keep filing stuff as questions and issues come up. Glad we could resolve this one :))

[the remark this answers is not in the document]
henryiii: It's not resolved, it's just a duplicate of an existing issue (https://github.com/charliermarsh/ruff/issues/2005) you helped me open a couple of days ago. :)

cibuildwheel/options.py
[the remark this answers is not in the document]
henryiii: Yes, old, ancient ruff was used. :P

I'll rerender this later with newer Ruff. Don't want to wipe my usefixtures changes.

cibuildwheel/macos.py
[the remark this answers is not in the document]
henryiii: It will go away (Ruff will actually remove it) when this does get fixed, so fine to leave it in here. But we could probably turn off F821 as well, since we are fully typed. It's probably mostly useful if you have code not checked by MyPy.

Czaki:
What is need to be done here?

joerick (approved):
looks good!

Czaki:
Ruff readme consistent with real state :) 
(cibuildwheel was already mentioned as a project that uses Ruff)

charliermarsh:
Oh sorry, I didn't add that, I should've checked first!

Czaki:
> Oh sorry, that was sent in a PR, I should've checked first!

We fixed it on our side :P

Issue #1415, opened 2023-02-10
When building an arm64 wheel on macos-12 runner on GitHub Actions, the artifact filename contains x86_64


azagniotov:
### Description

Hello Team!

As per subject. When building an arm64 wheel on `macos-12` runner on GitHub Actions and using cibuildwheel `macosx_arm64`, the final artifact filename contains `x86_64`, instead of `arm64`. 

Is the file naming logic working as intended or is it a bug or am I blatantly missing something here? If this is the latter, then I apologize for wasting your time 🙇🏼 Does the library assigns the underlying architecture type to the filename, despite the fact that I am building an `arm64` wheel?

To name the .whl file to contain `arm64`, I ended up leveraging `CIBW_REPAIR_WHEEL_COMMAND_MACOS` to rename the artifact after the repair process. This is kind of important for me, since I generate multiple artifacts on different platforms and I do not want artifact names to conflict.

Although I cannot provide links to the Ci config, as this is a private repository, I can paste the GitHub action job config here:

### CI config

```
jobs:
  wheels_arm64:
    name: ${{ matrix.os }}_${{ matrix.cibuildwheel_platform_id }}
    runs-on: ${{ matrix.os }}

    strategy:
      # Ensure that a wheel builder finishes even if another fails
      fail-fast: false
      matrix:
        include:

          # MacOS AMD 64
          - os: macos-12
            python: 310
            cibuildwheel_platform_id: macosx_arm64

    steps:
      - uses: actions/checkout@v3
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.12.0
        env:
          CIBW_REPAIR_WHEEL_COMMAND_MACOS: >
            delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} &&
            for file in {dest_dir}/*.whl ; do mv $file ${file//x86_64/arm64} ; done
          CIBW_BUILD: cp${{ matrix.python }}-${{ matrix.cibuildwheel_platform_id }}
          CIBW_ARCHS_MACOS: "arm64"
          CIBW_BUILD_VERBOSITY: 1

        with:
          package-dir: ./python
          output-dir: wheelhouse
          config-file: "{package}/pyproject.toml"

      - name: "Store artifacts"
        ...
        ...
```

### Build log

```

2023-02-10T07:09:19.5900730Z Download action repository 'actions/setup-python@v4' (SHA:d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435)
[...]

PR #1416, opened 2023-02-10
Add information about lack of possibility of cross compile Poetry

Czaki:
Based on discussion form #1415 I think that it is a good idea to add information about such a problem to the documentation.


docs/faq.md
```diff
@@ -20,7 +20,13 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
 
 ### Building macOS wheels for Apple Silicon {: #apple-silicon}
 
-`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners. With the introduction of Apple Silicon, you now have several choices for wheels for Python 3.8+:
+`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners.
+
+!!! note
+    Currently it is impossible to cross-compiling project that are using Poetry as a build backend.
```
henryiii, line 26:
```suggestion
    Currently it is impossible to cross-compile project that uses Poetry as a build backend.
```

docs/faq.md
```diff
@@ -20,7 +20,13 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
 
 ### Building macOS wheels for Apple Silicon {: #apple-silicon}
 
-`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners. With the introduction of Apple Silicon, you now have several choices for wheels for Python 3.8+:
+`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners.
+
+!!! note
+    Currently it is impossible to cross-compile project that uses Poetry as a build backend.
+    Reladed  issue [here](https://github.com/python-poetry/poetry/issues/7107)
```
azagniotov, line 27: Hello Team, with regards to https://github.com/pypa/cibuildwheel/issues/1415 that caused the current PR, I would like to clarify that I can to cross-compile project that uses Poetry as a build backend. Perhaps it is my understanding but this documentation change is a little confusing (at least to me)

To elaborate, I can compile
1. `arm64` compatible wheel on MacOS (`x86_64`) runner
2. `arm64` compatible wheel on Linux (`x86_64`) runner

In case of#1, the compiled wheel filename just contained `x86_64` in it, instead of `arm64`, which is why I raised the ticket. But, a Apple Silicon M1 mac client was successfully able to download, install and execute the wheel in another Poetry-based project.

Czaki: This is put in the section about cross-compiling MacOS wheels. Cross-compilation of Linux ARM works differently than cross-compilation on MacOS. 

And the wrong name of the produced wheel is, for me something that makes the building wheel not work. 

If you have instructions that will allow you to build the wheel in a way that allows you to upload them directly from CI to PyPI without manual intervention, then this warning such be replaced. 

At this moment, I do not have such a project and M1 machine that will allow validating such instruction.

azagniotov: Perhaps my comment intention was not clear and I could have articulated myself more clearly - I did not mean that the warning should not be added at all. Thank you for clarifying your thoughts and intention.

TeoZosa: > If you have instructions that will allow you to build the wheel in a way that allows you to upload them directly from CI to PyPI without manual intervention, then this warning such be replaced.

Chiming in since I maintain a project which came across this same issue. 

What are we defining as "manual intervention"? Looking at #1415, it looks like @azagniotov included instructions. which set the `CIBW_REPAIR_WHEEL_COMMAND_MACOS` env var ([which is exactly how my project resolves this issue as well](https://github.com/TeoZosa/structlog-sentry-logger/blob/ec3a79ca66b815990daae4b15f077c3231f95490/.github/workflows/release.yml#L213-L219)):

```yaml
...
    steps:
      - uses: actions/checkout@v3
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.12.0
        env:
          CIBW_REPAIR_WHEEL_COMMAND_MACOS: >
            delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} &&
            for file in {dest_dir}/*.whl ; do mv $file ${file//x86_64/arm64} ; done
...
```

Czaki: This solution is really bad as it skip whole delocate steep (as delocate will look fot x86 libs and dependencies)

azagniotov: Sorry, just to clarify - where do you see that the delocate wheel step is skipped? 🤔 At least in the my config snippet reposted by Teo the delocate wheel step is present

Czaki: Is efectivelly skiped. 

delicate is called on file with `x86_64` in name, so it searches the wheel for x86_64 binary blobs to check if all dependencies are collected, proper os version is set etc. As there is no such code then delocate ends with success.  

But no check of correctness on the arm64 code is done. 

Then you rename wheel tox x86_64 to arm64. To have the correct workflow you need to first rename the wheel file and then execute delocate on renamed file.

azagniotov: @Czaki you raise a good point re `To have the correct workflow you need to first rename the wheel file and then execute delocate on renamed file`. <= thank you for your guidance and observation, I will look into this 👍 Much appreciated

azagniotov: @Czaki I followed your pointer and swapped the order of operations where:
1. The wheel was renamed first
2. The execute delocate command was on the renamed wheel


```
CIBW_REPAIR_WHEEL_COMMAND_MACOS: |
echo "Target delocate archs: {delocate_archs}"

ORIGINAL_WHEEL={wheel}
if [ "${{ matrix.cibuildwheel_platform_id }}" == "macosx_arm64" ];
then
  echo "Renaming .whl file when architecture is 'macosx_arm64'"
  RENAMED_WHEEL=${ORIGINAL_WHEEL//x86_64/arm64}

  echo "Wheel will be renamed to $RENAMED_WHEEL"
  mv $ORIGINAL_WHEEL $RENAMED_WHEEL
else
  RENAMED_WHEEL=$ORIGINAL_WHEEL
fi

echo "Running delocate-wheel command on $RENAMED_WHEEL"
delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v $RENAMED_WHEEL

echo "DONE!!!"
```

The output is as follows:
```
Repairing wheel...
  
  + echo "Target delocate archs: arm64"
  
  ORIGINAL_WHEEL=/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-xg2x4afn/cp310-macosx_arm64/built_wheel/my_project-0.0.1-cp310-cp310-macosx_12_0_x86_64.whl
  if [ "macosx_arm64" == "macosx_arm64" ];
  then
    echo "Renaming .whl file when architecture is 'macosx_arm64'"
    RENAMED_WHEEL=${ORIGINAL_WHEEL//x86_64/arm64}
  
    echo "Wheel will be renamed to $RENAMED_WHEEL"
    mv $ORIGINAL_WHEEL $RENAMED_WHEEL
  else
    RENAMED_WHEEL=$ORIGINAL_WHEEL
  fi
  
  echo "Running delocate-wheel command on $RENAMED_WHEEL"
  delocate-wheel --require-archs arm64 -w /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-xg2x4afn/cp310-macosx_arm64/repaired_wheel -v $RENAMED_WHEEL
  
  echo "DONE!!!"
  
  Target delocate archs: arm64
  Renaming .whl file when architecture is 'macosx_arm64'
  Wheel will be renamed to /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-xg2x4afn/cp310-macosx_arm64/built_wheel/my_project-0.0.1-cp310-cp310-macosx_12_0_arm64.whl
  Running delocate-wheel command on /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-xg2x4afn/cp310-macosx_arm64/built_wheel/my_project-0.0.1-cp310-cp310-macosx_12_0_arm64.whl
  Fixing: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/cibw-run-xg2x4afn/cp310-macosx_arm64/built_wheel/my_project-0.0.1-cp310-cp310-macosx_12_0_arm64.whl
  DONE!!!
```

The repairing wheel step completed successfully also in this case, and the compiled wheel was successfully installed on Apple M1 and invoked at runtime.

Thoughts? 🙇🏼

Czaki: Thanks. 
For now I do not know if just link to this response or embed whole script in the documentation.

azagniotov: Thanks, @Czaki . 

I personally lean towards a more defensive approach and I would just include a link to this discussion where appropriate as oppose to including my code snippet in the official docs. 

But, that being said, I defer to you and your team regarding this. Please do what you think is best for the Poetry-based community and such a widely used library (cibuildwheel).

Thank you for your time on this, regardless.

Czaki: I have updated the description to be more correct. Did it look better?

docs/faq.md
```diff
@@ -20,7 +20,18 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
...
 
-`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners. With the introduction of Apple Silicon, you now have several choices for wheels for Python 3.8+:
+`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners.
+
+!!! note
+    Currently it is not fully supported to cross-compile project that uses Poetry as a build backend.
+    Reladed  issue [here](https://github.com/python-poetry/poetry/issues/7107).
+    Some packages can be built with arm64 wheels, but their file names will be incorrect,
+    with the platform tag showing `x86_64` instead of `arm64`.
+    As a workaround, the file can be renamed before running delocate to repair the wheel.
```
henryiii, line 30: The internal tags will still be broken? You'd need https://github.com/pypa/wheel/pull/422 I believe.

docs/faq.md
```diff
@@ -20,7 +20,18 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
 
 ### Building macOS wheels for Apple Silicon {: #apple-silicon}
 
-`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners. With the introduction of Apple Silicon, you now have several choices for wheels for Python 3.8+:
+`cibuildwheel` supports cross-compiling `universal2` and `arm64` wheels on `x86_64` runners.
+
+!!! note
+    Currently it is not fully supported to cross-compile project that uses Poetry as a build backend.
+    Reladed  issue [here](https://github.com/python-poetry/poetry/issues/7107).
```
henryiii, line 27:
```suggestion
    Related issue [here](https://github.com/python-poetry/poetry/issues/7107).
```

docs/faq.md
```diff
@@ -20,7 +20,18 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
...
+    Some packages can be built with arm64 wheels, but their file names will be incorrect,
```
henryiii, line 28:
```suggestion
    Some packages can be built with arm64 wheels, but their tags will be incorrect,
```

docs/faq.md
henryiii: I'd remove this renaming suggestion and just say the tags have to be corrected. An example is in https://github.com/scikit-build/cmake-python-distributions/blob/master/scripts/repair_wheel.py.

Czaki: Now better?

joerick:
Sorry for the delay on this @Czaki - I'm sorta hoping/waiting til we can get https://github.com/pypa/wheel/pull/422 in so we can give some concrete advice on this, because the file rename advice is imperfect as an approach and has 'gotchas' that make it harder to explain. Hopefully that wheel feature will come soon!

Czaki:
Lets keep it open until pypa/wheel was released so people can find it if they cannot wait.

henryiii:
wheel 0.40 is out!

Czaki:
So could you prepare proper instructions? I'm not on time with wheel. Then I will close this PR.

henryiii:
If we want it to be done in the cibuildwheel steps (like repair wheel?), we should probably wait until it's in our pinned deps.

henryiii:
It's in the pinned deps in the latest release, we can update docs now.

joerick:
I've had a pass on this, including a general improvement to the whole section, comments welcome

henryiii (approved):
Technically, you can’t build universal natively, not sure what that would mean. I guess emulation for the Intel part then fusing? But that’s okay, not worth changing it (as it can be tested).

joerick:
Yeah, good point. That statement has been in the docs for a while, but I'll have a think on how to reword that.

docs/faq.md
```diff
@@ -20,61 +20,77 @@ Linux wheels are built in [`manylinux`/`musllinux` containers](https://github.co
...
     Only CPython 3.8 and newer support `universal2` and `arm64` wheels.
 
+!!! note
+    If your project uses **Poetry** as a build backend, cross-compiling on macOS [does not currently work](https://github.com/python-poetry/poetry/issues/7107). In some cases arm64 wheels can be built but their tags will be incorrect, with the platform tag showing `x86_64` instead of `arm64`.
+
+    As a workaround, the tag can be fixed before running delocate to repair the wheel. The [`wheel tags`](https://wheel.readthedocs.io/en/stable/reference/wheel_tags.html) command is ideal for this. See [this workflow](https://gist.github.com/anderssonjohan/49f07e33fc5cb2420515a8ac76dc0c95#file-build-pendulum-wheels-yml-L39-L53) for an example usage of `wheel tags`.
+
+Hopefully, cross-compilation is a temporary situation. Once we have widely
+available Apple Silicon CI runners, we can build and test `arm64` and
+`universal2` wheels natively. That's why `universal2`/`arm64` wheels require opt-in by setting `CIBW_ARCHS_MACOS`.
```
joerick, line 92:
```suggestion
Hopefully, cross-compilation is a temporary situation. Once we have widely available Apple Silicon CI runners, we can build `arm64` wheels natively. Until then, cross-compiling `arm64` or `universal2` wheels requires opt-in by setting `CIBW_ARCHS_MACOS` on Intel runners.
```

PR #1428, opened 2023-03-07
ci: fix appveyor (Python 3.8+ needed for macOS)

henryiii:
MacOS default image has changed.


Czaki (approved):

joerick (approved):
Thanks. I added an update to the CI table that tracks the Python versions too.

PR #1430, opened 2023-03-10
fix: pip config options were broken

### LINKED ISSUE #1429: `--config-settings` is passed as `--config-setting`
### Description

Whenever I try to use the feature to pass `--config-settings` (either through environment variable or `pyproject.toml`), I am met with a `no such option: --config-setting` error in the logs.

A cursory look at the pip documentation seems to suggest the argument name actually is `--config-settings`:
https://pip.pypa.io/en/stable/cli/pip_wheel/#cmdoption-config-settings

It looks like the options that should be passed as `--config-settings` are actually being passed as `--config-setting` (missing the `s`):

https://github.com/pypa/cibuildwheel/blob/acdbe87e3d002598b10435bb6f17bae758418ba8/cibuildwheel/util.py#L213

### Build log

_No response_

### CI config

_No response_

henryiii:
Closes #1429. Only "build" backend supported config options, since pip uses a different name (with an s).

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -208,9 +208,9 @@ def get_build_verbosity_extra_flags(level: int) -> list[str]:
         return []
 
 
-def split_config_settings(config_settings: str, *, plural: bool) -> list[str]:
+def split_config_settings(config_settings: str, frontend: Literal["pip", "build"]) -> list[str]:
     config_settings_list = shlex.split(config_settings)
-    s = "s" if plural else ""
+    s = "s" if frontend == "pip" else ""
     return [f"--config-setting{s}={setting}" for setting in config_settings_list]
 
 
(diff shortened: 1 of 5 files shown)


cibuildwheel/util.py
```diff
@@ -208,9 +208,10 @@ def get_build_verbosity_extra_flags(level: int) -> list[str]:
         return []
 
 
-def split_config_settings(config_settings: str) -> list[str]:
+def split_config_settings(config_settings: str, *, plural: bool) -> list[str]:
```
joerick, line 211: the 'plural' argument here is a little odd... Do you think it would be better to pass in the build backend instead (build or pip) and use that to decide the option name?

joerick (commented):
Thank you for this! Looks good, I just had a suggestion on function params

cibuildwheel/util.py
henryiii: I started that way, but then went this way as a little nicer than passing a string around. But I don't have a strong preference. If a third backend was added someday, I guess mypy would force handling the extra literal. I'd hope this goes away in the future when we can rely on the (not yet released) next version of pip, which supports `-C`, like build.

joerick (approved):
Thanks!

PR #1442, opened 2023-03-25
GitHub with QEMU only for AArch64

Marco-Sulla:


joerick:
why?

Marco-Sulla:
Well, because this way you don't use QEMU for Intel archs. I took inspiration from matplotlib:
[matplotlib workflow](https://github.com/matplotlib/matplotlib/blob/1d0fc0c0ae7aa95efa4239bc83f9f97391946cf5/.github/workflows/cibuildwheel.yml#L63)

Maybe should I add also the other non-native archs, and not only arm?

henryiii:
It’s pretty common to break it up for larger projects so that everything can finish inter the limit. But I don’t think it’s necessary for a small example, and you will not be using QEMU for Intel architectures, that only enables the use of other architectures.

Marco-Sulla:
@henryiii I can understand the need to leave the details about fine-grained configuration to the help of the specific CI/CD system you use. 

Maybe can I add a separate, real-world example for Github, that's quite mainstream? I'd need it :-)

henryiii:
That's partially why we have an extensive list of example projects. There are a lot of different ways to do this, and some very large projects have to split up by Python version as well. Where do you draw the line? I would say that an example needs to be simple. I do have a blog post that shows you how to split up by python version GHA automatically. https://iscinumpy.dev/post/cibuildwheel-2-10-0/

This change drops macOS and Windows, and all other archs. There is also no harm in enabling QEMU if you don't use it.

Marco-Sulla:
> That's partially why we have an extensive list of example projects

@henryiii Where they are listed? In the docs I see the list of Working examples, but I suppose they are not.

joerick:
As @henryiii says, there are many different ways to do this, and I'd rather the example config we provide users is as simple as possible. Everyone has different needs and optimisation techniques will vary.

PR #1455, opened 2023-04-04
tests: use uppercase envvar on Windows

henryiii:
Quick fix for https://github.com/charliermarsh/ruff/pull/3876 - also, besides consistency, I like this a bit better, because it's easy to assume that envvar's are case sensitive when you see them in camel case. I was confused by that once on Windows (I had a nice idea that was broken by them not being case sensitive on just this one OS).

### DIFF
--- a/test/test_windows.py
+++ b/test/test_windows.py
@@ -12,9 +12,9 @@ basic_project = test_projects.new_c_project()
 
 
 def skip_if_no_msvc(arm64=False):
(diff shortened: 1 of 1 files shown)


Czaki (approved):

joerick:
Wow, TIL windows env vars are case insensitive. Does ruff know that this is a windows test? it can't right? must be applying the same rule across all platforms.

joerick (approved):

henryiii:
Yes, the rule (SIM112) is just to always use all-caps envvars. The only reason we have one that's not is because Windows doesn't care. :)

PR #1456, opened 2023-04-04
Add Emscripten support

### LINKED ISSUE #1454: Support for wasm32-emscripten wheels
@henryiii has discussed adding a wasm32-emscripten target to cibuildwheel in the past. I would like to help with this. Currently it is not super invasive to build such wheels.

Here's numpy's CI support for Emscripten using `pyodide build`:
https://github.com/numpy/numpy/blob/main/.github/workflows/emscripten.yml

Rust + Python projects can directly use maturin:
https://github.com/pydantic/pydantic-core/blob/main/Makefile#L46

There's currently not all that many examples of projects building against Emscripten in their CI, but I would like to change that!. I think a lot of maintainers are enthusiastic about the concept.

hoodmane:
Resolves #1454. Work in progress!

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -15,6 +15,7 @@ import cibuildwheel.linux
 import cibuildwheel.macos
 import cibuildwheel.util
 import cibuildwheel.windows
+import cibuildwheel.pyodide
 from cibuildwheel.architecture import Architecture, allowed_architectures_check
 from cibuildwheel.logger import log
 from cibuildwheel.options import CommandLineArguments, Options, compute_options
@@ -42,7 +43,7 @@ def main() -> None:
 
     parser.add_argument(
         "--platform",
-        choices=["auto", "linux", "macos", "windows"],
+        choices=["auto", "linux", "macos", "windows", "pyodide"],
         default=None,
         help="""
             Platform to build for. Use this option to override the
@@ -295,6 +296,8 @@ def build_in_directory(args: CommandLineArguments) -> None:
                 cibuildwheel.windows.build(options, tmp_path)
             elif platform == "macos":
                 cibuildwheel.macos.build(options, tmp_path)
+            elif platform == "pyodide":
+                cibuildwheel.pyodide.build(options, tmp_path)
             else:
                 assert_never(platform)
     finally:
@@ -357,6 +360,10 @@ def get_build_identifiers(
         )
     elif platform == "macos":
         python_configurations = cibuildwheel.macos.get_python_configurations(
+            build_selector, architectures
+        )
+    elif platform == "pyodide":
+        python_configurations = cibuildwheel.pyodide.get_python_configurations(
             build_selector, architectures
         )
     else:
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -8,12 +8,12 @@ from enum import Enum
 
 from .typing import Final, Literal, PlatformName, assert_never
 
-PRETTY_NAMES: Final = {"linux": "Linux", "macos": "macOS", "windows": "Windows"}
+PRETTY_NAMES: Final = {"linux": "Linux", "macos": "macOS", "windows": "Windows", "pyodide": "Pyodide"}
 
 ARCH_SYNONYMS: Final[list[dict[PlatformName, str | None]]] = [
-    {"linux": "x86_64", "macos": "x86_64", "windows": "AMD64"},
-    {"linux": "i686", "macos": None, "windows": "x86"},
-    {"linux": "aarch64", "macos": "arm64", "windows": "ARM64"},
+    {"linux": "x86_64", "macos": "x86_64", "windows": "AMD64", "pyodide": "wasm32"},
+    {"linux": "i686", "macos": None, "windows": "x86", "pyodide": "wasm32"},
+    {"linux": "aarch64", "macos": "arm64", "windows": "ARM64", "pyodide": "wasm32"},
 ]
 
 
@@ -38,6 +38,9 @@ class Architecture(Enum):
     x86 = "x86"
     AMD64 = "AMD64"
     ARM64 = "ARM64"
+
+    # WebAssembly
+    wasm32 = "wasm32"
 
     # Allow this to be sorted
     def __lt__(self, other: Architecture) -> bool:
@@ -113,6 +116,7 @@ class Architecture(Enum):
             },
             "macos": {Architecture.x86_64, Architecture.arm64, Architecture.universal2},
             "windows": {Architecture.x86, Architecture.AMD64, Architecture.ARM64},
+            "pyodide": {Architecture.wasm32}
         }
         return all_archs_map[platform]
 
--- a/cibuildwheel/pyodide.py
+++ b/cibuildwheel/pyodide.py
@@ -1,0 +1,354 @@
+import contextlib
+from filelock import FileLock
+from typing import Sequence, Tuple, cast
+from pathlib import Path
+from .architecture import Architecture
+from dataclasses import dataclass
+from .options import Options
+import os
+from .typing import Literal, PathOrStr, assert_never
+from .environment import ParsedEnvironment
+import sys
+import shutil
+
+from .util import (
+    CIBW_CACHE_PATH,
+    AlreadyBuiltWheelError,
+    BuildFrontend,
+    BuildSelector,
+    extract_zip,
+    NonPlatformWheelError,
+    call,
+    detect_ci_provider,
+    download,
+    find_compatible_wheel,
+    get_build_verbosity_extra_flags,
+    get_pip_version,
+    install_certifi_script,
+    pyodide_build_script,
+    prepare_command,
+    read_python_configs,
+    shell,
+    split_config_settings,
+    test_fail_cwd_file,
+    unwrap,
+    virtualenv,
+)
+from .logger import log
+
+
+@dataclass(frozen=True)
+class PythonConfiguration:
+    version: str
+    identifier: str
+    pyodide_version: str
+    emscripten_version: str
+
+
+def install_emscripten(tmp: Path, version: str) -> Path:
+    url = "https://github.com/emscripten-core/emsdk/archive/main.zip"
+    installation_path = CIBW_CACHE_PATH / ("emsdk-" + version)
+    emsdk_path = installation_path / "emsdk-main/emsdk"
+    emcc_path = installation_path / "emsdk-main/upstream/emscripten/emcc"
+    with FileLock(str(installation_path) + ".lock"):
+        if installation_path.exists():
+            return emcc_path
+        emsdk_zip = tmp / "emsdk.zip"
+        download(url, emsdk_zip)
+        installation_path.mkdir()
+        extract_zip(emsdk_zip, installation_path)
+        os.chmod(emsdk_path, 0o777)
+        call(emsdk_path, "install", version)
+        call(emsdk_path, "activate", version)
+
+    return emcc_path
+
+
+def setup_python(
+    tmp: Path,
+    python_configuration: PythonConfiguration,
+    dependency_constraint_flags: Sequence[PathOrStr],
+    environment: ParsedEnvironment,
+    build_frontend: BuildFrontend,
+) -> dict[str, str]:
+    implementation_id = python_configuration.identifier.split("-")[0]
+    log.step(f"Installing Python {implementation_id}...")
+    # TODO?
+
+    python_version = implementation_id.removeprefix("cp")
+    python_name = f"python{python_version[0]}.{python_version[1:]}"
+    base_python = Path(shutil.which(python_name))
+
+    log.step("Setting up build environment...")
+    venv_path = tmp / "venv"
+    env = virtualenv(base_python, venv_path, dependency_constraint_flags)
+    venv_bin_path = venv_path / "bin"
+    assert venv_bin_path.exists()
+    env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
+
+    # upgrade pip to the version matching our constraints
+    # if necessary, reinstall it to ensure that it's available on PATH as 'pip'
+    call(
+        "python",
+        "-m",
+        "pip",
+        "install",
+        "--upgrade",
+        "pip",
+        *dependency_constraint_flags,
+        env=env,
+        cwd=venv_path,
+    )
+
+    env = environment.as_dictionary(prev_environment=env)
+
+    # check what pip version we're on
+    assert (venv_bin_path / "pip").exists()
+    call("which", "pip", env=env)
+    call("pip", "--version", env=env)
+    which_pip = call("which", "pip", env=env, capture_stdout=True).strip()
+    if which_pip != str(venv_bin_path / "pip"):
+        print(
+            "cibuildwheel: pip available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert pip above it.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    # check what Python version we're on
+    call("which", "python", env=env)
+    call("python", "--version", env=env)
+    which_python = call("which", "python", env=env, capture_stdout=True).strip()
+    if which_python != str(venv_bin_path / "python"):
+        print(
+            "cibuildwheel: python available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python above it.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    log.step("Installing build tools...")
+    call(
+        "pip",
+        "install",
+        "--upgrade",
+        "auditwheel-emscripten",
+        "build[virtualenv]",
+        f"pyodide-build=={python_configuration.pyodide_version}",
+        *dependency_constraint_flags,
+        env=env,
+    )
+
+    log.step("Installing emscripten...")
+    emcc_path = install_emscripten(tmp, python_configuration.emscripten_version)
+    import sys
+    
+    env["PATH"] = os.pathsep.join(
+        [
+            env["PATH"],
+            str(emcc_path.parent)
+        ]
+    )
+
+    # log.step("Installing xbuildenv")
+    # cache_dir = CIBW_CACHE_PATH / ("emsdk-" + python_configuration.emscripten_version)
+    # call("pyodide", "xbuildenv", "install", "--download", cwd=cache_dir, env=env)
+    # env["PYODIDE_ROOT"] = cache_dir / ".pyodide-xbuildenv/xbuildenv/pyodide-root/"
+
+    return env
+
+
+def get_python_configurations(
+    build_selector: BuildSelector, architectures: set[Architecture]
+) -> list[PythonConfiguration]:
+    full_python_configs = read_python_configs("pyodide")
+
+    python_configurations = [PythonConfiguration(**item) for item in full_python_configs]
+    python_configurations = [c for c in python_configurations if build_selector(c.identifier)]
+    return python_configurations
+
+
+def build(options: Options, tmp_path: Path) -> None:
+    python_configurations = get_python_configurations(
+        options.globals.build_selector, options.globals.architectures
+    )
+
+    try:
+        before_all_options_identifier = python_configurations[0].identifier
+        before_all_options = options.build_options(before_all_options_identifier)
+
+        if before_all_options.before_all:
+            log.step("Running before_all...")
+            env = before_all_options.environment.as_dictionary(prev_environment=os.environ)
+            before_all_prepared = prepare_command(
+                before_all_options.before_all, project=".", package=before_all_options.package_dir
+            )
+            shell(before_all_prepared, env=env)
+
+        built_wheels: list[Path] = []
+
+        for config in python_configurations:
+            build_options = options.build_options(config.identifier)
+            if build_options.build_frontend == "pip":
+                print("Don't support pip frontend", file=sys.stderr)
+                # sys.exit(1)
+            log.build_start(config.identifier)
+
+            identifier_tmp_dir = tmp_path / config.identifier
+            identifier_tmp_dir.mkdir()
+            built_wheel_dir = identifier_tmp_dir / "built_wheel"
+
+            dependency_constraint_flags: Sequence[PathOrStr] = []
+            if build_options.dependency_constraints:
+                dependency_constraint_flags = [
+                    "-c",
+                    build_options.dependency_constraints.get_for_python_version(config.version),
+                ]
+
+            env = setup_python(
+                identifier_tmp_dir / "build",
+                config,
+                dependency_constraint_flags,
+                build_options.environment,
+                build_options.build_frontend,
+            )
+
+            compatible_wheel = find_compatible_wheel(built_wheels, config.identifier)
+            if compatible_wheel:
+                log.step_end()
+                print(
+                    f"\nFound previously built wheel {compatible_wheel.name}, that's compatible with {config.identifier}. Skipping build step..."
+                )
+                built_wheel = compatible_wheel
+            else:
+                if build_options.before_build:
+                    log.step("Running before_build...")
+                    before_build_prepared = prepare_command(
+                        build_options.before_build, project=".", package=build_options.package_dir
+                    )
+                    shell(before_build_prepared, env=env)
+
+                log.step("Building wheel...")
+                built_wheel_dir.mkdir()
+
+                verbosity_flags = get_build_verbosity_extra_flags(build_options.build_verbosity)
+                extra_flags = split_config_settings(
+                    build_options.config_settings, build_options.build_frontend
+                )
+
+                # if build_options.build_frontend == "pip":
+                #     pass  # Nope
+                # elif build_options.build_frontend == "build":
+
+                if True:
+                    verbosity_setting = " ".join(verbosity_flags)
+                    extra_flags += (f"--config-setting={verbosity_setting}",)
+                    build_env = env.copy()
+                    if build_options.dependency_constraints:
+                        constraints_path = (
+                            build_options.dependency_constraints.get_for_python_version(
+                                config.version
+                            )
+                        )
+                        # Bug in pip <= 21.1.3 - we can't have a space in the
+                        # constraints file, and pip doesn't support drive letters
+                        # in uhi.  After probably pip 21.2, we can use uri. For
+                        # now, use a temporary file.
+                        if " " in str(constraints_path):
+                            assert " " not in str(identifier_tmp_dir)
+                            tmp_file = identifier_tmp_dir / "constraints.txt"
+                            tmp_file.write_bytes(constraints_path.read_bytes())
+                            constraints_path = tmp_file
+
+                        build_env["PIP_CONSTRAINT"] = str(constraints_path)
+                        build_env["VIRTUALENV_PIP"] = get_pip_version(env)
+                        print("build_options.package_dir", build_options.package_dir)
+                        print(built_wheel_dir)
+                        call(
+                            "pyodide",
+                            "build",
+                            build_options.package_dir,
+                            extra_flags,
+                            env=build_env,
+                        )
+                        output = next((build_options.package_dir / "dist").glob("*.whl"))
+                        shutil.move(output, built_wheel_dir)
+
+                else:
+                    assert_never(build_options.build_frontend)
+
+                built_wheel = next(built_wheel_dir.glob("*.whl"))
+
+                if built_wheel.name.endswith("none-any.whl"):
+                    raise NonPlatformWheelError()
+
+
+            if build_options.test_command and build_options.test_selector(config.identifier):
+
+                log.step(
+                    "Testing wheel..."
+                )
+
+                venv_dir = identifier_tmp_dir / "venv-test"
+                # set up a virtual environment to install and test from, to make sure
+                # there are no dependencies that were pulled in at build time.
+
+                # --no-download??
+                call("pyodide", "venv", venv_dir, env=env)
+
+
+                virtualenv_env = env.copy()
+                virtualenv_env["PATH"] = os.pathsep.join(
+                    [
+                        str(venv_dir / "bin"),
+                        virtualenv_env["PATH"],
+                    ]
+                )
+
+                # check that we are using the Python from the virtual environment
+                call("which", "python", env=virtualenv_env)
+
+                if build_options.before_test:
+                    before_test_prepared = prepare_command(
+                        build_options.before_test,
+                        project=".",
+                        package=build_options.package_dir,
+                    )
+                    shell(before_test_prepared, env=virtualenv_env)
+
+                # install the wheel
+                call(
+                    "pip",
+                    "install",
+                    f"{built_wheel}{build_options.test_extras}",
+                    env=virtualenv_env,
+                )
+
+                # test the wheel
+                if build_options.test_requires:
+                    call(
+                        "pip", "install", *build_options.test_requires, env=virtualenv_env
+                    )
+
+                # run the tests from a temp dir, with an absolute path in the command
+                # (this ensures that Python runs the tests against the installed wheel
+                # and not the repo code)
+                test_command_prepared = prepare_command(
+                    build_options.test_command,
+                    project=Path(".").resolve(),
+                    package=build_options.package_dir.resolve(),
+                )
+
+                test_cwd = identifier_tmp_dir / "test_cwd"
+                test_cwd.mkdir(exist_ok=True)
+                (test_cwd / "test_fail.py").write_text(test_fail_cwd_file.read_text())
+
+                shell(test_command_prepared, cwd=test_cwd, env=virtualenv_env)
+
+            # we're all done here; move it to output (overwrite existing)
+            if compatible_wheel is None:
+                with contextlib.suppress(FileNotFoundError):
+                    (build_options.output_dir / built_wheel.name).unlink()
+
+                shutil.move(str(built_wheel), build_options.output_dir)
+                built_wheels.append(build_options.output_dir / built_wheel.name)
+    finally:
+        pass
--- a/cibuildwheel/resources/build-platforms.toml
+++ b/cibuildwheel/resources/build-platforms.toml
@@ -115,3 +115,9 @@ python_configurations = [
   { identifier = "pp38-win_amd64", version = "3.8", arch = "64", url = "https://downloads.python.org/pypy/pypy3.8-v7.3.11-win64.zip" },
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.11-win64.zip" },
 ]
+
+[pyodide]
+python_configurations = [
+  { identifier = "cp310-emscripten_3_1_27", version = "3.10.2", pyodide_version = "0.22.1", emscripten_version = "3.1.27" },
+  # { identifier = "cp311-emscripten_3_1_32", version = "3.11.2", pyodide_version = "0.23.0", emscripten_version = "3.1.32" },
+]
--- a/test/test_emscripten.py
+++ b/test/test_emscripten.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import os
+import subprocess
+import textwrap
+
+import pytest
+
+from . import test_projects, utils
+
+basic_project = test_projects.new_c_project()
+
+
+@pytest.mark.parametrize("use_pyproject_toml", [True, False])
+def test_emscripten_build(tmp_path, use_pyproject_toml):
+
+    if use_pyproject_toml:
+        basic_project.files["pyproject.toml"] = textwrap.dedent(
+            """
+            [build-system]
+            requires = ["setuptools"]
+            build-backend = "setuptools.build_meta"
+            """
+        )
+
+    project_dir = tmp_path / "project"
+    basic_project.generate(project_dir)
+
+    # build the wheels
+    actual_wheels = utils.cibuildwheel_run(
+        project_dir,
+        add_args=["--platform", "pyodide"],
+    )
+
+    # check that the expected wheels are produced
+    expected_wheels = [
+        'spam-0.1.0-cp310-cp310-emscripten_3_1_27_wasm32.whl',
+    ]
+
+    print("actual_wheels", actual_wheels)
+    print("expected_wheels", expected_wheels)
+
+    assert set(actual_wheels) == set(expected_wheels)
--- a/cibuildwheel/logger.py
+++ b/cibuildwheel/logger.py
@@ -34,6 +34,8 @@ PLATFORM_IDENTIFIER_DESCRIPTIONS: Final = {
     "macosx_x86_64": "macOS x86_64",
     "macosx_universal2": "macOS Universal 2 - x86_64 and arm64",
     "macosx_arm64": "macOS arm64 - Apple Silicon",
+    "emscripten_3_1_27" : "Pyodide v0.22.x",
+    "emscripten_3_1_32" : "Pyodide v0.23.x",
 }
 
 
--- a/cibuildwheel/resources/constraints-python310.txt
+++ b/cibuildwheel/resources/constraints-python310.txt
@@ -18,11 +18,11 @@ typing-extensions==4.5.0
     # via delocate
 virtualenv==20.21.0
     # via -r cibuildwheel/resources/constraints.in
-wheel==0.40.0
+# wheel==0.40.0
     # via -r cibuildwheel/resources/constraints.in
 
 # The following packages are considered to be unsafe in a requirements file:
 pip==23.0.1
     # via -r cibuildwheel/resources/constraints.in
-setuptools==67.6.1
+# setuptools==67.6.1
     # via -r cibuildwheel/resources/constraints.in
--- a/cibuildwheel/typing.py
+++ b/cibuildwheel/typing.py
@@ -41,4 +41,4 @@ else:
 
 
 PlatformName = Literal["linux", "macos", "windows"]
-PLATFORMS: Final[set[PlatformName]] = {"linux", "macos", "windows"}
+PLATFORMS: Final[set[PlatformName]] = {"linux", "macos", "windows", "pyodide"}
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -17,6 +17,7 @@ from enum import Enum
 from functools import lru_cache
 from pathlib import Path, PurePath
 from time import sleep
+from zipfile import ZipFile
 from typing import (
     Any,
     ClassVar,
@@ -66,6 +67,7 @@ __all__ = [
 resources_dir: Final[Path] = Path(__file__).parent / "resources"
 
 install_certifi_script: Final[Path] = resources_dir / "install_certifi.py"
+pyodide_build_script: Final[Path] = resources_dir / "pyodide_build_script.py"
 
 test_fail_cwd_file: Final[Path] = resources_dir / "testing_temp_dir_file.py"
 
@@ -336,6 +338,10 @@ def download(url: str, dest: Path) -> None:
             if i == repeat_num - 1:
                 raise
             sleep(3)
+
+def extract_zip(zip_src: Path, dest: Path) -> None:
+    with ZipFile(zip_src) as zip_:
+        zip_.extractall(dest)
 
 
 class DependencyConstraints:
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -10,7 +10,6 @@ from dataclasses import dataclass
 from functools import lru_cache
 from pathlib import Path
 from typing import Sequence
-from zipfile import ZipFile
 
 from filelock import FileLock
 from packaging.version import Version
@@ -28,6 +27,7 @@ from .util import (
     NonPlatformWheelError,
     call,
     download,
+    extract_zip,
     find_compatible_wheel,
     get_build_verbosity_extra_flags,
     get_pip_version,
@@ -87,11 +87,6 @@ def get_python_configurations(
     ]
 
     return python_configurations
-
-
-def extract_zip(zip_src: Path, dest: Path) -> None:
-    with ZipFile(zip_src) as zip_:
-        zip_.extractall(dest)
 
 
 @lru_cache(maxsize=None)
--- a/test/test_projects/c.py
+++ b/test/test_projects/c.py
@@ -48,8 +48,7 @@ from setuptools import setup, Extension
 {{ setup_py_add }}
 
 libraries = []
-if sys.platform.startswith('linux'):
-    libraries.extend(['m', 'c'])
+
 
 setup(
     ext_modules=[Extension(


hoodmane:
@henryiii could you approve the github actions workflows?

hoodmane:
@henryiii is the test failure a flake?

```
failed to download pip version 20.0.2, pip download exit code 1
The conflict is caused by:
    The user requested pip==20.0.2
    The user requested (constraint) pip==20.0.2
To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
WARNING: Skipping page https://pypi.org/simple/pip/ because the GET request got Content-Type: .The only supported Content-Type is text/html
ERROR: Cannot install pip==20.0.2 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

cibuildwheel/resources/constraints-pyodide-0_22.txt
```diff
@@ -0,0 +1,28 @@
...
+    # via delocate
+virtualenv==20.21.0
+    # via -r cibuildwheel/resources/constraints.in
+wheel==0.40.0
+    # via -r cibuildwheel/resources/constraints.in
+
+# The following packages are considered to be unsafe in a requirements file:
+pip==23.0.1
+    # via -r cibuildwheel/resources/constraints.in
+setuptools==65.5.1
```
hoodmane, line 27: I manually changed the setuptools pin in these two files to 65.5.1...

cibuildwheel/resources/pyodide_python.sh
```diff
@@ -0,0 +1,192 @@
+#!/bin/sh
```
rth, line 1: Can't download it from the Pyodide repo?

docs/options.md
```diff
@@ -180,14 +180,15 @@ wheels on manylinux2010, and manylinux2014 wheels for any newer Python
...
-Options: `auto` `linux` `macos` `windows`
+Options: `auto` `linux` `macos` `windows` `pyodide`
 
 Default: `auto`
 
 `auto` will auto-detect platform using environment variables, such as `TRAVIS_OS_NAME`/`APPVEYOR`/`CIRCLECI`.
 
 - For `linux`, you need [Docker or Podman](#container-engine) running, on Linux, macOS, or Windows.
 - For `macos` and `windows`, you need to be running on the respective system, with a working compiler toolchain installed - Xcode Command Line tools for macOS, and MSVC for Windows.
+- For `pyodide` you need to be on an x86 linux and have both Python 3.10 and Python 3.11 installed.
```
rth, line 191: [x86-64](https://en.wikipedia.org/wiki/X86-64) technically. x86 is 32 bit linux.

rth (commented):
Thanks a lot for doing this work! It's certainly more work that I anticipated (without knowing much about cibuildwheel) :)

henryiii:
> @henryiii is the windows_x86 failure a flake?


Yes. Some networking issue that is hard to track down.

henryiii:
BTW, would you like me to rebase for you?

hoodmane:
I already merged main locally.

cibuildwheel/resources/pyodide_python.sh
hoodmane: This is a modified version of the script. The modification is in pyodide main branch but I failed to backport it into 0.23.1. I will release 0.23.2 with this and a couple of other changes and then remove this file.

hoodmane: Okay fixed now.

docs/options.md
hoodmane: Updated...

cibuildwheel/resources/pyodide_python.sh
hoodmane: Fixed now.

cibuildwheel/resources/constraints-pyodide-0_22.txt
hoodmane: Removed.

hoodmane:
Okay I think this is working now. Would appreciate a review @henryiii, @joerick

hoodmane:
Presumably we need documentation somewhere for the fact that Python 311 needs to be installed.

docs/options.md
```diff
@@ -180,14 +180,15 @@ wheels on manylinux2010, and manylinux2014 wheels for any newer Python
...
-Options: `auto` `linux` `macos` `windows`
+Options: `auto` `linux` `macos` `windows` `pyodide`
 
 Default: `auto`
 
 `auto` will auto-detect platform using environment variables, such as `TRAVIS_OS_NAME`/`APPVEYOR`/`CIRCLECI`.
 
 - For `linux`, you need [Docker or Podman](#container-engine) running, on Linux, macOS, or Windows.
 - For `macos` and `windows`, you need to be running on the respective system, with a working compiler toolchain installed - Xcode Command Line tools for macOS, and MSVC for Windows.
+- For `pyodide` you need to be on an x86-64 linux and have both Python 3.10 and Python 3.11 installed.
```
henryiii, line 191: Why 3.10?

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,342 @@
...
+                    build_env["PIP_CONSTRAINT"] = str(constraints_path)
+                build_env["VIRTUALENV_PIP"] = get_pip_version(env)
+                call(
+                    "pyodide",
+                    "build",
+                    build_options.package_dir,
+                    f"--outdir={built_wheel_dir}",
+                    *extra_flags,
+                    env=build_env,
+                    cwd=identifier_tmp_dir / "build",
```
joerick, line 271: Why do we change the cwd for this call?

test/test_emscripten.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.parametrize("use_pyproject_toml", [True, False])
+def test_emscripten_build(tmp_path, use_pyproject_toml):
+    if platform.machine() == "arm64":
+        pytest.skip("emsdk doesn't work correctly on arm64")
+
+    if not shutil.which("python3.10"):
+        pytest.skip("Python 3.10 not installed")
+
```
joerick, line 21: Is 3.10 really needed?

joerick:
Splendid work on this so far @hoodmane ! Quality looks really good. Comments above.

.github/workflows/test.yml
```diff
@@ -109,3 +109,36 @@ jobs:
     - name: Run the emulation tests
       run: |
         pytest --run-emulation test/test_emulation.py
+
+  test-emscripten:
```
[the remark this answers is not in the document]
joerick: It looks like currently, the platform is pyodide, but the wheel platform tag is emscripten, similar to how we have `linux`->`manylinux`/`musllinux` currently. That could work too, I just wanted to check that makes sense to everyone.  (If that's the case, let's use the 'pyodide' terminology primarily.)

hoodmane:
Thanks for the detailed review @joerick!

hoodmane:
> pyodide isn't auto-fixing out-of-tree wheels

You're talking about `replace_so_abi_tags()`? I think we can release a Pyodide 0.23.3 that calls that for out of tree wheels.

@ryanking13 do you see any reason not to call `replace_so_abi_tags()` when building out of tree? This probably fits into your project of making the out of tree builds more the same as the in-tree ones.

Of course it would be great if we could also fix the issue with Emscripten, but it seems potentially easier to change Pyodide.
https://github.com/emscripten-core/emscripten/pull/19301

test/test_emscripten.py
hoodmane: No, not anymore.

docs/options.md
hoodmane: Needs update.

cibuildwheel/architecture.py
```diff
@@ -10,12 +10,17 @@
...
+    "pyodide": "Pyodide",
+}
 
 ARCH_SYNONYMS: Final[list[dict[PlatformName, str | None]]] = [
-    {"linux": "x86_64", "macos": "x86_64", "windows": "AMD64"},
-    {"linux": "i686", "macos": None, "windows": "x86"},
-    {"linux": "aarch64", "macos": "arm64", "windows": "ARM64"},
+    {"linux": "x86_64", "macos": "x86_64", "windows": "AMD64", "pyodide": "wasm32"},
+    {"linux": "i686", "macos": None, "windows": "x86", "pyodide": "wasm32"},
+    {"linux": "aarch64", "macos": "arm64", "windows": "ARM64", "pyodide": "wasm32"},
```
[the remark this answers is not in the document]
hoodmane: >  they're not synonyms

Ah I see so the word "synonyms" means that for instance aarch64, arm64, and ARM64 are all synonymous architectures. But wasm32 is obviously not. It was unclear to me what the word "synonyms" was referring to, but the point here is that different platforms have different words for the same architecture.

The problem is I didn't read the comment where `ARCH_SYNONYMS` is used:
```
# we might need to rename the native arch to the machine we're running
# on, as the same arch can have different names on different platforms
```
So indeed, the sensible thing is to say `if platform is pyodide, return {wasm32}`.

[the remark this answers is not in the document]
hoodmane: Though maybe we should check that the native machine is not a windows since we actually can't build there. MacOS at least sometimes works.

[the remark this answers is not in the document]
hoodmane: Okay I see the check is the responsibility of `allowed_architectures_check` so we can just return `{wasm32}` unconditionally and add an extra check there.

.github/workflows/test.yml
```diff
@@ -109,3 +109,36 @@ jobs:
...
+      uses: ./
+      with:
+        package-dir: sample_proj
+        output-dir: wheelhouse
+      env:
+        CIBW_PLATFORM: pyodide
+
+    - name: Run the emulation tests
+      run: |
+        pytest test/test_emscripten.py
```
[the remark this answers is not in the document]
hoodmane: Sounds good to me. Should I say `python ./bin/run_tests.py` here?

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,342 @@
...
+    emsdk_path = installation_path / "emsdk-main/emsdk"
+    emcc_path = installation_path / "emsdk-main/upstream/emscripten/emcc"
+    with FileLock(str(installation_path) + ".lock"):
+        if installation_path.exists():
+            return emcc_path
+        emsdk_zip = tmp / "emsdk.zip"
+        download(url, emsdk_zip)
+        installation_path.mkdir()
+        extract_zip(emsdk_zip, installation_path)
+        os.chmod(emsdk_path, 0o777)
```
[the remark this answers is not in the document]
hoodmane: Without it I get:
```
PermissionError: [Errno 13] Permission denied: '/home/hood/.cache/cibuildwheel/emsdk-3.1.32/emsdk-main/emsdk'
```

Not really sure why... `zipinfo` shows that the proper permissions are set inside the zip file:
```console
$ zipinfo emsdk-main.zip | grep 'emsdk$'
-rwxr-xr-x  2.3 unx     1642 tx defN 23-Apr-03 18:09 emsdk-main/emsdk
```
And indeed if I unzip it with unzip I get the same permissions:
```console
$ unzip emsdk-main.zip 
Archive:  emsdk-main.zip
886a44a5febaf861d371bfdb191785467303ba63
   creating: emsdk-main/
[...]
$ stat emsdk-main -c %A
drwxrwxr-x
$ stat emsdk-main/emsdk -c %A
drwxrwxr-x
```

But `zipfile.extractall` doesn't preserve permissions:
```
$ python3.11 -c 'from zipfile import ZipFile; zf = ZipFile("emsdk-main.zip"); zf.extractall(".")'
$ stat emsdk-main -c %A
drwxrwxr-x
 stat emsdk-main/emsdk -c %A
-rw-rw-r--
```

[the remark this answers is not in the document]
hoodmane: https://github.com/python/cpython/issues/59999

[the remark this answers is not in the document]
hoodmane: Well I tried copying the workarounds for the zipfile permissions bug but it doesn't work. zipinfo reports the permissions of `emsdk-main` as `drwxrwxr-x` but `zinfo.external_attr >> 16` gives 0 for `emsdk-main`. I carefully read `zipinfo.c` and it does:
```C
xattr = (unsigned)((G.crec.external_file_attributes >> 16) & 0xFFFF);
```
and then proceeds to use it in the way you would expect. So... certainly zipinfo doesn't think `external_attr` should be null...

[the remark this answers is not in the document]
hoodmane: Okay I just said if `zinfo.external_attr >> 16 == 0` then leave the permissions alone. I still haven't read `zipfile.py` and `zipinfo.c` carefully enough to determine just why they disagree on the value of `external_attr` but it's probably not worth it to find out...

ryanking13:
> @ryanking13 do you see any reason not to call replace_so_abi_tags() when building out of tree? This probably fits into your project of making the out of tree builds more the same as the in-tree ones.

Sounds good. We should try to make out of tree and in tree builds identical.

cibuildwheel/pyodide.py
hoodmane: I suspect it's not needed

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,342 @@
...
+    major_minor = f"{majorminor[0]}.{majorminor[1:]}"
+    python_name = f"python{major_minor}"
+    which_python = shutil.which(python_name)
+    if which_python is None:
+        print(
+            f"Error: CPython {major_minor} is not installed.",
+            file=sys.stderr,
+        )
+        raise SystemExit(1)
+    return Path(which_python)
```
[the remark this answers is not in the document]
hoodmane: I would prefer to leave it as is for this PR and if you like we can switch to a different approach in a followup. I think it would add unneeded complexity to this PR. Plus a lot of common CI setups have Python 3.11 preinstalled.

.github/workflows/test.yml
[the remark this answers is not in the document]
hoodmane: Ok trying this...

.github/workflows/test.yml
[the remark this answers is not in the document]
hoodmane: There's an ongoing discussion about whether to call the platform emscripten or pyodide. See the discussion here:
https://github.com/pyodide/pyodide/pull/3823

Whether it should be like emscripten/linux ==> pyodide/manylinux or the other way around pyodide/linux ==> emscripten/manylinux is under debate. Of course neither is really like manylinux at all, it would be more like linux vs blessedlinux where blessedlinux wheels we can upload to pypi. The pypa folks I talked to at pycon seemed open to having Pyodide as a blessedemscripten platform.

cibuildwheel/__main__.py
```diff
@@ -176,6 +177,8 @@ def _compute_platform_only(only: str) -> PlatformName:
         return "macos"
     if "win_" in only or "win32" in only:
         return "windows"
+    if "emscripten_" in only or "pyodide" in only:
```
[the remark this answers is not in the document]
hoodmane: We are thinking about renaming the platform to pyodide. But we might not, it's not entirely clear what is going to happen. It is probably reasonable to remove this for now and add it later if we end up using it.

henryiii:
> You're talking about `replace_so_abi_tags()`? 

There are three fixes, any one of them would be enough:

1. Run `replace_so_abi_tags()` for out-of-tree builds. It's a bit of a hack, but it's a minimum to get things working.
2. Replace emcmake with pyodide's own implementation. This has to be pretty easy, it's a really simple script. This would allow scikit-build-core to get the extension right to begin with, and would support FindPython properly I think.
3. Fix emscripten to not strip this variable when using emcmake. That's a harder fix, and would require an upstream change.

I'd really like both 1 and 2 if possible. :)

.github/workflows/test.yml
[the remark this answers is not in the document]
joerick: > I was thinking that we would standardize a Pyodide ABI and then put Pyodide in the platform tags. It might even be good to go the other way: by default tag the wheel with `emscripten_version_abiflags_wasm32` and switch to `pyodide_version_wasm32` if it is tested to match. Then `emscripten` can be like `linux`: a bit under-determined as a platform tag. And then `pyodide` isn't really like `manylinux` it's more like a specific ABI that is blessed by the PEP.

FWIW, this makes sense to me. Which points towards "pyodide" as the platform name and the term we'd use 99% of the time, which sounds good.

cibuildwheel/resources/build-platforms.toml
```diff
@@ -115,3 +115,8 @@ python_configurations = [
   { identifier = "pp38-win_amd64", version = "3.8", arch = "64", url = "https://downloads.python.org/pypy/pypy3.8-v7.3.11-win64.zip" },
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.11-win64.zip" },
 ]
+
+[pyodide]
+python_configurations = [
+  { identifier = "cp311-emscripten_3_1_32", version = "3.11.2", pyodide_version = "0.23.2", emscripten_version = "3.1.32" },
```
joerick, line 121: So, if we think the output wheels will be called pyodide, shall we change the build identifier to `cp311-pyodide_wasm32`? I like to keep the build identifiers stable. Or do we need to wait for that decision do you think?

test/test_testing.py
```diff
@@ -85,6 +85,7 @@ def test(tmp_path):
             # mac/linux.
             "CIBW_TEST_COMMAND": "false || pytest {project}/test",
             "CIBW_TEST_COMMAND_WINDOWS": "COLOR 00 || pytest {project}/test",
+            "_PYODIDE_EXTRA_MOUNTS": str(tmp_path),
```
joerick, line 88: If adding this is always necessary, can we add `"_PYODIDE_EXTRA_MOUNTS": str(project_dir)` to the actual code, rather than adding it each time in the tests? It's possible that users might want to build from a temp dir too sometimes, and cibulidwheel itself builds from temp dirs when building from an sdist I think.

hoodmane: I will try that. I think we could add it only if `str(project_dir).startswith("/tmp")`. 

Another option I have been considering is just mounting `/tmp/*` in the command line runner. Or perhaps the best thing would be to locate and fix bug in the nodefs implementation that prevents us from mounting `/tmp` itself...

cibuildwheel/resources/build-platforms.toml
hoodmane: Yeah we could do that.

test/test_abi_variants.py
```diff
@@ -156,6 +157,7 @@ def test():
 )
 
 
+@utils.skip_if_pyodide(reason="Doesn't work for some reason")
```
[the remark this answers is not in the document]
hoodmane: Well Pyodide *does* support abi3, it is just pointless: you can build a `cp311-abi3-emscripten_3_1_32_wasm32` wheel but it won't work with the Python 3.10 builds of Pyodide because they used a different version of Emscripten. But the wheel is created exactly in the normal way with the limited ABI etc and it gets loaded just fine too. For instance some of the rust wheels are abi3. So I don't think abi3 is the trouble here.

.github/workflows/test.yml
```diff
@@ -139,6 +139,8 @@ jobs:
       env:
         CIBW_PLATFORM: pyodide
 
-    - name: Run the emulation tests
+    - name: Run tests with 'CIBW_PLATFORM' set to 'pyodide'
       run: |
-        pytest test/test_emscripten.py
+        python ./bin/run_tests.py
+      env:
+        CIBW_PLATFORM: pyodide
```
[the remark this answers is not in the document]
hoodmane: Strange... On ubuntu 20.04 the slowest two tests are:
```
703.00s call     test/test_before_test.py::test
626.81s call     test/test_testing.py::test_extras_require
```
We skip `test_before_test` so that saves 11 minutes by itself. But `test_extras_require` only takes 45 seconds! For comparison on osx `test_extras_require` takes 210 seconds. I wonder how the time breaks down between building and testing. Probably mostly building?

test/test_custom_repair_wheel.py
```diff
@@ -32,13 +31,29 @@ def test(tmp_path, capfd):
     project_dir = tmp_path / "project"
     basic_project.generate(project_dir)
 
-    with pytest.raises(subprocess.CalledProcessError):
+    num_builds = len(utils.cibuildwheel_get_build_identifiers(project_dir))
+    err = None
```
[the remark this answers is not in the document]
hoodmane: I did not know about `nullcontext`, interesting.

test/test_abi_variants.py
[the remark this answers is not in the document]
henryiii: FYI, ABI3 is not all that common, but wheels that avoid the CABI are somewhat common. IMO you need to have the pyodide major version in the pyodide tag, so that `py3-none-pyodide(_22)_wasm32` will work correct - without the pyodide version, you can't tell where it can be loaded.

test/test_before_test.py
```diff
@@ -37,6 +37,7 @@ def test_prefix(self):
 """
 
 
+@utils.skip_if_pyodide(reason="Building with pip doesn't work with Pyodide")
```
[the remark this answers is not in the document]
hoodmane: `python -m pip` and `pip` both work okay, but not if `pip` tries to build the package. There are actually a handful of different stacked problems that come up in this test, but I will try to fix it...

.github/workflows/test.yml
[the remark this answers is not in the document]
joerick: I just did a little experiment locally. Most of our tests don't involve testing, but for the ones that do, the time is spent roughly 70% in the testing phase. I'd _love_ our test suite to be faster, but it's a lot of shelling out to external tools, which in turn do lots of I/O, so gains are hard to come-by.

test/test_before_test.py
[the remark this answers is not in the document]
hoodmane: Okay I think I fixed this.

test/test_abi_variants.py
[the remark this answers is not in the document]
hoodmane: Well it currently seems to just ignore the request for it to be tagged `py3-none-emscripten_wasm32` and tags it `cp311-cp311-emscripten_wasm32`

rth:
What's the current status of this work? Is there any chance to merge it with an experimental status in the near future (since CI passes) and then iron out the potential build issues for individual packages as they come along?

For instance, I would really love to be able to do this for building dev scikit-learn wheels, which already work according to the issue description.

henryiii:
(PS: I'm happy to do the update & rebase, but don't want to force push something if @hoodmane is already working on it - just let me know)

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
henryiii: I'm in favor of this method, at least until something like PyBI comes along. Saving time is worthwhile, I think; at least for now, requiring Python 3.11 isn't that bad for most CIs, and this is already something you have to setup manually.

henryiii:
Ahh, there are 75 commits & multiple merges. Guess I'll just merge too.

henryiii:
I'm curious to see if this fails CI. I know there's a bug which I've just fixed locally but would like to see if anything in CI catches it.

henryiii:
Okay, "no space left on device" wasn't the failure I was expecting...

henryiii:
@hoodmane, do you know why tests are failing to set up a venv?

FWIW, I'm testing it locally with awkward-array like this:

```bash
docker run -v $PWD:/cibuildwheel -v $PWD/../../scikit-hep/awkward:/awkward -w /awkward/awkward-cpp --rm -it python:3.11 bash
pip install /cibuildwheel
cibuildwheel --platform pyodide
```

And I'm getting:

```
FAILED: _ext.cpython-311-x86_64-linux-gnu.so
: && /tmp/tmp5ii37twm/c++ -fPIC -std=c++14 -std=c++14 -std=c++14 -std=c++14 -s DISABLE_EXCEPTION_CATCHING=0 -O3 -DNDEBUG  -flto -shared  -o _ext.cpython-311-x86_64-linux-gnu.so CMakeFiles/_ext.dir/src/python/_ext.cpp.o CMakeFiles/_ext.dir/src/python/content.cpp.o CMakeFiles/_ext.dir/src/python/forth.cpp.o CMakeFiles/_ext.dir/src/python/io.cpp.o  libawkward-static.a  libawkward-cpu-kernels-static.a && cd /awkward/awkward-cpp/build/cpython-311 && emstrip /awkward/awkward-cpp/build/cpython-311/_ext.cpython-311-x86_64-linux-gnu.so
  File "/tmp/tmp5ii37twm/c++", line 668, in <module>
    compiler_main()
  File "/tmp/tmp5ii37twm/c++", line 664, in compiler_main
    sys.exit(handle_command(args, build_args))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/tmp5ii37twm/c++", line 654, in handle_command
    returncode = subprocess.run(new_args).returncode
                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/subprocess.py", line 1024, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.11/subprocess.py", line 1917, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: 'em++'
[174/174] Linking CXX shared library libawkward.so
em++: warning: EXPORTED_FUNCTIONS is not valid with LINKABLE set (normally due to SIDE_MODULE=1/MAIN_MODULE=1) since all functions are exported this mode.  To export only a subset use SIDE_MODULE=2/MAIN_MODULE=2 [-Wunused-command-line-argument]
em++: warning: ignoring unsupported linker flag: `-soname` [-Wlinkflags]
ninja: build stopped: subcommand failed.
```

(This works outside of cibuildwheel on awkward-array, by the way, now that 0.23.3 is out)

rth:
> OSError: [Errno 7] Argument list too long: 'em++'

Seems to be the same issue as in https://github.com/pyodide/pyodide/issues/3427#issuecomment-1374418990

henryiii:
Ahh, yes, that's there in:
https://github.com/scikit-hep/awkward/blob/2d259cc54e3787f1d285a57274e5ce5748ff110f/.github/workflows/docs.yml#L112

Not sure how I'm supposed to pass that with cibuildwheel, though.

henryiii:
@hoodmane would adding an environment variable version of `--exports whole_archive` be a good idea for 0.22.4?

hoodmane:
I think we could do that.

cibuildwheel/resources/build-platforms.toml
```diff
@@ -137,3 +137,8 @@ python_configurations = [
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.12-win64.zip" },
   { identifier = "pp310-win_amd64", version = "3.10", arch = "64", url = "https://downloads.python.org/pypy/pypy3.10-v7.3.12-win64.zip" },
 ]
+
+[pyodide]
+python_configurations = [
+  { identifier = "cp311-pyodide_wasm32", version = "3.11.2", pyodide_version = "0.23.3", emscripten_version = "3.1.32" },
```
henryiii, line 143:
```suggestion
  { identifier = "cp311-pyodide_wasm32", version = "3.11.2", pyodide_version = "0.23.4", emscripten_version = "3.1.32" },
```

rth:
Hmm, so the CI now fails building the `pp310-manylinux_i686` wheel, unless I misread the logs. Not very clear to me what's the actual error..

mayeut:
The error was "No space left", this should now be fixed in main.

celestinoxp:
pyodide updated its version to 0.24 on 13-September-2023...

joerick:
I merged with main to update this to the new build-frontend option.

cibuildwheel/__main__.py
```diff
@@ -45,7 +46,7 @@ def main() -> None:
 
     parser.add_argument(
         "--platform",
-        choices=["auto", "linux", "macos", "windows"],
+        choices=["auto", "linux", "macos", "windows", "pyodide"],
```
martinRenou, line 49:
```suggestion
        choices=["auto", "linux", "macos", "windows", "emscripten"],
```

martinRenou (commented):
Really nice work! Thank you for starting this!

Part from the emscripten-forge team, I could really see us providing the ability for our users to use pip-installed packages into emscripten environments.

For this reason, I have a small suggestion about the name of the platform, something more generic like `emscripten-32` could suit our use-case as well!

hoodmane:
@martinRenou I see valid arguments in both directions for the platform name. I think calling it Pyodide is slightly better: the platform includes not just the ABI defined by the Emscripten version but also what libraries are statically linked into the main Python executable. These conventions are currently defined as "whatever Pyodide does." If other people decide to match these conventions for their own Emscripten Pythons it will work correctly. But I think calling it just "Emscripten" could mislead people? It's not completely clear.

martinRenou:
>  If other people decide to match these conventions for their own Emscripten Pythons it will work correctly

@derthorsten correct me if I'm wrong but I feel like this is totally fine?

This may be the right opportunity for pyodide and emscripten-forge to start to  converge on a general solution for Python packaging in wasm?

ryanking13:
> the platform includes not just the ABI defined by the Emscripten version but also what libraries are statically linked into the main Python executable.

We (Pyodide) link only minimum static libraries to run Python to the main Python module, so if others do this way, too, then it is probably okay. I am not sure how other flags (e.g. `WASM_BIGINT`) should be matched too.

hoodmane:
> This may be the right opportunity for pyodide and emscripten-forge to start to converge on a general solution for Python packaging in wasm?

Well we keep discussing this but it's never all that clear what this entails specifically. Perhaps it would be a good idea to meet semiregularly, though so far we have never seemed to manage that.

> converge on a general solution for Python packaging

If you mean "write down a definition for the platform", I agree that would be a nice thing to do. This is a prerequisite to getting to PyPI uploads.

Another point is that if we don't resolve https://discuss.python.org/t/im-stepping-down-as-sponsor-of-emscripten/41064 the ecosystem support for Emscripten will end up in a pretty confusing state, with support in various packaging tools but being an officially unsupported platform for the interpreter itself.

agriyakhetarpal:
>  numpy -- builds wheel but hangs 99% of the way through test suite

> ril -- rust compilation errors

Hi, this might be low on priority right now for the author(s) and reviewer(s) here, but I wish to point out that this PR can probably be revised, since the NumPy test suite should now pass. `scikit-image` is almost there as well, out-of-tree, and awaits a review, albeit it uses the Node.js test runner rather than the Pyodide-provided virtual environment (in which case the `CIBW_TEST_COMMAND_PYODIDE` environment variable logic might have to be adjusted), and builds with Meson there are succeeding (scikit-image/scikit-image#7350). I am not sure about `ril`, however.

Please let me know if I could help drive this forward in any way, in spite of the fact that my experience with the ecosystem in general might be a bit limited! – both Pyodide and Emscripten have gone through a few minor and patch releases upstream respectively at the time of writing.

henryiii:
Let's try rebasing as soon as the next pyodide release is out, which should be soon.

henryiii:
FYI, I changed the pinning to just install directly, no constraints, as it adds a ton of pins to random packages that could affect other tools. Once we get it going, we should make a new constraints file for pyodide. Might even make sense to isolate `pyodide-build`, as it's likely to conflict with packages.

I wonder if it would make sense to make the `--platform` emscripten, and use `cp312-pyodide_wasm32` for the identifier? I think that would match our plan for sys.platform vs. wheel tag.

hoodmane:
cc @ryanking13 we should also consider reducing the number of dependencies of pyodide-build.

henryiii:
Not a fan of Rich repr’s. They get chopped just to draw a box.

hoodmane:
I know, right? Looks nice but the required fixup when you copy is not worth it.

hoodmane:
Oh it's even worse here it isn't wrapping it's just dropping the text entirely... Maybe you could set term width really large?

hoodmane:
@henryiii pyodide tests pass, can we merge?

docs/options.md
```diff
@@ -247,14 +247,15 @@ environment variables will completely override any TOML configuration.
...
-Options: `auto` `linux` `macos` `windows`
+Options: `auto` `linux` `macos` `windows` `pyodide`
 
 Default: `auto`
 
 `auto` will build wheels for the current platform.
 
 - For `linux`, you need [Docker or Podman](#container-engine) running, on Linux, macOS, or Windows.
 - For `macos` and `windows`, you need to be running on the respective system, with a working compiler toolchain installed - Xcode Command Line tools for macOS, and MSVC for Windows.
+- For `pyodide` you need to be on an x86-64 linux runner and run cibuildwheel from Python 3.11.
```
henryiii, line 258:
```suggestion
- For `pyodide` you need to be on an x86-64 linux runner and run cibuildwheel from Python 3.12.
```

test/test_from_sdist.py
```diff
@@ -53,6 +53,7 @@ def cibuildwheel_from_sdist_run(sdist_path, add_env=None, config_file=None):
 # tests
 
 
+@utils.skip_if_pyodide(reason="TODO: fix!")
```
henryiii, line 56: Does this just need to be updated to 312? Also a few others below.

henryiii (commented):
I need to add the constraints too, then I think it's probably ready.

cibuildwheel/resources/build-platforms.toml
```diff
@@ -166,3 +166,8 @@ python_configurations = [
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.16-win64.zip" },
   { identifier = "pp310-win_amd64", version = "3.10", arch = "64", url = "https://downloads.python.org/pypy/pypy3.10-v7.3.16-win64.zip" },
 ]
+
+[pyodide]
+python_configurations = [
+  { identifier = "cp312-pyodide_wasm32", version = "3.12.1", pyodide_version = "0.26.0a6", emscripten_version = "3.1.58" },
```
henryiii, line 172: @hoodmane Is the wheel tag `pyodide_0_26_wasm32`?

hoodmane: I guess maybe it would better be `312_0` or something... If @hugovk switches CPython releases to match the calendar year then they would sort of come to us. But 312 would under calver actually be Python 2023, but since the Python release is in August and we try to update ~6 months behind we'll always be a different year.

But it's probably best not to worry too much we can switch to a different format if need be.

henryiii: I'm just trying to pick a good name for the constraints file. I think I'll use Python version, then.

test/test_from_sdist.py
hoodmane: I got the obvious ones, there may be more tests that would work if they were against 312 but I'd say we should deal with that in followup work.

henryiii:
I need to do one final commit, the `-x` should be restored, and I have some tweaks to the README. @mayeut or @Czaki would you like to do a quick review? Otherwise I'll merge.

henryiii:
By the way, I think it might be interesting to revisit the more-sharing PR from a while back; with this there will be 4 platforms, and iOS and Android are likely incoming later in the year, so that could be 6 platforms by the end of the year instead of the 3 we have now.

Also, we might want to look at the action to see if there's a good way to allow the platform to be set on it.

hoodmane:
I'll go ahead and try to release 0.26.0 asap. Gotta write release notes.

henryiii:
Okay, this can be bumped when you do.

noxfile.py
```diff
@@ -79,22 +81,44 @@ def update_constraints(session: nox.Session) -> None:
...
 
+    build_platforms = nox.project.load_toml(resources / "build-platforms.toml")
+    pyodides = build_platforms["pyodide"]["python_configurations"]
+    for pyodide in pyodides:
+        python_version = ".".join(pyodide["version"].split(".")[:2])
+        pyodide_version = pyodide["pyodide_version"]
+        output_file = resources / f"constraints-pyodide{python_version.replace('.', '')}.txt"
+        tmp_file = Path(session.create_tmp()) / "constraints-pyodide.in"
+        tmp_file.write_text(
+            f"auditwheel-emscripten\nbuild[virtualenv]\npyodide-build=={pyodide_version}"
```
mayeut, line 109:
```suggestion
            f"auditwheel-emscripten\nbuild[virtualenv]\npip\npyodide-build=={pyodide_version}"
```

This would require re-running `nox -s update_constraints`

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,385 @@
...
+    # upgrade pip to the version matching our constraints
+    # if necessary, reinstall it to ensure that it's available on PATH as 'pip'
+    call(
+        "python",
+        "-m",
+        "pip",
+        "install",
+        "--upgrade",
+        "pip",
+        env=env,
```
mayeut, line 128:
```suggestion
        *dependency_constraint_flags,
        env=env,
```

This requires pip being added to pyodide constraints.

mayeut (changes requested):
I left a few suggestions and some questions that should be addressed I think.

mayeut:
> By the way, I think it might be interesting to revisit the more-shared code PR(s) from a while back.

The [branch is still available on my fork](https://github.com/mayeut/cibuildwheel/tree/issue-560) but hasn't been touched since the PR was closed.

.github/workflows/test.yml
```diff
@@ -148,3 +148,37 @@ jobs:
     - name: Run the emulation tests
       run: |
         pytest --run-emulation test/test_emulation.py
+
+  test-pyodide:
+    name: Test cibuildwheel building pyodide wheels
+    needs: lint
+    runs-on: ubuntu-24.04
```
mayeut, line 155: Do we want to tests other platforms (mostly to ensure some local builds could work) ?
Maybe just a reduced set on other platforms than linux.

test/test_before_test.py
```diff
@@ -35,25 +39,44 @@ def test_prefix(self):
 """
 
 
+@utils.skip_if_pyodide(reason="TODO: fix!")
```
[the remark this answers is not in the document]
mayeut: Is the failure related to the pytest invocation ? I had to change it to `python -m pytest` on a local build of a project I tested this PR on ?

docs/options.md
```diff
@@ -247,14 +247,15 @@ environment variables will completely override any TOML configuration.
...
-Options: `auto` `linux` `macos` `windows`
+Options: `auto` `linux` `macos` `windows` `pyodide`
 
 Default: `auto`
 
 `auto` will build wheels for the current platform.
 
 - For `linux`, you need [Docker or Podman](#container-engine) running, on Linux, macOS, or Windows.
 - For `macos` and `windows`, you need to be running on the respective system, with a working compiler toolchain installed - Xcode Command Line tools for macOS, and MSVC for Windows.
+- For `pyodide` you need to be on an x86-64 linux runner and run cibuildwheel from Python 3.12.
```
mayeut, line 258:
```suggestion
- For `pyodide` you need to be on an x86-64 linux runner and `python3.12` must be available in `PATH`.
```

I did test on macOS arm64 using cibuildwheel running python 3.11.
I'm not sure we want to encourage using anything else than linux x86_64 for CI but it might be worth mentioning that local builds can be done on other platforms if some are really supported.

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,385 @@
...
+
+            if build_options.test_command and build_options.test_selector(config.identifier):
+                log.step("Testing wheel...")
+
+                venv_dir = identifier_tmp_dir / "venv-test"
+                # set up a virtual environment to install and test from, to make sure
+                # there are no dependencies that were pulled in at build time.
+
+                # --no-download??
+                call("pyodide", "venv", venv_dir, env=env)
```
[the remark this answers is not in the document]
mayeut: One of the things that did not work out of the box on a local build test was that I had a broken `node` in my PATH. This led to tests failure when creating the test virtual environment.
That means that the node version from emscripten is not used here. We should probably use that node version to limit differences in build environments.

hoodmane:
Thanks for the review @mayeut!

.github/workflows/test.yml
hoodmane: Building on windows absolutely does not work and will not work until Python gets proper cross builds. Building on macs generally sort of works but we don't test it in any way upstream and cannot test it because mac runners are too expensive. So I think we should just mark it as unsupported. Mac builds would be possible to support if someone decided to fund CI resources for it. But it's likely that nobody will ever care enough.

Or were you thinking that we'd use a docker image or something on these platforms? It's not worth it I don't think. We could check for win32 and make a hard error saying it won't work, and check for Darwin and warn saying we don't actually test this configuration and it is known to be buggy but it sometimes works.

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
hoodmane: We would not want to do this. Generally the node from Emscripten is quite old. It currently includes node 16, but the earliest node that Pyodide fully supports is node v18 (in node v16 http requests won't work correctly unless you apply a manual fix that we've dropped from Pyodide). Starting from node v20 we can make `loop.run_until_complete()` work correctly which will allow us to test packages that use asyncio much more naturally.

test/test_before_test.py
```diff
@@ -14,7 +16,8 @@ def test_version(self):
         # assert that the Python version as written to pythonversion_bt.txt in the CIBW_BEFORE_TEST step
         # is the same one as is currently running.
         # because of use symlinks in MacOS run this test is also need
-        version_file = 'c:\\pythonversion_bt.txt' if sys.platform == 'win32' else '/tmp/pythonversion_bt.txt'
+        tmp_dir = 'c:\\' if sys.platform == 'win32' else '/tmp/my-tmp-dir/'
```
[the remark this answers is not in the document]
mayeut: addressed in #1834

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
mayeut: ok so using the node version from Emscripten is out. Should we add what's needed to install a pinned version of node (e.g. latest LTS version) to prevent environment changes then ?

.github/workflows/test.yml
mayeut: It's probably worth erroring out on Windows & adding a warning on macOS then.

I was pleased to see this working out-of-the-box on macOS, maybe I was just lucky then ;-)

Adding a docker image for these platforms only is not worth it. If going the docker way, then I'd say it should be the only option.

hoodmane: Well I think macOS is generally working, but it's hard to know since we aren't testing it in our CI.

cibuildwheel/resources/defaults.toml
```diff
@@ -46,3 +46,5 @@ repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"
 repair-wheel-command = "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}"
 
 [tool.cibuildwheel.windows]
+
+[tool.cibuildwheel.pyodide]
```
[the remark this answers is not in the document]
henryiii: It’s integrated into `pyodide build`. And because of that it’s actually a dependency of pyodide-build and we don’t really have to list it separately.

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,385 @@
...
+@dataclass(frozen=True)
+class PythonConfiguration:
+    version: str
+    identifier: str
+    pyodide_version: str
+    emscripten_version: str
+
+
+def install_emscripten(tmp: Path, version: str) -> Path:
+    url = "https://github.com/emscripten-core/emsdk/archive/main.zip"
```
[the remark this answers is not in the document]
henryiii: We could pull the version from the same toml file. Will wait for @hoodmane to comment, I’d guess it was left over from development?

cibuildwheel/resources/defaults.toml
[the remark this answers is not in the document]
Czaki: Should we document it somehow? To explain that there is no repair command for windows, but it is builtin for pyodide?

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
hoodmane: Sure it is fine to pull from the toml file. emsdk is just an installer for emscripten and its dependencies, most commits add metadata to install a new version of emscripten. I don't think installing from the main branch would ever be wrong because they never commit anything that breaks installing files. But if people prefer to select the tag corresponding to our emscripten version that should work fine too...

cibuildwheel/resources/build-platforms.toml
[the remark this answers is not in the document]
hoodmane: No, not unless Pyodide bumps the emscripten version to 3.1.60 first. We have to match Pyodide's emscripten version.

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
hoodmane: Sounds reasonable.

agriyakhetarpal:
If it helps, @hoodmane, I can try this out with a few bigger packages that have their out-of-tree builds working, would be better to catch any bugs early – please let me know!

agriyakhetarpal:
If it helps, @hoodmane and @henryiii, I can try this out with a few bigger packages that have their out-of-tree builds working, would be better to catch any bugs early – please let me know!

hoodmane:
Feel free to test anything out. Though I'd like to land this one after this set of reviews it's already been open for over a year. We can fix bugs in smaller follow up prs.

noxfile.py
henryiii: I'll swap auditwheel-emscripten for pip, then. That's currently required by and only used through pyodide build, so listing it again here isn't needed.

henryiii:
Is there a problem with pytest and pyodide? I was trying this out and got a really weird error from ImageMagick and:

```
(help output from ImageMagick)
import: delegate library support not built-in '' (X11) @ error/import.c/ImportImageCommand/1302.
/private/var/folders/_8/xtbws09n017fbzdx9dmgnyyr0000gn/T/cibw-run-f4v4u6s4/cp312-pyodide_wasm32/venv-test/bin/pytest: line 5: from: command not found
/private/var/folders/_8/xtbws09n017fbzdx9dmgnyyr0000gn/T/cibw-run-f4v4u6s4/cp312-pyodide_wasm32/venv-test/bin/pytest: line 7: syntax error near unexpected token `('
/private/var/folders/_8/xtbws09n017fbzdx9dmgnyyr0000gn/T/cibw-run-f4v4u6s4/cp312-pyodide_wasm32/venv-test/bin/pytest: line 7: `    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])'
...
subprocess.CalledProcessError: Command 'pytest --benchmark-disable /Users/henryschreiner/git/scikit-hep/boost-histogram/tests' returned non-zero exit status 2.
```

But `python -m pytest` worked. I think that's what was seen in the unit tests, too; the `pytest` executable script is broken.

.github/workflows/test.yml
henryiii: macOS works fine for me locally. I think I fixed or at least reported bugs a year or two ago about macOS, and it seem okay now. I'd prefer to not put a warning in for macOS unless people report issues. Error on Windows (not on the subsystem for Linux, I'd assume) is fine, though.

test/test_before_test.py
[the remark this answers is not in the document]
henryiii: I tried this and got:

```
subprocess.CalledProcessError: Command '['pyodide', 'venv', '/tmp/cibw-run-73m32l4l/cp312-pyodide_wasm32/venv-test']' returned non-zero exit status 1.
```

cibuildwheel/resources/defaults.toml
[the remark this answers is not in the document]
henryiii: Added a bit to the docs.

test/test_before_test.py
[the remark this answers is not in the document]
henryiii: Made it an xfail, that's better than skipif if the test needs fixing.

cibuildwheel/pyodide.py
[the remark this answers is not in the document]
mayeut: Ok , I'll look into adding a commit for this then.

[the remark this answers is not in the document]
hoodmane: Thanks for the help @mayeut!

hoodmane:
> Could pyodide build read from tool.pyodide.build in pyproject.toml, by the way, perhaps?

Sounds like a great idea I opened a dedicated issue.

hoodmane:
@henryiii `getWasmTableEntry(...) is not a function` is a quite strange error, I'd like to reproduce it. Could you give me a bit more detailed description of what you did?

mayeut:
I'll fix the conflicts I just introduced.

mayeut:
The branch has been rebased/forced push, let's see how it goes.

henryiii:
> @henryiii `getWasmTableEntry(...) is not a function` is a quite strange error, I'd like to reproduce it. Could you give me a bit more detailed description of what you did?

https://github.com/scikit-hep/boost-histogram/pull/935

test/test_environment.py
```diff
@@ -92,10 +92,22 @@ def test_overridden_path(tmp_path, capfd):
...
     assert len(os.listdir(output_dir)) == 0
     captured = capfd.readouterr()
-    assert "python available on PATH doesn't match our installed instance" in captured.err
+    assert "python available on PATH doesn't match our installed instance" in captured.err.replace(
+        "venv", "installed"
+    )
 
 
-@pytest.mark.parametrize("build_frontend", ["pip", "build"])
+@utils.skip_if_pyodide(reason="TODO: fix!")
```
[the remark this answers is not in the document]
mayeut: I'm working on getting this fixed (and the tests that broke with the rebase).

.github/workflows/test.yml
mayeut: In fact there's already a test running on macOS-13. I will enable the test on macOS-14 as it's what I used to fix all the tests I could.

I will add the error message on Windows.

test/test_before_test.py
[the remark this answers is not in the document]
mayeut: the test is fixed

test/test_environment.py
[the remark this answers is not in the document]
mayeut: this iw now fixed

test/test_abi_variants.py
```diff
@@ -184,9 +192,9 @@ def test_abi_none(tmp_path, capfd):
         project_dir,
         add_env={
             "CIBW_TEST_REQUIRES": "pytest",
-            "CIBW_TEST_COMMAND": "pytest {project}/test",
+            "CIBW_TEST_COMMAND": "python -m pytest {project}/test",
```
hoodmane, line 195: It would be great to dig into why this change is needed, since it's indicative of a bug in the pyodide virtual environment code.

test/test_environment.py
```diff
@@ -121,7 +120,7 @@ def test_overridden_pip_constraint(tmp_path, build_frontend):
         setup_py_add=textwrap.dedent(
             """
             import pytz
-            assert pytz.__version__ == "2022.4"
+            assert pytz.__version__ == "2022.4", f"{pytz.__version__!r} != '2022.4'"
```
hoodmane, line 123: Did this just work already?

test/test_pure_wheel.py
```diff
@@ -24,7 +24,11 @@ def a_function():
 """
 
 
-@utils.skip_if_pyodide(reason="Doesn't work")
+@pytest.mark.xfail(
+    condition=utils.platform == "pyodide",
+    reason="pyodide build re-tag platform 'none' as 'pyodide_2024_0_wasm32'",
```
hoodmane, line 29: Ideally this should also have a Pyodide bug so we can track this and fix it.

unit_test/main_tests/main_platform_test.py
```diff
@@ -264,3 +264,16 @@ def test_only_overrides_env_vars(monkeypatch, intercepted_build_args, envvar_nam
     assert options.globals.build_selector.skip_config == ""
     assert options.platform == "linux"
     assert options.globals.architectures == Architecture.all_archs("linux")
+
+
+def test_pyodide_on_windows(monkeypatch, capsys):
```
hoodmane, line 269: Fantastic! Thanks!

test/test_emscripten.py
```diff
@@ -1,20 +1,44 @@
...
+if __name__ == "__main__":
+    check_node()
+"""
 
 
 @pytest.mark.parametrize("use_pyproject_toml", [True, False])
 def test_pyodide_build(tmp_path, use_pyproject_toml):
-    if platform.machine() == "arm64":
-        pytest.skip("emsdk doesn't work correctly on arm64")
```
hoodmane: I wonder if we took in some of @henryiii's fixes and that is what made this work. I also think emsdk had some bugs on arm64 last time I tried to use it. I had a macbook pro as my work laptop for a while and it couldn't build Pyodide and I couldn't figure out why.

test/test_environment.py
mayeut: the a change in pyodide.py in commit https://github.com/pypa/cibuildwheel/pull/1456/commits/4ac060d60361a36b2f33cbff3d16b87ee8357558 fixed this test. I just added some debug string/help if this test fails again.

test/test_emscripten.py
```diff
@@ -1,20 +1,44 @@
...
+def check_node():
+    node = shutil.which("node")
+    assert node is not None, "node is None"
+    node_path = Path(node).resolve(strict=True)
+    cibw_cache_path = Path(sys.argv[1]).resolve(strict=True)
+    assert cibw_cache_path in node_path.parents, f"{cibw_cache_path} not a parent of {node_path}"
+    node_js = run_js("globalThis.process.execPath")
+    assert node_js is not None, "node_js is None"
+    node_js_path = Path(node_js).resolve(strict=True)
+    assert node_js_path == node_path, f"{node_js_path} != {node_path}"
```
hoodmane, line 30: Could you add comments to explain what is going on here? This is pretty confusing.

mayeut: done

test/test_environment.py
hoodmane: Okay I really appreciate you helping because the last time I tried to fix these I just got so confused.

bin/update_nodejs.py
hoodmane, line 1: Thanks so much for adding all of this @mayeut. It would have taken me a long time to write and probably not have worked as well. Did you copy it from somewhere? It looks very detailed.

mayeut: I browsed downloads from nodejs to check how we could adapt our update scripts and found their index.json so it was only a matter of copying/adapting from our existing update scripts for virtualenv/python.

test/test_abi_variants.py
hoodmane: Opened a Pyodide bug to track this https://github.com/pyodide/pyodide/issues/4802

test/test_pure_wheel.py
hoodmane: Will be resolved by https://github.com/pyodide/pyodide/pull/4803.

docs/setup.md
```diff
@@ -107,6 +107,13 @@ You can override the cache folder using the ``CIBW_CACHE_PATH`` environment vari
     Download link: https://www.python.org/ftp/python/3.6.8/python-3.6.8-macosx10.9.pkg
     ```
 
+### Pyodide (WebAssembly) builds (experimental)
+
+Pre-requisite: you need to have a matching host version of Python (unlike all
+other cibuildwheel platforms) and Node 20. Linux host highly recommended; macOS hosts may
```
henryiii, line 113: I think the Node 20 part isn't required anymore?

henryiii (approved):
Two things I'd like to see fixed before release: `pytest` needs to work, most people use that vs. `python -m pytest`. I think this might be on us, as it seems to work if you do it by hand.

The other is the error I'm seeing in boost-histogram, but pretty sure that's a Pyodide bug, since it shows up without cibuildwheel too.

cibuildwheel/pyodide.py
```diff
@@ -0,0 +1,398 @@
...
+                # set up a virtual environment to install and test from, to make sure
+                # there are no dependencies that were pulled in at build time.
+
+                virtualenv_env = env.copy()
+                virtualenv_env["PATH"] = os.pathsep.join(
+                    [
+                        str(ensure_node(config.node_version)),
+                        virtualenv_env["PATH"],
+                    ]
+                )
```
mayeut, line 340: Should we insert node in PATH in `setup_python` rather than this late in the process ?

hoodmane:
Hopefully can fix boost-histogram in Pyodide 0.26.1 =)

test/test_abi_variants.py
mayeut: It seems that the issue only happens on macOS (either arm64 or x86_64). I tested on linux (aarch64/x86_64) and there were no issues.

mayeut:
> pytest needs to work, most people use that vs. python -m pytest. I think this might be on us, as it seems to work if you do it by hand.

@henryiii, did you do it by hand on macOS too or are you referring to the workflow in boost-histogram ? (I'm seeing a different behavior under cibuildwheel depending on whether we're running on Linux or macOS).

test/test_abi_variants.py
mayeut: I added a commit that workaround the issue in our tests and a link to the issue in the docs as we're saying that macOS is not fully supported.

mayeut (approved):
With the `pytest` invocation stuff being worked-out (mostly).
I don't have any more reservations to this being merged.

The question I have regarding node might be resolved later.

mayeut:
@henryiii, can you review the last commits please ?

cibuildwheel/resources/build-platforms.toml
```diff
@@ -166,3 +166,8 @@ python_configurations = [
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.16-win64.zip" },
   { identifier = "pp310-win_amd64", version = "3.10", arch = "64", url = "https://downloads.python.org/pypy/pypy3.10-v7.3.16-win64.zip" },
 ]
+
+[pyodide]
+python_configurations = [
+  { identifier = "cp312-pyodide_wasm32", version = "3.12.1", pyodide_version = "0.26.0a6", emscripten_version = "3.1.58", node_version = "v20" },
```
henryiii, line 172:
```suggestion
  { identifier = "cp312-pyodide_wasm32", version = "3.12.1", pyodide_version = "0.26.0", emscripten_version = "3.1.58", node_version = "v20" },
```

mayeut: local tests are failing. Not sure why yet.

mayeut: local failures were due to the parallel initialization of xbuildenv. Now fixed (& tests updated for 0.26.0).

hoodmane:
I guess it's failing because of the color / formatting in `pyodide build -h`.

hoodmane:
Hmm test passes locally...

test/test_build_frontend_args.py
```diff
@@ -19,12 +19,11 @@ def test_build_frontend_args(tmp_path, capfd, frontend_name):
     project.generate(project_dir)
 
     # the build will fail because the frontend is called with '-h' - it prints the help message
+    add_env = {"CIBW_BUILD_FRONTEND": f"{frontend_name}; args: -h"}
+    if utils.platform == "pyodide":
+        add_env["TERM"] = "dumb"  # disable color / style
```
hoodmane, line 24: Hopefully this works? I have no idea why it passes locally though??

hoodmane: Seems that it fixed it indeed =)

mayeut: it seems too have been style/color as you pointed out.

henryiii: Maybe something is checking for the GitHub actions var and assuming it is in a CI run that supports color?

henryiii (approved):
IMO, this is ready to merge, we can do followup patches if we find any improvements. It looks like the two outstanding issues might both be on the pyodide side, so we might not want to make a release until 0.26.1.

Unless anyone has any reservations, will merge tomorrow.

henryiii:
> Edit: I guess the manual one was on Linux,

Yes, https://github.com/scikit-hep/boost-histogram/pull/938 (linux) makes it through the "pytest" command and dies on the other bug. But we were seeing this issue in cibuildwheel's tests, even on Linux.

henryiii:
Thanks everyone!

PR #1458, opened 2023-04-05
chore: add structurally typed GenericPythonConfiguration

hoodmane:
Use structural typing for the return value of get_python_configurations so that we don't have to specify a union of all possible `PythonConfiguration` types.

This makes it slightly more convenient to add a new platform.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -9,6 +9,7 @@ import textwrap
 import typing
 from pathlib import Path
 from tempfile import mkdtemp
+from collections.abc import Sequence, Set
 
 import cibuildwheel
 import cibuildwheel.linux
@@ -344,9 +345,9 @@ def print_preamble(platform: str, options: Options, identifiers: list[str]) -> N
 
 
 def get_build_identifiers(
-    platform: PlatformName, build_selector: BuildSelector, architectures: set[Architecture]
+    platform: PlatformName, build_selector: BuildSelector, architectures: Set[Architecture]
 ) -> list[str]:
-    python_configurations: typing.Sequence[GenericPythonConfiguration]
+    python_configurations: Sequence[GenericPythonConfiguration]
 
     if platform == "linux":
         python_configurations = cibuildwheel.linux.get_python_configurations(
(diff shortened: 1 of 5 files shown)


henryiii (approved):

cibuildwheel/typing.py
```diff
@@ -42,3 +42,9 @@
 
 PlatformName = Literal["linux", "macos", "windows"]
 PLATFORMS: Final[set[PlatformName]] = {"linux", "macos", "windows"}
+
+
+class GenericPythonConfiguration(Protocol):
+    @property
+    def identifier(self) -> str:
+        pass
```
henryiii, line 50:
```suggestion
        ...
```

Typing uses `...` conventionally.

cibuildwheel/__main__.py
```diff
@@ -341,11 +346,7 @@ def print_preamble(platform: str, options: Options, identifiers: list[str]) -> N
 def get_build_identifiers(
     platform: PlatformName, build_selector: BuildSelector, architectures: set[Architecture]
 ) -> list[str]:
-    python_configurations: (
-        list[cibuildwheel.linux.PythonConfiguration]
-        | list[cibuildwheel.windows.PythonConfiguration]
-        | list[cibuildwheel.macos.PythonConfiguration]
-    )
+    python_configurations: typing.Sequence[GenericPythonConfiguration]
```
henryiii, line 349:
```suggestion
    python_configurations: Sequence[GenericPythonConfiguration]
```

henryiii: Ahh, can't add `from collections.abc import Sequence` as a suggestion. This is fine for now, but ideally we should do that (also `Set`, from there, for input arguments, is better than a concrete and mutable `set`).

hoodmane: oh yeah it hadn't occurred to me that `set[A]` is similar to `MutableSet[A]`? but still more specific? But `Dict` is the same as `dict` and `Map` / `MutableMap` are different?

henryiii: `Dict` is just a older way to say `dict[...]`. I think `typing.Set` / `set` are also similar, while `collections.abc.{Map,MutableMap,Set,MutableSet}` are more general structural types. Input parameters should generally be as loose as possible, and output parameters should be as specific as possible (usually concrete).

PR #1460, opened 2023-04-07
chore: factor out _compute_platform from build_in_directory

hoodmane:
build_in_directory was ~140 lines long which is a bit long for my taste. It also has a bunch of nested if statements which I don't prefer. The first ~60 lines are computing the platform, so I factored those out into helper methods which have clearer intent and less indentation.


mayeut (approved):

henryiii (approved):

hoodmane:
Could someone merge this?

PR #1461, opened 2023-04-07
chore: Add a PlatformModule type

hoodmane:
This avoids the repetition of the platform checks in __main__.py by defining a dataclass with `get_python_configurations` and `build`.


mayeut (approved):

hoodmane:
Could someone merge this?

joerick:
I've just been playing with this and I think it would be simpler to use the module itself as the 'interface' object. We can type it with a Protocol I believe, like so

```python
class PlatformModule(typing.Protocol):
    # note that as per PEP544, the self argument is ignored when the protocol
    # is applied to a module
    def get_python_configurations(self, build_selector: BuildSelector, architectures: Set[Architecture]) -> typing.Sequence[GenericPythonConfiguration]:
        ...
    def build(self, options: Options, tmp_path: Path) -> None:
        ...
```

cibuildwheel/__main__.py
```diff
@@ -244,6 +244,16 @@ def _compute_platform(args: CommandLineArguments) -> PlatformName:
     return _compute_platform_ci()
 
 
+def get_platform_interface(platform: PlatformName) -> PlatformInterface:
+    if platform == "linux":  # noqa: SIM116
+        return cibuildwheel.linux.interface
+    elif platform == "windows":
+        return cibuildwheel.windows.interface
+    elif platform == "macos":
```
henryiii, line 252:
```suggestion
    if platform == "windows":
        return cibuildwheel.windows.interface
    if platform == "macos":
```

Given the final `assert_never` is unindented (which I prefer), these should be if's, not elif's.

(if this is still valid after trying @joerick's suggestion)

joerick:
I have a version that looks to be working locally, I can push here if you like?

joerick:
I'm off to bed so I've pushed that, hope that's okay. I think it's a bit simpler without the extra object.

cibuildwheel/__main__.py
```diff
@@ -244,6 +244,28 @@ def _compute_platform(args: CommandLineArguments) -> PlatformName:
...
+    ) -> Sequence[GenericPythonConfiguration]:
+        ...
+
+    def build(self, options: Options, tmp_path: Path) -> None:
+        ...
+
+
+def get_platform_module(platform: PlatformName) -> PlatformModule:
+    if platform == "linux":
+        return cibuildwheel.linux
```
hoodmane, line 261: Nice! I did not realize you could do this.

hoodmane: When I tried, I left out the self parameter and got complaints.

joerick (approved):

PR #1462, opened 2023-04-07
Set the default build_frontend to be "default" (which is still "pip")

hoodmane:
Instead of using pip as the default, set it to default. This way each platform can choose independently what to do as the default behavior.

This is intended as part of adding Pyodide support #1456. Pyodide cannot use pip as the build frontend. It always uses a modified pypa/build. So if the build_frontend is explicitly set to pip we want to raise an error. But if it's unset, then we should be able to tell and just do what we want.

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -71,6 +71,15 @@ test_fail_cwd_file: Final[Path] = resources_dir / "testing_temp_dir_file.py"
 
 BuildFrontend = Literal["pip", "build"]
 
+
+def build_frontend_or_default(
+    setting: BuildFrontend | Literal["default"], default: BuildFrontend
+) -> BuildFrontend:
+    if setting == "default":
+        return default
+    return setting
+
+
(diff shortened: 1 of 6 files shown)


cibuildwheel/util.py
```diff
@@ -71,6 +71,15 @@
...
 
+
+def build_frontend_or_default(
+    setting: BuildFrontend | Literal["default"], default: BuildFrontend
+) -> BuildFrontend:
+    if setting == "default":
+        return default
+    return setting
+
+
```
mayeut, line 82: minor detail, if switching the default from pip to build in the future, we'd now have to change multiple files to ensure consistency.
I wonder if taking the `config.identifier` rather than a `default` BuildFrontend and managing the actual default BuildFrontend in this function would be better or worse ?

hoodmane: What about just giving the default argument a default value `default = "pip"`?

mayeut (approved):

joerick (approved):

PR #1469, opened 2023-04-10
Remove pydantic from example lists

### LINKED ISSUE #1463: pydantic doesn't seem to use cibuildwheel
pydantic is listed as an example here:
https://cibuildwheel.readthedocs.io/en/stable/working-examples/
grepping for cibuildwheel doesn't seem to show anything (@samuelcolvin ?)

hoodmane:
Resolves #1463


Czaki:
I prefer only to update the link to point to 1.10 branch.

hoodmane:
Does the script that generates the table from `projects.yml` support linking to a branch?

joerick:
I'm happy to remove since it's not actively in use. Plenty of sample projects to choose from!

PR #1474, opened 2023-04-17
fix: verbosity isn't supported for pypa/build

henryiii:
This was incorrectly adding an empty config setting, and it was not correct if included for build - it's a pip wheel only setting. Reported in #1471.


Czaki (approved):

ianthomas23:
Thanks for fixing this so quickly.

PR #1475, opened 2023-04-17
chore: minor cleanup

henryiii:
Adding a bit of cleanup from looking at #1297 again. Drops the unmaintained `[mypy]` extra.

- chore: clean up mypy a little bit
- chore: clean up pylint/black a bit

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -256,6 +256,7 @@ class PlatformModule(typing.Protocol):
         ...
 
 
+# pylint: disable-next=inconsistent-return-statements
 def get_platform_module(platform: PlatformName) -> PlatformModule:
     if platform == "linux":
         return cibuildwheel.linux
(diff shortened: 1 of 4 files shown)


cibuildwheel/__main__.py
```diff
@@ -256,6 +256,7 @@ def build(self, options: Options, tmp_path: Path) -> None:
         ...
 
 
+# pylint: disable-next=inconsistent-return-statements
 def get_platform_module(platform: PlatformName) -> PlatformModule:
```
henryiii, line 260: FYI, I think the need for RET 503 below might be a bug in Ruff: https://github.com/charliermarsh/ruff/issues/3304

henryiii: It's been fixed, the noqa can go away in the next Ruff release :)

joerick (approved):

PR #1476, opened 2023-04-18
chore: ruff and typing updates

henryiii:
This adds a couple of Ruff checks, and updates typing. The deprecated typing generic aliases are replaced by proper `collections.abc` counterparts, checked by an added Ruff check. And list/set/dict have been updated to Protocols when used as an input.

- chore: touch up Ruff, add exe check
- chore: add Ruff logging format check
- chore: update typing to collections.abc
- chore: update typing to be generic on function args


joerick (commented):
I'm a little confused by the typing.Something -> collections.abc.Something change. Are abstract base classes the same as protocols? My understanding is that a protocol is something that anything can conform to, as long as it has the right interface and regardless of its class hierarchy, but an ABC needs to be subclassed to be considered a match. By that logic Protocols seems more appropriate, I think.

Hmm. I just found [PEP 585](https://peps.python.org/pep-0585/#parameters-to-generics-are-available-at-runtime), it seems I have some reading to do!

joerick:
Ahh, I see that these types were always aliases to collections.abc, even before Python 3.9. So my point around Protocols is invalid. Still not totally sure I understand why typing works this way, but I see that this PR does indeed fix the use of the deprecated types, so it seems good to me.

joerick (approved):

PR #1486, opened 2023-04-29
chore: update to virtualenv<20.22.0

mayeut:
I'd like not to drop 3.6 right away (& possibly up-to centos 7 EOL or OpenSSL 1.1.1 EOL) as it is still used in around 15% of wheel builds (on linux at least).

### DIFF
--- a/bin/update_virtualenv.py
+++ b/bin/update_virtualenv.py
@@ -53,6 +53,10 @@ def git_ls_remote_versions(url) -> list[VersionTuple]:
             if version.is_prerelease:
                 log.info("Ignoring pre-release %r", str(version))
                 continue
+            # Do not upgrade past 20.22.0 to keep python 3.6 compat
+            if version >= Version("20.22.0"):
+                log.info("Ignoring %r which is not compatible with python 3.6", str(version))
+                continue
             versions.append(VersionTuple(version, version_string))
         except InvalidVersion:
             log.warning("Ignoring ref %r", ref)


henryiii:
Don’t we need the new version of virtualenv for python 3.12? https://github.com/pypa/virtualenv/pull/2558

Can’t users needing an old version of Python use an old version of cibuildwheel, an old version of manylinux, etc? How many of that 15% are actually wheels that are still being updated, and how many are old wheels for projects that have already dropped 3.6?

mayeut:
> How many of that 15% are actually wheels that are still being updated

That's for the projects published in the past 28 days (I rebuilt the graph for https://mayeut.github.io/manylinux-timeline/ changing the sliding window to 28 days).

> Don’t we need the new version of virtualenv for python 3.12 ?

We can revisit at this time if needed (admittedly, very soon).

Just for comparison, when 2.7 & 3.5 were dropped in July 2021, the usage was around 6% & 9% (of around 700 packages for a 28 days period at that time vs around 1300 packages nowadays for a 28 days period)
If we look at the consumer side (which does not take into account wheels still being updated), it was 5% & 2% vs 10%

mayeut:
> Don’t we need the new version of virtualenv for python 3.12 ?
> > We can revisit at this time if needed (admittedly, very soon).

The answer is we don't need it: https://github.com/mayeut/cibuildwheel/actions/runs/4850698405/jobs/8643858588

joerick:
I'm okay with holding on a little longer with Python 3.6 if you feel it's important @mayeut . But I wonder if there might be a bit of a chicken-egg situation here, where some users are gonna keep using 3.6 until they realise they need to upgrade for a package, and packagers are gonna keep releasing 3.6 wheels while cibuildwheel supports it, and cibuildwheel/manylinux aren't gonna drop 3.6 while users... you get the picture :)

henryiii:
See also https://github.com/pypa/cibuildwheel/issues/1375. I do know there's a strong desire to support "everything cibuildwheel/manylinux supports" (I've been influenced by that before). But it's okay for users who really need to support old versions to use old versions of things, including old cibuildwheel/manylinux.

Windows working might not be a guarantee of what will happen on Linux or macOS.

We should continue to support whatever manylinux does, I think, but if it starts forcing caps on users who don't care about 3.6, that could start increasing the burden of supporting it.

henryiii:
PS: Not sure about "right away" - we are already a year and four months after 3.6's EoL. :)

joerick (approved):

Issue #1487, opened 2023-04-30
Project setup.cfg causes errors in cibuildwheel installing other packages


rogerbinns:
### Description

My project is a C extension.  Of note it has no dependencies on any other package for building or testing.

To customize the github actions binary build I added a `setup.cfg` to it with some additional options defined.

    [build]
    fetch = True
    enable_all_extensions = True
    enable = COLUMN_METADATA

This is causing an error when cibuildwheel is installing other packages because the current working directory is my project and my `setup.cfg` is affecting their installation.  In particular the `fetch` option causes their installation to fail because they have no knowledge of it (and it is my option not theirs).  

Here is an example of the problem occurring:

````
Installing build tools...
  
  + pip install --upgrade setuptools wheel -c 'C:\Program Files (x86)\pipx\.cache\f150de73f83b030\Lib\site-packages\cibuildwheel\resources\constraints-python36.txt'
  Requirement already satisfied: setuptools in c:\users\runneradmin\appdata\local\temp\cibw-run-6rvfifu1\cp36-win32\build\venv\lib\site-packages (59.6.0)
  Requirement already satisfied: wheel in c:\users\runneradmin\appdata\local\temp\cibw-run-6rvfifu1\cp36-win32\build\venv\lib\site-packages (0.37.1)
  ERROR: Exception:
[ ... frames elided ...]
  distutils.errors.DistutilsOptionError: error in setup.cfg: command 'build' has no such option 'fetch'
Error: Command ['pip', 'install', '--upgrade', 'setuptools', 'wheel', '-c', 'C:\\Program Files (x86)\\pipx\\.cache\\f150de73f83b030\\Lib\\site-packages\\cibuildwheel\\resources\\constraints-python36.txt'] failed with code 2. None
````

I did a second run where I changed the setup.cfg to not be present until just before the build so that it didn't affect the build packages being installed (using CIBW_BEFORE_BUILD).  The test step then fell over:

````
Testing wheel...
  
      + pip install virtualenv -c /constraints.txt
  ERROR: Exception:
[ ... frames elided ...]
  distutils.errors.DistutilsOptionError: error in setup.cfg: command 'build' has no such option 'fetch'
````

I tried removing the setup.cfg in CIBW_BEFORE_TEST but that is run after the step above so it is too late.

[...]

PR #1488, opened 2023-05-01
fix: install our dependencies outside the project directory

henryiii:
Testing a suggestion from #1487.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -313,11 +313,15 @@ def build_in_container(
             )
 
             testing_temp_dir = PurePosixPath(
-                container.call(["mktemp", "-d"], capture_output=True).strip()
+                container.call(["mktemp", "-d"], capture_output=True, cwd="/").strip()
             )
             venv_dir = testing_temp_dir / "venv"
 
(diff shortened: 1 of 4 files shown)


henryiii:
There are serious networking issues with GHA currently, so needs to be rebuilt later.

henryiii:
@pypa/cibuildwheel-team, what do you think of this patch? This is a fix for a weird problem when you have custom build command options in `setup.cfg` that causes Python < 3.10 to break when running pip in the build directory - which breaks pip installs, virtual environment creation, and modern `pyproject.toml` based build isolation. By making sure pip is not run in the build directory, the failure no longer happens on `pip install <cibuildwheel's stuff>`, making it easier for the package author to work around and make it work (though all the other stuff is still broken, like adding a `pyproject.toml`).

I'm neutral on this - I'm not very interested in making sure cibuldwheel works when `pip install`, `python -m venv`, and adding a `pyproject.toml` are all broken, but this seems like a low-cost "fix" that will help make a workaround possible and will keep cibuildwheel out of the tracebacks. :)

joerick:
I'm -1 on this, it adds complexity to our code without offering a functionality benefit. If we were to add it, I'd like to include a test also, but adding a test that ensures broken projects still make it through seems a bit perverse!

PR #1494, opened 2023-05-08
Fix a few types to keep pyright happy

joerick:
I use VS Code and it doesn't like these unspecified Final types, so I'm specifying fully to get rid of the red squigglies.


henryiii (approved):
I’ve noticed this too and update them when I notice it.

Czaki (approved):

PR #1498, opened 2023-05-12
chore: Simplify host platform logic

hoodmane:
There is some duplicated code for deciding what the host platform is. This groups it together in util.py.

I was slightly confused in some places whether `startswith("win")` or `== "win32"` was desired. I added a new `HOST_IS_WIN32` variable since I think it's as explicit as possible that the intended check is specifically for win32 and not for windows in general. Since `HOST_IS_WINDOWS` is for that.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -23,6 +23,8 @@ from cibuildwheel.options import CommandLineArguments, Options, compute_options
 from cibuildwheel.typing import PLATFORMS, GenericPythonConfiguration, PlatformName
 from cibuildwheel.util import (
     CIBW_CACHE_PATH,
+    HOST_IS_WINDOWS,
+    HOST_PLATFORM,
     BuildSelector,
     CIProvider,
     Unbuffered,
@@ -164,7 +166,7 @@ def main() -> None:
     finally:
         # avoid https://github.com/python/cpython/issues/86962 by performing
         # cleanup manually
-        shutil.rmtree(temp_dir, ignore_errors=sys.platform.startswith("win"))
+        shutil.rmtree(temp_dir, ignore_errors=HOST_IS_WINDOWS)
         if temp_dir.exists():
             log.warning(f"Can't delete temporary folder '{temp_dir}'")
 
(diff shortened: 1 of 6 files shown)


cibuildwheel/__main__.py
```diff
@@ -197,19 +199,15 @@ def _compute_platform_ci() -> PlatformName:
             file=sys.stderr,
         )
         sys.exit(2)
-    if sys.platform.startswith("linux"):
-        return "linux"
-    elif sys.platform == "darwin":
-        return "macos"
-    elif sys.platform == "win32":
```
hoodmane, line 204: This used to say `sys.platform == "win32"` but this changes it to `startswith("win")`. Am I introducing a bug with this behavior change? If so I'd like to add a comment explaining why we use `== "win32"` here but in architecture we use `startswith("win")`.

cibuildwheel/architecture.py
```diff
@@ -74,20 +74,14 @@ def parse_config(config: str, platform: PlatformName) -> set[Architecture]:
     def auto_archs(platform: PlatformName) -> set[Architecture]:
         native_machine = platform_module.machine()
 
-        # Cross-platform support. Used for --print-build-identifiers or docker builds.
-        host_platform: PlatformName = (
-            "windows"
-            if sys.platform.startswith("win")
-            else ("macos" if sys.platform.startswith("darwin") else "linux")
-        )
```
hoodmane, line 82: I made this PR because it was bothering me that the logic here does not exactly match the logic in `_compute_platform_ci()`.

cibuildwheel/util.py
```diff
@@ -93,7 +93,24 @@ def build_frontend_or_default(
...
+        return "windows"
+    else:
+        return None
+
+
+HOST_PLATFORM: Final[PlatformName | None] = get_host_platform()
+
+HOST_IS_WINDOWS: Final[bool] = get_host_platform() == "windows"
+
+HOST_IS_WIN32: Final[bool] = sys.platform == "win32"
```
hoodmane, line 113: This might be a bit excessive but I think it makes the intended distinction between these two checks 100% clear.

hoodmane:
Well this seems to pass the tests. Maybe there should be more unit test coverage for the difference between `sys.platform == win32` and `sys.platform.startswith('win')`...

joerick:
This is one of those cases where backward compatibility was preferred over correctness - the 'win32' value applies even on 64-bit windows. See here: 

https://github.com/python/cpython/blob/b15a1a6ac6ea0d7792036e639e90f0e51400c2ee/PC/pyconfig.h#L321-L333

(Found via [this SO answer](https://stackoverflow.com/a/2145582/382749))

So there's no functional difference between `sys.platform.startswith("win")` and `sys.platform == 'win32'`.

henryiii:
It's a recommended practice to always use `sys.platform.startswith` - for some platforms it matters, and it should be possible to "extend" this without breaking checks. MyPy, etc. support this. So my recommendation is never to check via `==`.

I personally don't like pulling this check out - I'd rather it just always be `sys.platform.startswith("win")`. It used to be better to do that statically, though these days you can use `Final` and it should get assigned a `Literal` boolean value (though, there's a chance some type checkers will complain about that, some don't seem to like Final deduction!). But I like the explicit way to do it - no need to look up some magic constant when you can specify exactly what you mean in the same number of lines.

joerick:
Agreed on the literal `sys.platform.startswith("win")` over a constant - it's clear and means I don't have to look up a constant definition if i'm in doubt. I'm okay with the `startswith()` convention too - I guess if we had used that everywhere there wouldn't be the 32/64 confusion in the first place!

hoodmane:
Okay will close this and do another which switches to using `startswith("win")` throughout since I think that is the consensus here.

PR #1499, opened 2023-05-15
Add create_args suboption to CIBW_CONTAINER_ENGINE

### LINKED ISSUE #1384: Okay to add arbitrary flags to docker create?
### Description

Hi all,

I used your project to create manylinux2014 wheels for a [project](https://github.com/nvidia-riva/riva-asrlib-decoder#pre-built-wheels) that uses CUDA. This in itself is quite interesting because I needed to (1) keep a stable ABI on the pybind11 boundaries, (2) not link to any cuda shared libraries, (3) make sure that the wheel size stayed small, and (4) the version of gcc used in manylinux2014 has some incapabilities with. This workflow is definitely worth of sharing more broadly since a lot of CUDA projection adoption is hampered by a lack of pre-built binaries. But that is not the current topic.

Building wheels is perfectly fine for the build process, which uses only CPUs, but I noticed failures when I tried to test because GPUs (which were required by the test cases) on the host machine weren't accessible within the container. The fix, in short, was to add "---gpus all" to the end of the flags here: https://github.com/pypa/cibuildwheel/blob/27fc88e6385a995e61a87ee4b903bed263e6a6e2/cibuildwheel/oci_container.py#L88-L92

I made a quick hack where I added an extra variable (CIBW_EXTRA_DOCKER_CREATE_ARGS, which I set to "--gpus all") to support my use case. Unfortunately, there's no clever way to make docker automatically see from what I've found other than changing the  (I would be happy to be corrected). Is this a change that you would be interested in accepting? I can see how allowing users to add arbitrary commands to the docker create command could be a bit fishy (e.g., they could override `--volume=/:/host` if they were silly), but cibuildwheel does seem like an expert's tools anyway. I would greatly appreciate any feedback form the maintainers.

### Build log

_No response_

### CI config

_No response_

joerick:
Fix #1384

Adds the ability to give additional command line arguments to the container engine (docker or podman). These are supplied inside the existing CIBW_CONTAINER_ENGINE option in a backward-compatible format so we don't increase the options surface area, and because, in theory at least, these options could vary in future depending on the container engine used.

### DIFF
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -192,7 +192,7 @@ class OCIContainer:
         if from_path.is_dir():
             self.call(["mkdir", "-p", to_path])
             subprocess.run(
-                f"tar cf - . | {self.engine} exec -i {self.name} tar --no-same-owner -xC {shell_quote(to_path)} -f -",
+                f"tar cf - . | {self.engine.name} exec -i {self.name} tar --no-same-owner -xC {shell_quote(to_path)} -f -",
                 shell=True,
                 check=True,
                 cwd=from_path,
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -11,6 +11,7 @@ import sys
 import typing
 import uuid
 from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
 from pathlib import Path, PurePath, PurePosixPath
 from types import TracebackType
 from typing import IO, Dict
@@ -19,7 +20,16 @@ from ._compat.typing import Literal
 from .typing import PathOrStr, PopenBytes
 from .util import CIProvider, detect_ci_provider
 
-ContainerEngine = Literal["docker", "podman"]
+ContainerEngineName = Literal["docker", "podman"]
+
+
+@dataclass(frozen=True)
+class OCIContainerEngineConfig:
+    name: ContainerEngineName
+    create_args: Sequence[str] = ()
+
+
+DEFAULT_ENGINE = OCIContainerEngineConfig("docker")
 
 
 class OCIContainer:
@@ -57,7 +67,7 @@ class OCIContainer:
         image: str,
         simulate_32_bit: bool = False,
         cwd: PathOrStr | None = None,
-        engine: ContainerEngine = "docker",
+        engine: OCIContainerEngineConfig = DEFAULT_ENGINE,
     ):
         if not image:
             msg = "Must have a non-empty image to run."
@@ -84,13 +94,14 @@ class OCIContainer:
 
         subprocess.run(
             [
-                self.engine,
+                self.engine.name,
                 "create",
                 "--env=CIBUILDWHEEL",
                 f"--name={self.name}",
                 "--interactive",
                 "--volume=/:/host",  # ignored on CircleCI
                 *network_args,
+                *self.engine.create_args,
                 self.image,
                 *shell_args,
             ],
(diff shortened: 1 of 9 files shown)


unit_test/oci_container_test.py
```diff
@@ -21,7 +21,7 @@
 pm = platform.machine()
 if pm == "x86_64":
     DEFAULT_IMAGE = "quay.io/pypa/manylinux2014_x86_64:2020-05-17-2f8ac3b"
-elif pm == "aarch64":
+elif pm in ["aarch64", "arm64"]:
```
joerick, line 24: Similarly, this was just me getting tests running on my local machine

joerick:
I don't expect this to change much, but I'll mark as draft until I can get the tests to pass!

joerick:
Travis failure looks unrelated. Ready for review!

cibuildwheel/oci_container.py
```diff
@@ -11,15 +11,44 @@
...
+@dataclass(frozen=True)
+class OCIContainerEngineConfig:
+    name: ContainerEngineName
+    create_args: Sequence[str] = ()
+
+    @staticmethod
+    def from_config_string(config_string: str) -> OCIContainerEngineConfig:
+        config_dict = parse_key_value_string(config_string, ["name"])
+        name = " ".join(config_dict["name"])
+        if name not in ["docker", "podman"]:
```
henryiii, line 35:
```suggestion
        if name not in {"docker", "podman"}:
```

Ultra minor stylistic preference - this is logically closer to the correct meaning - `in`/`not in` are ideally set operations.

They both compile to exactly the same bytecode when used on literals like this, however, so take it or leave it. ;)

cibuildwheel/oci_container.py
```diff
@@ -11,15 +11,44 @@
...
+            msg = f"unknown container engine {name}"
+            raise ValueError(msg)
+
+        name = typing.cast(ContainerEngineName, name)
+        # some flexibility in the option name to cope with TOML conventions
+        create_args = config_dict.get("create_args") or config_dict.get("create-args") or []
```
henryiii, line 41: Do we do this anywhere else? I'm pretty sure we decided not to, and instead just have nice error messages if it fails. There's not much reason to add this when you don't have to worry about backward compatibility. It's better to have one correct way instead of multiple nearly identical ways. It's easier for a third party to parse, for example, if there's one and only one way to do it. For example, I am often code-searching to see what people are doing in their pyproject.toml. It makes it much harder to grep if there's not a clear way to spell it.

This is a bit different since it's combing from inside an environment variable or from pyproject.toml. I'm not as worried about it when it comes from an environment variable, but it would at least be nice to not have ambiguity when it comes from pyproject.toml.

unit_test/oci_container_test.py
henryiii, line 24:
```suggestion
elif pm in {"aarch64", "arm64"}:
```

Same minor comment.

henryiii (approved):

cibuildwheel/oci_container.py
joerick: This wasn't intended to be a change in policy for options, it's more that because the way options parsing works, the option is transformed from a dict to a string in the parser, so it would be a bit more code and more fiddly to be strict about it, and I didn't think it was worth it :) . I'm documenting the kebab-case in TOML and snake_case in env vars, so I figured accidental usage would be rare.

henryiii: Okay, fair enough.

PR #1500, opened 2023-05-15
Use PyPI Trusted Publisher for releases

joerick:
Basically copying https://github.com/scikit-build/scikit-build-core/pull/306
And following https://docs.pypi.org/trusted-publishers/adding-a-publisher/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,9 @@ jobs:
     needs: [dist]
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
+    environment: pypi
+    permissions:
+      id-token: write
 
     steps:
     - uses: actions/download-artifact@v3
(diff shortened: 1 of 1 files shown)


henryiii (approved):
Also https://scikit-hep.org/developer/gha_pure#distribution-pure-python-wheels :)

Remember to set up on PyPI.

Czaki:
Did you define the pypi environment in repository settings?

henryiii:
It will create it the first time it’s used. Though if you want to control who can release, you should create or edit it. :)

webknjaz:
> control who can release

Or put a cooldown timer, for example...

.github/workflows/release.yml
```diff
@@ -27,6 +27,9 @@ jobs:
     needs: [dist]
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
+    environment: pypi
```
webknjaz, line 30: A nicer syntax could be
```suggestion
    environment:
        name: pypi
        url: https://pypi.org/p/cibuildwheel
```
which makes the URL show up in the workflow graph UI (and some other places). I like rendering the versioned URL even, when I have a version variable available...

PR #1511, opened 2023-06-02
Minor docs fix

ianthomas23:
This is a minor docs fix. It looks like it was originally a copy-paste-modify but with the modify stage missing.

Whilst I am here, thanks very much for such a useful project.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -550,7 +550,7 @@ a table of items, including arrays.
     single values.
 
 Platform-specific environment variables also available:<br/>
-`CIBW_BEFORE_ALL_MACOS` | `CIBW_BEFORE_ALL_WINDOWS` | `CIBW_BEFORE_ALL_LINUX`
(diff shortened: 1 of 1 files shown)


joerick (approved):
Thanks!

docs/options.md
```diff
@@ -550,7 +550,7 @@ a table of items, including arrays.
     single values.
```
henryiii, line 550: Not needed for this PR, but this is not true anymore, pip supports arrays, as long as we don't stop it in cibuildwheel (which I don't think we do), it should just work.

ianthomas23: Yes, pip support for multiple config settings with the same key was added in 23.1. Here's a link to the changelog: https://pip.pypa.io/en/stable/news/#v23-1

henryiii: That reminds me, this would still be true for 3.6.

PR #1520, opened 2023-06-12
unify config in `pyproject.toml`

Borda:


Borda:
well this is strange; no idea where this comes from :(
```
ValueError: option names {'--run-podman'} already added
  File "/tmp/cirrus-ci-build/./bin/run_tests.py", line 31, in <module>
    subprocess.run(unit_test_args, check=True)
  File "/usr/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,

Borda:
> You need to follow PEP 621, not just copy values over. Several tools can do this for you. I use `pipx run hatch new --init` to convert.

I did create it by following the official docs...

> Already trying this sort of thing in #1297.

ok, then I ll close this one as duplicate :flamingo:

PR #1525, opened 2023-06-17
feature: add PyPy 3.10

mayeut:
This adds PyPy 3.10 and updates PyPy to 7.3.12 for PyPy 3.9

c.f. discussion in #1524

PR #1526, opened 2023-06-17
chore: drop python 3.7 for cibuildwheel driver

mayeut:
This drops python 3.7 for cibuildwheel.
Building python 3.6 / 3.7 packages is still supported.


henryiii (approved):
This is a week early, but it makes sense to do with a minor point bump, which #1525 requires too, so doing them together makes sense.

henryiii:
Could you bump ruff too? Maybe I can from my phone, actually…

joerick (approved):
Lovely stuff, thank you @mayeut. Let's just hold on merging this for a few days in case we need to get a patch release out in the meantime.

henryiii:
3.7's EoL is today.

.pre-commit-config.yaml
```diff
@@ -34,9 +34,9 @@ repos:
   rev: v1.3.0
   hooks:
   - id: mypy
-    name: mypy 3.7 on cibuildwheel/
+    name: mypy 3.8 on cibuildwheel/
     exclude: ^cibuildwheel/resources/.*py$
-    args: ["--python-version=3.7"]
+    args: ["--python-version=3.8"]
     additional_dependencies: &mypy-dependencies
```
henryiii, line 40: FYI, the markdown-it-py cap below can be removed, it was there to support Python 3.7.

mayeut: I'd rather be explicit here if it's OK.

henryiii: I mean “markdown-it-py<3” can be removed. I can’t comment on the correct line.

mayeut: ok, I misunderstood completely.
will do

henryiii:
I'll probably retry #1297 after this goes in.

henryiii:
> for a few days

It's been three weeks. I'd like to get something out before SciPy next week.

joerick:
My apologies, I was holding out for a minor release but had forgotten that #1525 went in a couple of weeks back.

PR #1532, opened 2023-06-22
docs: use trusted publisher for GHA docs/example

### LINKED ISSUE #1522: [docs] Secretless publishing to PyPI
### Description

I noticed that the GHA docs have a snippet with an updated version of [`pypi-publish`] and suggest using username/password there. The username field has always defaulted to `__token__` so I've hidden that from the readme some time ago.
And now that PyPI supports trusted publishing, the docs here should stop suggesting people to make new non-expireable tokens. Instead, drop that password field.
Though, this requires not using that old tag. I specifically use `release/v1` so the docs don't get outdated regularly and just hint to the users that they may use pins instead. I suggest you do the same.

[`pypi-publish`]: https://github.com/marketplace/actions/pypi-publish

henryiii:
Close #1522. Simplified the example a bit.

### DIFF
--- a/examples/github-deploy.yml
+++ b/examples/github-deploy.yml
@@ -48,6 +48,8 @@ jobs:
     permissions:
       id-token: write
     if: github.event_name == 'release' && github.event.action == 'published'
+    # or, alternatively, upload to PyPI on every tag starting with 'v' (remove on: release above to use this)
+    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     steps:
       - uses: actions/download-artifact@v3
         with:
--- a/examples/github-deploy.yml
+++ b/examples/github-deploy.yml
@@ -1,14 +1,14 @@
 name: Build and upload to PyPI
 
-# Build on every branch push, tag push, and pull request change:
-on: [push, pull_request]
-# Alternatively, to publish when a (published) GitHub Release is created, use the following:
-# on:
-#   push:
-#   pull_request:
-#   release:
-#     types:
-#       - published
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types:
+      - published
 
 jobs:
   build_wheels:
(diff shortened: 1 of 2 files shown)


examples/github-deploy.yml
```diff
@@ -44,10 +44,10 @@ jobs:
...
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    # upload to PyPI on every tag starting with 'v'
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
-    # alternatively, to publish when a GitHub Release is created, use the following rule:
-    # if: github.event_name == 'release' && github.event.action == 'published'
+    environment: pypi
+    permissions:
+      id-token: write
+    if: github.event_name == 'release' && github.event.action == 'published'
```
joerick, line 50: I'd rather keep the reference for release-on-tag, some projects don't bother with Github releases (e.g. django, pip, tornado). But they're in the minority, so switching the default to release-based makes sense

```suggestion
    environment: pypi
    permissions:
      id-token: write
    if: github.event_name == 'release' && github.event.action == 'published'
    # or, alternatively, upload to PyPI on every tag starting with 'v' (remove on: release above to use this)
    # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
```

joerick (approved):
Looks good, thanks

examples/github-deploy.yml
henryiii: Do environments work properly with release on tag, @webknjaz? I was rather assuming things like review deployments might be tied to releases.

henryiii: I see it mentioned in the docs for the action, so it must be okay. I think "environments" and "releases" are not really as integrated as I thought (which might be why I can make a release then still have to review the deployment).

webknjaz: @henryiii not sure why you thought they depend on the events.. I'm not aware of any problems. PyPI only checks the env name, and on the GH side you may have extra protections attached.

By the way, I like to use the full syntax and specify env URLs so they are better embedded in the UI.

Issue #1504, opened 2023-05-23
Allow test command to access {wheel} path


anthrotype:
### Description

I would like to use [tox](https://tox.wiki/en/latest/) to run the tests against the wheel that cibuildwheel has just built.
Tox has an `--installpkg` option that takes the path to a precompiled wheel to be installed, in alternative to its default mode whereby it builds the package from source and installs it within its own virtual environment before running the tests.

I know cibuildwheel has a special `{wheel}` placeholder that one can use in a custom repair-wheel command; however it appears that the cibuildwheel test_command (CIBW_TEST_COMMAND env variable) does not support the `{wheel}` special placeholder, only `{package}` and `{project}` seem to be supported from the test command.

Would you accept a PR that enables the cibuildwheel test command to also interpolate the built wheel's full path using the same `{wheel}` placeholder used for repair command?

I hacked this on my fork here: 

https://github.com/pypa/cibuildwheel/compare/main...anthrotype:cibuildwheel:test_command_wheel?expand=1

... and I have verified it working on the CI for a project I work on:
https://github.com/fonttools/fonttools-wheels/blob/332fbd94f1186efd02d18c692a5ed207df293bb3/.github/workflows/ci.yml#L20

Thanks for considering this!

joerick:
Seems okay to me! Using cibuildwheel in this way you'd have to accept that we're gonna create the venv and install the wheel regardless, even though you're not gonna use that install. But if that isn't an issue for you, I don't see why not.

PR #1533, opened 2023-06-22
fix: allow access to {wheel} in more places

henryiii:
Taken by clicking the link in #1504.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -349,6 +349,7 @@ def build_in_container(
                 build_options.test_command,
                 project=container_project_path,
                 package=container_package_dir,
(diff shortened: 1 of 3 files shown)


joerick:
It's interesting to see which CI providers merge HEAD automatically and which don't. I was surprised by that CI feature when I first saw it but I've come to like it, as it prevents issues like this and also it's essentially testing the code that _would_ exist if you merged this PR, so catches bugs that pop up after the merge.

joerick (approved):

henryiii:
Cirrus CI, Circle CI, and ReadTheDocs doesn’t either (though not obvious above).

PR #1538, opened 2023-06-28
Support GraalPy

timfel:
This PR adds support for GraalPy.

A couple of questions:

1. GraalPy is not included in manylinux images, so I added a facility to download Pythons there as well. Does that seem ok?
2. I couldn't find a way to exclude GraalPy by default - I'm not sure, but maybe it would be good to be conservative here and let people opt-in to building wheels on GraalPy via the `CIBW_BUILD` variable explicitly.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,6 +20,7 @@ jobs:
 
 - job: macos_311
   pool: {vmImage: 'macOS-13'}
+  timeoutInMinutes: 120
   steps:
     - task: UsePythonVersion@0
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ pr:
 
 jobs:
 - job: linux_311
-  timeoutInMinutes: 120
+  timeoutInMinutes: 180
   pool: {vmImage: 'Ubuntu-22.04'}
   steps:
     - task: UsePythonVersion@0
--- a/cibuildwheel/resources/build-platforms.toml
+++ b/cibuildwheel/resources/build-platforms.toml
@@ -18,6 +18,7 @@ python_configurations = [
   { identifier = "pp39-manylinux_x86_64", version = "3.9", path_str = "/opt/python/pp39-pypy39_pp73" },
   { identifier = "pp310-manylinux_x86_64", version = "3.10", path_str = "/opt/python/pp310-pypy310_pp73" },
   { identifier = "pp311-manylinux_x86_64", version = "3.11", path_str = "/opt/python/pp311-pypy311_pp73" },
+  { identifier = "gp242-manylinux_x86_64", version = "3.11", path_str = "/opt/python/graalpy311-graalpy242_311_native" },
   { identifier = "cp38-manylinux_aarch64", version = "3.8", path_str = "/opt/python/cp38-cp38" },
   { identifier = "cp39-manylinux_aarch64", version = "3.9", path_str = "/opt/python/cp39-cp39" },
   { identifier = "cp310-manylinux_aarch64", version = "3.10", path_str = "/opt/python/cp310-cp310" },
@@ -57,6 +58,7 @@ python_configurations = [
   { identifier = "pp39-manylinux_aarch64", version = "3.9", path_str = "/opt/python/pp39-pypy39_pp73" },
   { identifier = "pp310-manylinux_aarch64", version = "3.10", path_str = "/opt/python/pp310-pypy310_pp73" },
   { identifier = "pp311-manylinux_aarch64", version = "3.11", path_str = "/opt/python/pp311-pypy311_pp73" },
+  { identifier = "gp242-manylinux_aarch64", version = "3.11", path_str = "/opt/python/graalpy311-graalpy242_311_native" },
   { identifier = "pp38-manylinux_i686", version = "3.8", path_str = "/opt/python/pp38-pypy38_pp73" },
   { identifier = "pp39-manylinux_i686", version = "3.9", path_str = "/opt/python/pp39-pypy39_pp73" },
   { identifier = "pp310-manylinux_i686", version = "3.10", path_str = "/opt/python/pp310-pypy310_pp73" },
@@ -143,6 +145,8 @@ python_configurations = [
   { identifier = "pp310-macosx_arm64", version = "3.10", url = "https://downloads.python.org/pypy/pypy3.10-v7.3.19-macos_arm64.tar.bz2" },
   { identifier = "pp311-macosx_x86_64", version = "3.11", url = "https://downloads.python.org/pypy/pypy3.11-v7.3.19-macos_x86_64.tar.bz2" },
   { identifier = "pp311-macosx_arm64", version = "3.11", url = "https://downloads.python.org/pypy/pypy3.11-v7.3.19-macos_arm64.tar.bz2" },
+  { identifier = "gp242-macosx_x86_64", version = "3.11", url = "https://github.com/oracle/graalpython/releases/download/graal-24.2.0/graalpy-24.2.0-macos-amd64.tar.gz" },
+  { identifier = "gp242-macosx_arm64", version = "3.11", url = "https://github.com/oracle/graalpython/releases/download/graal-24.2.0/graalpy-24.2.0-macos-aarch64.tar.gz" },
 ]
 
 [windows]
@@ -171,6 +175,7 @@ python_configurations = [
   { identifier = "pp39-win_amd64", version = "3.9", arch = "64", url = "https://downloads.python.org/pypy/pypy3.9-v7.3.16-win64.zip" },
   { identifier = "pp310-win_amd64", version = "3.10", arch = "64", url = "https://downloads.python.org/pypy/pypy3.10-v7.3.19-win64.zip" },
   { identifier = "pp311-win_amd64", version = "3.11", arch = "64", url = "https://downloads.python.org/pypy/pypy3.11-v7.3.19-win64.zip" },
+  { identifier = "gp242-win_amd64", version = "3.11", arch = "64", url = "https://github.com/oracle/graalpython/releases/download/graal-24.2.0/graalpy-24.2.0-windows-amd64.zip" },
 ]
 
 [pyodide]
--- a/test/test_0_basic.py
+++ b/test/test_0_basic.py
@@ -1,3 +1,4 @@
+import re
 import textwrap
 
 import pytest
@@ -41,7 +42,22 @@ def test(tmp_path, build_frontend_env, capfd):
     # Verify pip warning not shown
     captured = capfd.readouterr()
     for stream in (captured.err, captured.out):
-        assert "WARNING: Running pip as the 'root' user can result" not in stream
+        # manylinux-interpreters ensure will run pip on GraalPy as the root
+        # user once, so we will get the warning below. We make sure it is only
+        # because of manylinux and nothing else. We do this by checking that
+        # the error message only occurs directly after a call to graalpy's
+        # ensurepip in the manylinux `/opt/` directory
+        start = 0
+        bad_msg = "WARNING: Running pip as the 'root' user can result"
+        assert_msg = f"{bad_msg} is only tolerated from manylinux's ensure for graalpy"
+        for m in re.finditer(bad_msg, stream):
+            idx = stream.rfind("/opt/", start, m.end())
+            start = m.end()
+            m2 = re.match(
+                "/opt/.*/graalpy.*/bin/python -m ensurepip --default-pip", stream[idx : m.start()]
+            )
+            assert m2 is not None, assert_msg
+            assert idx + m2.end() + 1 == m.start(), assert_msg
         assert "A new release of pip available" not in stream
 
 
--- a/test/test_0_basic.py
+++ b/test/test_0_basic.py
@@ -1,4 +1,3 @@
-import re
 import textwrap
 
 import pytest
@@ -42,22 +41,7 @@ def test(tmp_path, build_frontend_env, capfd):
     # Verify pip warning not shown
     captured = capfd.readouterr()
     for stream in (captured.err, captured.out):
-        # manylinux-interpreters ensure will run pip on GraalPy as the root
-        # user once, so we will get the warning below. We make sure it is only
-        # because of manylinux and nothing else. We do this by checking that
-        # the error message only occurs directly after a call to graalpy's
-        # ensurepip in the manylinux `/opt/` directory
-        start = 0
-        bad_msg = "WARNING: Running pip as the 'root' user can result"
-        assert_msg = f"{bad_msg} is only tolerated from manylinux's ensure for graalpy"
-        for m in re.finditer(bad_msg, stream):
-            idx = stream.rfind("/opt/", start, m.end())
-            start = m.end()
-            m2 = re.match(
-                "/opt/.*/graalpy.*/bin/python -m ensurepip --default-pip", stream[idx : m.start()]
-            )
-            assert m2 is not None, assert_msg
-            assert idx + m2.end() + 1 == m.start(), assert_msg
+        assert "WARNING: Running pip as the 'root' user can result" not in stream
         assert "A new release of pip available" not in stream
 
 
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -41,8 +41,8 @@ def test(tmp_path):
     project_with_before_build_asserts.generate(project_dir)
 
     before_build = (
-        """python -c "import sys; open('{project}/pythonversion_bb.txt', 'w').write(sys.version)" && """
-        f'''python -c "import sys; open('{{project}}/pythonprefix_bb.txt', 'w').write({SYS_PREFIX})"'''
+        """python -c "import sys; f = open('{project}/pythonversion_bb.txt', 'w'); f.write(sys.version); f.close()" && """
+        f'''python -c "import sys; f = open('{{project}}/pythonprefix_bb.txt', 'w'); f.write({SYS_PREFIX}); f.close()"'''
     )
     frontend = "build"
     if utils.platform != "pyodide":
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -41,8 +41,8 @@ def test(tmp_path):
     project_with_before_build_asserts.generate(project_dir)
 
     before_build = (
-        """python -c "import sys; f = open('{project}/pythonversion_bb.txt', 'w'); f.write(sys.version); f.close()" && """
-        f'''python -c "import sys; f = open('{{project}}/pythonprefix_bb.txt', 'w'); f.write({SYS_PREFIX}); f.close()"'''
+        """python -c "import pathlib, sys; pathlib.Path('{project}/pythonversion_bb.txt').write_text(sys.version)" && """
+        f'''python -c "import pathlib, sys; pathlib.Path('{{project}}/pythonprefix_bb.txt').write_text({SYS_PREFIX})"'''
     )
     frontend = "build"
     if utils.platform != "pyodide":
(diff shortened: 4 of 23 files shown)


timfel:
FYI the failures we have right now are because of my question regarding 2 above: I couldn't find a way to exclude GraalPy by default, and the tests all find it included in the list of default builds. If anyone can point me to how to exclude GraalPy from the default builds, I think those would all pass.

henryiii:
> GraalPy is not included in manylinux images

Would it make sense to include it?

> I couldn't find a way to exclude GraalPy by default

Maybe it's time to think about adding this. It comes up occasionally. I'd like to also remove pypy37 and pypy38 from the default, as they are no longer supported by upstream.

joerick:
> > GraalPy is not included in manylinux images
> 
> Would it make sense to include it?

It would certainly be neater if it was included. Also, I've never heard of GraalPy, so I can't assess wether it's inclusion and maintenance here would be worth it for users. If Manylinux were to adopt it, that would make the decision much easier!

> > I couldn't find a way to exclude GraalPy by default
> 
> Maybe it's time to think about adding this. It comes up occasionally. I'd like to also remove pypy37 and pypy38 from the default, as they are no longer supported by upstream.

Hm. Yes, agreed. It would also allow us to soft-drop EoL CPythons, while keeping the functionality. The simplest way to do this would be to set a default for CIBW_BUILD. E.g. the default could be `CIBW_BUILD=cp* pp{39,310}-*`. One downside of this would be some projects ([example](https://sourcegraph.com/github.com/google/neuroglancer/-/blob/.github/workflows/build.yml?L118)) set CIBW_BUILD to `*`, for example, in a matrix where some other configurations are set. The fix in such cases, I suppose, would be to set to empty string instead `""`. On the upside, it does change the default as we want, and doesn't add additional complexity.

joerick:
Converting to draft until manylinux support lands.

mayeut:
manylinux support has landed. It will be available in cibuildwheel once the update workflow runs & gets merge.

timfel:
I've updated the PR to use the manylinux support, but in the meantime `uv` has become included in manylinux and used by default, and that doesn't support GraalPy, yet. See also https://github.com/astral-sh/uv/pull/5141

timfel:
uv now supports GraalPy on `main`, once the next release is out we need to include it in manylinux images and then I'll revisit this PR

mayeut:
Unfortunately, the new version of `uv` still does not work on manylinux.
@timfel, you can opt-out from using uv with graalpy for now, c.f. #1941 for a sample opting out PyPy 3.8 on Windows.

joerick:
@timfel, I don't know if you're still actively working on this, but we have a v3.0 release coming up, which would be a perfect time to land a feature like this!

timfel:
> @timfel, I don't know if you're still actively working on this, but we have a v3.0 release coming up, which would be a perfect time to land a feature like this!

What is the timeframe for 3.0? We are still working on this, but needed a few fixes in tox, uv, and virtualenv to go in first. The virtualenv fix is still not merged because we can only pass it with the new GraalPy release that is due out in March

joerick:
Ah, I didn't realise there were blockers.

> What is the timeframe for 3.0?

No definitive timeline, (when it's ready ;) ) but probably some time over the next few weeks.

joerick:
March is probably too far out for us to hold the 3.0 release for specifically. Not a problem, it can arrive in a point update, I just thought it would be neat :)

henryiii:
> March is probably too far out for us to hold the 3.0 release for specifically

I think it could make it, actually. :)

henryiii:
I think cibuildwheel 3.0 might be done around May 6, could this be ready before then, perhaps?

timfel:
> I think cibuildwheel 3.0 might be done around May 6, could this be ready before then, perhaps?

I'm picking this up now, hopefully the virtualenv PR is merged soon and then there are no issues remaining

henryiii:
https://github.com/pypa/virtualenv/releases/tag/20.30.0

timfel:
> https://github.com/pypa/virtualenv/releases/tag/20.30.0

I'm fixing things in [my fork](https://github.com/timfel/cibuildwheel/actions/workflows/test.yml) for now to avoid loading your CI unduly. I'm also waiting for manylinux to push a new image with the latest graalpy after they [merged the latest update](https://github.com/pypa/manylinux/pull/1762)

timfel:
So, I have the Linux and macOS builds [green](https://github.com/timfel/cibuildwheel/actions/runs/14199215909), but I'm at a bit of a loss as to why the Windows test is failing.

In the Github runner, I see a [slightly different stderr context]((https://github.com/timfel/cibuildwheel/actions/runs/14199215909/job/39793071743#step:15:1089) than in the Azure runner [here](https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=7120&view=logs&j=6867f353-080e-5b5a-fb38-3ea557023d8c&t=34fa7c82-514c-5af7-becb-1141c13c5d74&l=2077), but the error message is the same
```
Build failed because a wheel named spam-0.1.0-cp311-cp311-win_amd64.whl was already generated in the current run.
```

What I don't understand is why these would be generated twice? I find it hard to reproduce. Locally I sometimes get the error, sometimes not, and I'm not sure what a good place would be for a `breakpoint()`. Any hints for what I could look for or what I could try?

henryiii:
Wild guess: is it trying to build for both 32-bit and 64-bit on Windows, but actually producing the 64-bit wheels both times?

henryiii:
Does GraalPy support the stable ABI? I see one test failing for that, I'd assume it's not supported. I also see `<frozen graalpy.pip_hook>:48: RuntimeWarning: You are using an untested version of pip. GraalPy provides patches and workarounds for a number of packages when used with compatible pip versions. We recommend to stick with the pip version that ships with this version of GraalPy.` but maybe that's on one of the pinning tests?

henryiii:
It looks to me that `gp*` is not producing a graalpy wheel, but instead the host Python's wheel.

```
+ python --version
GraalPy 3.11.7 (Oracle GraalVM Native 24.2.0)
...
+ python -m build 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-2\\popen-gw2\\test0\\project' --wheel '--outdir=C:\\Users\\runneradmin\\AppData\\Local\\Temp\\cibw-run-sxknk4dt\\gp242-win_amd64\\built_wheel' --no-isolation
* Getting build dependencies for wheel...
stored_version 3.13.2 (tags/v3.13.2:4f8bb39, Feb  4 2025, 15:23:48) [MSC v.1942 64 bit (AMD64)]
sys.version 3.13.2 (tags/v3.13.2:4f8bb39, Feb  4 2025, 15:23:48) [MSC v.1942 64 bit (AMD64)]
...
Successfully built spam-0.1.0-cp313-cp313-win_amd64.whl
```

timfel:
> Could this be an issue with `build`? Since this is a Windows issue, I guess I can't just use setup-python to test it in pypa/build's test suite.

Yes, GraalPy has a few bugs on Windows when using build. I got past most of them now I think, failed at 98% of the tests just earlier so I hope I'll have this ready by tomorrow.

timfel:
Ok, everything [green](https://github.com/timfel/cibuildwheel/actions/runs/14248869324/job/39936650117) now in Github actions. Let's see if circleci and zure agree :)

henryiii:
Looks like it's hitting a "couldn't find uv" error. Our tests run with or without uv.

timfel:
> Looks like it's hitting a "couldn't find uv" error. Our tests run with or without uv. Is uv required? We could skip if it's not present.

Ah, yeah, I linked https://github.com/oracle/graalpython/issues/491 - we have a bug when build uses virtualenv *and* isolation for building only on Windows, so my workaround was to force using uv with GraalPy on Windows. I guess we'll have to skip then.

timfel:
Thank you @henryiii for fixing the pipeline :) Just to make sure no one is waiting here for us now, is there anything else I should do about this PR right now, or is it fine to just wait for review comments?

cibuildwheel/resources/build-platforms.toml
```diff
@@ -18,6 +18,7 @@ python_configurations = [
   { identifier = "pp39-manylinux_x86_64", version = "3.9", path_str = "/opt/python/pp39-pypy39_pp73" },
   { identifier = "pp310-manylinux_x86_64", version = "3.10", path_str = "/opt/python/pp310-pypy310_pp73" },
   { identifier = "pp311-manylinux_x86_64", version = "3.11", path_str = "/opt/python/pp311-pypy311_pp73" },
+  { identifier = "gp242-manylinux_x86_64", version = "3.11", path_str = "/opt/python/graalpy311-graalpy242_311_native" },
```
henryiii, line 21: Wheels have to be made for each minor version too? I guess this means we can't drop all the workarounds until we drop `gp242`?

henryiii (approved):

azure-pipelines.yml
```diff
@@ -16,18 +16,19 @@ jobs:
         docker run --rm --privileged docker.io/tonistiigi/binfmt:latest --install all
         python -m pip install dependency-groups
         python -m dependency_groups test | xargs python -m pip install -e.
-        python ./bin/run_tests.py
+        python ./bin/run_tests.py --num-processes 2
```
henryiii, line 19:
```suggestion
        python ./bin/run_tests.py
```

azure-pipelines.yml
```diff
@@ -16,18 +16,19 @@ jobs:
...
+  timeoutInMinutes: 120
   steps:
     - task: UsePythonVersion@0
       inputs:
         versionSpec: '3.11'
     - bash: |
         python -m pip install dependency-groups
         python -m dependency_groups test | xargs python -m pip install -e.
-        python ./bin/run_tests.py --num-processes 2
+        python ./bin/run_tests.py --num-processes 3
```
henryiii, line 31:
```suggestion
        python ./bin/run_tests.py
```

azure-pipelines.yml
```diff
@@ -39,4 +40,4 @@ jobs:
     - bash: |
         python -m pip install dependency-groups
         python -m dependency_groups test | xargs python -m pip install -e.
-        python ./bin/run_tests.py
+        python ./bin/run_tests.py --num-processes 2
```
henryiii, line 43:
```suggestion
        python ./bin/run_tests.py
```

henryiii (commented):
This can be batched with other changes if someone else requests changes.

cibuildwheel/resources/build-platforms.toml
timfel: We are doing patch releases, I missed the merge window for a patch release that goes out next week, but I'm confident I can get the required fixes in for our patch release in June, at which time we can drop this.

Versioning is a bit confusing right now, we used to do calendar versioning with two releases per year, but we are switching to align with the OpenJDK release schedule and will then have 2 releases per year. At that point every 4th release will be a long-term support release, releases in between will only receive support for 6 months. The upcoming release 25 (due out in Sep) will be an LTS, so we could then stick to that for 2 years and also have 26/27/28 each for 6 months until we go to 29 at which point 25 is out of support.

joerick: So in terms of ABI stability, a sample wheel I just produced was given the filename
`spam-0.1.0-graalpy311-graalpy242_311_native-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_28_x86_64.whl`. So it does look like there's only ABI stability within each minor release of GraalPy (i.e. across patch releases), is that right? 

(I just want to check the logic behind this build identifier, as most have the python version at the start, not the interpreter version. But if the ABI stability is tied to the GraalPy minor version, this makes sense to me.)

test/test_before_build.py
```diff
@@ -41,8 +41,8 @@ def test(tmp_path):
     project_with_before_build_asserts.generate(project_dir)
 
     before_build = (
-        """python -c "import sys; open('{project}/pythonversion_bb.txt', 'w').write(sys.version)" && """
-        f'''python -c "import sys; open('{{project}}/pythonprefix_bb.txt', 'w').write({SYS_PREFIX})"'''
+        """python -c "import sys; f = open('{project}/pythonversion_bb.txt', 'w'); f.write(sys.version); f.close()" && """
+        f'''python -c "import sys; f = open('{{project}}/pythonprefix_bb.txt', 'w'); f.write({SYS_PREFIX}); f.close()"'''
```
joerick, line 45: What's the reason for this change? I wouldn't expect a problem here given that the Python process is exiting immediately - is the behaviour on GraalPy different?

henryiii: As written, it's invalid Python, even though it's generally fine on CPython. Files are supposed to be explicitly closed always. PyPy on Windows often trips up on this. This isn't really correct either, since if there was an issue writing the file, it would not close. I think this should instead be

```python
import sys, pathlib; pathlib.Path('{project}/pythonversion_bb.txt').write_text(sys.version)
```

(Or it should be in a context manager, but the above keeps it in one line)

henryiii: (PS: I am not answering for GraalPy, just on the construct in general from the language standpoint)

cibuildwheel/resources/build-platforms.toml
henryiii: I don't think we've got a plan for how to handle this if PyPy ever has a new ABI other than 7.3?

I'd assume GraalPy would hopefully bump the Python version once a year, but that's still two releases per Python version, so it seems unavoidable.

test/test_before_build.py
timfel: The flushing of files on CPython happens because the refcount will drop to 0 and finalization will flush and close the file. A GC, however, may not ever run (especially for short running programs that are not under memory pressure) and thus finalization may never be triggered. This is what's happening for GraalPy in this case (well, sometimes you get lucky and it does flush, but it's not guaranteed). As @henryiii wrote, even the CPython documentation warns against not closing files explicitly, so I changed it here. (see https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files where it says
```
Warning Calling f.write() without using the with keyword or calling f.close() might result in the arguments of f.write() not being completely written to the disk, even if the program exits successfully.
```
)

timfel:
re https://github.com/pypa/cibuildwheel/pull/1538#discussion_r2035529220 @joerick 

Right now GraalPy only has ABI stability within minor version releases (the version you are reading is 24.2, patch releases are 24.2.1 and so on).

azure-pipelines.yml
timfel: Afaiks these and your other suggestions are in your commit, did I miss something?

henryiii: Yes, I rebased and removed this now that the fix in in main.

test/test_before_build.py
henryiii: I've converted all the `open()`s I could find without a context manager to use `.write_text`.

joerick:
Something we need to consider: how this impacts CI times. It's considerable-

|   | main | this PR |
|---|---|---|
| ubuntu-latest (3.13) | 37min | 65min |
| macOS-13 (3.13) | 32min | 43min |
| windows-latest (3.13) | 49min  | 77min |

Perhaps there's a way we can flag what's enabled for an integration test run - e.g. don't run the test suite on all the EnableGroups by default? The test duration is less important on `main`, but it does affect how fast we can move in PRs.

test/test_before_build.py
joerick: I always thought the OS would clean this up for you when the process exits. Although, maybe this does explain why some `shutil.rmtree` calls randomly fail on windows!

joerick:
Perhaps this is a good model to copy - pyodide-build has this code in their CI config to add flags to their tests based on PR labels https://github.com/pyodide/pyodide-build/commit/794ed7a12193020278e3034e11b7f709517b2e86

joerick:
I've drawn up that approach in #2357

cibuildwheel/resources/build-platforms.toml
timfel: Triggered by this discussion, we are changing the way handle ABI changes and will align with CPython in that we only break our ABI once a year, in the GraalPy release that happens around the same time of the CPython release in the latter half of the year. So our spring release will keep the ABI stable and then adopting new GraalPy ABI will happen in the same timeframe were cibuildwheels and other projects update to the new CPython version, too. Would that be ok?

timfel:
> I've drawn up that approach in #2357

This looks very reasonable to me :-) Will you let me know if/when I should rebase this PR on top of that one?

henryiii:
Will #2362 affect this?

timfel:
> * Will [ci: add Windows ARM #2362](https://github.com/pypa/cibuildwheel/pull/2362) affect this?

We don't release Windows arm binaries of GraalPy so far, so as long as it is not enabled for that platform it's fine.

henryiii:
Doesn't seem to be causing a problem after rebasing, so I think it's fine.

test/test_0_basic.py
```diff
@@ -41,7 +42,22 @@ def test(tmp_path, build_frontend_env, capfd):
...
+        bad_msg = "WARNING: Running pip as the 'root' user can result"
+        assert_msg = f"{bad_msg} is only tolerated from manylinux's ensure for graalpy"
+        for m in re.finditer(bad_msg, stream):
+            idx = stream.rfind("/opt/", start, m.end())
+            start = m.end()
+            m2 = re.match(
+                "/opt/.*/graalpy.*/bin/python -m ensurepip --default-pip", stream[idx : m.start()]
+            )
+            assert m2 is not None, assert_msg
+            assert idx + m2.end() + 1 == m.start(), assert_msg
```
mayeut, line 60: In light of the much longer CI times, I'm checking how not excluding graalpy from image warm-up impacts CI time on Linux, which would most likely make this workaround no-op:
```diff
diff --git a/test/conftest.py b/test/conftest.py
index 8c0c9504..f2b575df 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -56,9 +56,8 @@ def docker_warmup(request: pytest.FixtureRequest) -> None:
     images = [build_options.manylinux_images[arch] for arch in archs] + [
         build_options.musllinux_images[arch] for arch in archs
     ]
-    # exclude GraalPy as it's not a target for cibuildwheel
     command = (
-        "manylinux-interpreters ensure $(manylinux-interpreters list 2>/dev/null | grep -v graalpy) &&"
+        "manylinux-interpreters ensure-all &&"
         "cpython3.13 -m pip download -d /tmp setuptools wheel pytest"
     )
     for image in images:
```

mayeut:
> Something we need to consider: how this impacts CI times. It's considerable-

Updating warm-up on Linux images allows to reduce CI times on Linux.

|   | main | this PR | Linux warm-up |
|---|---|---|---|
| ubuntu-latest (3.13) | 39min | 66min | 50min |
| ubuntu-24.04-arm (3.13) | 24min | 42min | 34min |
| macOS-13 (3.13) | 29min | 48min | N/A |
| macOS-15 (3.13) | 21min | 29min | N/A |
| windows-latest (3.13) | 47min  | 65min | N/A |

joerick:
Let's apply @mayeut's patch here then, and if we can get the conflicts sorted I think this is ready. @timfel would you be able to do that? We'll merge #2357 afterwards.

timfel:
> Let's apply @mayeut's patch here then, and if we can get the conflicts sorted I think this is ready. @timfel would you be able to do that? We'll merge #2357 afterwards.

I'll look at the conflicts

test/test_0_basic.py
timfel: Applied in https://github.com/pypa/cibuildwheel/pull/1538/commits/a47e3542ba76beb32d8242dd043bf5e5961a5ed0 and I removed this workaround in the same commit.

henryiii:
Thanks!

PR #1542, opened 2023-07-09
Ignore docs changes on Azure Pipelines and CirrusCI

joerick:
Saves a bit of pointless CI time. Jenkins and Circle don't support it.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,4 +1,4 @@
-skip: changesIncludeOnly('docs/*')
+skip: changesIncludeOnly('docs/*') || changesIncludeOnly('.pre-commit-config.yaml')
 
 run_tests: &RUN_TESTS
   install_cibuildwheel_script:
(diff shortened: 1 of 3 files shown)


Czaki:
Maybe this https://github.com/bjd2385/dynamic-continuation-orb

.cirrus.yml
```diff
@@ -1,3 +1,5 @@
+skip: changesIncludeOnly('docs/*')
```
henryiii, line 1: What about `.pre-commit-config.yaml`?

henryiii (approved):

.cirrus.yml
joerick: good idea!

joerick:
> Maybe this https://github.com/bjd2385/dynamic-continuation-orb

I didn't look at this, but there was also a first-party 'orb' that did something similar. But it seemed way more complicated than what I was looking for, I think it's more for multi-project monorepo setups.

If you can figure it out, please do!

PR #1564, opened 2023-08-07
fix: --only shouldn't require prerelease pythons setting

### LINKED ISSUE #1560: --only shouldn't require CIBW_PRERELEASE_PYTHONS
### Description

Using `--only` to target a pre-release Python shouldn't require `CIBW_PRERELEASE_PYTHONS` be set.

### Build log

_No response_

### CI config

_No response_

henryiii:
Fix #1560.

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -450,6 +450,7 @@ class Options:
             build_config = args.only
             skip_config = ""
             architectures = Architecture.all_archs(self.platform)
+            prerelease_pythons = True
 
         build_selector = BuildSelector(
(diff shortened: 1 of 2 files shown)


joerick (approved):

mayeut:
should a test be added to `unit_test/main_tests/main_platform_test.py` ?

joerick:
Yeah, that's probably a good idea. It would need a mock of `BuildSelector.PRERELEASE_SKIP`. Given that I'll probably release 2.15 today without any prerelease_skip versions there's no reason to rush this.

PR #1565, opened 2023-08-07
feat: build using 3.12 RC by default

henryiii:
Requires #1553.


joerick:
Thanks for the quick work on this @henryiii, and @mayeut on the manylinux upstreams :)

EwoutH:
If everything goes well with the 3.13 rc1 release, we can do this again tomorrow!

PR #1569, opened 2023-08-10
chore: use 2x faster pre-commit mirror

henryiii:


Czaki:
2x faster? Installation or also execution?

henryiii:
Execution. It’s mypyc compiled, since it pulls the PyPI version. The other one build from source, which does not compile with mypyc.

For reference:

* https://black.readthedocs.io/en/latest/integrations/source_version_control.html
* https://github.com/psf/black/issues/3405
* https://github.com/psf/black/pull/3828

Czaki (approved):

mayeut (approved):

PR #1570, opened 2023-08-11
Update Tornado's entry in working-examples.md

bdarnell:
We've moved to github actions and added windows support. Also add a note about our use of the stable ABI which may be the most interesting part of our config.


Czaki (dismissed):

joerick:
thanks! But, this file is generated. 

> this section is generated by bin/projects.py. Don't edit it directly, instead, edit docs/data/projects.yml

Would you mind updating the source file?

bdarnell:
Sorry, I was led astray by RTD's "edit on github" button. Closing this one in favor of #1571

PR #1571, opened 2023-08-12
Update Tornado's entry in projects.yml

bdarnell:
We've moved to github actions and added windows support. Also add a note about our use of the stable ABI which may be the most interesting part of our config.

Replaces #1570


joerick (approved):
Brilliant, thank you!

henryiii (approved):

Czaki (approved):

PR #1582, opened 2023-08-21
Doc: Update info for CIBW_TEST_EXTRAS

mwtoews:
Update setuptools link to new URL, and add `pyproject.toml` as a source file (as already done in the example).


henryiii:
Thanks!

PR #1588, opened 2023-08-26
Add the ability to pass extra flags to a build frontend through CIBW_BUILD_FRONTEND

### LINKED ISSUE #1580: [FR] `CIBW_BUILD_ISOLATION` or similar
It would be nice if we could have a `CIBW_BUILD_ISOLATION=false`, equivalent to `python -m build --no-isolation` and `pip wheel --no-deps --no-build-isolation`.

The context for this FR is testing: some projects (like build-backends or build-backend plugins, e.g. `setuptools-rust`) have integration tests that use `cibuildwheel`; and during those tests we want to install the working versions of one of the build requirements from the source directory instead of fetching from PyPI.

The only alternative that we have nowadays is `PIP_NO_BUILD_ISOLATION` (no `build` equivalent as far as I know). However, I found that `PIP_NO_BUILD_ISOLATION` is tricky to use (in PyO3/setuptools-rust#352, it took me a long time to figure out how to make it work and solve the errors). I can highlight the following reasons:

1. It is not supported by `build`
2. You have to set it inside `CIBW_ENVIRONMENT` (forgetting to do that is a silly mistake if you understand how `cibuildwheel` works, but nevertheless, very easy to forget and let it happen/difficult for beginners to figure it out)
3. `PIP_NO_BUILD_ISOLATION` is counter-intuitive: see pypa/pip#5735
4. When `PIP_NO_BUILD_ISOLATION=false` is set, every call to `pip install` (e.g. the ones in `CIBW_BEFORE_BUILD` to install the build dependencies) will also be subject to the "no isolation" rule.
   This make a supposedly simple operation (i.e., installing build dependencies using a `pip install` command) error prone. For example, the following `CIBW_BEFORE_BUILD` may result in errors (depending on the versions of Python/`setuptools`/`wheel`/`pip` already present on the build container/VM): 

   ```yaml
   CIBW_BEFORE_BUILD: pip install -U pip>=23.2.1 setuptools>=68.0.0 wheel>=0.41.1 -e .
   ```
   (Instead to make it work, after a few ours of debugging, you might have to split the command into a succession of independent `pip install`s).

   ~It would be better if `CIBW_BEFORE_BUILD` is not affected by `PIP_NO_BUILD_ISOLATION` (it gives less margins for surprise/errors).~
   `CIBW_BUILD_ISOLATION=false` would be better because it would not affect `CIBW_BEFORE_BUILD` (less margin for surprise/errors).

An alternative to `CIBW_BUILD_ISOLATION=true`, would be allowing customising the build command `CIBW_BUILD_CMD`, as suggested in #442 (however that comes with a different set of considerations/issues as discussed in the other issue).

joerick:
Fixes #1580.

Adds the ability to pass extra arguments to the build frontend using an option like `CIBW_BUILD_FRONTEND=pip; args: --no-build-isolation`. The mini-syntax is the same as CIBW_CONTAINER_ENGINE. 

I liked this approach because it adds extra flexibility to the existing CIBW_BUILD_FRONTEND option in a backward-compatible way. Also, options are always going to be different between pip and build, so it makes sense to nest those within this option - if you change the build frontend, you have to change the args too.

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -505,7 +505,11 @@ class Options:
             test_extras = self.reader.get("test-extras", sep=",")
             build_verbosity_str = self.reader.get("build-verbosity")
 
-            build_frontend_str = self.reader.get("build-frontend", env_plat=False)
+            build_frontend_str = self.reader.get(
+                "build-frontend",
+                env_plat=False,
+                table={"item": "{k}:{v}", "sep": "; ", "quote": shlex.quote},
+            )
             build_frontend: BuildFrontendConfig | None
             if not build_frontend_str or build_frontend_str == "default":
                 build_frontend = None
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -27,7 +27,7 @@ from .typing import PLATFORMS, PlatformName
 from .util import (
     MANYLINUX_ARCHS,
     MUSLLINUX_ARCHS,
-    BuildFrontend,
+    BuildFrontendConfig,
     BuildSelector,
     DependencyConstraints,
     TestSelector,
@@ -92,7 +92,7 @@ class BuildOptions:
     test_requires: list[str]
     test_extras: str
     build_verbosity: int
-    build_frontend: BuildFrontend | Literal["default"]
+    build_frontend: BuildFrontendConfig | None
     config_settings: str
 
     @property
@@ -488,7 +488,6 @@ class Options:
         with self.reader.identifier(identifier):
             before_all = self.reader.get("before-all", sep=" && ")
 
-            build_frontend_str = self.reader.get("build-frontend", env_plat=False)
             environment_config = self.reader.get(
                 "environment", table={"item": '{k}="{v}"', "sep": " "}
             )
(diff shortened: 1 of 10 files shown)


cibuildwheel/options.py
```diff
@@ -506,17 +505,16 @@ def build_options(self, identifier: str | None) -> BuildOptions:
...
             build_verbosity_str = self.reader.get("build-verbosity")
 
-            build_frontend: BuildFrontend | Literal["default"]
-            if build_frontend_str == "build":
-                build_frontend = "build"
-            elif build_frontend_str == "pip":
-                build_frontend = "pip"
-            elif build_frontend_str == "default":
-                build_frontend = "default"
+            build_frontend_str = self.reader.get("build-frontend", env_plat=False)
```
dlech, line 508: Looks like this needs a `table` kwarg. When I tried this with a `pyproject.toml`, I got the following crash:

```console
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\extra\work\pywinrt\.venv\Scripts\cibuildwheel.exe\__main__.py", line 7, in <module>  
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\__main__.py", line 144, in main
    build_in_directory(args)
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\__main__.py", line 301, in build_in_directory
    print_preamble(platform=platform, options=options, identifiers=identifiers)
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\__main__.py", line 353, in print_preamble
    options_summary = textwrap.indent(options.summary(identifiers), "  ")
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\options.py", line 648, in summary
    build_options = self.build_options(identifier=None)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\options.py", line 508, in build_options
    build_frontend_str = self.reader.get("build-frontend", env_plat=False)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\extra\work\pywinrt\.venv\Lib\site-packages\cibuildwheel\options.py", line 357, in get
    raise ConfigOptionError(msg)
cibuildwheel.options.ConfigOptionError: "'build-frontend' does not accept a table"
```

dlech:
```
build-frontend = "build; args: --skip-dependency-check --no-isolation"
```
does work though, so if that was the intention, then the docs just need to be corrected.

joerick: Thanks for testing this out! That reveals a hole in my testing coverage too. I'll add a test and a fix.

dlech: Thanks for the fix. I changed back to the toml inline table syntax and it is working now.

joerick:
Would love a review on this @henryiii if you have time?

henryiii (approved):
I think this is fine. No build isolation does seem common/useful enough to make its own flag, but having this flexibility is useful (for example, you can select whether to validate the non-isolated environment).

PR #1599, opened 2023-09-06
fix: do not use `linux32` when unnecessary

mayeut:
The `linux32` command is always used for linux i686.
This makes cross building from macOS arm64 impossible at the moment (might depend on docker configuration).

This PR proposes to check the architecture seen inside the container before using (or not) the `linux32` command.

### DIFF
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -110,7 +110,7 @@ class OCIContainer:
         if detect_ci_provider() == CIProvider.travis_ci and platform.machine() == "ppc64le":
             network_args = ["--network=host"]
 
-        simulate_32_bit = self.enforce_32_bit
+        simulate_32_bit = False
         if self.enforce_32_bit:
             # If the architecture running the image is already the right one
             # or the image entrypoint takes care of enforcing this, then we don't need to
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -111,13 +111,14 @@ class OCIContainer:
             network_args = ["--network=host"]
 
         simulate_32_bit = self.enforce_32_bit
-        container_machine = call(
-            self.engine.name, "run", "--rm", self.image, "uname", "-m", capture_stdout=True
-        ).strip()
-        if container_machine not in {"x86_64", "aarch64"}:
-            # either the architecture running the image is already the right one
-            # or the image entrypoint took care of this
-            simulate_32_bit = False
+        if self.enforce_32_bit:
+            # If the architecture running the image is already the right one
+            # or the image entrypoint takes care of enforcing this, then we don't need to
+            # simulate this
+            container_machine = call(
+                self.engine.name, "run", "--rm", self.image, "uname", "-m", capture_stdout=True
+            ).strip()
+            simulate_32_bit = container_machine != "i686"
 
         shell_args = ["linux32", "/bin/bash"] if simulate_32_bit else ["/bin/bash"]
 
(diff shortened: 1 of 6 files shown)


cibuildwheel/oci_container.py
```diff
@@ -108,7 +108,16 @@ def __enter__(self) -> OCIContainer:
...
             network_args = ["--network=host"]
 
-        shell_args = ["linux32", "/bin/bash"] if self.simulate_32_bit else ["/bin/bash"]
+        simulate_32_bit = self.simulate_32_bit
+        container_machine = call(
+            self.engine.name, "run", "--rm", self.image, "uname", "-m", capture_stdout=True
+        ).strip()
+        if container_machine not in {"x86_64", "aarch64"}:
+            # either the architecture running the image is already the right one
+            # or the image entrypoint took care of this
```
joerick, line 117: I think that the ENTRYPOINT isn't used when you specify a program via `run` or `create`?

cibuildwheel/oci_container.py
```diff
@@ -108,7 +108,16 @@ def __enter__(self) -> OCIContainer:
         if detect_ci_provider() == CIProvider.travis_ci and platform.machine() == "ppc64le":
             network_args = ["--network=host"]
 
-        shell_args = ["linux32", "/bin/bash"] if self.simulate_32_bit else ["/bin/bash"]
+        simulate_32_bit = self.simulate_32_bit
```
joerick, line 111: As written here, the parameter might be better named `ensure_32_bit`, as now it only does so when it's required.

joerick (commented):
Interesting. Is this a fix for the '-mnative' error on macOS?

mayeut:
> Is this a fix for the '-mnative' error on macOS?

No, it's a fix for a deadlock following a call to linux32:
```
WARNING: The requested image's platform (linux/386) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
3fca67068b8d0aa05202598ccb3d9e80a27327038b85ede10fb6ab3f73b7543e
    + /bin/true
linux32: Failed to set personality to linux32: Invalid argument
```

> I think that the ENTRYPOINT isn't used when you specify a program via run or create?

It's definitely used on run, the program specified via run is passed to the entrypoint. I'll check it's passed via create also and add a test for this.

Turning this into draft until I get time to address all comments.

joerick:
> It's definitely used on run, the program specified via run is passed to the entrypoint. I'll check it's passed via create also and add a test for this.

Ah, perhaps I am confused about the difference between CMD and ENTRYPOINT.

mayeut:
I checked https://docs.docker.com/engine/reference/commandline/create/ and since the ENTRYPOINT can be overridden (same as with run), it's used so the confusion might indeed come from ENTRYPOINT vs CMD (which is indeed overriden when specifying a custom command).

mayeut:
seems cirrus-ci windows x86 is having troubles, found https://github.com/cirruslabs/cirrus-ci-docs/issues/1213

cibuildwheel/oci_container.py
```diff
@@ -108,7 +108,17 @@ def __enter__(self) -> OCIContainer:
         if detect_ci_provider() == CIProvider.travis_ci and platform.machine() == "ppc64le":
             network_args = ["--network=host"]
 
-        shell_args = ["linux32", "/bin/bash"] if self.simulate_32_bit else ["/bin/bash"]
+        simulate_32_bit = self.enforce_32_bit
```
joerick, line 111: I think this is a little clearer
```suggestion
        simulate_32_bit = False
```

joerick (approved):
looks good to me!

cibuildwheel/oci_container.py
mayeut: this is indeed a little clearer

PR #1613, opened 2023-09-13
fix(setup.py): look inside if name == main block

### LINKED ISSUE #1611: setup.py setup() not detected in __name__ == '__main__' block
### Description

My setup.py setup() includes:

    python_requires=">=3.8"

However cibuildwheel still tries and fails to compile under Python 3.6.

I understand there is [CIBW_BUILD / CIBW_SKIP](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip) but that is then duplicating the python requires information.

I can add a \[project\] section to pyproject.toml but that leads to a lot of problems because it ends up fighting with setup() parameters and they **really** don't like it.

I believe cibuildwheel should establish the Python version support automatically whether it comes from setuptools or pyproject.toml, and not try to build on unsupported versions.  My [pyproject.toml](https://github.com/rogerbinns/apsw/blob/master/pyproject.toml) is:

````
[build-system]
requires = ["setuptools"]
````

### Build log

https://github.com/rogerbinns/apsw/actions/runs/6175182758/job/16761477543

### CI config

https://github.com/rogerbinns/apsw/actions/runs/6175182758/workflow

henryiii:
Fixes #1611. We could slide down a slippery slope here, but this one is common and simple (vs. supporting `**kwargs` or something like that). And the code will look a lot better once we drop Python 3.9 from our runners in a few years.

### DIFF
--- a/cibuildwheel/projectfiles.py
+++ b/cibuildwheel/projectfiles.py
@@ -8,6 +8,43 @@ from pathlib import Path
 from ._compat import tomllib
 
 
+def get_parent(node: ast.AST | None, depth: int = 1) -> ast.AST | None:
+    for _ in range(depth):
+        node = getattr(node, "parent", None)
+    return node
+
+
+def is_main(parent: ast.AST | None) -> bool:
+    if parent is None:
+        return False
+
+    # This would be much nicer with 3.10's pattern matching!
+    if not isinstance(parent, ast.If):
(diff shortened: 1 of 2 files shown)


joerick (approved):
ASTs are kinda fun, so, yeah, i'm happy with it :)

PR #1621, opened 2023-09-19
Improve parse_key_value_string to allow colons in values.

### LINKED ISSUE #1600: `CIBW_CONTAINER_ENGINE` splits on `:` when it shouldn't
### Description

`CIBW_CONTAINER_ENGINE="docker; create_args: -p=8080:8080"`

gets parsed into

```
  container_engine: 
    name: docker
    create_args: ['-p=8080', ':', '8080']
```

problem is here https://github.com/pypa/cibuildwheel/blob/e5d0cc0b0d37d20a053b85112e4e1f137079c85e/cibuildwheel/util.py#L710

potential fix from [here](https://stackoverflow.com/a/61737902/9045206):
```python
shlexer = shlex.shlex(key_value_string, posix=True, punctuation_chars=True)
shlexer.wordchars += ':'
```

with this change, `CIBW_CONTAINER_ENGINE="docker; create_args : -p=8080:8080"` (note the space after `create_args`) parses to `fields=[['docker'], ['create_args', ':', '-p=8080:8080']]`.

joerick:
Fix #1600

Admittedly this is quite a strange format, but we started along the shlex-style options with CIBW_ENVIRONMENT so we'll always have it in some respect. I still think it's quite natural to use, especially for specifying command line arguments.

Also adds unit tests to prevent regressions.

### DIFF
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -731,23 +731,26 @@ def parse_key_value_string(
 
     all_field_names = [*positional_arg_names, *kw_arg_names]
 
-    shlexer = shlex.shlex(key_value_string, posix=True, punctuation_chars=";:")
+    shlexer = shlex.shlex(key_value_string, posix=True, punctuation_chars=";")
     shlexer.commenters = ""
+    shlexer.whitespace_split = True
     parts = list(shlexer)
     # parts now looks like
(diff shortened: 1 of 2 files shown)


henryiii (approved):

henryiii:
Should we put this in too?

PR #1622, opened 2023-09-19
feat: add schema and validate-pyproject support

henryiii:
Adding a json schema. This supports validate-pyproject, allowing it to check a `tool.cibuildwheel` section. I wrote a Python 3.10+ `bin/generate_schema.py` to make the schema as I didn't see a good way to do the OS sections and overrides by hand. I mostly tried to replicate the current behavior, though we technically could be more strict with a schema than in reality, since it's optional to validate it.

Testing by validating the defaults toml, a few manual examples, and running all the docs examples though.

Once we have this, we could also contribute it to SchemaStore, which would make it work in vscode, etc.


bin/generate_schema.py
```diff
@@ -0,0 +1,233 @@
...
+
+overrides["items"]["properties"]["select"]["oneOf"] = string_array
+overrides["items"]["properties"] |= limited.copy()
+
+del limited["archs"]
+
+not_linux = limited.copy()
+
+del not_linux["environment-pass"]
+del not_linux["container-engine"]
```
joerick, line 216: container-engine is actually only acceptable as a global option

bin/generate_schema.py
```diff
@@ -0,0 +1,233 @@
...
+items:
+  type: object
+  required: ["select"]
+  additionalProperties: false
+  properties:
+    select: {}
+"""
+)
+
+limited = schema["properties"].copy()
```
joerick, line 204: Perhaps this variable would be more clear named as `non_global_options`?

bin/generate_schema.py
```diff
@@ -0,0 +1,233 @@
...
+del limited["build"]
+del limited["skip"]
+
```
joerick, line 207: Also, `test-skip` and `archs` are global. Archs cannot be overridden in an override.

noxfile.py
```diff
@@ -113,6 +113,14 @@ def update_proj(session: nox.Session) -> None:
     )
 
 
+@nox.session(reuse_venv=True)
+def generate_schema(session: nox.Session) -> None:
+    session.install("pyyaml", "ruamel.yaml")
```
joerick, line 118: I think that ruamel.yaml isn't used?
```suggestion
    session.install("pyyaml")
```

joerick (commented):
This is really cool, I didn't know about this. Getting auto-complete in vscode would be amazing.

henryiii:
Also added a mechanism to test all the docs examples. Several fixes of the schema came from that. :)

docs/options.md
```diff
@@ -917,9 +917,7 @@ Platform-specific environment variables are also available:<br/>
     CIBW_REPAIR_WHEEL_COMMAND: >
       python scripts/repair_wheel.py -w {dest_dir} {wheel} &&
       python scripts/check_repaired_wheel.py -w {dest_dir} {wheel}
-    ```
```
henryiii, line 920: FYI, I could handle this, I suppose, but it's easier to just leave these as one block, like all the other examples.

joerick (approved):
Looks good! Testing with the docs samples is a very cool idea.

PR #1629, opened 2023-09-28
refactor: use refs instead of copies

henryiii:
Refactoring this to be a bit simpler and reduce the number of lines in the schema by a lot (around 800). The only loss is the custom titles for the OS-specific variants, but I think that's fine.

### DIFF
--- a/bin/generate_schema.py
+++ b/bin/generate_schema.py
@@ -206,7 +206,10 @@ items:
 """
 )
 
-non_global_options = copy.deepcopy(schema["properties"])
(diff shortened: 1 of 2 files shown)


henryiii:
Only thing failing is the Cirrus CI jobs, which were out of free time when the PR was created.

henryiii:
@joerick, okay to merge?

joerick (approved):
I'm learning more than I'm critiquing here, but the use of refs looks like a good idea!

PR #1630, opened 2023-10-01
feat: use manylinux-interpreters tool if present

mayeut:
This will allow dropping some EOL (and maybe some none EOL) python interpreters in manylinux images which would allow to reduce its size (faster for a vast majority of user, a small slow-down for users of those interpreters).

Updates to keep tests times down are not here yet, this is just to get things started.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -117,10 +117,20 @@ def check_all_python_exist(
     *, platform_configs: Iterable[PythonConfiguration], container: OCIContainer
 ) -> None:
     exist = True
+    has_manylinux_interpreters = True
     messages = []
+
+    try:
+        # use capture_output to keep quiet
+        container.call(["manylinux-interpreters", "--help"], capture_output=True)
+    except subprocess.CalledProcessError:
+        has_manylinux_interpreters = False
+
     for config in platform_configs:
         python_path = config.path / "bin" / "python"
         try:
+            if has_manylinux_interpreters:
+                container.call(["manylinux-interpreters", "ensure", config.path.name])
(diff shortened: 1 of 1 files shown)


cibuildwheel/linux.py
```diff
@@ -117,10 +117,20 @@ def check_all_python_exist(
     *, platform_configs: Iterable[PythonConfiguration], container: OCIContainer
 ) -> None:
     exist = True
+    has_manylinux_interpreters = True
```
henryiii, line 120:
```suggestion
    has_manylinux_interpreters = False
```

cibuildwheel/linux.py
```diff
@@ -117,10 +117,20 @@ def check_all_python_exist(
...
     messages = []
+
+    try:
+        # use capture_output to keep quiet
+        container.call(["manylinux-interpreters", "--help"], capture_output=True)
+    except subprocess.CalledProcessError:
+        has_manylinux_interpreters = False
```
henryiii, line 127:
```suggestion
    with contextlib.suppress(subprocess.CalledProcessError):
        # use capture_output to keep quiet
        container.call(["manylinux-interpreters", "--help"], capture_output=True)
        has_manylinux_interpreters = True
```

(also add contextlib import above)

henryiii (commented):
Small suggestion - Defaulting to true then changing to false seems a little odd, maybe this is more normal? Also, `which manylinux-interpreters` could be used? It even seems like you could make this one step by just suppressing the ensure call?

henryiii (approved):
The comments are minor, I'm also happy with this as is for now. I assume it's good to get in so that manylinux can eventually require it.

PR #1638, opened 2023-10-06
fix(schema): small bug + prepare for SchemaStore

henryiii:
Fixing a small bug discovered when running this via SchemaStore's tests that fastjsonschema apparently doesn't care about (extra list). Also have a way to specify the full schema (with tool.cibuildwheel) so that cibuildwheel.toml/.cibuildwheel.toml can be checked. (Currently not sure how to have a subschema that doesn't have a file also, but this will at least move cibuildwheel forward, because we do).

https://github.com/SchemaStore/schemastore/pull/3287

### DIFF
--- a/bin/generate_schema.py
+++ b/bin/generate_schema.py
@@ -1,10 +1,15 @@
 #!/usr/bin/env python
 
+import argparse
 import copy
 import json
 from typing import Any
 
 import yaml
+
+parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


henryiii:
Accepted: https://json.schemastore.org/cibuildwheel.json

Not sure how long it takes to sync to editors.

joerick (approved):
Great!

PR #1643, opened 2023-10-14
chore: limit Cirrus CI usage

mayeut:
New usage limits are now enforced on Cirrus CI: https://cirrus-ci.org/faq/#are-there-any-limits
Limit runs to the minimum.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,4 +1,4 @@
-skip: changesIncludeOnly('docs/*', '.pre-commit-config.yaml')
+only_if: changesInclude('.cirrus.yml') || ($BRANCH == "main" && !changesIncludeOnly('docs/*', '.pre-commit-config.yaml'))
 
 run_tests: &RUN_TESTS
   install_cibuildwheel_script:


Czaki (approved):

joerick (approved):
Thanks. It does kinda suck that this takes out our only arm64 mac CI run, but I can't see a way around it right now. GHA and Circle arm64 runners are paid, I think.

Czaki:
@joerick I see that circle.ci has option to extend support of opensource. did you try to contact here: https://circleci.com/talk-to-us/?source-button=Request%20early%20access%20-%20Open%20Source%20usage%20plan 

Did we know which platform may be cheapest (and if it will be ok for you to apply for some grant for cover such option?)?

joerick:
@czaki, good point. Actually, that reminds me that we already have some kind of open-source exemption with CircleCI. So it would be worth trying their existing m1 runners, we might already have access.

henryiii:
Could be worth asking if we need something for any CI provider; we are providing them more potential customers by supporting them, after all.

PR #1650, opened 2023-10-26
chore: test and use 3.12 as host

henryiii:
Missed updating this earlier.


joerick (approved):
It would be good to update CI.md too, just so we keep a record of our testing matrix

PR #1665, opened 2023-11-18
Use quotes when silencing warning

hugovk:
We found (re: https://github.com/python-pillow/Pillow/pull/7552#pullrequestreview-1732529361) we needed to add quotes to `CIBW_TEST_SKIP: "*-macosx_arm64"` to silence the warnings.

And this matches the recommendation in docs: https://cibuildwheel.readthedocs.io/en/stable/options/#test-skip


joerick (approved):
v cool that Pillow is moving towards cibuildwheel :)

PR #1672, opened 2023-11-21
feat: Adds disable_host_mount suboption to container-engine

### LINKED ISSUE #1669: Allow overriding of  docker volume argument
### Description

The current ci_buildwheel runs a docker container with `/` mounted  to '/host'. volume This can perhaps be made dynamic so that it can be overridden  when passed as argument. 

### Build log

_No response_

### CI config

_No response_

joerick:
Fix #1669.

### DIFF
--- a/bin/generate_schema.py
+++ b/bin/generate_schema.py
@@ -61,9 +61,9 @@ properties:
     oneOf:
       - enum: [docker, podman]
       - type: string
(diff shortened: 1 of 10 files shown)


joerick:
Huh. It looks like Gitlab has the same behaviour as Circle regarding the /host mount. I'll add an exception for that too.

PR #1675, opened 2023-11-24
fix: Don't override PIP_CONSTRAINT if set by the user

joerick:
Following on from #1667, change the logic so that we retain user-set PIP_CONSTRAINT options, by concatting them with our own constraints


docs/options.md
```diff
@@ -673,10 +673,13 @@ Platform-specific environment variables are also available:<br/>
 !!! note
     cibuildwheel always defines the environment variable `CIBUILDWHEEL=1`. This can be useful for [building wheels with optional extensions](faq.md#building-packages-with-optional-c-extensions).
 
+!!! note
+    To do its work, cibuildwheel internally sets the options `VIRTUALENV_PIP`, `DIST_EXTRA_CONFIG`, `SETUPTOOLS_EXT_SUFFIX`, `PIP_DISABLE_PIP_VERSION_CHECK`, `PIP_ROOT_USER_ACTION`. Your assignments to these options might be overridden.
```
webknjaz, line 677: @joerick looks like `PIP_CONSTRAINT` is missing from this list.

cibuildwheel/macos.py
```diff
@@ -409,7 +409,9 @@ def build(options: Options, tmp_path: Path) -> None:
                                 config.version
                             )
                         )
-                        build_env["PIP_CONSTRAINT"] = constraint_path.as_uri()
+                        build_env["PIP_CONSTRAINT"] = (
+                            constraint_path.as_uri() + " " + build_env.get("PIP_CONSTRAINT", "")
+                        )
```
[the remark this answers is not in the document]
joerick: Hmm, I'm not sure. In this case wouldn't it be better to tell the user to set `CIBW_DEPENDENCY_VERSIONS=latest` instead? Or set it to your own lockfile? It feels a bit like a foot-gun if setting PIP_CONSTRAINT removes the build determinism feature from cibuildwheel.

docs/options.md
joerick: My intention was that it's not here because we don't override it, we extend it.

webknjaz: Then, maybe it's worth noting that it ends up being not exactly what the end-user sets?

cibuildwheel/macos.py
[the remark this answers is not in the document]
joerick: Well, that option isn't exactly what you're after I think. That option is used by cibuildwheel to pin versions of build tools on Mac and Windows. It isn't used on Linux builds, because there, the versions of tools is instead governed by which manylinux image is used.

I guess this is why I'm keen for [a more first-party supported way](https://github.com/pypa/build/issues/292) to do this - as we could specifically target the versions of build-backends, not the surrounding tools like pip or virtualenv.

I'm now thinking... maybe the better answer is for cibuildwheel to stop including setuptools in our constraint file. We don't pin other build backends, and the platform inconsistency above means that the setuptools pin is used inside mac & windows isolated builds, but not on linux (where I suppose the latest is fetched by `build` when it sets up the build venv).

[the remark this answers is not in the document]
joerick: > maybe the better answer is for cibuildwheel to stop including setuptools in our constraint file

On reflection, this does seem like a good idea. The only issue with this idea is that projects without pyproject.toml are gonna lose our pin. But projects **with** pyproject.toml aren't getting their build-backend pinned anymore anyway, so it's not a huge issue.

I think we'd also want to remove the auto-install of setuptools and wheel. That feels a bit legacy at this point, given that setuptools is no longer the only game in town, and that most(?) projects are on pyproject.toml.

I'm currently putting together a PR for this, let's see how it looks...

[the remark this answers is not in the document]
joerick: Just pushed the above in PR #1725. But, at the same time, if you're doing your own pinning via PIP_CONSTRAINT you can just disable cibuildwheel's pinning using `CIBW_DEPENDENCY_VERSIONS=latest`.

webknjaz:
@joerick it just occurred to me — what if `cibuildwheel` would take the user-provided `$PIP_CONSTRAINT`, then create the following file:
```
-c user-provided-constraints.txt
-c cibuildwheel-provided-constraints.txt
```
perhaps, removing entries from `cibuildwheel-provided-constraints.txt` that are defined in the user-provided-constraints.

As an alternative, `cibuildwheel` could write (of symlink?) its selected constraints file for the current session to a predictable location on each invocation, letting the end-users include it into their constraints via `-c`.

Would that be usable or would it make things more problematic?

joerick:
> what if `cibuildwheel` would take the user-provided `$PIP_CONSTRAINT`, then create the following file:

That would have the same behaviour as this PR. In this PR, the users' constraints are concat'd with our constraints.

> perhaps, removing entries from `cibuildwheel-provided-constraints.txt` that are defined in the user-provided-constraints.

Since #1725, I hope there isn't a big problem with conflicts. I'd rather see how this plays out before adding more complexity to how we handle this particular problem. 

It's worth noting that the 'removing duplicate entries' idea is good but isn't foolproof, because each of the packages also have their own requirements specifiers that could conflict with what's set in the constraints. I'd rather just tell the minority of users that have conflicts to use DEPENDENCY_VERSIONS=latest and they can pin `pip`, `build`, `delocate`, `auditwheel` themselves, if it's important to them.

joerick:
@henryiii or @mayeut, a review here would be great. I'd like to get some of these PRs merged if we can, they've been hanging around for a while.

cibuildwheel/windows.py
```diff
@@ -455,7 +455,9 @@ def build(options: Options, tmp_path: Path) -> None:
                             tmp_file.write_bytes(constraints_path.read_bytes())
                             constraints_path = tmp_file
 
-                        build_env["PIP_CONSTRAINT"] = str(constraints_path)
+                        build_env["PIP_CONSTRAINT"] = (
+                            str(constraints_path) + " " + build_env.get("PIP_CONSTRAINT", "")
+                        )
```
henryiii, line 460:
```suggestion
                        constraint = build_env.get("PIP_CONSTRAINT", "")
                        build_env["PIP_CONSTRAINT"] = f"{constraints_path} {constraint}"
```

cibuildwheel/macos.py
henryiii, line 414:
```suggestion
                        constraint = build_env.get("PIP_CONSTRAINT", "")
                        build_env["PIP_CONSTRAINT"] = f"{constraint_path.as_uri()} {constraint}"
```

henryiii: Actually, I don't like that the above adds a space to the end if PIP_CONSTRAINT is not set. We could add `.strip()`, or just build it correctly:

```suggestion
                        constraints = [constraint_path.as_uri(), build_env.get("PIP_CONSTRAINT", None)]
                        build_env["PIP_CONSTRAINT"] = " ".join(filter(constraints, None))
```

cibuildwheel/windows.py
henryiii:
```suggestion
                        constraints = [str(constraints_path), build_env.get("PIP_CONSTRAINT", None)]
                        build_env["PIP_CONSTRAINT"] = " ".join(filter(constraints, 
```

webknjaz:
> It's worth noting that the 'removing duplicate entries' idea is good but isn't foolproof, because each of the packages also have their own requirements specifiers that could conflict with what's set in the constraints. I'd rather just tell the minority of users that have conflicts to use DEPENDENCY_VERSIONS=latest and they can pin `pip`, `build`, `delocate`, `auditwheel` themselves, if it's important to them.

True, I also realized this later on, after posting that comment ;)

cibuildwheel/windows.py
joerick: Yeah, this is a bit unclear. I'm feeling something like this:
```python
                        our_constraints = str(constraints_path)
                        user_constraints = build_env.get("PIP_CONSTRAINT")
                        build_env["PIP_CONSTRAINT"] = " ".join(
                            c for c in [our_constraints, user_constraints] if c
                        )
```

henryiii: Sure, that's basically what I wrote below. `filter(item, None)` filters None's out, by the way, or you can the c for c if c pattern.

joerick: I think it's `filter(None, a_list)`, but yeah, never been used map or filter in python though, list comprehensions are more readable and powerful imo!

joerick:
ready for another review henryiii

henryiii (approved):

PR #1679, opened 2023-11-27
Stop suggesting adding `wheel` to `pyproject.toml`

webknjaz:
This was a historic mistake in the `setuptools`' docs that many projects copied. But `wheel` is already a dependency injected by `setuptools` when needed. So let's not suggest people to depend on it when producing sdists.


joerick (approved):

henryiii (approved):

joerick:
thanks @webknjaz !

PR #1686, opened 2023-11-30
doc: add clarification regarding linux docker environment

doronz88:


joerick:
thanks. though it looks like your text editor has broken lots of formatting, could you revert those changes?

joerick (approved):

PR #1687, opened 2023-12-05
Edit path when building sdist package

### LINKED ISSUE #1683: [FR] Expose `{package}` placeholder to the build stage
### Description

I mentioned before that I set the `PIP_CONSTRAINT` env var when building wheels, to improve the reproducibility (https://github.com/pypa/cibuildwheel/issues/1666).

I got that integrated into yarl, and it works well when I use my https://github.com/re-actors/checkout-python-sdist action to check out the project from sdist instead of Git.

Later on, I learned that cibuildwheel can consume `*.tar.gz` files directly, so I figured why not try that out. There's no huge practical gain in my case, since the cibuildwheel action combined with my action already achieves this. And this doesn't reduce the number of steps in the job, just replaces one thing with another. But I wanted to see it in action.

Long story short, it didn't actually work. The first obstacle was figuring out how to pass an sdist into the action. I read the source and found out that there's an input called `package-dir` for passing sdists :exploding_head: (https://github.com/pypa/cibuildwheel/issues/1682). I replaced my action with a simple download, but I only had a wildcard for the tarball name — and the action quotes the input internally so it wouldn't be auto-expanded. So I had to add some supporting code to look up the actual sdist filename (which is fine — I wanted to do that at some point anyway).

I thought, that would be the end of it, but it crashed on the build step, with all the above setup! Apparently, since I was no longer checking out the project to the CWD, the relative path in the `PIP_CONSTRAINT` variable was pointing to a non-existent location :man_shrugging:
Oh, well, I thought I'd find something in the docs. And I did find mentions of some placeholders. I tried out `{project}` (confusing where it's supposed to point to) and `{package}` but the internal `pip install` was still reporting a “file not found”, with those placeholders remaining non-rendered, as is.

Later, I found the notes at the very bottom of the options page, mentioning that not all settings interpolate values. And realized that maybe, it's just not implemented.

So here I am, filing this feature request to make it work. While doing so, I realized that while implementing this (with the placeholder pointing to a temporary directory where the sdist is unpacked) will likely fix it for me (unless, new issues arise at later stages, like having to switch the tests dir path to `{package}`, I suppose).
But then, is it really worth it? Is it the best UX? After all, the thing I had was already doing what I needed, following KISS / DRY and typical *NIX composability considerations. Maybe, cibuildwheel (the action, not the PyPI dist!) should really delegate this to `checkout-python-sdist` instead of complicating the setup. Or, maybe, it should just call the action internally, bypassing the corresponding inputs there. WDYT?

The PR is here, if you're curious: https://github.com/aio-libs/yarl/pull/967. Though, I'll probably keep using my action that is a bit more generic, and I use it in other jobs (like tests) as well.

### Build log

_No response_

### CI config

_No response_

OlenaYefymenko:
This patch changes the working directory from the temp to the project when building sdist package. This resolves issues with relative paths in configuration files.

Resolves #1683

### DIFF
--- a/test/test_from_sdist.py
+++ b/test/test_from_sdist.py
@@ -65,8 +65,8 @@ def test_simple(tmp_path):
         """
         import os
 
-        assert os.path.exists("setup.py")
-        assert os.path.exists("{package}/setup.py")
+        assert os.path.exists('setup.py')
+        assert os.path.exists('{package}/setup.py')
         """,
     )
     setup_py_assertion_cmd = f'python3 -c "{setup_py_assertion_snippet !s}"'
--- a/test/test_from_sdist.py
+++ b/test/test_from_sdist.py
@@ -61,10 +61,23 @@ def test_simple(tmp_path):
     sdist_dir.mkdir()
     sdist_path = make_sdist(basic_project, sdist_dir)
 
+    setup_py_assertion_snippet = textwrap.dedent(
+        """
+        import os
+
+        assert os.path.exists("setup.py")
+        assert os.path.exists("{package}/setup.py")
+        """,
+    )
+    setup_py_assertion_cmd = f"python3 -c '{setup_py_assertion_snippet !s}'"
+
     # build the wheels from sdist
     actual_wheels = cibuildwheel_from_sdist_run(
         sdist_path,
-        add_env={"CIBW_BUILD": "cp39-*"},
+        add_env={
+            "CIBW_BEFORE_BUILD": setup_py_assertion_cmd,
+            "CIBW_BUILD": "cp39-*",
+        },
     )
 
     # check that the expected wheels are produced
--- a/test/test_from_sdist.py
+++ b/test/test_from_sdist.py
@@ -69,7 +69,7 @@ def test_simple(tmp_path):
         assert os.path.exists("{package}/setup.py")
         """,
     )
-    setup_py_assertion_cmd = f"python3 -c '{setup_py_assertion_snippet !s}'"
+    setup_py_assertion_cmd = f'python3 -c "{setup_py_assertion_snippet !s}"'
 
     # build the wheels from sdist
     actual_wheels = cibuildwheel_from_sdist_run(
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -160,7 +160,7 @@ def main() -> None:
         # This is now the new package dir
         args.package_dir = project_dir.resolve()
 
-        with chdir(temp_dir):
+        with chdir(project_dir):
             build_in_directory(args)
     finally:
         # avoid https://github.com/python/cpython/issues/86962 by performing


joerick:
CIBW_BEFORE_BUILD runs from the CWD, so we could check we are in the project dir by doing something like `CIBW_BEFORE_BUILD=python3 -c 'import os; assert os.path.exists("setup.py")'`.

webknjaz:
_UPD:_ adding tests may take a while since Olena is a beginner and needs to learn a thing or two about the ecosystem. I was helping her out to figure out how to run tests locally and learned that following the CONTRIBUTING instructions doesn't work — the way the tests are being executed requires something like `GITHUB_ACTIONS=true nox -s tests` to pretend the env is a CI. Otherwise, the underlying invocation of `cibuildwheel` in tests errors out with a return code of 2 and an error message saying that it couldn't detect a CI.

@joerick have you considered synchronizing the CIs with the way the invocations are supposed to run locally to avoid such discrepancies? As in run nox everywhere the same (ish) way? One interesting approach I saw in some other projects was having a dedicated "dev env" CI job that essentially runs the tools in a way the contributors would do it locally.

joerick:
Apologies, that doc does probably need a note about this. I wouldn't recommend setting the GITHUB_ACTIONS like you mention, rather what I do is set `CIBW_PLATFORM` e.g. `CIBW_PLATFORM=macos` or `CIBW_PLATFORM=linux` (both run locally on my Mac, because the linux one runs in docker).

I don't use the nox integration personally, because the test suite takes a while to run, so I generally want to be more specific. I will generally do `pytest unit_test` or if I want to run an integration test, I might do `CIBW_PLATFORM=linux pytest test -k <test_name>`.

Just one caveat to the above - for local runs with `CIBW_PLATFORM=macos`, your system might be missing some versions of Python - you can install them from Python.org, the error message should include the link to download.

joerick:
> @joerick have you considered synchronizing the CIs with the way the invocations are supposed to run locally to avoid such discrepancies? As in run nox everywhere the same (ish) way? One interesting approach I saw in some other projects was having a dedicated "dev env" CI job that essentially runs the tools in a way the contributors would do it locally.

Interesting idea... though really the problem here is that the docs are incomplete, I don't think we can test for that.

webknjaz:
> I don't use the nox integration personally, because the test suite takes a while to run, so I generally want to be more specific. I will generally do `pytest unit_test` or if I want to run an integration test, I might do `CIBW_PLATFORM=linux pytest test -k <test_name>`.

@joerick I checked the `noxfile.py` and it does allow you to pass arbitrary args to pytest. So you can do `nox -s tests -- unit_test` or `nox -s tests -- test -k <test_name>`. This is a commonly known approach that is usually implemented with `tox` too. Could you try it out? This should eliminate the need for you to run `pytest` directly.

webknjaz:
> > @joerick have you considered synchronizing the CIs with the way the invocations are supposed to run locally to avoid such discrepancies? As in run nox everywhere the same (ish) way? One interesting approach I saw in some other projects was having a dedicated "dev env" CI job that essentially runs the tools in a way the contributors would do it locally.
> 
> Interesting idea... though really the problem here is that the docs are incomplete, I don't think we can test for that.

The idea with workflows tools like `tox` / `nox` is that you absolutely can make the invocations nearly the same locally and in CI without having to compromise the ability to invoke the underlying tools with customizations. This is why the posargs supplied after `--` are usually bypassed to those integrated invocations in the respective configs.

Note that both tools have ways to list the configured sessions. They are `tox -av` or `nox --list` — and it's a convenient centralized place to document what different sessions implements. For tox, there's a dedicated setting for the env description, and in the case of nox, these are taken from docstrings.

If the contributing doc says to check out the sessions in the workflow tool, with a few more concrete examples, and the CI runs the same, that's usually the best way of making sure there's no difference between what different people run.

A more sophisticated solution could be auto-generating a portion of the contributing doc from the nox file.

henryiii:
FYI, you can also use `CIBW_PLATFORM=linux nox -Rs tests -- test -k <test_name>` which will avoid reinstalling anything and be just as fast as running it manually (at least the second time it runs).

joerick:
I've just pushed a PR #1698 with an improvement to this page. My opinion is that the direct pytest invocation is better, as I'm nearly always using some pytest-specific option, so it's extra level of indirection and mental overhead to also think about the `nox` layer, and a little confusing to thing about passing options to `nox` that behind the scenes get passed through to `pytest`. Also, that's the method I always use, so I'm more likely to notice if it changes.

Regarding using `nox` in CI, it's complicated by the fact that we do a few tricks in CI (like CI-specific flags for docker/podman checks, parallel execution) that we probably wouldn't pass on to the end-user. Perhaps those features from `bin/run_tests.py` could be added to the nox task. That might be okay, but I'm more of the opinion that the problem here is just missing documentation.

webknjaz:
@joerick while I understand your sentiment, people are quite used to passing arguments to the wrapped commands. This is quite idiomatic in the tox world (and by extension, nox too). The whole idea is that this ends up always using the correct environment that is set up properly and the same way as in the CI. Tox also has `--devenv` for exposing disconnected envs, not sure nox has the same.
Either way, you stated that *you*do things this way, which is fine because you're proficient and know enough implementation details. But it's hugely different for an arbitrary contributor. This is why, their starting point should be as close to the CI as possible. This doesn't prevent folks from doing advanced stuff but lowers the entry barrier for the beginners, which I think is important.

As for the custom CLI options, a better place for them is pytest. I can take a look into that, if that's something you'd like to explore.

test/test_from_sdist.py
```diff
@@ -61,10 +61,23 @@ def test_simple(tmp_path):
...
 
+    setup_py_assertion_snippet = textwrap.dedent(
+        """
+        import os
+
+        assert os.path.exists("setup.py")
+        assert os.path.exists("{package}/setup.py")
+        """,
+    )
+    setup_py_assertion_cmd = f"python3 -c '{setup_py_assertion_snippet !s}'"
```
webknjaz, line 72: I think Windows requires the CLI arguments to be wrapped with double quotes and breaks on single quotes. This should work for both Windows and *nix:
```suggestion
    setup_py_assertion_cmd = f'python3 -c "{setup_py_assertion_snippet !s}"'
```

test/test_from_sdist.py
```diff
@@ -61,10 +61,23 @@ def test_simple(tmp_path):
     sdist_dir.mkdir()
     sdist_path = make_sdist(basic_project, sdist_dir)
 
+    setup_py_assertion_snippet = textwrap.dedent(
+        """
+        import os
+
+        assert os.path.exists("setup.py")
+        assert os.path.exists("{package}/setup.py")
```
webknjaz, line 69: Try this additionally:
```suggestion
        assert os.path.exists('setup.py')
        assert os.path.exists('{package}/setup.py')
```

webknjaz:
@henryiii @joerick approve the GHA run?

webknjaz:
No CI failures so far 🤞 Waiting for the completion of the rest of the jobs.

webknjaz:
@joerick @henryiii the CI is fully green. Would you like some commits to be polished before merging?

henryiii (approved):

henryiii:
Nah, squash and merge is fine unless files both move and change.

joerick:
Thanks all for this!

PR #1698, opened 2023-12-13
Add docs for how to run tests in development

joerick:
Ref #1687

Adding additional information about running tests. I've chosen to spell this out using a venv and pytest because I've found that I often want to use extra pytest options, like `-k` and `--capture`, not to mention the CIBW_PLATFORM variable, so it feels like it's not all that useful to make out that it's as easy as `nox -s tests`.


docs/contributing.md
```diff
@@ -12,8 +12,49 @@ If you have an idea for a modification or feature, it's probably best to raise a
...
+pytest unit_test
+```
+
+There are a few custom options to enable different parts of the test suite - check `pytest unit_test --help` for details.
+
+#### Integration tests
+
+The integration test suite is big - it can take more than 30 minutes to run the whole thing. So normally you'd choose specific tests to run locally, and rely on the project's CI for the rest.
+
+To run them locally, you must pass the platform you want to test via the `CIBW_PLATFORM` environment variable. The `-k` pytest option can be used to select the specific tests. For example, you could do:
```
webknjaz, line 45: I think it might be reasonable to expose some sort of a CLI option in pytest that would allow autodetecting the current system.

joerick: I've been thinking more about this and it seems to me that really the root cause is that cibuildwheel refuses to run at all without the `--platform` argument. That was deliberate originally, because it would assume a CI environment and install stuff globally, but we've gradually removed all those bits and now I think it's just a speed bump without much reason. We should consider removing it.

joerick: Just got that done, it's https://github.com/pypa/cibuildwheel/pull/1727

joerick:
I'll take another look at this after #1727 is resolved.

joerick:
Took another go at this. The diff's a bit of a mess - it's easier to read the rendered version: https://cibuildwheel--1698.org.readthedocs.build/en/1698/contributing/#development

joerick:
@henryiii any input on this before it's merged? I've decided to go with `nox` as the primary way to document these, and since we don't need CIBW_PLATFORM any more, things are simpler generally. And, nox is gonna get much faster very soon :)

Once this one is merged, I'm keen to get a release out.

PR #1700, opened 2023-12-16
chore(CI): test on GHA macOS-14 runner

mayeut:
PyPA is now on an enterprise plan.
Let's use the new macos-13-xlarge runner in order to test on macOS arm64.


mayeut:
Seems macos-13-xlarge is only available as a paid option (Included minutes cannot be used for larger runners.)
It might be worth discussing this with @pradyunsg who'll have a chat early next year with GitHub folks.

joerick:
@pradyunsg, quick note from us at cibuildwheel, if you do have another discussion with Github about support of PyPA, it would be great if we could get access to 'larger' runners under the plan. 

Testing on macOS arm64 is pretty spotty currently, we only have CirrusCI that supports it and we can't afford to run that on PRs, only on main, as otherwise we run out of quota every month.

Github [recently released hosted runners for macOS arm64](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/), but they're of the 'larger' type, which are priced per-minute and seemingly not included in the current plan.

So access to these runners would be great, if at all possible!

henryiii:
Looks like M1's are now available generally:

https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-latest, macos-11]
+        os: [ubuntu-20.04, windows-latest, macos-11, macos-13-xlarge]
```
henryiii, line 41:
```suggestion
        os: [ubuntu-20.04, windows-latest, macos-11, macos-14]
```

henryiii:
We can rebase on #1743 once it's in, no need to merge this PR before a patch release, unless you want to.

joerick:
Lets get the examples and docs updated first. We might as well provide a config that does native compilation on both x86_64 and arm64 while both runners are available IMO.

examples/github-minimal.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macos-11]
+        os: [ubuntu-20.04, windows-2019, macos-12, macos-14]
```
joerick, line 11: I do wonder if it looks a bit like a mistake to list both these versions here. Worth a comment?

henryiii: I think I would, since you can't read from this that one is Intel and one ARM. Why not `macOS-13`, though?

docs/faq.md
```diff
@@ -68,43 +68,26 @@ Opinions vary on which of arch-specific or `universal2` wheels are best - some p
...
 
-!!! note
-    Your runner needs Xcode Command Line Tools 12.2 or later to build `universal2` or `arm64`.
+If your CI provider doesn't offer arm64 runners yet, or you want to create `universal2`, you'll have to cross-compile. Cross-compilation can be enabled by adding extra archs to the [`CIBW_ARCHS_MACOS` option](options.md#archs) - e.g. `CIBW_ARCHS_MACOS="x86_64 universal2"`. Cross-compilation is provided by Xcode toolchain v12.2+.
+
+Regarding testing,
 
-    Only CPython 3.8 and newer support `universal2` and `arm64` wheels.
+- On an arm64 runner, it is possible to test x86_64 wheels an and both parts of a universal2 wheel using Rosetta 2 emulation.
+- On an x86_64 runner, arm64 code can be compiled but it can't be tested. `cibuildwheel` will raise a warning to notify you of this - these warnings be be silenced by skipping testing on these platforms: `CIBW_TEST_SKIP: "*_arm64 *_universal2:arm64"`.
```
lelit, line 84: s/be be/can be/

mayeut, line 84:
```suggestion
- On an x86_64 runner, arm64 code can be compiled but it can't be tested. `cibuildwheel` will raise a warning to notify you of this - these warnings can be silenced by skipping testing on these platforms: `CIBW_TEST_SKIP: "*_arm64 *_universal2:arm64"`.
```

joerick: good spot! this took me 30 seconds to actually find 😅

examples/github-minimal.yml
joerick: have changed it to macos-13, along with our tests

docs/faq.md
```diff
@@ -68,43 +68,26 @@ Opinions vary on which of arch-specific or `universal2` wheels are best - some p
...
+#### Cross-compiling
 
-!!! note
-    Your runner needs Xcode Command Line Tools 12.2 or later to build `universal2` or `arm64`.
+If your CI provider doesn't offer arm64 runners yet, or you want to create `universal2`, you'll have to cross-compile. Cross-compilation can be enabled by adding extra archs to the [`CIBW_ARCHS_MACOS` option](options.md#archs) - e.g. `CIBW_ARCHS_MACOS="x86_64 universal2"`. Cross-compilation is provided by Xcode toolchain v12.2+.
+
+Regarding testing,
 
-    Only CPython 3.8 and newer support `universal2` and `arm64` wheels.
+- On an arm64 runner, it is possible to test x86_64 wheels an and both parts of a universal2 wheel using Rosetta 2 emulation.
```
radarhere, line 83:
```suggestion
- On an arm64 runner, it is possible to test x86_64 wheels and both parts of a universal2 wheel using Rosetta 2 emulation.
```

joerick (approved):

PR #1705, opened 2023-12-19
docs: update examples for v4 of `upload-artifact` and `download-artifact` actions

alugowski:
See discussion: https://github.com/pypa/cibuildwheel/issues/1699

### DIFF
--- a/docs/deliver-to-pypi.md
+++ b/docs/deliver-to-pypi.md
@@ -56,8 +56,9 @@ well as several useful actions. Alongside your existing job(s) that runs cibuild
     - name: Build SDist
       run: pipx run build --sdist
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
+        name: sdist
         path: dist/*.tar.gz
 ```
 
@@ -74,10 +75,10 @@ This requires setting this GitHub workflow in your project's PyPI settings (for 
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
     steps:
-    - uses: actions/download-artifact@v3
+    - uses: actions/download-artifact@v4
       with:
-        name: artifact
         path: dist
+        merge-multiple: true
 
     - uses: pypa/gh-action-pypi-publish@release/v1
 ```
--- a/docs/deliver-to-pypi.md
+++ b/docs/deliver-to-pypi.md
@@ -58,7 +58,7 @@ well as several useful actions. Alongside your existing job(s) that runs cibuild
 
     - uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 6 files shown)


alugowski:
Working workflow using this setup: https://github.com/alugowski/fast_matrix_market/blob/main/.github/workflows/wheels.yml

joerick (approved):
Looks good to me. Thank you for putting this together. @henryiii I see you're across this so would you mind double-checking this too?

henryiii:
Looks mostly good, just two thoughts:

1. We could make the names generically, using `cibw-${{ github.job }}-${{ strategy.job-index }}`.
2. We could avoid downloading all artifacts, and only download ours using `pattern: cibw-*`.

I think the current PR is probably simplest, as usually (and in this example) there aren't extra artifacts to worry about. We could mention this somewhere, though.

henryiii (approved):

njzjz:
> We could make the names generically, using `cibw-${{ github.job }}-${{ strategy.job-index }}`.

I tried this, but the name looks too generic when I want to download the wheel for a specific platform.

image

It's hard to know which one is built for Linux.

alugowski:
Would `cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}` and `cibw-sdist` be a good complexity/uniqueness compromise?

I thought just `matrix.os` is enough, but the GH docs use that as the example and I've seen comments complaining, as if they didn't know it was just a suggestion. I think the places where `github.job` would help disambiguate are few, and those folks would know to add it.

Also I just noticed the `examples/` directory that has like 4 workflows that should be updated too. Once a call is made on this I can update those.

alugowski:
>     1. We could make the names generically, using `cibw-${{ github.job }}-${{ strategy.job-index }}`.

Based on feedback above I went with `cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}`, but happy to change.

>     2. We could avoid downloading all artifacts, and only download ours using `pattern: cibw-*`.

Done.

I also updated the example `.yml` files.

joerick:
I like the renamed artifacts with the `cibw` prefix, it also make it clearer what 'merge-multiple' means in the example.

I'll let @henryiii have another look, but LGTM.

henryiii (approved):
LGTM!

alexlancaster:
The PR seems to be missing an update to the GitHub action in the https://github.com/pypa/cibuildwheel/blob/main/README.md ?

alugowski:
> The PR seems to be missing an update to the GitHub action in the "Example setup " in https://github.com/pypa/cibuildwheel/blob/main/README.md ?

Good catch.

https://github.com/pypa/cibuildwheel/pull/1708

PR #1719, opened 2024-01-06
Refactor error handling to use exceptions

joerick:
cibuildwheel has up until now handled most errors by printing an error message to sys.stderr and calling sys.exit. Others were handled using Logger.error, depending on the context. We also had return codes, but these weren't explicitly defined anywhere.

This makes that convention more explicit and codified. Now to halt the program, the correct thing to do is to throw a cibuildwheel.errors.FatalError exception - that is caught in main() and printed before exiting. The existing behaviour was kept - if an error occurs within a build step (probably something to do with the build itself), the Logger.error() method is used. Outside of a build step (e.g. a misconfiguration), the behaviour is still to print 'cibuildwheel: <message>'

I also took the opportunity to add a debugging option `--debug-traceback` (and `CIBW_DEBUG_TRACEBACK`), which you can enable to see a full traceback on errors.

~(I've deactivated the flake8-errmsg lint rule, as it was throwing loads of errors and these error messages aren't generally seen in a traceback context)~


cibuildwheel/windows.py
```diff
@@ -578,7 +574,6 @@ def build(options: Options, tmp_path: Path) -> None:
 
             log.build_end()
     except subprocess.CalledProcessError as error:
-        log.step_end_with_error(
-            f"Command {error.cmd} failed with code {error.returncode}. {error.stdout}"
-        )
-        sys.exit(1)
+        raise errors.FatalError(
+            f"Command {error.cmd} failed with code {error.returncode}. {error.stdout or ''}"
+        ) from error
```
henryiii, line 579: IMO, it's still easier to read _and_ fewer lines with:

```suggestion
        msg = f"Command {error.cmd} failed with code {error.returncode}. {error.stdout or ''}"
        raise errors.FatalError(msg) from error
```

Which `ruff --fix` would do for you automatically with that rule on (might require unsafe fixes).

cibuildwheel/util.py
```diff
@@ -379,7 +379,7 @@ def from_config_string(config_string: str) -> BuildFrontendConfig:
         config_dict = parse_key_value_string(config_string, ["name"], ["args"])
         name = " ".join(config_dict["name"])
         if name not in {"pip", "build"}:
-            msg = f"Unrecognised build frontend {name}, only 'pip' and 'build' are supported"
+            msg = f"Unrecognised build frontend '{name}', only 'pip' and 'build' are supported"
```
henryiii, line 382:
```suggestion
            msg = f"Unrecognised build frontend {name!r}, only 'pip' and 'build' are supported"
```

cibuildwheel/options.py
```diff
@@ -767,6 +763,7 @@ def _get_pinned_container_images() -> Mapping[str, Mapping[str, str]]:
 def deprecated_selectors(name: str, selector: str, *, error: bool = False) -> None:
     if "p2" in selector or "p35" in selector:
         msg = f"cibuildwheel 2.x no longer supports Python < 3.6. Please use the 1.x series or update {name}"
-        print(msg, file=sys.stderr)
-        if error:
-            sys.exit(4)
+        if not error:
+            print(msg, file=sys.stderr)
+        else:
+            raise errors.DeprecationError(msg)
```
henryiii, line 769:
```suggestion
        if error:
            raise errors.DeprecationError(msg)

        print(msg, file=sys.stderr)
```

cibuildwheel/windows.py
joerick: Yeah, happy to compromise on this one, especially if ruff can do it for me!

webknjaz: FWIW I'd use a better variable name, to fix what Ruff would suggest...

cibuildwheel/errors.py
webknjaz, line 1: Worth adding a module docstring explaining the exception hierarchy and the intended use/architecture?

cibuildwheel/linux.py
```diff
@@ -395,8 +386,8 @@ def build(options: Options, tmp_path: Path) -> None:  # noqa: ARG001
             check=True,
             stdout=subprocess.DEVNULL,
         )
-    except subprocess.CalledProcessError:
-        print(
+    except subprocess.CalledProcessError as e:
```
webknjaz, line 389: nitpick: use a better variable name? Like `proc_err`, maybe?

cibuildwheel/errors.py
joerick: good idea.

cibuildwheel/errors.py
```diff
@@ -0,0 +1,20 @@
+class FatalError(SystemExit):
```
joerick, line 1: Btw, I couldn't decide whether to subclass SystemExit or Exception here, any strong opinions?

cibuildwheel/linux.py
joerick: I don't see a huge benefit, there's only one statement in the `except` clause. I will standardise it to look like the others though.

cibuildwheel/errors.py
joerick: On reflection, I'll change it to Exception to avoid the confusion around SystemExit's `code` attribute.
```suggestion
class FatalError(Exception):
```

joerick:
A review here would be good too. It's not critical, maybe just a little code quality improvement. I'm not too attached to it though, we could drop the idea if the existing convention approach is preferred.

cibuildwheel/options.py
```diff
@@ -770,6 +767,7 @@ def _get_pinned_container_images() -> Mapping[str, Mapping[str, str]]:
 def deprecated_selectors(name: str, selector: str, *, error: bool = False) -> None:
     if "p2" in selector or "p35" in selector:
         msg = f"cibuildwheel 2.x no longer supports Python < 3.6. Please use the 1.x series or update {name}"
-        print(msg, file=sys.stderr)
         if error:
-            sys.exit(4)
+            raise errors.DeprecationError(msg)
+
+        print(msg, file=sys.stderr)
```
mayeut, line 773: Shouldn't this use `logger.warning` ?

mayeut (approved):

henryiii (approved):

cibuildwheel/errors.py
webknjaz: If it needs to behave like `SystemExit`, I'd inherit from `BaseException`. Have you considered this?

webknjaz: (since it's raised in places where `sys.exit()` used to be, I'd expect this, I think)

cibuildwheel/errors.py
```diff
@@ -0,0 +1,27 @@
+"""
+Errors that can cause the build to fail. Each subclass of FatalError has
+a different return code, by defining them all here, we can ensure that they're
+semantically clear and unique.
+"""
+
+
+class FatalError(Exception):
```
henryiii, line 8:
```suggestion
class FatalError(BaseException):
```

henryiii:
Let's release after this. I've been trying out macOS free threading and `build[uv]` in some draft PR's,, and have been happy with the results.

cibuildwheel/errors.py
webknjaz: Yep, that's what I meant :)

PR #1725, opened 2024-01-20
Don't preinstall setuptools and wheel, remove from constraints

joerick:
Reasons to do this:

- setuptools is no longer the only game in town. When cibuildwheel was created, setuptools was the only way to make wheels. these days, there are many build backends so special casing setuptools doesn't make so much sense
- we don't need to preinstall `wheel` any more, setuptools will do that itself when it needs it.
- projects with pyproject.toml (i.e. the vast majority of projects going forward) don't benefit from the pin anyway, because both `pip` and `build` create an isolated environment to install `build-system.requires` into.
- this makes it easier for end users (such as @webknjaz) to control build-system versions using `PIP_CONSTRAINT`, because there's less potential for a user-specified version to conflict with one of our pins.
  - (long-term, I'd love `build` to [support this directly](https://github.com/pypa/build/issues/292), and then we could have a proper way to keep pyproject.toml dependencies loose and forward-compatible while retaining build determinism. But PIP_CONSTRAINT is the best we have for now. PIP_CONSTRAINT isn't ideal because it affects other things, like the test virtualenv).

This PR also adds a pin for pypa/build. It appears that was missing before, but we should pin it.


test/test_dependency_versions.py
```diff
@@ -70,13 +72,13 @@ def test_pinned_versions(tmp_path, python_version, build_frontend_env):
...
-        constraint_filename = "constraints.txt"
-        build_pattern = "[cp]p39-*"
+        constraint_filename = "constraints-python310.txt"
+        build_pattern = "[cp]p310-*"
 
     constraint_file = cibuildwheel.util.resources_dir / constraint_filename
     constraint_versions = get_versions_from_constraint_file(constraint_file)
 
-    for package in ["pip", "setuptools", "wheel", "virtualenv"]:
+    for package in ["pip"]:
```
webknjaz, line 81: Why not drop the loop entirely?

test/test_dependency_versions.py
```diff
@@ -27,7 +27,9 @@
 
         print('Gathered versions', versions)
 
-        for package_name in ['pip', 'setuptools', 'wheel']:
+        packages_to_test = ['pip']
```
webknjaz, line 30: Why not drop the loop?

joerick: Sorry, I should have dropped a comment here. The loop was kept as a reminder that this test is just a little bit in flux going forward. Checking just the version of pip is not ideal, but it's the only package that's both in our constraints and accessible from a setup.py. 

This test could be refactored to use CIBW_BEFORE_BUILD to do the asserts, but that would only give us the extra ability to check the version of delocate on macOS, nothing is preinstalled on windows. I didn't think it was worth it, especially because CIBW_DEPENDENCY_VERSIONS feels less useful than it once was because it no longer affects the actual build.

joerick: I guess, it's possible that if we do get the ability to control build-system.requires dependencies some day, we might change CIBW_DEPENDENCY_VERSIONS to do that, and remove the ability to fiddle with the tool versions, or relegate it to a more 'under-the-hood' option.

henryiii:
Would it make sense to get a release out before this? PyPy had a bug that broke NumPy's CI, and they had to pin on an older version, and we updated PyPy weeks ago, but haven't released yet. This seems like a bigger change that should sit in main a bit, but we are well past due on a a release with the latest PyPy.

test/test_dependency_versions.py
webknjaz: I see. May I suggest adding that as a code comment?

webknjaz (approved):

joerick:
> This seems like a bigger change that should sit in main a bit, but we are well past due on a a release with the latest PyPy

Apologies, I missed the PyPy stuff over the holidays. I'll get a release out today/tomorrow.

webknjaz:
@joerick since that release is out, should this be merged now?

test/test_dependency_versions.py
joerick: reading this back now, there's a lot of what-ifs... i'll just remove the loop :)

henryiii (approved):

PR #1727, opened 2024-01-21
Don't require `--platform` when running on a development machine

joerick:
Historically, we required `--platform` when running cibuildwheel outside of CI, because cibuildwheel would attempt to globally install Python interpreters and some packages.

That was fixed in https://github.com/pypa/cibuildwheel/pull/974 a couple of years ago. As a result, there's no reason to require --platform these days, the `auto` behaviour is perfectly fine for dev machines.

It also makes running tests locally simpler, one can just do `pytest test/test_0_basic.py` (or `nox -s tests -- test/test_0_basic.py`) and not have to worry about setting CIBW_PLATFORM most of the time.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -192,7 +192,9 @@ def _compute_platform_auto() -> PlatformName:
         return "windows"
     else:
         print(
-            'cibuildwheel: Unable to detect platform from "sys.platform". cibuildwheel doesn\'t support building wheels for this platform. You might be able to build for a different platform using the --platform argument. Check --help output for more information.',
+            'cibuildwheel: Unable to detect platform from "sys.platform". cibuildwheel doesn\'t '
+            "support building wheels for this platform. You might be able to build for a different "
+            "platform using the --platform argument. Check --help output for more information.",
             file=sys.stderr,
         )
         sys.exit(2)
@@ -277,7 +279,9 @@ def build_in_directory(args: CommandLineArguments) -> None:
     # Add CIBUILDWHEEL environment variable
     os.environ["CIBUILDWHEEL"] = "1"
 
-    # Python is buffering by default when running on the CI platforms, giving problems interleaving subprocess call output with unflushed calls to 'print'
+    # Python is buffering by default when running on the CI platforms, giving
+    # problems interleaving subprocess call output with unflushed calls to
+    # 'print'
     sys.stdout = Unbuffered(sys.stdout)  # type: ignore[assignment]
 
     # create the cache dir before it gets printed & builds performed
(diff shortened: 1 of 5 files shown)


cibuildwheel/__main__.py
```diff
@@ -206,8 +192,7 @@ def _compute_platform_ci() -> PlatformName:
         return "windows"
     else:
         print(
-            'cibuildwheel: Unable to detect platform from "sys.platform" in a CI environment. You can run '
-            "cibuildwheel using the --platform argument. Check --help output for more information.",
+            'cibuildwheel: Unable to detect platform from "sys.platform". cibuildwheel doesn\'t support building wheels for this platform. You might be able to build for a different platform using the --platform argument. Check --help output for more information.',
```
webknjaz, line 195: nitpick: I liked the variant with implicit string concat better since it's easier to consume shorter lines in code/diffs..

unit_test/main_tests/main_platform_test.py
```diff
@@ -10,25 +10,26 @@
...
-    assert exit.value.code == 2
-    _, err = capsys.readouterr()
+    main()
 
-    assert "cibuildwheel: Unable to detect platform." in err
-    assert "cibuildwheel should run on your CI server" in err
+    options = intercepted_build_args.args[0]
+
+    # check that the platform was auto detected to build for the current system
+    if sys.platform.startswith("linux"):
```
webknjaz, line 25: nit: this feels like too much logic in a test. Any chance to simplify it?

joerick: I know what you mean, it's essentially just repeating the logic in the main program. But I would like to be sure that both `auto` and unspecified are doing the right thing. A little extra verbosity in the test seems like a small price to pay IMO.

henryiii (approved):

unit_test/main_tests/main_platform_test.py
webknjaz: In that case, why not try solving this on the level of parametrization? It'd be even clearer if it shows up in the test report.

webknjaz (approved):

unit_test/main_tests/main_platform_test.py
Czaki: Even not parametrization but simeple dict could solve this:
```python
platform_dkt = {"linux": "linux", "darwin": "macos", "win32": "windows"}

assert sys.platform in platform_dkt, f"Unknown platform: {sys.platform}"
assert platform_dkt[sys.platform] == options.platform
```

joerick: While I appreciate the focus on craft and code quality, there becomes a point when it's not worth our time to debate such minor points. Especially in tests, where the control flow complexity is so low, because each test is an isolated execution. All that matters in a test is the actions that are taken and the asserts that are reached - if these are clear to the reader, it's fine. If they grow to become unmaintainable, it's easy to rewrite them.

cibuildwheel/__main__.py
joerick: updated

PR #1734, opened 2024-01-27
Don't bother pinning OS images for Github Actions

### LINKED ISSUE #1612: Github actions docs use different versions in matrix.os
### Description

There is a github action shown in the main doc at https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions

It points to a more complete example at https://github.com/pypa/cibuildwheel/blob/main/examples/github-deploy.yml

They have differences in matrix.os:

     os: [ubuntu-20.04, windows-2019, macos-11]

     os: [ubuntu-22.04, windows-2022, macos-11]

There are a few other spots where matrix.os is present like [here](https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation).  It would be really nice if the doc only had the recommend set which is presumably the newer versions.

### Build log

_No response_

### CI config

_No response_

joerick:
Fix #1612

The hassle of updating these pins (both for us and our users, most copy our documentation) seems to outweigh the benefit of the pin. The truth is that they're not really pins anyway, Github update them all the time, and they don't work for historical repeatability because old images are retired all the time.


henryiii (approved):

henryiii:
`ubuntu-latest` is a good idea (we don't use anything really here anyway), but we probably should keep `macos-*` specified so that the images don't turn in AS by surprise. Once that does happen, then it's fine to use `macos-latest` (for AS).

joerick:
ah yeah, good point. I'll make that change.

README.md
```diff
@@ -88,7 +88,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-11]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
```
mayeut, line 91:
```suggestion
        os: [ubuntu-latest, windows-latest, macOS-13]
```

henryiii: Should this have 14 too?

joerick: The change to add macos-14 is in #1700. I wasn't sure why macos-13 never made it to `-latest`, I figured it might not be as battle-tested as macos-12. Also since we run our unit tests on macos-12 that seems like the best option to recommend to me.

joerick: happy to go to macos-13 if we update our tests to that too though.

.github/workflows/test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         os: [ubuntu-20.04, windows-latest, macos-13, macos-14]
```
mayeut, line 41:
```suggestion
        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
```

joerick:
Thanks for doing that merge.

PR #1735, opened 2024-01-27
Add keyvi as an example that combines cibuildwheel with the ccache action to speedup runtime.

hendrikmuhs:
Adding keyvi as example, it combines cibuildwheel with the ccache action to speedup runtime.


joerick:
thanks! but this file is generated, you should add to docs/data/projects.yml

hendrikmuhs:
Thanks @joerick, I reverted my change and put it into docs/data/projects.yml

henryiii:
The docs will get regenerated on the next (weekly) update after merging.

henryiii:
PS: I'm also probably going to add a schema check for docs/data/projects.yml soon, which will make these PRs get checked even before the docs run.

PR #1739, opened 2024-01-29
chore: check schemas

henryiii:
This does some validation on schemas.

Mentioned in #1735. Can't validate Azure Pipelines yet due to https://github.com/SchemaStore/schemastore/issues/3560.


joerick (approved):
Nice!

PR #1741, opened 2024-01-29
Add support for PowerShell 7.3 in GHA

jborean93:
Fixes compatibility with PowerShell 7.3 on Windows through GitHub Actions when one of the options for cibuildwheel is empty.

I'm not sure how you want to go about this but I believe this should solve the issue with the migration of GHA to PowerShell 7.4 that is happening now.

Fixes: https://github.com/pypa/cibuildwheel/issues/1740

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -46,7 +46,13 @@ runs:
       if: runner.os != 'Windows'
 
     # Windows needs powershell to interact nicely with Meson
+    # $PSNativeCommandArgumentPassing was introduced in pwsh 7.3 and the
+    # legacy behaviour is needed for backwards compatibility with how this
+    # was called in the past.
     - run: >
+        if ($PSNativeCommandArgumentPassing) {
+            $PSNativeCommandArgumentPassing = 'Legacy'
+        };
         pipx run
         --python "${{ steps.python.outputs.python-path }}"
         --spec "${{ github.action_path }}"


jborean93:
I've used https://github.com/pythongssapi/python-gssapi/pull/342 to test out these changes.

+ Runner image `20240122.1.0` (PowerShell 7.2) - https://github.com/pythongssapi/python-gssapi/actions/runs/7703855440/job/20995280412?pr=342
+ Runner image `20240128.1.0` (PowerShell 7.4) - https://github.com/pythongssapi/python-gssapi/actions/runs/7703855440/job/20995280547?pr=342

Both work with the changes in the PR.

henryiii (approved):
I think this is fine for a quick fix. Hopefully in a few months everyone will be on 7.3+ and we can simply this to look much more like the bash version, getting rid of the weird double quoting finally.

jborean93:
I would assume you would still want this fix due to self hosted runners that might be running with 7.2. 7.2 EOL is November 8th 2024 https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.4#powershell-end-of-support-dates

jborean93:
Thanks for the reviews!

henryiii:
I'd probably not bother to remove this workaround until 2025 or so. Just nice to think we can eventually remove it.

PR #1742, opened 2024-01-30
fix: simplify action quoting on Windows

### LINKED ISSUE #1740: GitHub Actions Windows Wheels - Invalid --only='""', must be a build selector with a known platform
### Description

I've getting random errors when attempting to build Windows wheels using the `pypa/cibuildwheel@2.16.4` action.

```
Run pypa/cibuildwheel@v2.16.4
Run actions/setup-python@v5
Installed versions
Run pipx run --python "C:\hostedtoolcache\windows\Python\3.12.1\x64\python.exe" --spec "D:\a\_actions\pypa\cibuildwheel\v2.16.4" cibuildwheel "." --output-dir '"wheelhouse"' --config-file '""' --only '""'
  pipx run --python "C:\hostedtoolcache\windows\Python\3.12.1\x64\python.exe" --spec "D:\a\_actions\pypa\cibuildwheel\v2.16.4" cibuildwheel "." --output-dir '"wheelhouse"' --config-file '""' --only '""'
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    CIBW_ARCHS: all
    CIBW_TEST_SKIP: *_arm64
    CIBW_BUILD: cp38-win32
    CIBW_BUILD_VERBOSITY: 1
creating virtual environment...
determining package name from 'D:\\a\\_actions\\pypa\\cibuildwheel\\v2.16.4'...
creating virtual environment...
installing cibuildwheel from spec 'D:\\a\\_actions\\pypa\\cibuildwheel\\v2.16.4'...
Invalid --only='""', must be a build selector with a known platform
Error: Process completed with exit code 1.
```

On a build that works I can see the same command is being run with the `--only='""'` parameter.

```
Run pipx run --python "C:\hostedtoolcache\windows\Python\3.12.1\x64\python.exe" --spec "D:\a\_actions\pypa\cibuildwheel\v2.16.4" cibuildwheel "." --output-dir '"wheelhouse"' --config-file '""' --only '""'
  pipx run --python "C:\hostedtoolcache\windows\Python\3.12.1\x64\python.exe" --spec "D:\a\_actions\pypa\cibuildwheel\v2.16.4" cibuildwheel "." --output-dir '"wheelhouse"' --config-file '""' --only '""'
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    CIBW_ARCHS: all
    CIBW_TEST_SKIP: *_arm64
    CIBW_BUILD: cp38-win_amd64
    CIBW_BUILD_VERBOSITY: 1
```

The weird thing is that this only happens sometimes, other times the build works just fine. I know the `2.15.0` version is also affected. I only started seeing this in the last day. For example here is an older build that failed 2 times while on the 3rd it worked again https://github.com/pythongssapi/python-gssapi/actions/runs/7697806880

image

I'm unsure whether it's my code or something else but this has worked for me for the past year or so whereas now it seems to be popping up. At a guess it might be 

### Build log

https://github.com/pythongssapi/python-gssapi/actions/runs/7703392610/job/20993685914?pr=341

### CI config

https://github.com/pythongssapi/python-gssapi/blob/18f3e2bc5119a3621c7bc14e307ab70ce011e34f/.github/workflows/ci.yml#L123-L129

henryiii:
I'm not sure the quoting was necessary to make this complicated, I don't see a reason in #1346. This is option 1 to fix #1740. Would also close #1741, which is option 2 (personally, if option 1 doesn't work, option 3 seems better than option 2).

This is needed to support Powershell 7.4, which is beginning to ship on GHA.

~~The action is not exercised in tests, so we'll need someone to check it. I can later today, perhaps.~~ (It seems to be in the tests, actually)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -52,8 +52,8 @@ runs:
         --spec "${{ github.action_path }}"
         cibuildwheel
         "${{ inputs.package-dir }}"
-        --output-dir '"${{ inputs.output-dir }}"'
-        --config-file '"${{ inputs.config-file }}"'
-        --only '"${{ inputs.only }}"'
+        --output-dir "${{ inputs.output-dir }}"
+        --config-file "${{ inputs.config-file }}"
+        --only "${{ inputs.only }}"
       shell: pwsh
       if: runner.os == 'Windows'


henryiii:
Ahh, so the problem is that `""` resolves to nothing, while `'""'` resolved to the empty string (pre-7.4), but now resolves to the string `""`. That's why the quoting was that way. Is there a recommended alternative to pass the empty string in powershell 7.3/7.4?

henryiii:
I don't understand what is eating those empty args. If I run powershell 7.4.1 on my Mac, I see `python3 -c "import sys; print(sys.argv)" ""` produces `['-c', '']`, as I'd expect, and `pipx` passes them through correctly too. YAML shouldn't be eating them inside a block, so I'm not sure where they are going.

PR #1743, opened 2024-01-30
fix: download pipx for action, allow support for M1

henryiii:
Takes an extra second or two, plus it could interfere with the environment a bit if someone activates the same version of Python. But not too bad.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -33,12 +33,12 @@ runs:
 
     # macos-14 (M1) may be missing pipx (due to it not having CPython)
     - run: |
-        '${{ steps.python.outputs.python-path }}' -m pip install pipx
+        "${{ steps.python.outputs.python-path }}" -m pip install pipx
       shell: bash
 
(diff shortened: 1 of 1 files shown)


joerick:
@henryiii, hope it's okay, I pushed a couple commits to get this working. I hope that actions/setup-python wouldn't put spaces in the path! I want to do a quick release to get that Powershell fix in and I thought that we might as well let people try out the new macos-14 runners. We can do the official support with examples/docs in #1700.

joerick:
Erm, actually, on second thoughts maybe that's not a good assumption  given 'Program Files' and all that.

henryiii:
Great, thanks! Had to teach. That's what I was thinking, fixing the powershell plus macOS AS runners would be a good quick patch release.

henryiii:
This won't break Travis, so fine to merge IMO.

joerick:
brill, thanks, releasing now...

PR #1745, opened 2024-01-31
chore(action): do not use pipx

mayeut:
If #1743 is just meant to be temporary while waiting on https://github.com/actions/runner-images/issues/9256, then that's ok. 

However, if we need this in the long run, I'd rather not change the worker default environment. It's not necessarily an issue for github-hosted workers which are ephemeral but it could became an issue for self-hosted.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -31,16 +31,48 @@ runs:
         python-version: "3.8 - 3.12"
         update-environment: false
 
-    # macos-14 (M1) may be missing pipx (due to it not having CPython)
-    - run: |
-        "${{ steps.python.outputs.python-path }}" -m pip install pipx
+    - id: cibw
+      run: |
+        # Install cibuildwheel
+        "${{ steps.python.outputs.python-path }}" -u << "EOF"
+        import os
+        import shutil
+        import sys
+        import venv
+
+        from pathlib import Path
+        from subprocess import run
+
+
+        class EnvBuilder(venv.EnvBuilder):
+            def __init__(self):
+                super().__init__()
+
+            def setup_scripts(self, context):
+                pass
+
+            def post_setup(self, context):
+                super().post_setup(context)
+                self.bin_path = Path(context.env_exe).parent
+                run([sys.executable, "-m", "pip", "--python", context.env_exe, "install", r"${{ github.action_path }}"], check=True)
+
+
+        print("::group::Install cibuildwheel")
+        venv_path = Path(r"${{ runner.temp }}") / "cibw"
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -57,12 +85,12 @@ runs:
...
-        & "${{ steps.python.outputs.python-path }}" -m pipx run
-        --spec "${{ github.action_path }}"
+        & "${{ steps.cibw.outputs.python-path }}" -m
         cibuildwheel
         "${{ inputs.package-dir }}"
         --output-dir '"${{ inputs.output-dir }}"'
         --config-file '"${{ inputs.config-file }}"'
-        --only '"${{ inputs.only }}"'
+        --only '"${{ inputs.only }}"';
+        exit $LASTEXITCODE
```
henryiii, line 94: Why is this needed? Is it currently not exiting correctly?

henryiii:
Let's see if they re-add pipx, which I'd really hope they do. I should have limited the workaround to a subset of runners (macos-14 only? macos-13-xlarge? Ideally only AS arch only?)

henryiii:
(It can also break later steps if people are doing things with setup-python and need a clean environments that doesn't have pipx's dependencies installed. `cibuildwheel` tends to be mostly standalone, which is nice, but not true for nox (where I haven't proposed a fix for this yet))

action.yml
mayeut: This was needed locally when testing different things with `pwsh` but GHA appends this automatically so  not needed:
https://github.com/actions/runner/blob/1d82031a2cc4bb95ace076743c6cce96ca729881/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs#L62-L64

henryiii:
Should we take these out of draft? I don't think pipx / built-in Python is coming back soon, and am fine with the workaround. (Also https://github.com/wntrblm/nox/pull/768)

joerick (approved):
Looks good to me!

nightlark:
I noticed a CI job is failing again on macOS due to the missing pipx when updating to the latest cibuildwheel release: https://github.com/nightlark/swig-pypi/pull/123

Is there a recommended alternative (aside from migrating to scikit-build-core now)? Currently using pyproject.toml `tool.cibuildwheel.before-all` with pipx to ensure suitable cmake and ninja versions are installed.

henryiii:
Migrating sounds good. :) But GHA is shipping pipx in the next image deployment. And you can pre-install pipx yourself in the mean time.

PR #1750, opened 2024-02-01
test macos-x86_64 and arm64 in different venv dir

njzjz:
pip cannot install wheels with different architecture into the same venv directory. (See https://github.com/pypa/cibuildwheel/issues/1746#issuecomment-1920164503)

Downstream test:
Before this PR: failed [log](https://github.com/tongzhugroup/mddatasetbuilder/actions/runs/7733603508/job/21086015132)
After this PR: pass [log](https://github.com/tongzhugroup/mddatasetbuilder/actions/runs/7748040866/job/21129838429)


joerick (commented):
Ah yes, that is an unfortunate bug! Thank you for the fix! A regression test would be great if you have time to create it.

joerick:
I'll look at adding a unit test now...

mayeut (approved):

PR #1752, opened 2024-02-03
chore(ci): speed-up GHA tests

mayeut:


mayeut:
Maybe we should just get `os.cpu_count()` or `len(os.sched_getaffinity(0))` as a default & override where it needs to be override ? (probably only TravisCI)

Czaki:
Last time when I try this it crashed because some collision in pip install using pip cache. Maybe the current timeout maybe also connected with this.

joerick:
> Last time when I try this it crashed because some collision in pip install using pip cache. Maybe the current timeout maybe also connected with this.

This was investigated in https://github.com/pypa/cibuildwheel/issues/1254, I think we worked around this in https://github.com/pypa/cibuildwheel/pull/1301 , so shouldn't be an issue any more.

joerick:
Not huge speedups, but it looks reasonable. Any particular reason to hard-code the concurrency on travis/azure rather than letting `os.cpu_count()` do it?

henryiii:
Let's merge this for the speedups, then revisit if we can use a smarter method to get the num threads.

henryiii (approved):

PR #1766, opened 2024-02-20
Add information about new delocate error "Library dependencies do not satisfy target MacOS" 

Czaki:
After 3 years, I finally contributed to delocate with [PR](https://github.com/matthew-brett/delocate/pull/198) is adding validation of platform tag against real requirements. 

As many users open issues in this repository even when should in some other package, I think is worth to at least opening PR for better visibility. 

This PR should not be merged before the next delocate release.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -355,6 +355,13 @@ sh "/Applications/Python 3.8/Install Certificates.command"
 
 Then cibuildwheel will detect that it's installed and use it instead. However, you probably don't want to build x86_64 wheels on this Python, unless you're happy with them only supporting macOS 11+.
 
+### macOS: Library dependencies do not satisfy target MacOS
+
+Since delocate 0.10.8 there is added verification that the library binary dependencies match the target macOS version. This is to prevent the situation where a wheel platform tag is lower than the actual minimum macOS version required by the library. To resolve this error you need to build the library to the same macOS version as the target wheel (for example using `MACOSX_DEPLOYMENT_TARGET` environment variable).
+Alternatively, you could set a higher `MACOSX_DEPLOYMENT_TARGET` to drop support for older macOS versions.
+
+This error may happen when you install a library using a package manager like Homebrew, which compiles the library for the macOS version of the build machine. This is not suitable for wheels, as the library will only work on the same macOS version as the build machine. You should compile the library yourself, or use a precompiled binary that matches the target macOS version.
+
 ### Windows: 'ImportError: DLL load failed: The specific module could not be found'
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -355,6 +355,13 @@ sh "/Applications/Python 3.8/Install Certificates.command"
 
 Then cibuildwheel will detect that it's installed and use it instead. However, you probably don't want to build x86_64 wheels on this Python, unless you're happy with them only supporting macOS 11+.
 
+### macOS: Library dependencies do not satisfy target MacOS
+
+Since delocate 0.10.8 there is added verification that the library binary dependencies match the target macOS version. This is to prevent the situation where a wheel platform tag is lower than the actual minimum macOS version required by the library. To resolve this error you need to build the library to the same macOS version as the target wheel (for example using `MACOSX_DEPLOYMENT_TARGET` environment variable).
```
Czaki, line 360: I'm only guessing here next number. It should be verified before potential merge.

docs/faq.md
```diff
@@ -355,6 +355,13 @@ sh "/Applications/Python 3.8/Install Certificates.command"
...
+Alternatively, you could set a higher `MACOSX_DEPLOYMENT_TARGET` to drop support for older macOS versions.
```
joerick, line 361:
```suggestion
Alternatively, you could set `MACOSX_DEPLOYMENT_TARGET` in `CIBW_ENVIRONMENT` to correctly label the wheel as incompatible with older macOS versions.
```

joerick (commented):
Nice work on this @Czaki ! A big improvement for sure.

henryiii:
I set this to draft to make it less tempting to merge before delocate is ready. :)

docs/faq.md
henryiii:
```suggestion
Since delocate 0.11.0 there is added verification that the library binary dependencies match the target macOS version. This is to prevent the situation where a wheel platform tag is lower than the actual minimum macOS version required by the library. To resolve this error you need to build the library to the same macOS version as the target wheel (for example using `MACOSX_DEPLOYMENT_TARGET` environment variable).
```

henryiii:
CC https://github.com/pypa/cibuildwheel/pull/1796.

henryiii:
Can it also detect of the arch is wrong? Missing Apple Silicon parts on a universal or arm wheel?

Czaki:
> Can it also detect of the arch is wrong? Missing Apple Silicon parts on a universal or arm wheel?

Not in the current state. When creating PR, I assume that some of the inner binary could be single arch only. But will raise an error if there is no binary with a given arch here: https://github.com/matthew-brett/delocate/blob/526cc494e4c5750380a6d6ad3d31a2040ff63c60/delocate/delocating.py#L798

It is possible to extend the check to validate archs, but will require to use a dependency graph in place of simple scan files. 

As delocate has been released 2 days ago, I mark it as ready for review.

joerick (approved):

PR #1768, opened 2024-02-28
Set SYSTEM_VERSION_COMPAT=0 during pip install on macos

laggykiller:
Fixes https://github.com/pypa/cibuildwheel/issues/1767

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -579,11 +579,7 @@ def build(options: Options, tmp_path: Path) -> None:
                         shell_with_arch(before_test_prepared, env=virtualenv_env)
 
                     # install the wheel
-                    if (
-                        sys.version_info <= (3, 8)
-                        and testing_arch == "x86_64"
-                        and python_arch == "x86_64"
-                    ):
+                    if is_cp38 and python_arch == "x86_64":
                         virtualenv_env_install_wheel = virtualenv_env.copy()
                         virtualenv_env_install_wheel["SYSTEM_VERSION_COMPAT"] = "0"
                         log.notice(
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -578,15 +578,24 @@ def build(options: Options, tmp_path: Path) -> None:
                         )
                         shell_with_arch(before_test_prepared, env=virtualenv_env)
 
-                    # setting virtualenv_env used for pip install wheel
-                    # set SYSTEM_VERSION_COMPAT=0 to ensure x86_64 cp38 python on
-                    # arm64 runner can get correct macos version and allow installation
-                    # of wheels with MACOSX_DEPLOYMENT_TARGET >= 11.0
-                    # https://github.com/pypa/cibuildwheel/issues/1767
-                    virtualenv_env_install_wheel = virtualenv_env.copy()
-                    virtualenv_env_install_wheel["SYSTEM_VERSION_COMPAT"] = "0"
-
                     # install the wheel
+                    if testing_arch == "x86_64" and is_cp38 and python_arch == "x86_64":
+                        virtualenv_env_install_wheel = virtualenv_env.copy()
+                        virtualenv_env_install_wheel["SYSTEM_VERSION_COMPAT"] = "0"
+                        log.notice(
+                            unwrap(
+                                """
+                                Setting SYSTEM_VERSION_COMPAT=0 to ensure x86_64 CPython 3.8 on
+                                arm64 runner can get correct macOS version and allow installation
+                                of wheels with MACOSX_DEPLOYMENT_TARGET >= 11.0.
+                                See https://github.com/pypa/cibuildwheel/issues/1767
+                                for the details.
+                                """
+                            )
+                        )
+                    else:
+                        virtualenv_env_install_wheel = virtualenv_env
+
                     call_with_arch(
                         "pip",
                         "install",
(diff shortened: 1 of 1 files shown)


joerick (commented):
I think you're modifying the wrong environment here, it should be the previous invocation labelled 'install the wheel'. Having said that it's probably worth setting the env var for this invocation too.

laggykiller:
Done, please check!

joerick (commented):
It would also be good to limit this to CPython 3.8, it's not required to do this on CPython 3.9+ and on 3.6 and 3.7 this is an error that users will also get, so it's worth alerting packagers to.

laggykiller:
> on 3.6 and 3.7 this is an error that users will also get, so it's worth alerting packagers to.

Sorry can you clarify? Do you mean on python 3.6 and 3.7, we also have to set SYSTEM_VERSION_COMPAT=0 as user would also encounter the same problem (e.g. On arm64 runner building x86_64 wheels, triggering the problem when they install wheel with target >=11.0)? Or only set SYSETEM_VERSION_COMPAT=0 on python 3.8?

henryiii:
Python 3.6 and 3.7 were never ported to ARM on macOS. (3.8 was only partially ported).

laggykiller:
@henryiii theoretically, can someone crosscompile cp36/cp37 wheel on arm64 macos runner, and trigger the bug we are discussing? (Probably not likely if anyone would actually do this irl though)

laggykiller:
Done adding the additional check, please review.

laggykiller:
Turns out the bug can indeed occur if someone crosscompile cp36/37 x86_64 wheel son arm64 macos runner: https://github.com/laggykiller/rlottie-python/actions/runs/8086369591/job/22095975016

I have changed to checking if python version <= 3.8, please review.

laggykiller:
> the same issue would happen on an x86_64 runner too

Can confirm about this

cibuildwheel/macos.py
```diff
@@ -579,17 +579,41 @@ def build(options: Options, tmp_path: Path) -> None:
                         shell_with_arch(before_test_prepared, env=virtualenv_env)
 
                     # install the wheel
+                    if (
+                        sys.version_info <= (3, 8)
+                        and testing_arch == "x86_64"
+                        and python_arch == "x86_64"
+                    ):
```
joerick, line 586:
```suggestion
                    if is_cp38 and python_arch == "x86_64":
```

sys.version_info is the wrong Python version, we want the builder version, not the host

I also don't think we should do the workaround for versions other than 3.8, as users of these wheels probably won't be able to install them (without setting SYSTEM_VERSION_COMPAT themselves). Wheel maintainers should be made aware of this.

Additionally, `testing_arch == "x86_64"` is redundant when `python_arch == "x86_64"`.

cibuildwheel/macos.py
```diff
@@ -579,17 +579,41 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                        virtualenv_env_install_wheel = virtualenv_env.copy()
+                        virtualenv_env_install_wheel["SYSTEM_VERSION_COMPAT"] = "0"
+                        log.notice(
+                            unwrap(
+                                """
+                                Setting SYSTEM_VERSION_COMPAT=0 to ensure x86_64 CPython <= 3.8 on
+                                arm64 runner can get correct macOS version and allow installation
+                                of wheels with MACOSX_DEPLOYMENT_TARGET >= 11.0.
+                                See https://github.com/pypa/cibuildwheel/issues/1767
+                                for the details.
```
joerick, line 596:
```suggestion
                                Setting SYSTEM_VERSION_COMPAT=0 to ensure CPython 3.8 can get 
                                correct macOS version and allow installation of wheels with 
                                MACOSX_DEPLOYMENT_TARGET >= 11.0. 
                                See https://github.com/pypa/cibuildwheel/issues/1767 for the 
                                details.
```

laggykiller:
All agreed and committed, please check!

joerick (approved):

cibuildwheel/macos.py
```diff
@@ -579,17 +579,37 @@ def build(options: Options, tmp_path: Path) -> None:
                         shell_with_arch(before_test_prepared, env=virtualenv_env)
 
                     # install the wheel
+                    if is_cp38 and python_arch == "x86_64":
```
alexlancaster, line 582: It looks like this change only fixed the MacOS builds for the `cp38` case.  My builds for `cp36` and `cp37`, e.g.: https://github.com/alexlancaster/pypop/actions/runs/9866159045/job/27244457994#step:4:298 fail.

I had to modify my `pyproject.toml` to include the `SYSTEM_VERSION_COMPAT=0` as a workround, but I feel this should be in cibuildwheel, and the comment here: https://github.com/pypa/cibuildwheel/pull/1768#issuecomment-1970914303 seems to imply that the fix should also apply to <`cp38` builds too, but it doesn't appear to.

joerick: Users of Python 3.6 and 3.7 can't install wheels that target macOS newer than 11.0. So you can build them, but users can't install them - well, they technically could, but they'd have to set SYSTEM_VERSION_COMPAT=0 too when they install them, otherwise pip will see the wheels as targeting a macOS that's newer than what's running.

alexlancaster: I see, but isn't it possible that Python 3.6 or 3.7 could be running on MacOS 12 or later? or did Python drop packages for those Python versions after 11.0?

joerick: See this comment for further explanation- https://github.com/pypa/cibuildwheel/pull/1768#issuecomment-1970914303

PR #1773, opened 2024-03-02
[draft] Sketched out an explicit QEMU mode

jonded94:
This is a very primitive sketch of forcing `cibuildwheel` to execute a given docker build container in the platform mode that corresponds to the wanted wheel architecture. This should only have an effect when doing non-native platform wheel builds through QEMU.

Possibly interesting since I noticed that all build containers run commands are given no further platform information, even if an image could not be executed natively and *has* to be run through QEMU. This should work because of the binfmt_misc feature of the linux kernel, but for some reason I saw this crash catastrophically in this issue in a Github Action: https://github.com/pypa/cibuildwheel/issues/1771.

This would also remove these warning messages that docker prints out:
```
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested
```

Note that this is just a sketch and not intended to be merged. Let me know what you think about this in general.

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -441,12 +441,16 @@ def build(options: Options, tmp_path: Path) -> None:  # noqa: ARG001
                         this_build_platform = docker_platform
                         break
                 else:
-                    raise ValueError(f"Unexpected build platform ’{build_step.platform_tag}'.")
+                    msg = f"Unexpected build platform: {build_step.platform_tag}."
+                    raise ValueError(msg)
 
                 container_engine_config = OCIContainerEngineConfig(
                     name=options.globals.container_engine.name,
-                    create_args=tuple(options.globals.container_engine.create_args)
-                    + ("--platform", this_build_platform),
+                    create_args=(
+                        *options.globals.container_engine.create_args,
+                        "--platform",
+                        this_build_platform,
+                    ),
                     disable_host_mount=options.globals.container_engine.disable_host_mount,
                 )
             else:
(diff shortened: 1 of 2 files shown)


joerick (commented):
Thanks for putting this together. my thoughts-

- no need to put this behind a new option, IMO. We can add the --platform always
- we shouldn't fiddle with the container_engine_config, instead, pass the architecture into OCIContainer, and map it as appropriate in there
- For arm64, I see you specified `linux/arm64/v8`, I wonder if we can be more vague and leave out the variant here and just say e.g. `linux/arm64`

jonded94:
Implemented your suggested changes. Unfortunately I'm getting failed tests in CircleCI when setting `--platform` everywhere:
```
"--platform" is only supported on a Docker daemon with experimental features enabled
```

joerick:
> Also, this is failing: https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=5690&view=logs&j=65f37ddb-1a35-539a-d91f-4ab7f01a34d0&t=677308e5-f586-5eb1-7f4b-3c9712904030&l=1767. Seems like a `i686` image is actually built in `amd64` mode?

Hmm. this kind of detail makes me less sure that manually specifying the platform is the right approach.

> ```
> "--platform" is only supported on a Docker daemon with experimental features enabled
> ```

It's not experimental since Docker API v1.32, guess Circle are lagging pretty bad there. But given this route is still in the process of rolling out, I think a better option would be for you to use `create_args` on `container-engine` for this. I can make a change to container-engine so it's overridable per-platform, which should make it easy to specify the right flags using overrides. It could be a useful feature for other use-cases too.

joerick:
It's now possible since #1792 to use `create_args` in `container-engine` for this, so I'll close this out for now. I'd like to add `--platform` across the board one day, but it would be better to wait until it's out of experimental status with all CI providers first.

PR #1778, opened 2024-03-05
chore: use uv pip compile

henryiii:
This cuts the time down from 40 seconds to 4 seconds, and doesn't require pre-installing every version of Python. It also is easy to run without nox too, since it's now just a simple command that doesn't require a Python version present.

It also adds a lock on importlib-metadata on older Pythons; build does require importlib-metadata on older Pythons, so guessing pip-tools just missed this?

This fixes running the noxfile with uv, too, and silences an incorrect warning from pylint.


henryiii:
The manylinux tags latest thing is broken on other PRs too.

joerick:
reported that issue over on manylinux- https://github.com/pypa/manylinux/issues/1581

henryiii:
Is it only manylinux1? We could just skip if this happens.

noxfile.py
```diff
@@ -60,32 +59,36 @@ def check_manifest(session: nox.Session) -> None:
...
+
+    if getattr(session.virtualenv, "venv_backend", "") != "uv":
+        session.install("uv")
+
+    for minor_version in range(7, 13):
+        python_version = f"3.{minor_version}"
+        env = os.environ.copy()
+        # CUSTOM_COMPILE_COMMAND is a pip-compile option that tells users how to
+        # regenerate the constraints files
+        env["CUSTOM_COMPILE_COMMAND"] = f"nox -s {session.name}"
```
joerick, line 76: Not critical, but it looks like this isn't having an effect on uv yet, found an issue here- https://github.com/astral-sh/uv/issues/1527

joerick (approved):
can we also specify in noxfile.py that nox should prefer uv, if it's available?

henryiii:
We can, but it would require the latest version of nox.

noxfile.py
henryiii: It's added now. Though IMO it's not really required, as unlike before, you don't need any special setup to compile it.

henryiii: It's called UV_CUSTOM_PIP_COMPILE, even though PR and release notes get it wrong. https://github.com/astral-sh/uv/pull/2554

PR #1783, opened 2024-03-08
ci: pin manylinux1 with missing tag

henryiii:
This should fix our CI while waiting on https://github.com/pypa/manylinux/issues/1581.


joerick (approved):
Thanks @henryiii.

PR #1785, opened 2024-03-11
Remove manylinux1 docker pin

joerick:
Reverts the workaround in #1783


joerick:
Kicked off another update task https://github.com/pypa/cibuildwheel/actions/runs/8230411204

Issue #1771, opened 2024-03-01
Linux ARM64 wheel build with QEMU fails because cibuildwheel tries execute container in x86_64 mode


jonded94:
### Description

I'm currently working on a pull request (https://github.com/chatnoir-eu/chatnoir-resiliparse/pull/34) enabling Linux ARM64 wheel builds for a certain package.

They use a special manylinux container since they need some other dependencies available in their build environment. I already prepared containers for both x86_64 and aarch64 that are available here (built from [this](https://github.com/jonded94/chatnoir-resiliparse/blob/implement-docker-arm-build/Dockerfile) Dockerfile and [this](https://github.com/jonded94/chatnoir-resiliparse/blob/implement-docker-arm-build/.github/workflows/build-docker.yml) workflow):
- https://github.com/jonded94/chatnoir-resiliparse/pkgs/container/resiliparse-manylinux_2_28_x86_64
- https://github.com/jonded94/chatnoir-resiliparse/pkgs/container/resiliparse-manylinux_2_28_aarch64

Trying both of these out locally works without a problem; I'm working on a Linux x86_64 machine and ran the aarch64 docker container through QEMU.

The wheel builds on Linux x86_64 work without a problem, but unfortunately the aarch64 one crashes: https://github.com/jonded94/chatnoir-resiliparse/actions/runs/8109883876/job/22165995057

```
Starting container image ghcr.io/jonded94/resiliparse-manylinux_2_28_aarch64...
  
  info: This container will host the build for cp38-manylinux_aarch64, cp39-manylinux_aarch64, cp310-manylinux_aarch64, cp311-manylinux_aarch64, cp312-manylinux_aarch64...
  Unable to find image 'ghcr.io/jonded94/resiliparse-manylinux_2_28_aarch64:latest' locally
  latest: Pulling from jonded94/resiliparse-manylinux_2_28_aarch64
  no matching manifest for linux/amd64 in the manifest list entries
```

[...]

PR #1792, opened 2024-03-15
Make `container-engine` a build (non-global) option

joerick:
Lets users specify different container-engine settings for different builds using TOML overrides. Especially useful with the create_args setting, such as #1771.


cibuildwheel/oci_container.py
```diff
@@ -71,6 +71,18 @@ def options_summary(self) -> str | dict[str, str]:
                 "disable_host_mount": str(self.disable_host_mount),
             }
 
+    def __eq__(self, value: object) -> bool:
```
henryiii, line 74: Why is this being defined? It looks like what data classes generate for you to me. It should at least have a commend explaining why this is required.

henryiii (commented):
This also needs to be updated in the schema. I can do that if it's helpful.

cibuildwheel/oci_container.py
joerick: hm. good point. I defined `__hash__` because it was needed for the build steps computation, so instinctively defined this too. Actually, if I just define the create_args as a tuple, I can get rid of both and let dataclass do it for me.

joerick:
I've had a go at that @henryiii . Pretty simple to do, thanks to the tooling. I don't know how to update the schemastore, if that's necessary? Actually, maybe that should be part of the release process?

henryiii (approved):

PR #1815, opened 2024-05-09
feat: add Python 3.13 beta 1

mayeut:
Let's try to start building with python 3.13


henryiii:
```
AssertionError: would build wheel with unsupported tag ('cp313', 'cpwin32', 'win32')
```

(Coming from `wheel`)

agronholm:
Tag parsing is done via the vendored `packaging` library, and I explicitly made the 0.43.0 release to incorporate that in `wheel`.

agronholm:
Can you confirm which version of `wheel` was even used? Where does the test suite install that?

.github/workflows/test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         pipx run nox -s pylint
 
   test:
-    name: Test cibuildwheel on ${{ matrix.os }}
+    name: Test cibuildwheel on ${{ matrix.os }} (${{ matrix.python_version }})
     needs: lint
     runs-on: ${{ matrix.os }}
     strategy:
```
henryiii, line 39:
```suggestion
    strategy:
      fail-fast: false
```

henryiii:
It seems to be only on Windows, and it's these four tests:

```
FAILED test/test_0_basic.py::test[pip] - subprocess.CalledProcessError: Command '['C:\\hostedtoolcache\\windows\\Python\\3.12.3\\x64\\python.exe', '-m', 'cibuildwheel', '--prerelease-pythons', '--output-dir', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpu5qo0yto', '.']' returned non-zero exit status 1.
FAILED test/test_before_all.py::test_cwd - subprocess.CalledProcessError: Command '['C:\\hostedtoolcache\\windows\\Python\\3.12.3\\x64\\python.exe', '-m', 'cibuildwheel', '--prerelease-pythons', '--output-dir', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpdb__fxb6', '.']' returned non-zero exit status 1.
FAILED test/test_before_build.py::test_cwd - subprocess.CalledProcessError: Command '['C:\\hostedtoolcache\\windows\\Python\\3.12.3\\x64\\python.exe', '-m', 'cibuildwheel', '--prerelease-pythons', '--output-dir', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpbhu69bgj', '.']' returned non-zero exit status 1.
FAILED test/test_before_all.py::test - subprocess.CalledProcessError: Command '['C:\\hostedtoolcache\\windows\\Python\\3.12.3\\x64\\python.exe', '-m', 'cibuildwheel', '--prerelease-pythons', '--output-dir', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpugwbyfy3', '.']' returned non-zero exit status 1.
```

I see a (3.12) venv seeded with 0.37.1, though the isolated build _shouldn't_ pick up this cached version from what I understand. I think it's only failing on Windows, I'm rerunning with fail-fast off to check, though.

mayeut:
The issue comes from cibuildwheel.
The environment is pre-seeded, there's no pin on setuptools / wheel anymore (https://github.com/pypa/cibuildwheel/commit/9879937c6374b3fa6c29a578be3170751d61b80b) but the project does not have a pyproject.toml so pre-seeded with outdated versions is used.

Reproduced there: https://github.com/mayeut/sandbox/actions/runs/9029260391

mayeut:
@henryiii, I'll force push and overwrite your commits with something else.

mayeut:
maybe we ought to get the fixes / changes in the last 3 commits in another PR to discuss those changes ?

henryiii:
Might make sense, yes. I don't think we should be injecting or pinning setuptools/wheel; if someone is building on Python 3.12 or 3.13 and doesn't have a pyproject.toml, I think they should be required to pip install setuptools and wheel in `CIBW_BEFORE_BUILD`. (Or add a pyproject.toml).

test/test_dependency_versions.py
```diff
@@ -61,9 +67,6 @@ def test_pinned_versions(tmp_path, python_version, build_frontend_env):
             pytest.skip("macOS arm64 does not support Python 3.6")
         constraint_filename = "constraints-python36.txt"
         build_pattern = "[cp]p36-*"
-    elif python_version == "3.7":
```
henryiii, line 64: Why is this removed?

mayeut: Because there is no parametrization on `python_version == "3.7"` and if there was one added, then the rest of the function does not account for it so I'd rather have either all or nothing and went with nothing.

mayeut: I reworked this differently in #1818, I'll resolve the conversation here. Let's discuss it in the new PR if need be.

henryiii (approved):

PR #1818, opened 2024-05-10
fix: respect constraints when building with pip

mayeut:
When building a project that has a `pyproject.toml`, constraints are not respected when building with the `pip` frontend.

This commit fixes this by using the same tricks as for the `build` frontend.


henryiii (approved):
I don't like constraints affecting isolated builds (how does one excite which of the half a dozen or so build backends are constrained?), but we should be consistent between pip and build.

PR #1819, opened 2024-05-10
fix: do not pre-seed setuptools / wheel in virtual environment

mayeut:
Since https://github.com/pypa/cibuildwheel/commit/9879937c6374b3fa6c29a578be3170751d61b80b, setuptools & wheel are not pinned/installed anymore.
We should not pre-seed the virtual environment with those either.

### DIFF
--- a/cibuildwheel/resources/constraints-python36.txt
+++ b/cibuildwheel/resources/constraints-python36.txt
@@ -48,3 +48,5 @@ zipp==3.6.0
 # The following packages are considered to be unsafe in a requirements file:
 pip==21.3.1
     # via -r cibuildwheel/resources/constraints.in
+setuptools==59.6.0
+    # via -r cibuildwheel/resources/constraints.in
--- a/cibuildwheel/util.py
+++ b/cibuildwheel/util.py
@@ -540,6 +540,7 @@ def _ensure_virtualenv() -> Path:
 
 
 def _parse_constraints_for_virtualenv(
+    seed_packages: list[str],
     dependency_constraint_flags: Sequence[PathOrStr],
 ) -> dict[str, str]:
     """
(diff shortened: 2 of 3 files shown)


henryiii (approved):

henryiii:
This is only in draft due to needing https://github.com/pypa/cibuildwheel/pull/1818 (merged)?

mayeut:
> this is only in draft due to needing https://github.com/pypa/cibuildwheel/pull/1818 (merged)?

yes

EwoutH:
Any chance [this error](https://github.com/scipy/scipy/actions/runs/10340013845/job/28620088192#step:7:508) has to do with this chance? It appeared when updating from cibuildwheel 2.17.0 to 2.20.0, and this PR was included in 2.18.0.

If so, what would be the best way to resolve it?

mayeut:
yes, it seems to be related.
you need to install your dependencies. for wheel cli, this can most likely run with pipx.

EwoutH:
I think we fixed it by simply installing it explicitly before build: https://github.com/scipy/scipy/commit/f20a0b4572ab153b3ac3d2c16d7b523bf8c200d3.

To confirm, it was intended that we now need to do this explicitly? (since that could be considered a breaking change, which would go against [semantic versioning](https://semver.org/) in a minor release)

mayeut:
This PR is a fix. What you were using was not intended to work after https://github.com/pypa/cibuildwheel/commit/9879937c6374b3fa6c29a578be3170751d61b80b. 2.17.0 probably should have been called 3.0.0 but it's too late for that.

EwoutH:
[image](https://xkcd.com/1172/)

No worries, it's always difficult how to handle fixes to unintended and undocumented behavior. Thanks for getting back and explaining the situation :)

henryiii:
Cibuildwheel does not follow SemVer (many other Python libraries do not, including Python itself). That’s why we don’t allow you to pin to a major version in GHA, but only to a minor version. We do try very hard not to break anyone in patch versions. Most projects will not break in minor versions, but we add and remove python versions, which definitely can break your workflow. If you don’t support Python 3.13, 2.20 is a breaking release. If you need Python 3.6, I’m sure we will eventually drop it. Etc.

I would consider this basically a user bug; if you needed wheel, you should’ve been making sure it is installed. Modern setuptools doesn’t even need wheel anymore. It looks like you were doing exactly the right thing, though, pinning to a minor version and then updating regularly!

PS: in general, please do not rely on setuptools or wheel being present! They should be treated like any other dependency. They need to be removed from the default environments; in some places they were removed in 3.12 (venv and virtualenv), and some other places in 3.13 (manylinux), but it is happening! It speeds up environment creation if they aren’t needed, and gives the user more version freedom if they are.

PR #1821, opened 2024-05-12
docs: fix options page

henryiii:
The options page is broken, and mkdocs "helpfully" hides errors and just puts them on the output page!

MkDocs macros breaks if the Python environment isn't activated. ¯\_(ツ)_/¯

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,5 +11,5 @@ build:
     - asdf install uv latest
     - asdf global uv latest
     - uv venv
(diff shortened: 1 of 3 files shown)


henryiii:
I was able to get the `stable` readthedocs to rebuild by making a branch for a couple of minutes called `stable`. Should go back to the latest release automatically after the next release.

henryiii:
Ahh, we'll have to keep a `stable` branch to get readthedocs to show the right docs until we make a patch/minor release.

PR #1829, opened 2024-05-18
fix(tests): linux tests from macOS/Windows arm64

mayeut:
When running on arm64, macOS reports machine as `arm64` and Windows as `ARM64`. We need to override this to `aarch64` when running linux tests.


henryiii (approved):

mayeut:
I'll merge this without waiting for travis ci status (which is now limited to 1 concurrent build...).

PR #1830, opened 2024-05-18
chore: un-pin virtualenv update

mayeut:
virtualenv has been pinned to 20.21.1 in order to be able to create python 3.6 venvs.

This commit allows to pin virtualenv for given versions of pythons which in turns re-allows updates as a default.

### DIFF
--- a/test/test_dependency_versions.py
+++ b/test/test_dependency_versions.py
@@ -111,7 +111,6 @@ def test_dependency_constraints_file(tmp_path, build_frontend_env):
             """
             pip=={pip}
             delocate=={delocate}
-            importlib-metadata<3,>=0.12; python_version < "3.8"
             """.format(**tool_versions)
         )
     )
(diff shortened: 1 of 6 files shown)


mayeut:
converting to draft while I debug Windows builds.

test/test_dependency_versions.py
```diff
@@ -111,7 +111,6 @@ def test_dependency_constraints_file(tmp_path, build_frontend_env):
             """
             pip=={pip}
             delocate=={delocate}
-            importlib-metadata<3,>=0.12; python_version < "3.8"
```
mayeut, line 114: Failing tests on windows came from this pin.
I'm not sure exactly why since I can't really debug on Windows.
All I can say is that it was causing some backtracking in the pip resolver in order to install an old version of virtualenv (2.16.2).

henryiii (approved):

Issue #1002, opened 2022-01-28
Pyodide Wheels


DaAwesomeP:
### Description

I'm not sure if this is out of the scope of this project, but the ability to build Pyodide wheels would be wonderful: https://pyodide.org/en/stable/development/new-packages.html

henryiii:
The numpy.org webpage now has a pyodide-powered scientific stack in the browser. Wow.

I have no idea ATM what is involved with building a pyodide wheel, and I'd assume there really aren't "platforms" for them, and I think a lot would likely be needed upstream to standardize tags and such, and cibuildwheel likely _might_ not be the place to produce them, but I really have no idea. I don't even know if they are in traditional files. If it was something cibuildwheel could do, it would probably be a new "platform".

henryiii:
You can see this mentioned in the roadmap here: https://pyodide.org/en/stable/project/roadmap.html

https://github.com/pyodide/pyodide/pull/655 & https://github.com/pyodide/pyodide/pull/2027 are linked there. Both of which seem to be merged.

DaAwesomeP:
> The numpy.org webpage now has a pyodide-powered scientific stack in the browser. Wow.

I know, it's pretty cool!

> Both of which seem to be merged.

Yeah, I was under the impression that they are already using wheels. This documentation seems to hint that you can create wheels: https://pyodide.org/en/stable/development/new-packages.html

henryiii:
Yes, they do make wheels now. Looking into the process. I just found that page, actually. :)

hoodmane:
For the platform compatibility question see https://github.com/emscripten-core/emscripten/issues/15917

henryiii:
Pretty sure there’s no way to load the a non pure wheel from anything but pyodide’s own build process at the moment. I was able to compile for it though locally (took a couple of hours to build CPython and all)

https://github.com/pyodide/pyodide/discussions/2167

hoodmane:
If you build a compatible non-pure wheel in some other way, it's possible to load it with `pyodide.loadPackage(url)`.

DaAwesomeP:
Maybe more feasible and makes more sense to implement, this is now beginning to be added as an official architecture in CPython: https://github.com/python/cpython/blob/main/Tools/wasm/README.md

rth:
[...]

PR #1831, opened 2024-05-18
feat: add support for free-threaded (no-gil) Python 3.13

mayeut:
Support has been added for Linux & Windows awaiting macOS installers (IMHO, not blocking to ship this experimental feature)

Users have to opt-in this feature using the new `free-threaded-support` option of the config file (probably preferable over the environment variable which is supported as well).


henryiii:
Nice! Are we doing to go ahead and ship experimental support for Linux & Windows early, or wait on macOS builds with beta 2? I also think macOS might be tricky, as I'm not what it will look like installing both, I guess they aren't supposed to collide? (Maybe that's why it's taking till beta 2!)

cibuildwheel/windows.py
```diff
@@ -224,18 +229,21 @@ def setup_python(
...
     if implementation_id.startswith("cp"):
         native_arch = platform_module.machine()
+        free_threaded = implementation_id.endswith("t")
         if python_configuration.arch == "ARM64" != native_arch:
             # To cross-compile for ARM64, we need a native CPython to run the
             # build, and a copy of the ARM64 import libraries ('.\libs\*.lib')
             # for any extension modules.
             python_libs_base = install_cpython(
-                python_configuration.version, python_configuration.arch
+                python_configuration.version, python_configuration.arch, free_threaded
```
mayeut, line 238: Maybe we should just pass the `python_configuration.identifier` here and let `install_cpython` deal with this ?
If other things than free threaded (e.g. debug, debug+ft, ...), It might be easier to deal with at a lower level ?
Thoughts ?

mayeut:
> Are we doing to go ahead and ship experimental support for Linux & Windows early, or wait on macOS builds with beta 2? I also think macOS might be tricky, as I'm not what it will look like installing both, I guess they aren't supposed to collide? (Maybe that's why it's taking till beta 2!)

I think we should ship experimental support without waiting on macOS.
The only thing that's missing is the opt-in mechanism & I think it should ship with the first version allowing free threaded builds (c.f. https://github.com/pypa/cibuildwheel/issues/1657#issuecomment-2119140641)

As for how the macOS build will turn out, the only info I found in the cpython repo was https://github.com/python/cpython/issues/112430#issuecomment-1939514000 (didn't find anything on discourse, there might be some other threads out there).
If I understand correctly, it will be an installation option. This probably means we'll have to rework the macOS bit to force a re-install (only in CI of course) if the free-threaded build is missing in an existing installation. But it might get trickier:
> Ideally, we'd want to allow both variants to be able to to be installed side-by-side; while that should mostly work today, there are some loose ends that would need to be taken care of involving path names and the like. If it turns out to be too disruptive to support co-existing installs, the installer package could simply inhibit the installation of the normal variant if the user selects the optional free-threaded variant.

cibuildwheel/windows.py
```diff
@@ -519,6 +524,10 @@ def build(options: Options, tmp_path: Path) -> None:
                 # check that we are using the Python from the virtual environment
                 call("where", "python", env=virtualenv_env)
 
+                # TODO remove me once virtualenv provides pip>=24.1b1
+                if config.version.startswith("3.13."):
+                    call("python", "-m", "pip", "install", "--pre", "-U", "pip", env=virtualenv_env)
```
mayeut, line 529: This does not work with `call("python", "-m", "pip", "install", "pip>=24.1b1", env=virtualenv_env)`
Maybe something to do with `shell=True` being used on Windows ?

cibuildwheel/windows.py
mayeut: I went with `python_configuration`, so marking as resolved.

henryiii:
What if we swap the default, at least while it’s under a beta flag, then decide if we keep it opt-out instead of opt-in when the RC drops?

henryiii:
@ned-deily, will it be possible to install both variants at the same time on macOS like on windows/linux?

henryiii (approved):

henryiii:
Should we call this 2.17.1 or 2.18.0? It's experimental and opt-in, so I'm slightly biased toward the first.

mayeut:
with #1792, we might want to go with 2.19.0 anyway ?

henryiii:
Yes, possibly. Though I think that's also only opt-in, something that was disallowed before is now allowed, but existing builds don't change.

It's sad that #1002 isn't quite ready (even if it was finished, still waiting on the pyodide release, which will likely be delayed I'd expect with some wheel tag changes planned). GraalPy might not be too far away, etc. Minor version bumps require everyone to update their two-digit pins. In that sense, I'm not sure this is a minor bump, as nothing changed or is added without user changes.

mayeut:
> I'm not sure this is a minor bump, as nothing changed or is added without user changes.

Let's go with 2.18.1 then ! Maybe being too pedantic regarding semver is not a good idea here.

> GraalPy might not be too far away

This one is waiting at least on pip 24.1

I haven't looked in the pyodide one in a long time but saw there were some commits.

henryiii:
> This one is waiting at least on pip 24.1

Well, technically free-threading was too. ;)

henryiii:
Not going to bother with Travis, it's not even started the job.

mayeut:
> Well, technically free-threading was too. ;)

yes, that's true. but bumping pip to a beta version for cp 3.13 is definitely acceptable (all is in BETA) while bumping pip for python 3.10 is probably less so (and I don't want to add specific code for this, personal opinion)

ned-deily:
> install both variants at the same time on macOS 

Yes

ned-deily:
FYI: 3.13.0b2 is now released (https://discuss.python.org/t/3-13-0b2-now-available/55056) and the macOS installer provided for it on python.org now (finally!) includes an optional and experimental python3.13t free-threaded build along with the traditional build. We are using https://github.com/python/cpython/issues/120098 to track macOS installer-specific issues.

henryiii:
Thanks, great work!

Ohh, 10.13 minimum, now too! :heart:

henryiii:
@ned-deily The builds are not visible at https://www.python.org/api/v2/downloads/release/?is_published=true yet (https://www.python.org/api/v2/downloads/release/?is_published=true&name=Python%203.13.0b2 is empty). It's also not showing up at https://www.python.org/downloads/macos/ for example. It does show up here: https://www.python.org/download/pre-releases/. We use the first link above to update our Python versions, so it would be nice if those update, is that planned? I'd rather not manually tinker on our update scripts if it's going to fix itself soon.

ned-deily:
> The builds are not visible at https://www.python.org/api/v2/downloads/release/?is_published=true yet (https://www.python.org/api/v2/downloads/release/?is_published=true&name=Python%203.13.0b2 is empty). It's also not showing up at https://www.python.org/downloads/macos/ for example. It does show up here: https://www.python.org/download/pre-releases/. We use the first link above to update our Python versions, so it would be nice if those update, is that planned? I'd rather not manually tinker on our update scripts if it's going to fix itself soon.

@Yhg1s ?

Yhg1s:
> @Yhg1s ?

Should be visible now.

henryiii:
Superb, thanks!

PR #1834, opened 2024-05-25
chore(tests): do not use hard-coded paths in tests

mayeut:
This would help reduce the amount of specific code for pyodide in #1456

### DIFF
--- a/test/test_before_build.py
+++ b/test/test_before_build.py
@@ -14,26 +14,21 @@ project_with_before_build_asserts = test_projects.new_c_project(
 
         # assert that the Python version as written to pythonversion_bb.txt in the CIBW_BEFORE_BUILD step
         # is the same one as is currently running.
-        version_file = 'c:\\pythonversion_bb.txt' if sys.platform == 'win32' else '/tmp/pythonversion_bb.txt'
-        with open(version_file) as f:
+        with open('pythonversion_bb.txt') as f:
             stored_version = f.read()
         print('stored_version', stored_version)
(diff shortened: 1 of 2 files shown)


mayeut:
@henryiii, would you mind reviewing this one ?

I don't mind waiting for pyodide PR to land first but I'd rather have it the other way around though (and I'm willing to fix conflicts there if it helps).

test/test_before_build.py
```diff
@@ -14,26 +14,21 @@
...
+        #  Works around path-comparison bugs caused by short-paths on Windows e.g.
+        #  vssadm~1 instead of vssadministrator
 
-        # windows/mac are case insensitive
-        stored_path = os.path.realpath(stored_executable).lower()
-        current_path = os.path.realpath(sys.executable).lower()
-
-        # TODO: This is not valid in an virtual environment
-        assert stored_path == current_path, '{0} != {1}'.format(stored_path, current_path)
+        assert os.stat(stored_prefix) == os.stat(sys.prefix)
```
henryiii, line 31: Why not os.path.samefile?

henryiii (approved):

test/test_before_build.py
henryiii: There is also at least one more existing os.stat compare that would read better this way, imo. But could be a follow up because this matches the existing one(s).

mayeut: This check was already used in test_before_test.py so I reused it here.
I will change it for samefile which seems better suited.

PR #1835, opened 2024-05-25
chore: rework single python tests

mayeut:
Seeing the many tests that were changed in #1456 to move from one python version to another, we might want a helper for that so that we can do the change in a single location.

Using a single python version is extended to the following tests in order to reduce CI load:
- test_before_all.py
- test_before_build.py:test_cwd
- test_build_skip.py (by updating the skip pattern for the build)
- test_cpp_standards.py (even though "python version independent" wasn't quite true for PyPy at some point)
- test_environment.py
- test_musllinux_X_Y_only.py
- test_testing.py: test_extras_require

There are also 1 other commits to reduce the load on CI for [test_abi3](https://github.com/pypa/cibuildwheel/pull/1835/commits/29e56b5344a0d7b9ca1882db7ceee0b5fba58da3).

### DIFF
--- a/test/test_abi_variants.py
+++ b/test/test_abi_variants.py
@@ -20,7 +20,7 @@ limited_api_project = test_projects.new_c_project(
         CAN_USE_ABI3 = IS_CPYTHON and not Py_GIL_DISABLED
         cmdclass = {}
         extension_kwargs = {}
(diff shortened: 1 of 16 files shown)


henryiii (approved):

mayeut:
I'll apply the same review suggestion as in #1834 (I re-used the same pattern once more in `test/test_before_all.py`)

PR #1838, opened 2024-05-27
chore: better determinism for the test virtual environment

mayeut:
While reviewing #1456, the way that the test virtual environment was seeded was bugging me.

We're not really enforcing anything other than "do not download something new" but, it did not prevent using some version already in virtualenv cache (unlikely in CI but still possible, more likely for local builds).

This PR ensures that the test virtual environment gets seeded with determinism w.r.t. pip, setuptools & wheel versions.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from typing import Literal, Tuple
 
 from filelock import FileLock
+from packaging.version import Version
 
 from ._compat.typing import assert_never
 from .architecture import Architecture
@@ -359,6 +360,7 @@ def build(options: Options, tmp_path: Path) -> None:
                 build_options.environment,
                 build_frontend.name,
             )
+            pip_version = get_pip_version(env)
 
             compatible_wheel = find_compatible_wheel(built_wheels, config.identifier)
             if compatible_wheel:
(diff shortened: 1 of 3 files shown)


cibuildwheel/macos.py
```diff
@@ -560,9 +562,12 @@ def build(options: Options, tmp_path: Path) -> None:
                     call_with_arch = functools.partial(call, *arch_prefix)
                     shell_with_arch = functools.partial(call, *arch_prefix, "/bin/sh", "-c")
 
-                    # Use --no-download to ensure determinism by using seed libraries
-                    # built into virtualenv
-                    call_with_arch("python", "-m", "virtualenv", "--no-download", venv_dir, env=env)
+                    # Use pip version from the initial env to ensure determinism
+                    venv_args = ["--no-periodic-update", f"--pip={pip_version}"]
+                    # In Python<3.12, setuptools & wheel are installed as well, use virtualenv embedded ones
+                    if Version(config.version) < Version("3.12.0a0"):
```
henryiii, line 568: Why 3.12.0a0? 3.12 would be fine, we don’t have pre-releases for 3.12 and this is just a two element version anyway?

mayeut: It's just a habit to exclude pre-release also, but yeah, I can change that.

henryiii (approved):

PR #1841, opened 2024-05-28
chore: remove pip <21.3 workaround

henryiii:
This shouldn't be needed anymore.


mayeut (approved):

mayeut:
Now that virtualenv is used to create build environments, it shouldn't be needed.

Note: pip version embedded in python 3.9 on windows still uses pip < 21.3

PR #1842, opened 2024-05-28
fix: include VIRTUAL_ENV variable

henryiii:
Include the `VIRTUAL_ENV` variable when running "in" a virtualenv. This will be useful when we start trying out uv.

### DIFF
--- a/test/test_testing.py
+++ b/test/test_testing.py
@@ -58,6 +58,7 @@ class TestSpam(TestCase):
 
         self.assertTrue(path_contains(sys.prefix, sys.executable))
         self.assertTrue(path_contains(sys.prefix, spam.__file__))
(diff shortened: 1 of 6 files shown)


mayeut (approved):

test/test_testing.py
```diff
@@ -58,6 +58,7 @@ def test_virtualenv(self):
 
         self.assertTrue(path_contains(sys.prefix, sys.executable))
         self.assertTrue(path_contains(sys.prefix, spam.__file__))
+        self.assertIn("VIRTUAL_ENV", os.environ)
```
mayeut, line 61: Just thinking after I've merged the PR...
Should a test like this one be added for build environments as well ? (except on Linux)

PR #1851, opened 2024-06-03
Handle case where output_dir does not already exist on macos & windows

### LINKED ISSUE #1850: Macos-14 builds fail if output_dir does not already exist AND testing is enabled
### Description

Building for Macos14 (in github actions = arm64) fails when attempting to move the wheel after testing if the output directory does not already exist.

Contrary to the documentation `pathlib.Path.unlink()` throws a `NotADirectoryError` in this case, and apparently only on this platform. (I will raise an Issue in cpython for that as well - at least to update the docs ...)

Compare:
- Failed run: https://github.com/MusicalNinjaDad/FizzBuzz/actions/runs/9351587993/job/25744497746
```
============================== 11 passed in 0.05s ==============================
    File "/Users/runner/work/_temp/cibw/bin/cibuildwheel", line 8, in <module>
      sys.exit(main())
               ^^^^^^
    File "/Users/runner/work/_temp/cibw/lib/python3.12/site-packages/cibuildwheel/__main__.py", line 143, in main
      build_in_directory(args)
    File "/Users/runner/work/_temp/cibw/lib/python3.12/site-packages/cibuildwheel/__main__.py", line 315, in build_in_directory
      platform_module.build(options, tmp_path)
    File "/Users/runner/work/_temp/cibw/lib/python3.12/site-packages/cibuildwheel/macos.py", line 643, in build
      (build_options.output_dir / repaired_wheel.name).unlink()
    File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 1342, in unlink
      os.unlink(self)
  NotADirectoryError: [Errno 20] Not a directory: '/Users/runner/work/FizzBuzz/FizzBuzz/wheelhouse/fizzbuzz-2.1.0-cp39-cp39-macosx_11_0_arm64.whl'
  Error: Process completed with exit code 1.
```
- Previously passed (no tests): https://github.com/MusicalNinjaDad/FizzBuzz/actions/runs/9108520076/job/25039459108
- Subsequently passed with workaround (added `&& mkdir wheelhouse` to `tool.cibuildwheel.macos.before-all`): https://github.com/MusicalNinjaDad/FizzBuzz/actions/runs/9354284188/job/25746806904

### Build log

https://github.com/MusicalNinjaDad/FizzBuzz/actions/runs/9351587993/job/25744497746

### CI config

https://github.com/MusicalNinjaDad/FizzBuzz/tree/pr48

MusicalNinjaDad:
### [Updated 2024-06-04]

In cases where the output_dir does not already exist, the attempt to move the tested wheel to the output_dir actually ends up creating a file named `output_dir` rather than `outputdir/wheelname.whl`

This leads to the build process failing on macos and windows if the directory does not exist, or is removed as part of a before-build clean.

This PR updates the handling of the final wheels to:
1. explicitly resolve the output directory path
2. explicitly create the output directory if it does not already exist
3. then move the wheel file

It also cleans up the handling of errors by making use of built in pathlib.Path functionality rather than with contextlib.suppress

It has been tested in Github actions CI against a specific project & commit which provided a consistent, reproducible error described in #1850.

Fixes #1850 

### Original
Pathlib apparently raises a `NotADirectoryError` instead of a `FileNotFoundError` on macos. This causes builds to fail after testing if the output_dir does not already exist.

This PR fixes #1850 by using `missing_ok=True` introduced with pathlib 3.8 and trusting that Pathlib are able to handle the different errors provided on different OS.

For the sake of consistency it also updates windows and pyodide to follow the same pattern at the same point in the code.

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -641,8 +641,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                with contextlib.suppress(FileNotFoundError):
-                    (build_options.output_dir / repaired_wheel.name).unlink()
+                (build_options.output_dir / repaired_wheel.name).unlink(missing_ok=True)
 
                 shutil.move(str(repaired_wheel), build_options.output_dir)
                 built_wheels.append(build_options.output_dir / repaired_wheel.name)
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -640,7 +640,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                (build_options.output_dir / repaired_wheel.name).unlink(missing_ok=True)
+                build_options.output_dir.joinpath(repaired_wheel.name).unlink(missing_ok=True)
 
                 shutil.move(str(repaired_wheel), build_options.output_dir)
                 built_wheels.append(build_options.output_dir / repaired_wheel.name)
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import contextlib
 import functools
 import os
 import platform
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -36,6 +36,7 @@ from .util import (
     get_build_verbosity_extra_flags,
     get_pip_version,
     install_certifi_script,
+    move_file,
     prepare_command,
     read_python_configs,
     shell,
@@ -640,16 +641,10 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                output_dir = build_options.output_dir.resolve()
-                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
-                output_wheel.unlink(missing_ok=True)
-
-                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
-                output_dir.mkdir(parents=True, exist_ok=True)
-
-                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
-                # explicit str() needed for Python 3.8
-                shutil.move(str(repaired_wheel), str(output_wheel))
+                output_wheel = build_options.output_dir.joinpath(repaired_wheel.name)
+                moved_wheel = move_file(repaired_wheel, output_wheel)
+                if moved_wheel != output_wheel.resolve():
+                    log.warning("{repaired_wheel} was moved to {moved_wheel} instead of {output_wheel}")
                 built_wheels.append(output_wheel)
 
             # clean up
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -640,10 +640,16 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                build_options.output_dir.joinpath(repaired_wheel.name).unlink(missing_ok=True)
+                output_dir = build_options.output_dir.resolve()
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
 
-                shutil.move(str(repaired_wheel), build_options.output_dir)
-                built_wheels.append(build_options.output_dir / repaired_wheel.name)
+                # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+                
+                # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                shutil.move(repaired_wheel, output_wheel)
+                built_wheels.append(output_wheel)
 
             # clean up
             shutil.rmtree(identifier_tmp_dir)
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -644,13 +644,13 @@ def build(options: Options, tmp_path: Path) -> None:
                 output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
                 output_wheel.unlink(missing_ok=True)
 
-                # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
                 output_dir.mkdir(parents=True, exist_ok=True)
 
-                # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
                 shutil.move(
                     str(repaired_wheel), str(output_wheel)
-                )  # explicit str() needed for mypy3.8
+                )  # explicit str() needed for Python 3.8 - can change to Paths when we drop 3.8 support
                 built_wheels.append(output_wheel)
 
             # clean up
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -648,9 +648,8 @@ def build(options: Options, tmp_path: Path) -> None:
                 output_dir.mkdir(parents=True, exist_ok=True)
 
                 # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
-                shutil.move(
-                    str(repaired_wheel), str(output_wheel)
-                )  # explicit str() needed for Python 3.8 - can change to Paths when we drop 3.8 support
+                # explicit str() needed for Python 3.8
+                shutil.move(str(repaired_wheel), str(output_wheel))
                 built_wheels.append(output_wheel)
 
             # clean up
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -646,7 +646,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
                 # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
                 output_dir.mkdir(parents=True, exist_ok=True)
-                
+
                 # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
                 shutil.move(repaired_wheel, output_wheel)
                 built_wheels.append(output_wheel)
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -648,7 +648,7 @@ def build(options: Options, tmp_path: Path) -> None:
                 output_dir.mkdir(parents=True, exist_ok=True)
 
                 # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
-                shutil.move(repaired_wheel, output_wheel)
+                shutil.move(str(repaired_wheel), str(output_wheel)) # explicit str() needed for mypy3.8
                 built_wheels.append(output_wheel)
 
             # clean up
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -648,7 +648,9 @@ def build(options: Options, tmp_path: Path) -> None:
                 output_dir.mkdir(parents=True, exist_ok=True)
 
                 # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
-                shutil.move(str(repaired_wheel), str(output_wheel)) # explicit str() needed for mypy3.8
+                shutil.move(
+                    str(repaired_wheel), str(output_wheel)
+                )  # explicit str() needed for mypy3.8
                 built_wheels.append(output_wheel)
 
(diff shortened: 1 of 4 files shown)


MusicalNinjaDad (commented):
Note: I am not able to run integration tests on Macos, windows or pyodide as it is not my native system. If you would like me to try to ramp up a dedicated test project and github actions workflow to test this change, I can ...

cibuildwheel/windows.py
```diff
@@ -543,8 +542,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (remove if already exists)
             if compatible_wheel is None:
-                with suppress(FileNotFoundError):
-                    (build_options.output_dir / repaired_wheel.name).unlink()
+                (build_options.output_dir / repaired_wheel.name).unlink(missing_ok=True)
```
henryiii, line 545:
```suggestion
                build_options.output_dir.joinpath( repaired_wheel.name).unlink(missing_ok=True)
```

Maybe this style is marginally nicer?

henryiii:
> pyodide as it is not my native system

Pyodide isn't anyone's native system. :)

No worry about running tests locally, that's what CI is for.

cibuildwheel/macos.py
```diff
@@ -641,8 +640,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                with contextlib.suppress(FileNotFoundError):
-                    (build_options.output_dir / repaired_wheel.name).unlink()
+                (build_options.output_dir / repaired_wheel.name).unlink(missing_ok=True)
```
henryiii, line 643:
```suggestion
                build_options.output_dir.joinpath(repaired_wheel.name).unlink(missing_ok=True)
```

cibuildwheel/pyodide.py
```diff
@@ -395,8 +394,7 @@ def build(options: Options, tmp_path: Path) -> None:
 
             # we're all done here; move it to output (overwrite existing)
             if compatible_wheel is None:
-                with contextlib.suppress(FileNotFoundError):
-                    (build_options.output_dir / repaired_wheel.name).unlink()
+                (build_options.output_dir / repaired_wheel.name).unlink(missing_ok=True)
```
henryiii, line 397:
```suggestion
                build_options.output_dir.joinpath(repaired_wheel.name).unlink(missing_ok=True)
```

henryiii (approved):
I think this is a nice clean up, and I'm not sure it fixes the problem (as I can't reproduce exactly), but I like the cleanup anyway, and a chance to fix a problem is nice too.

MusicalNinjaDad:
Thanks henryii,

I'm going to put together a quick CI in my fork which tries building the offending repo to check whether this actually works to solve the issue. (Which also hit windows after the workaround worked for macos ...)

henryiii:
You should be able to just change the cibuildwheel install/use to this PR. Though you might also try rebuilding to see if it a constant failure, or an intermittent one.

"Hit Windows" - does that mean this workaround hit Windows, or without the workaround hit Windows? It's a little troubling if the workaround causes failures on Windows...

MusicalNinjaDad:
The issue (without workaround) hit windows.

It's consistent - I've reproduced it here: https://github.com/MusicalNinjaDad/cibuildwheel/actions/runs/9361752026/job/25769267931

I only added `mkdir wheelhouse` to `macos.before-all`

Looking at it now - it's actually a different NotADirectoryError on the win setup ...
I'll take a look at that once I've confirmed this fixes macos

MusicalNinjaDad:
OK - so my debugging found the issue (and then I saw [serhiy-storchaka](https://github.com/serhiy-storchaka) comment on python/cpython#119993

I've got a fix done, currently final test ongoing. Then I just need to remove the debug logging and push it to this branch ...

cibuildwheel/macos.py
```diff
@@ -641,11 +640,18 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
+
+                # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+
+                # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                shutil.move(
+                    str(repaired_wheel), str(output_wheel)
+                )  # explicit str() needed for mypy3.8
```
henryiii, line 653:
```suggestion
                )  # explicit str() needed for Python 3.8
```

https://docs.python.org/3/library/shutil.html#shutil.move

cibuildwheel/macos.py
```diff
@@ -641,11 +640,18 @@ def build(options: Options, tmp_path: Path) -> None:
...
-                shutil.move(str(repaired_wheel), build_options.output_dir)
-                built_wheels.append(build_options.output_dir / repaired_wheel.name)
+                output_dir = build_options.output_dir.resolve()
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
+
+                # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+
+                # using os.move() as Path.rename() is not guaranteed to work across filesystem boundaries
```
henryiii, line 650:
```suggestion
                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
```

cibuildwheel/macos.py
```diff
@@ -641,11 +640,18 @@ def build(options: Options, tmp_path: Path) -> None:
...
-                with contextlib.suppress(FileNotFoundError):
-                    (build_options.output_dir / repaired_wheel.name).unlink()
-
-                shutil.move(str(repaired_wheel), build_options.output_dir)
-                built_wheels.append(build_options.output_dir / repaired_wheel.name)
+                output_dir = build_options.output_dir.resolve()
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
+
+                # os.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
```
henryiii:
```suggestion
                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
```

cibuildwheel/macos.py
MusicalNinjaDad: Thanks Henry! (This is why I don't work without tests, pipelines and linters in place ... and even then they don't check the comments.)

cibuildwheel/macos.py
```diff
@@ -641,11 +640,18 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
+
+                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+
+                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                shutil.move(
+                    str(repaired_wheel), str(output_wheel)
+                )  # explicit str() needed for Python 3.8 - can change to Paths when we drop 3.8 support
```
henryiii, line 653:
```suggestion
                # explicit str() needed for Python 3.8
                shutil.move(str(repaired_wheel), str(output_wheel))
```

cibuildwheel/windows.py
```diff
@@ -543,11 +542,18 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                output_wheel = output_dir.joinpath(repaired_wheel.name).resolve()
+                output_wheel.unlink(missing_ok=True)
+
+                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+
+                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                shutil.move(
+                    str(repaired_wheel), str(output_wheel)
+                )  # explicit str() needed for Python 3.8 - can change to Paths when we drop 3.8 support
```
henryiii, line 555:
```suggestion
                # explicit str() needed for Python 3.8
                shutil.move(str(repaired_wheel), str(output_wheel))
```

henryiii:
I moved the comments, trailing comments cause this to take 3 lines vs. 2. Also removed the comment about removing, as that's implied IMO anytime you mention a Python version in a comment. I always do something like `git grep -E "Python 3\.[89]"` to look for these sorts of comments.

cibuildwheel/macos.py
```diff
@@ -641,11 +640,17 @@ def build(options: Options, tmp_path: Path) -> None:
...
 
-                shutil.move(str(repaired_wheel), build_options.output_dir)
-                built_wheels.append(build_options.output_dir / repaired_wheel.name)
+                # shutil.move() will rename the file to what we were expecting to be the parent directory if we don't ensure it exists
+                output_dir.mkdir(parents=True, exist_ok=True)
+
+                # using shutil.move() as Path.rename() is not guaranteed to work across filesystem boundaries
+                # explicit str() needed for Python 3.8
+                shutil.move(str(repaired_wheel), str(output_wheel))
+                built_wheels.append(output_wheel)
```
mayeut, line 653: IMHO, we should add a `move` function  in `util.py` which takes care of unlinking, creating the parent directory and actually moving the wheel. It could be reused on all 3 platforms (workaround might be needed for local builds on macOS with pyodide for exemple).

MusicalNinjaDad: Hi mayeut,
That's a good idea.
Also a good catch: I missed updating pyodide as I wasn't building for it on my test project.

I'd be happy to put the change together later today.

What's the issue that you would see with local macos-pyodide builds?

MusicalNinjaDad: I'd imagine the util function has a signature along the lines of: `movewheel(wheel: Path) -> output_wheel: Path`, directly accesses `build_options` as this is global but leaves `built_wheels.append` at the calling site for clarity.

Does that make sense?

mayeut: The function signature should be something like `move(src: Path, dst: Path) -> None`

MusicalNinjaDad: How about this? (See latest push). I'm happy with the shape of the function and it follows shutil.move closely (which also returns the destination path).

I'm slightly less happy with the import inside the function but felt that logging the move was valuable enough to pay that price.

I've [tested against macos & windows](https://github.com/MusicalNinjaDad/cibuildwheel/actions/runs/9403124615). I couldn't test pyodide as the test project I'm using is rust-based and I haven't got pyodide/emscripten working for that yet ...

henryiii: What sort of issues are you hitting? It should Just Work™ in most cases, so there's a chance it's hitting a Pyodide or cibuildwheel bug.

henryiii: We can probably split utils up later to avoid the circular import, it's pretty big. Not something to worry about for this PR.

Why is the warning at every usage? Would it make sense to warn inside the function if it doesn't copy where it's supposed to?

cibuildwheel/util.py
```diff
@@ -359,6 +360,40 @@ def extract_tar(tar_src: Path, dest: Path) -> None:
...
+
+    # Importing here as logger needs various functions from util -> circular imports
+    from .logger import log
+
+    src_file = src_file.resolve()
+    dst_file = dst_file.resolve()
+
+    if dst_file.is_dir():
+        # Cannot overwrite a directory with a file
+        raise IsADirectoryError
```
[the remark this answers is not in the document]
MusicalNinjaDad: Thanks Henryii ... Now I see why I missed the parens - I didn't add a message, and I'm used to ruff complaining about parens but no message based on flake-8's ruleset ...

henryiii (approved):

cibuildwheel/macos.py
MusicalNinjaDad: > What sort of issues are you hitting? It should Just Work™ in most cases, so there's a chance it's hitting a Pyodide or cibuildwheel bug.

rust seemed to be targetting the wrong target on the one time I tried "just switching it on" as part of testing this PR. Looking at the code and the implementation PR on cibuildwheel side, I get the impression that I "just" need to add a few config entries into pyproject.toml, find the best way to build in pipeline, check whether ADO artefact repos can accept the wheels, and set up some kind of test environment to play with the result...

It looks interesting enough that I'm likely to try playing with it in my sandbox project soon

MusicalNinjaDad: > Why is the warning at every usage? Would it make sense to warn inside the function if it doesn't copy where it's supposed to?

I'd consider it to be in the responsibility of the main script to double check that a called util did what was expected, that also covers the risk of later changes to utils etc.

Adding the warning to `util.move_file` feels like a no-op to me ... the only line that would effectively be checked is whether `shutil.move` did what was expected (?)

MusicalNinjaDad:
> You should be able to just change the cibuildwheel install/use to this PR.

FWIW: I've just done that and the relevant parts of the build have all passed, just waiting on the emulated linux architectures now ...

henryiii:
Thanks!

PR #1871, opened 2024-06-11
DOC: Document macOS 3.8 arm64 workaround

larsoner:
I just hit https://github.com/pypa/cibuildwheel/issues/1278 :

https://github.com/h5py/h5py/actions/runs/9472096444/job/26096834822#step:9:483

```
    error: dlopen(/Users/runner/work/h5py/h5py/cache/hdf5/1.12.2-arm64/lib/libhdf5.dylib, 0x0006): tried:

... '/Users/runner/work/h5py/h5py/cache/hdf5/1.12.2-arm64/lib/libhdf5.200.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64'))
```

Using the suggestion from https://github.com/pypa/cibuildwheel/issues/1278#issuecomment-2052099631 in https://github.com/h5py/h5py/pull/2444/commits/87cf54c343e70e43885aa876aa1fee131726cb2a fixes it:

https://github.com/h5py/h5py/actions/runs/9472215836/job/26097237985?pr=2444#step:9:179

@cretz do you want to open a quick PR instead so you can get credit (and blame 😄 ) for this change?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -372,14 +372,6 @@ So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`
   if: runner.os == 'macOS' && runner.arch == 'ARM64'
 ```
 
-There is also an 'experimental' installer available that's built natively for arm64 that you can install manually with something like:
-
-```bash
-curl -o /tmp/Python38.pkg https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg
-sudo installer -pkg /tmp/Python38.pkg -target /
-sh "/Applications/Python 3.8/Install Certificates.command"
-```
-
 Then cibuildwheel will detect that it's installed and use it instead. However, you probably don't want to build x86_64 wheels on this Python, unless you're happy with them only supporting macOS 11+.
 
 ### macOS: Library dependencies do not satisfy target MacOS
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -363,7 +363,7 @@ If you're building on an arm64 runner, you might notice something strange about 
 
 This is fine for simple C extensions, but for more complicated builds on arm64 it becomes an issue.
 
-So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`) on Python 3.8, before invoking cibuildwheel, you should install a native arm64 Python 3.8 interpreter on the runner. On GitHub actions you can set this up easily using the `setup-python` action with something like:
+So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`) on Python 3.8, before invoking cibuildwheel, you should install a native arm64 Python 3.8 interpreter on the runner:
 
 
 !!! tab "GitHub Actions"
@@ -375,7 +375,7 @@ So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`
       if: runner.os == 'macOS' && runner.arch == 'ARM64'
     ```
 
-!!! tab "Other CIs"
+!!! tab "Generic"
 
     ```bash
     curl -o /tmp/Python38.pkg https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -363,10 +363,16 @@ If you're building on an arm64 runner, you might notice something strange about 
 
 This is fine for simple C extensions, but for more complicated builds on arm64 it becomes an issue.
 
-So, if the cross-compilation is an issue for you, there is an 'experimental' installer available that's built natively for arm64.
+So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`) on Python 3.8, before invoking cibuildwheel, you should install a native arm64 Python 3.8 interpreter on the runner. On GitHub actions you can set this up easily using the `setup-python` action with something like:
 
-To use this installer and perform native CPython 3.8 building, before invoking cibuildwheel, install the universal2 version of Python on your arm64 runner, something like:
+```yaml
+- uses: actions/setup-python@v5
+  with:
+    python-version: 3.8
+  if: runner.os == 'macOS' && runner.arch == 'ARM64'
+```
 
+There is also an 'experimental' installer available that's built natively for arm64 that you can install manually with something like:
 
 ```bash
 curl -o /tmp/Python38.pkg https://www.python.org/ftp/python/3.8.10/python-3.8.10-macos11.pkg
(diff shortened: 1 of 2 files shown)


henryiii:
The reason this is this way is because you can't target 10.9-10.15 with 3.9, only 11+, unlike every other binary, including the 3.9+ universal builds. 3.8 was actually built with 11+, even on the Intel side.

Maybe we should change this though, especially now that ARM runners are common (and hopefully no one will care about Python 3.8 by the time all the Intel runners are gone)? We could error if MACOSX_DEPLOYMENT_TARGET was set to anything under 11 on 3.8 running from ARM building Universal or Intel wheels?

henryiii:
Actually, shouldn't this only affect targeting ARM, not Universal or Intel?

larsoner:
I'm not sure I understand the subtleties here, but in my case at least the error happens for me on GHA `runs-on: macos-14` targeting `CIBW_ARCHS=arm64` -- both failing and successful tracebacks / CI runs above or on that runner while trying to build the native arm64 wheel on the arm64 image.

> We could error if MACOSX_DEPLOYMENT_TARGET was set to anything under 11 on 3.8 running from ARM building Universal or Intel wheels?

For me I originally saw the error with `MACOSX_DEPLOYMENT_TARGET=10.9`. Then I un-set the env var , and got the same error. Then I explicitly set it to `11.0`, no change. (These are in earlier commits of https://github.com/h5py/h5py/pull/2444 actually.) In all of those cases, The problematic `10.9` installer was used and x86_64 was the resulting `platform.machine()`, but maybe I didn't set the env var the right way / the right place? In any case, this change fixed my build at least.

henryiii:
> GitHub Actions has encountered an internal error when running your job.

Hmm, guessing Actions is down?

cretz:
> @cretz do you want to open a quick PR instead so you can get credit (and blame 😄 ) for this change?

Nah, happy to see it exist, no credit needed.

henryiii:
We currently don't change this based on MACOSX_DEPLOYMENT_TARGET (and that would be a little tricky, I think), but ARM should always treat it as if it's 11, since that's the first macOS for ARM. So I'm thinking while our current system is sadly required for universal2, it's safe to do this for pure ARM.

I don't think we reinstall, so this could cause an issue if someone builds non-pure-ARM (including universal) and ARM in the same job.

@mayeut, thoughts?

henryiii:
Looks like the issue was due to the fact we don't reinstall Python if it exists:

https://github.com/pypa/cibuildwheel/issues/1278#issuecomment-1264400453

(FYI, due to that, you can also manually "fix" this by pre-installing Python.org universal installer before running cibuildwheel, IIRC, but not the cleanest solution)

larsoner:
Ahh sorry I didn't read the whole issue, just jumped to a potential solution and tried it 🤦 

> Step 1 I think then would be to reinstall Python if it's not the right variety. Looks like no one ever took that up.

Okay I did that in https://github.com/h5py/h5py/pull/2444/commits/51bb92d83c27a3c388ca19719d01a7f842472abc and it seems to work!

https://github.com/h5py/h5py/actions/runs/9472998912/job/26099471057

mayeut:
I'm not sure there's a clean solution. Force re-install might be one for CI with a warning on local builds.
https://cibuildwheel.pypa.io/en/stable/faq/#macos-building-cpython-38-wheels-on-arm64 is one solution but on GHA, using setup-python works on macOS-14 as the link on h5py shows. I'm sure it was already mentioned elsewhere but never got around to updating the tips & tricks section.

larsoner:
> I'm sure it was already mentioned elsewhere but never got around to updating the tips & tricks section.

Okay I've updated this PR just to be a doc update!

docs/faq.md
```diff
@@ -363,10 +363,16 @@ If you're building on an arm64 runner, you might notice something strange about
...
 
-To use this installer and perform native CPython 3.8 building, before invoking cibuildwheel, install the universal2 version of Python on your arm64 runner, something like:
+```yaml
+- uses: actions/setup-python@v5
+  with:
+    python-version: 3.8
+  if: runner.os == 'macOS' && runner.arch == 'ARM64'
+```
 
+There is also an 'experimental' installer available that's built natively for arm64 that you can install manually with something like:
```
henryiii, line 375: Isn't the workaround documented just below here already? The workaround shown installs the official build, rather than GitHub's recompilation with a higher MACOSX_DEPLOYMENT_TARGET. (Assuming they are recompiled, not sure they are anymore)

mayeut: It's the same experimental installer used on GHA form arm64.
arm64 on GHA always uses installers from python.org

henryiii: Would it speed up cibuildwheel for a user to pre-install these via setup-python, actually? Might be a bit of a micro-op, just thinking.

mayeut: > Would it speed up cibuildwheel for a user to pre-install these via setup-python, actually? Might be a bit of a micro-op, just thinking.

The only diff in speed I could see would be downloading from GitHub vs python.org for non pre-installed pythons. I don't think it would be noticeable given only 3.8 is not preinstalled on macos-14 (and might add another source of potential failure), we can see in the test workflow that for the action sample build, python 3.9+ cibuildwheel install time is 0s. The same goes for Python 3.11+ on x86_64.

larsoner: If they're functionally equivalent then I'm happy to just change this doc to use `setup-python` and remove the manual install steps. It's more standard practice across GHA to use `setup-python` so if it's functionally equivalent or better to use it, it seems like the docs should suggest that rather than `curl`/`sudo`/`sh` steps suggested currently. I'll change the text to remove the old way, but if that's not preferred feel free to close!

henryiii: The old way works on other CI systems, the GHA one is GHA only. I wonder if tabs here would be better? GHA and general.

larsoner: Good call, changed to tabs!

docs/faq.md
```diff
@@ -363,16 +363,25 @@ If you're building on an arm64 runner, you might notice something strange about
 
 This is fine for simple C extensions, but for more complicated builds on arm64 it becomes an issue.
 
-So, if the cross-compilation is an issue for you, there is an 'experimental' installer available that's built natively for arm64.
+So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`) on Python 3.8, before invoking cibuildwheel, you should install a native arm64 Python 3.8 interpreter on the runner. On GitHub actions you can set this up easily using the `setup-python` action with something like:
```
mayeut, line 366:
```suggestion
So, if you want to build macOS arm64 wheels on an arm64 runner (e.g., `macos-14`) on Python 3.8, before invoking cibuildwheel, you should install a native arm64 Python 3.8 interpreter on the runner:
```

docs/faq.md
```diff
@@ -363,16 +363,25 @@ If you're building on an arm64 runner, you might notice something strange about
...
-sh "/Applications/Python 3.8/Install Certificates.command"
-```
+    ```yaml
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.8
+      if: runner.os == 'macOS' && runner.arch == 'ARM64'
+    ```
+
+!!! tab "Other CIs"
```
mayeut, line 378:
```suggestion
!!! tab "Generic"
```

larsoner:
Thanks for the fixes, applied!

mayeut (approved):

PR #1876, opened 2024-06-12
set VSCMD_ARG_TGT_ARCH based on targetted architecture

### LINKED ISSUE #1861: cibuildwheel GitHub action fails due to duplicate skbuild build directory on Windows ARM64 build
### Description

I hit a build failure when using skbuild in conjunction with cibuildwheel. 

Here is the relevant part of the build log:

```console
     Configuring Project
      Working directory:
        D:\a\micropython-uncrustify\micropython-uncrustify\_skbuild\win-amd64-3.11\cmake-build
      Command:
        'C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-x7dj25c1\overlay\Lib\site-packages\cmake\data\bin/cmake.exe' 'D:\a\micropython-uncrustify\micropython-uncrustify' -G 'Visual Studio 17 2022' --no-warn-unused-cli '-DCMAKE_INSTALL_PREFIX:PATH=D:\a\micropython-uncrustify\micropython-uncrustify\_skbuild\win-amd64-3.11\cmake-install' -DPYTHON_VERSION_STRING:STRING=3.11.9 -DSKBUILD:INTERNAL=TRUE '-DCMAKE_MODULE_PATH:PATH=C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-x7dj25c1\overlay\Lib\site-packages\skbuild\resources\cmake' '-DPYTHON_EXECUTABLE:PATH=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dk_dqzxv\cp311-win_arm64\build\venv\Scripts\python.exe' '-DPYTHON_INCLUDE_DIR:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\python.3.11.9\tools\Include' '-DPYTHON_LIBRARY:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\pythonarm64.3.11.9\tools\libs\python311.lib' '-DPython_EXECUTABLE:PATH=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dk_dqzxv\cp311-win_arm64\build\venv\Scripts\python.exe' '-DPython_ROOT_DIR:PATH=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dk_dqzxv\cp311-win_arm64\build\venv' -DPython_FIND_REGISTRY:STRING=NEVER '-DPython_INCLUDE_DIR:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\python.3.11.9\tools\Include' '-DPython_LIBRARY:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\pythonarm64.3.11.9\tools\libs\python311.lib' '-DPython3_EXECUTABLE:PATH=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dk_dqzxv\cp311-win_arm64\build\venv\Scripts\python.exe' '-DPython3_ROOT_DIR:PATH=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dk_dqzxv\cp311-win_arm64\build\venv' -DPython3_FIND_REGISTRY:STRING=NEVER '-DPython3_INCLUDE_DIR:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\python.3.11.9\tools\Include' '-DPython3_LIBRARY:PATH=C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\nuget-cpython\pythonarm64.3.11.9\tools\libs\python311.lib' -T v143 -A ARM64 -DCMAKE_BUILD_TYPE:STRING=Release
    
    Not searching for unused variables given on the command line.
    CMake Error: Error: generator platform: ARM64
    Does not match the platform used previously: x64
    Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
      File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-x7dj25c1\overlay\Lib\site-packages\skbuild\setuptools_wrap.py", line 666, in setup
        env = cmkr.configure(
              ^^^^^^^^^^^^^^^
      File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-x7dj25c1\overlay\Lib\site-packages\skbuild\cmaker.py", line 357, in configure
        raise SKBuildError(msg)
```

We can see that this is targeting ARM64 Windows, but the build directory is `_skbuild\win-amd64-3.11\cmake-build`. This is the same build directory that was already used by the amd64 build in the same CI job. So cmake is giving the expected error.

I did some digging into the code and this is happening because skbuild is `distutils.util.get_platform()` to [get the platform](https://github.com/scikit-build/scikit-build/blob/4dab4576d7a480da7484cfc5c249c86f7d3ecde3/skbuild/constants.py#L47), so it is getting the host platform instead of the target platform.

This can be [overridden by setting the VSCMD_ARG_TGT_ARCH environment variable](https://github.com/pypa/setuptools/blob/f91fa3d9fc7262e0467e4b2f84fe463f8f8d23cf/setuptools/_distutils/util.py#L52).

So I added the following to my `pyproject.toml`:

```toml
# HACK: causes distutils to return the proper platform name so that scikit-build
# will not use the same build directory as the host platform build.
select = "*-win_arm64"
environment = "VSCMD_ARG_TGT_ARCH=arm64"
```

It would be nice if this "just worked" or was at least the workaround was documented in the various examples.

FWIW, using `VSCMD_ARG_TGT_ARCH` was also mentioned in https://github.com/pypa/cibuildwheel/pull/1144#issuecomment-1347189520

### Build log

https://github.com/dlech/micropython-uncrustify/actions/runs/9429791938/job/25976561751

### CI config

https://github.com/dlech/micropython-uncrustify/actions/runs/9429791938/workflow

MusicalNinjaDad:
This fixes #1861 and any other build setups which make use of `setuptools._distutils.util.get_platform()` on windows

### Approach
- Uses `dict` lookup pattern analog to many other places within the code
- Only sets the variable if it has not already been set manually
- Will overwrite a pre-set `VSCMD_ARG_TGT_ARCH=""`
- Tests `monkeypatch` the environment before calling `setuptools._distutils.util.get_platform()` and validating returned platform

### Test availability
- Tests are set to only run on windows
- monkeypatching `os.name` in a tight context worked locally (Win11 + WSL2 + Ubunutu 22.04), but fails in pipeline on github actions
- all tests pass on github actions windows runner but tests targeting arm64 fail on azure pipeline. I'm assuming this is a weird, azure-specific thing in their environment and have set the arm64 test to skip on azure

#### Potential refactoring idea for a future PR
- I'd like to refactor the various mapping lookups into a WindowsArchitecture `dataclass` - there are many places with the pattern `somevar = architecture_mapping_dict[arch]` which would be more readable as `somevar = targetarchitecure.somevar`

### DIFF
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -157,8 +157,8 @@ def setup_setuptools_cross_compile(
     # identifies the target, not the host
     vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
     current_tgt_arch = vscmd_arg_tgt_arch[python_configuration.arch]
-    if env.get("VSCMD_ARG_TGT_ARCH", current_tgt_arch) != current_tgt_arch:
-        msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch}, got {env['VSCMD_ARG_TGT_ARCH']}. Make sure you setup MSVC targeting the right architecture."
+    if (env.get("VSCMD_ARG_TGT_ARCH") or current_tgt_arch) != current_tgt_arch:
+        msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch!r}, got {env['VSCMD_ARG_TGT_ARCH']!r}. If you're setting up MSVC yourself (e.g. using vcvarsall.bat or msvc-dev-cmd), make sure to target the right architecture. Alternatively, run cibuildwheel without configuring MSVC, and let the build backend handle it."
         raise errors.FatalError(msg)
     env["VSCMD_ARG_TGT_ARCH"] = current_tgt_arch
 
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -155,8 +155,9 @@ def setup_setuptools_cross_compile(
 
     # Set environment variable so that setuptools._distutils.get_platform()
     # identifies the target, not the host
-    vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
-    env["VSCMD_ARG_TGT_ARCH"] = vscmd_arg_tgt_arch[python_configuration.arch]
+    if not env.get("VSCMD_ARG_TGT_ARCH"):
+        vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
+        env["VSCMD_ARG_TGT_ARCH"] = vscmd_arg_tgt_arch[python_configuration.arch]
 
     # (This file must be default/locale encoding, so we can't pass 'encoding')
     distutils_cfg.write_text(
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -155,9 +155,12 @@ def setup_setuptools_cross_compile(
 
     # Set environment variable so that setuptools._distutils.get_platform()
     # identifies the target, not the host
-    if not env.get("VSCMD_ARG_TGT_ARCH"):
-        vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
-        env["VSCMD_ARG_TGT_ARCH"] = vscmd_arg_tgt_arch[python_configuration.arch]
+    vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
+    current_tgt_arch = vscmd_arg_tgt_arch[python_configuration.arch]
+    if env.get("VSCMD_ARG_TGT_ARCH", current_tgt_arch) != current_tgt_arch:
+         msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch}, got {env['VSCMD_ARG_TGT_ARCH']}. Make sure you setup MSVC targetting the right architecture."
+         raise errors.FatalError(msg)
+    env["VSCMD_ARG_TGT_ARCH"] = current_tgt_arch
 
     # (This file must be default/locale encoding, so we can't pass 'encoding')
     distutils_cfg.write_text(
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -158,8 +158,8 @@ def setup_setuptools_cross_compile(
     vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
     current_tgt_arch = vscmd_arg_tgt_arch[python_configuration.arch]
     if env.get("VSCMD_ARG_TGT_ARCH", current_tgt_arch) != current_tgt_arch:
-         msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch}, got {env['VSCMD_ARG_TGT_ARCH']}. Make sure you setup MSVC targetting the right architecture."
-         raise errors.FatalError(msg)
+        msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch}, got {env['VSCMD_ARG_TGT_ARCH']}. Make sure you setup MSVC targeting the right architecture."
+        raise errors.FatalError(msg)
     env["VSCMD_ARG_TGT_ARCH"] = current_tgt_arch
 
     # (This file must be default/locale encoding, so we can't pass 'encoding')
--- a/cibuildwheel/windows.py
+++ b/cibuildwheel/windows.py
@@ -152,6 +152,12 @@ def setup_setuptools_cross_compile(
     # set the platform name
     map_plat = {"32": "win32", "64": "win-amd64", "ARM64": "win-arm64"}
     plat_name = map_plat[python_configuration.arch]
+
+    # Set environment variable so that setuptools._distutils.get_platform()
+    # identifies the target, not the host
+    vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
+    env["VSCMD_ARG_TGT_ARCH"] = vscmd_arg_tgt_arch[python_configuration.arch]
+
     # (This file must be default/locale encoding, so we can't pass 'encoding')
     distutils_cfg.write_text(
         textwrap.dedent(
--- a/unit_test/get_platform_test.py
+++ b/unit_test/get_platform_test.py
@@ -7,6 +7,7 @@ from typing import Dict
 import pytest
 import setuptools._distutils.util
 
+from cibuildwheel.util import CIProvider, detect_ci_provider
 from cibuildwheel.windows import PythonConfiguration, setup_setuptools_cross_compile
 
 # monkeypatching os.name is too flaky. E.g. It works on my machine, but fails in pipeline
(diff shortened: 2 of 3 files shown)


MusicalNinjaDad:
Failing check on azure pipelines - linux caused by a timeout waiting to download docker image: `docker: Error response from daemon: Get "[https://quay.io/v2/"](https://quay.io/v2/%22): net/http: request canceled (Client.Timeout exceeded while awaiting headers).` [logs](https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=6097&view=logs&j=65f37ddb-1a35-539a-d91f-4ab7f01a34d0&t=677308e5-f586-5eb1-7f4b-3c9712904030&l=2484)

MusicalNinjaDad:
@henryiii - I'd welcome your feedback on this suggestion please.

henryiii:
I've always been in favor of this, it's just been others that are worried that setting this variable will make some build backend confused about not having the others.

henryiii:
@joerick or @mayeut, thoughts?

mayeut:
I don't know enough to really have an informed opinion here.

cibuildwheel/windows.py
```diff
@@ -154,6 +154,13 @@ def setup_setuptools_cross_compile(
     # set the platform name
     map_plat = {"32": "win32", "64": "win-amd64", "ARM64": "win-arm64"}
     plat_name = map_plat[python_configuration.arch]
+
+    # Set environment variable so that setuptools._distutils.get_platform()
+    # identifies the target, not the host
+    if not env.get("VSCMD_ARG_TGT_ARCH"):
+        vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
+        env["VSCMD_ARG_TGT_ARCH"] = vscmd_arg_tgt_arch[python_configuration.arch]
```
joerick, line 162: `env.setdefault` would be more ideomatic I think, that's what we use in macos.py

joerick (commented):
I'm okay with this, it seems to me to be very similar to _PYTHON_HOST_PLATFORM/ARCHFLAGS that we set in macos.py.

I don't have any specific knowledge of the MS compiler to predict and second-order effects though. @zooba, since you created the cross-compilation mechanism, do you have an opinion on the use of VSCMD_ARG_TGT_ARCH for cross-compilation?

cibuildwheel/windows.py
henryiii: If this is already set by `vcvarsall.bat`, then setdefault wouldn't change it, and it would be wrong.

joerick: The behaviour in the diff is to only set if not set already. I was only suggesting a style change really. But...

Who is calling `vcvarsall.bat`? I don't think cibuildwheel calls it. Do some users call that before calling cibuildwheel?

henryiii: Ahh, didn't look up... 2 lines. Yes, set default is better. Though is this what we want? Would erroring out, for example, make more sense?

Users may set up MSVC before calling cibuildwheel. [here are 366 examples](https://github.com/search?type=code&q=path%3A.github+msvc-dev-cmd+cibuildwheel).

joerick: [`vcvarsall.bat`](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#use-the-developer-tools-in-an-existing-command-window) and [`msvc-dev-cmd`](https://github.com/ilammy/msvc-dev-cmd?tab=readme-ov-file#inputs) do allow themselves to be configured for different targets, including cross-compilation. So asking cibuildwheel to cross-compile after calling one of these tools for a different target feels like a misconfiguration IMO.

Yes, perhaps an error here would be better. If VSCMD_ARG_TGT_ARCH is set to something other than what we were gonna set, we should error out IMO. The user can then decide what they want to do - either stop using vcvarsall/msvc-dev-cmd or call it again with the right configuration before cross-compiling.

henryiii: Agree.

zooba: > Yes, perhaps an error here would be better. If VSCMD_ARG_TGT_ARCH is set to something other than what we were gonna set, we should error out IMO. 

Yep, if a user has run vcvarsall already, _that's_ the cross-compiling configuration done. If they want cibuildwheel to configure it, then they shouldn't do it themselves. Otherwise you really need to clear all (most of) the environment variables that were set so that the build backend knows to reconfigure.

cibuildwheel/windows.py
```diff
@@ -154,6 +154,16 @@ def setup_setuptools_cross_compile(
     # set the platform name
     map_plat = {"32": "win32", "64": "win-amd64", "ARM64": "win-arm64"}
     plat_name = map_plat[python_configuration.arch]
+
+    # Set environment variable so that setuptools._distutils.get_platform()
+    # identifies the target, not the host
+    vscmd_arg_tgt_arch = {"32": "x86", "64": "x64", "ARM64": "arm64"}
+    current_tgt_arch = vscmd_arg_tgt_arch[python_configuration.arch]
+    if env.get("VSCMD_ARG_TGT_ARCH", current_tgt_arch) != current_tgt_arch:
```
joerick, line 162: Tiny readability suggestion...
```suggestion
    if "VSCMD_ARG_TGT_ARCH" in env and env["VSCMD_ARG_TGT_ARCH"] != current_tgt_arch:
```

cibuildwheel/windows.py
```diff
@@ -154,6 +154,16 @@ def setup_setuptools_cross_compile(
...
+        msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch}, got {env['VSCMD_ARG_TGT_ARCH']}. Make sure you setup MSVC targeting the right architecture."
```
joerick, line 163: Given that preconfiguring MSVC is now incompatible with our default setting of building both 32 and 64 bit windows in one cibuildwheel session, I feel we should suggest that users don't preconfigure MSVC using vcvarsall/msvc-dev-cmd.

```suggestion
        msg = f"VSCMD_ARG_TGT_ARCH must be set to {current_tgt_arch!r}, got {env['VSCMD_ARG_TGT_ARCH']!r}. If you're setting up MSVC yourself (e.g. using vcvarsall.bat or msvc-dev-cmd), make sure to target the right architecture. Alternatively, run cibuildwheel without configuring MSVC, and let the build backend handle it."
```

MusicalNinjaDad: Thanks everyone, that's great feedback. I'll put it all together and finalise the PR on Thu/Fri - right now I'm snowed under until then with work & personal stuff. Hope that timeframe is OK for you...

cibuildwheel/windows.py
henryiii: This doesn't handle having it set to an empty variable, which apparently one of CI's does. 🙄 We should treat empty variables like unset variables, I think.

```suggestion
    if (env.get("VSCMD_ARG_TGT_ARCH") or current_tgt_arch) != current_tgt_arch:
```

cibuildwheel/windows.py
henryiii: Some build backends don't like it not being set up, I think. Usually people add this because something is broken. However, on an error message, I think this is useful.

henryiii: No rush, and I've been adding and accepting changes to see if I can get it working.

MusicalNinjaDad:
I think this is done now (?) - I've updated the tests to expect the FatalError and can't think of anything else to do in the main code

unit_test/get_platform_test.py
```diff
@@ -0,0 +1,93 @@
+# ruff: noqa: ARG001
```
mayeut, line 1:
```suggestion
```

joerick (approved):
Looks good to me!

Issue #1877, opened 2024-06-12
build[uv] uses wrong Python when testing on Windows


njzjz:
### Description

It worked yesterday but failed today.

See below. The ` cp311-win_amd64` build uses Python 3.7.9 to create the environment.

```
Building cp311-win_amd64 wheel
CPython 3.11 Windows 64bit
...
Testing wheel...
  
  + uv venv 'C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\venv-test' --python=python
  Using Python 3.7.9 interpreter at: C:\hostedtoolcache\windows\Python\3.7.9\x64\python3.exe
  Creating virtualenv at: C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\venv-test
  Activate with: C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\venv-test\Scripts\activate
  + where python
  C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\venv-test\Scripts\python.exe
  C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\build\venv\Scripts\python.exe
  C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe
  + uv pip install 'C:\Users\runneradmin\AppData\Local\Temp\cibw-run-90ozl8o3\cp311-win_amd64\repaired_wheel\deepmd_kit-3.0.0a1.dev211+g4a27dec7-py37-none-win_amd64.whl[cpu]'
    × No solution found when resolving dependencies:
    ╰─▶ Because only deepmd-kit[cpu]==3.0.0a1.dev211+g4a27dec7 is available and
        the current Python version (3.7.9) does not satisfy Python>=3.8, we can
        conclude that all versions of deepmd-kit[cpu] are incompatible.
        And because you require deepmd-kit[cpu], we can conclude that the
        requirements are unsatisfiable.
                                                              ✕ 6.49s
[...]

PR #1880, opened 2024-06-12
chore(tests): run a project with tests using `build[uv]` frontend

mayeut:
see #1877

### DIFF
--- a/test/test_pep518.py
+++ b/test/test_pep518.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import textwrap
 
+import pytest
+
 from . import test_projects, utils
 
 basic_project = test_projects.new_c_project(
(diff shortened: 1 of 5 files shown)


mayeut:
This will need some more work, it's failing for another reason than the one I want to check.

test/test_pep518.py
```diff
@@ -24,8 +24,7 @@
 requires = [
     "setuptools >= 42",
     "setuptools_scm[toml]>=4.1.2",
-    "wheel",
-    "requests==2.27.0"
+    "jmespath==0.10.0"
```
mayeut, line 27: The test was failing because `requests` is in the pyodide constraints file.

henryiii (approved):

PR #1881, opened 2024-06-12
fix: specify full path to non-venv Python

henryiii:
Fix https://github.com/pypa/cibuildwheel/issues/1877. Should be fixed by https://github.com/astral-sh/uv/pull/4289, but putting the correct path here is better anyway.


mayeut (approved):
I do have local failures with `test/test_macos_archs.py::test_universal2_testing_on_arm64[build]` before this patch. Don't know why CI did not fail...

I'll keep draft #1880 opened to work on it this week-end in order to extend tests with uv.
I now see the same failure locally with this test so that means we were not testing tests with uv on other platforms.

PR #1884, opened 2024-06-16
chore: remove cached_property from utils

mayeut:
This was a leftover from python 3.7 support removal.

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -6,10 +6,10 @@ import contextlib
 import dataclasses
 import difflib
 import enum
-import functools
 import shlex
 import textwrap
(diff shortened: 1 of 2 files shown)


cibuildwheel/options.py
```diff
@@ -6,10 +6,10 @@
 import dataclasses
 import difflib
 import enum
-import functools
 import shlex
 import textwrap
 from collections.abc import Callable, Generator, Iterable, Iterator, Set
+from functools import cached_property, lru_cache
```
henryiii, line 12: Minor nit: I prefer absolute imports for things that are not classes. Especially since these are decorators, they will always be on their own line, so `@functools.cached_property` / `@functools.lru_cache` are fine, IMO.

Issue #1887, opened 2024-06-17
Docs suggest invalid configuration combination


burgholzer:
### Description

The docs currently suggest that the new `build[uv]` feature can be used with extra arguments, such as
```toml
build-frontend = { name = "build[uv]", args = ["--no-isolation"] }
```

However, the above code will trigger an error
```
python -m build: error: argument --installer: not allowed with argument --no-isolation/-n
```

I just wanted to raise this here because the docs suggest this as a valid configuration. However, if you feel that it would make more sense to open an issue over at the https://github.com/pypa/build repo, given that it should actually be valid to request turning off build isolation when using build with the `uv` installer, please say so.

henryiii:
Ohh, interesting, hadn’t thought about that. We probably should special case this.

joerick:
Fixed in #1889.

PR #1889, opened 2024-06-17
fix: support --no-isolation with build[uv]

henryiii:
Fixing #1887. I didn't special case chained short flags, like `-xn` or `-nx`, maybe we should?

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -282,7 +282,7 @@ def build_in_container(
                 if not 0 <= build_options.build_verbosity < 2:
                     msg = f"build_verbosity {build_options.build_verbosity} is not supported for build frontend. Ignoring."
                     log.warning(msg)
-                if use_uv:
+                if use_uv and "--no-isolation" not in extra_flags and "-n" not in extra_flags:
                     extra_flags += ["--installer=uv"]
                 container.call(
                     [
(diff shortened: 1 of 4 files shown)


joerick:
It looks to me like you're just disabling uv support when `--no-isolation` is passed. Is this a misconfiguration then? As in "you can't specify `build[uv]` and `--no-isolation`, if you need `--no-isolation` you must use `build`"?

henryiii:
We still install everything via uv when `build[uv]` is set. It means “use build and uv”. The problem is that with no isolation, build itself isn’t using uv because it doesn’t install anything or make and environment. That’s why it errors on that flag.

joerick:
Ah, got it. Understood!

henryiii:
If I add a utility for this, it will clash with #1873, so let's do this as a quick fix, and then write a utility later that tries to be smarter for combined flags.

henryiii:
I think we could release after this. Would like to get b3 out!

mayeut:
> I didn't special case chained short flags, like -xn or -nx, maybe we should?

I don't think it's necessary but a note / caution shall be added in the docs then.

mayeut:
> I think we could release after this. Would like to get b3 out!

I cancelled Travis CI to get some time for manylinux. CentOS 7 reached End-Of-Life so mirrors shall be updated on manylinux2014. IMHO, we should wait for the new images before the release.

mayeut (approved):

PR #1950, opened 2024-08-01
Build CPython 3.13 by default

EwoutH:
Enable building with CPython 3.13 by default, no longer requiring `prerelease_pythons` flag to be set true.

Merge after https://github.com/pypa/cibuildwheel/pull/1949 is merged.

Builds upon last-year PRs: https://github.com/pypa/cibuildwheel/pull/1226 and https://github.com/pypa/cibuildwheel/pull/1565.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ What does it do?
 
 <sup>¹ PyPy is only supported for manylinux wheels.</sup><br>
 <sup>² Windows arm64 support is experimental.</sup><br>
-<sup>³ CPython 3.13 is available using the [`CIBW_PRERELEASE_PYTHONS`](https://cibuildwheel.pypa.io/en/stable/options/#prerelease-pythons) option. Free-threaded mode requires opt-in.</sup><br>
+<sup>³ CPython 3.13 is built by default using Python RCs, starting with cibuildwheel 2.20.</sup><br>
 <sup>⁴ Experimental, not yet supported on PyPI, but can be used directly in web deployment. Use `--platform pyodide` to build.</sup><br>
 
 - Builds manylinux, musllinux, macOS 10.9+, and Windows wheels for CPython and PyPy
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ What does it do?
 
 <sup>¹ PyPy is only supported for manylinux wheels.</sup><br>
 <sup>² Windows arm64 support is experimental.</sup><br>
-<sup>³ CPython 3.13 is built by default using Python RCs, starting with cibuildwheel 2.20.</sup><br>
+<sup>³ CPython 3.13 is built by default using Python RCs, starting with cibuildwheel 2.20. Free-threaded mode will still require opt-in using [`CIBW_FREE_THREADED_SUPPORT`](https://cibuildwheel.pypa.io/en/stable/options/#free-threaded-support).</sup><br>
 <sup>⁴ Experimental, not yet supported on PyPI, but can be used directly in web deployment. Use `--platform pyodide` to build.</sup><br>
 
 - Builds manylinux, musllinux, macOS 10.9+, and Windows wheels for CPython and PyPy
--- a/unit_test/linux_build_steps_test.py
+++ b/unit_test/linux_build_steps_test.py
@@ -74,7 +74,7 @@ def test_linux_container_split(tmp_path: Path, monkeypatch):
         "cp313-manylinux_x86_64",
     ]
     assert before_alls(build_steps[0]) == [""] * 4
-    assert container_engines(build_steps[0]) == [default_container_engine] * 3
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ What does it do?
 
 <sup>¹ PyPy is only supported for manylinux wheels.</sup><br>
 <sup>² Windows arm64 support is experimental.</sup><br>
-<sup>³ CPython 3.13 is available using the [`CIBW_PRERELEASE_PYTHONS`](https://cibuildwheel.pypa.io/en/stable/options/#prerelease-pythons) option. Free-threaded mode requires opt-in.</sup><br>
+<sup>³ CPython 3.13 is built by default using Python RCs, starting with cibuildwheel 2.20.</sup><br>
```
joerick, line 42: I believe free-threaded mode will still require opt-in using [CIBW_FREE_THREADED_SUPPORT](https://cibuildwheel.pypa.io/en/stable/options/#free-threaded-support)

EwoutH: Good point, added a note!

EwoutH:
I see some CI failing. I think one error is the `free_threaded` 3.13 now being expected to build, but doesn't, so I need to update that test.

The other errors I don't understand directly, so some help with them is appreciated!

Maintainers, feel free to directly modify this PR (push to this branch) by the way.

unit_test/linux_build_steps_test.py
```diff
@@ -71,8 +71,9 @@ def container_engines(step):
         "cp36-manylinux_x86_64",
         "cp311-manylinux_x86_64",
         "cp312-manylinux_x86_64",
+        "cp313-manylinux_x86_64",
```
EwoutH, line 74: Does `"cp313t-manylinux_x86_64"` also needs to be added here in some way?

unit_test/option_prepare_test.py
```diff
@@ -21,6 +21,7 @@
     "cp310",
     "cp311",
     "cp312",
+    "cp313",
```
EwoutH, line 24: Does both `"cp313"` and `"cp313t"` need to be added here and below?

unit_test/linux_build_steps_test.py
mayeut: if we want to add specific tests for free threaded python, let's do that in a following PR.

unit_test/option_prepare_test.py
mayeut: if we want to add specific tests for free threaded python, let's do that in a following PR.

EwoutH:
Ah I missed that one, thanks! So far it all looks green.

EwoutH:
There we go, all green!

henryiii (approved):

EwoutH:
Thanks for merging!

In what timeframe do you think a new release would be possible?

henryiii:
If it depends on me, tonight.

rgommers:
I found that on macOS and Windows it is still necessary to add `CIBW_PRERELEASE_PYTHONS: True`. Just thought I'd ask here - was that intended?

henryiii:
No, it shouldn’t be

henryiii:
Boost-histogram is building 3.13 windows wheels without the flag, just checked. Is this 3.13 in general or maybe 3.13t? We will be releasing a new patch release of pybind11 with PyMutex, then I’ll be trying that in boost-histogram.

joerick:
Pyinstrument is building 3.13 without the flag, too.

rgommers:
Okay, never mind then, it's probably something else (even if adding the env var fixed it for me). And it was regular `cp313`. For the record:
- [failing run](https://github.com/rgommers/pywt/actions/runs/10342715488)
- [passing run](https://github.com/rgommers/pywt/actions/runs/10342999180)
- [commit that made the difference](https://github.com/rgommers/pywt/commit/0f2eebb603a6741a8c2615f8a6ed42780d6cc6f3)

joerick:
You'll need v2.20.0 , you're on 2.19.0 :)

henryiii:
(2.20 is also the one with the ABI stable Python 3.13rc1)

rgommers:
Argh, made the updates to 2.20 but missed one indeed, for Windows. And the macOS ones were free-threaded. Sorry for the noise.

EwoutH:
Now that everyone had about 3 months to get their Python 3.13 wheels building, and 3.13 was released as stable, would this be a good time to talk about building Python 3.13 free-threaded (`cp313t`) wheels by default?

henryiii:
One thing that might have improved the experience would be if we could detect and enable automatically if the free-threaded classifiers were present. However, the free-threaded classifiers discussion devolved into bike-shedding and died. https://github.com/pypa/trove-classifiers/issues/177 & https://discuss.python.org/t/free-threading-trove-classifier/62406

PR #1974, opened 2024-08-16
chore(ci): podman is now pre-installed on GHA runners

mayeut:
remove specific code to install podman (which does not require specific APT sources since ubuntu 22.04)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,17 +52,6 @@ jobs:
 
     - uses: yezz123/setup-uv@v4
 
-    # Install podman on this CI instance for podman tests on linux
-    # Snippet from: https://github.com/redhat-actions/podman-login/blob/main/.github/workflows/example.yml
-    - name: Install latest podman
-      if: runner.os == 'Linux'
-      run: |
-        . /etc/os-release
-        echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
-        curl -sSfL "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key" | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -52,17 +52,6 @@ jobs:
...
-    # Install podman on this CI instance for podman tests on linux
-    # Snippet from: https://github.com/redhat-actions/podman-login/blob/main/.github/workflows/example.yml
-    - name: Install latest podman
-      if: runner.os == 'Linux'
-      run: |
-        . /etc/os-release
-        echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
-        curl -sSfL "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key" | sudo apt-key add -
-        sudo apt-get update
-        sudo apt-get -y install podman
```
henryiii, line 64: Where is Podman getting installed now?

mayeut: It gets installed from ubuntu repos directly. This is the [recommended (only?) way of installing podman on ubuntu](https://podman.io/docs/installation#ubuntu) since ubuntu 20.10. There are no updates for 22.04 in the custom repo and the repo for ubuntu 24.04 doesn't even exist (which will fail once ubuntu-24.04 gets tagged ubuntu-latest).

henryiii: I just didn’t see the apt install line below for it.

mayeut: It's pre-installed on GHA runners: [images/ubuntu/scripts/build/install-container-tools.sh](https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/install-container-tools.sh)

henryiii (approved):

.github/workflows/test.yml
henryiii: I missed the title...

PR #1980, opened 2024-08-22
mac: sanitize rpaths by default

minrk:
it is opt-in in delocate-wheel (I'd say it should probably be default there, too), but should probably be the default to avoid leaking CI paths into rpath of shipped binaries

This is consistent with default behavior of auditwheel.

upstream issue [suggests](https://github.com/matthew-brett/delocate/issues/101#issuecomment-1822178608) tools like cibuildwheel opt-in by default.

### DIFF
--- a/cibuildwheel/resources/cibuildwheel.schema.json
+++ b/cibuildwheel/resources/cibuildwheel.schema.json
@@ -759,7 +759,7 @@
             }
           ],
           "title": "CIBW_REPAIR_WHEEL_COMMAND",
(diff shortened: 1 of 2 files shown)


henryiii:
This was added in 3.8+ supporting versions and we match manylinux (so 3.6+ currently).

joerick:
I reckon this would be better added as the default at the delocate side, I made a comment here: https://github.com/matthew-brett/delocate/issues/101#issuecomment-2312005528

minrk:
Makes sense!

PR #1817, opened 2024-05-10
feat: move default `musllinux` build to `musllinux_1_2`

mayeut:
Statistics for musl (mostly Alpine but also some OpenWRT)

BigQuery sql query:
```sql
SELECT
  details.distro.name as distro_name,
  details.distro.version as distro_version,
  details.distro.libc.lib as libc_name,
  details.distro.libc.version as libc_version,
  COUNT(*) as download_count,
FROM `bigquery-public-data.pypi.file_downloads`
WHERE timestamp BETWEEN TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -6 DAY) AND TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL -1 DAY)
  AND details.installer.name = "pip"
  AND details.system.name = "Linux"
  AND file.filename LIKE "%musllinux%"
  AND (details.distro.libc.lib IS NULL OR details.distro.libc.lib != "glibc")
GROUP BY
  distro_name, distro_version, libc_name, libc_version
ORDER BY
  download_count DESC
LIMIT 100
```

Once [reduced](https://docs.google.com/spreadsheets/d/1FHOYgutOLUzZZHjbuDJYsGxIDKsjiUd_47CXzxJ13gE/edit?usp=sharing), this gives:

| version | percentage |
| ------------- | ------- | 
| musllinux_1_2 | 97.5 % |
| musllinux_1_1 | 2.5 % |

If we omit the filter on musllinux wheel downloads, this gives:

| version | percentage |
| ------------- | ------- | 
| musllinux_1_2 | 90.5 % |
| musllinux_1_1 | 9.5 % |

It's worth noting that musl libc 1.1 is EOL and Alpine Linux 3.12 also (support ended 2 years ago, 01 May 2022)

PR #1988, opened 2024-08-27
feat: move default `manylinux` build to `manylinux_2_28`

EwoutH:
### Summary
This PR moves the default `manylinux` build from `manylinux2014` to `manylinux_2_28`. The main motivation behind this is that CentOS 7, on which `manylinux2014` is based, became end-of-life end of June. It also helps projects by defaulting to more performant wheels that have improved support for modern microarchitectures.

### Motivation
The motivation to move the default `manylinux` build from `manylinux2014` to `manylinux_2_28` is two fold: On the one hand there's a necessity and on the other hand an oppertunity.
- CentOS 7 reached end-of-life in June 2024. This is already impacting users, as can be seem by steeply increased reports in the following (tracking) issues:
  - https://github.com/pypa/cibuildwheel/issues/1772
  - https://github.com/pypa/manylinux/issues/1641
- `manylinux_2_28` offers better performance and support for modern architectures, including AVX-512 support, new optimized mathematical functions, new floating point types, AArch64 and PowerPC optimizations, and [much more](https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=154ab22d7ca065af5233406927302bb7f6a66890;hb=3c03baca37fdcb52c3881e653ca392bba7a99c2b).
- Support for `manylinux_2_28` has steadily risen to 92.8% according to the [manylinux timeline](https://mayeut.github.io/manylinux-timeline/), and is expected to increase further over time.
  image
  - According to [glibc versions](https://repology.org/project/glibc/versions), the distro that don't support glibc 2.28 is mainly [Amazon Linux 2](https://aws.amazon.com/amazon-linux-2/), which I think they can handle (and a newer Amazon Linux 2023 image has been available for a while).

### Implementation
- Updated default manylinux image references in `cibuildwheel/options.py` and `cibuildwheel/resources/defaults.toml`
- Modified documentation in `docs/options.md` to reflect the new default
- Adjusted test expectations in `test/utils.py` and `unit_test/option_prepare_test.py`

This PR closely follows the default move to `musllinux_1_2` in #1817.

### Impact
- Projects using cibuildwheel will now build `manylinux_2_28` wheels by default
- Existing configurations specifying `manylinux2014` or earlier will continue to work as before
- Users can still override the default to use other manylinux versions if needed

Closes https://github.com/pypa/cibuildwheel/issues/1772.

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -649,8 +649,8 @@ class Options:
                     )
 
                     if not config_value:
-                        # default to manylinux2014
-                        image = pinned_images["manylinux2014"]
+                        # default to manylinux_2_28
+                        image = pinned_images["manylinux_2_28"]
                     elif config_value in pinned_images:
                         image = pinned_images[config_value]
                     else:
--- a/cibuildwheel/resources/defaults.toml
+++ b/cibuildwheel/resources/defaults.toml
@@ -23,14 +23,14 @@ test-extras = []
 
 container-engine = "docker"
 
-manylinux-x86_64-image = "manylinux2014"
-manylinux-i686-image = "manylinux2014"
-manylinux-aarch64-image = "manylinux2014"
-manylinux-ppc64le-image = "manylinux2014"
-manylinux-s390x-image = "manylinux2014"
-manylinux-pypy_x86_64-image = "manylinux2014"
-manylinux-pypy_i686-image = "manylinux2014"
-manylinux-pypy_aarch64-image = "manylinux2014"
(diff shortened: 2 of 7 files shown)


EwoutH:
The remaining errors are about `manylinux_2_28` image for `"i686"` not existing. There are three ways to resolve this:
- Keep the default `"i686"` build on `manylinux2014`
- Stop building `"i686"` by default, and thus requiring `"i686"` to opt-in with a specific `manylinux` version.

I think for this PR the first options is most logical, while maybe on the long term the second options should be discussed.

PR #2005, opened 2024-09-14
fix: do not use `uv` to setup python on windows when conditions are not met

mayeut:
`uv` was still used in `setup_python` on Windows even when `use_uv` was `False`.
This PR fixes `setup_python` to fallback to `build` frontend name instead of `build[uv]` in that case.
Added `can_use_uv` function to make sure wether or not `uv` is usable in a single place rather than duplicate the checks (and potentially forgetting one out of two if the conditions were to change).

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -193,6 +193,11 @@ def install_pypy(tmp: Path, url: str) -> Path:
     return installation_path / "bin" / "pypy3"
 
 
+def can_use_uv(python_configuration: PythonConfiguration) -> bool:
+    conditions = (Version(python_configuration.version) >= Version("3.8"),)
+    return all(conditions)
+
+
 def setup_python(
     tmp: Path,
     python_configuration: PythonConfiguration,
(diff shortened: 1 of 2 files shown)


cibuildwheel/macos.py
```diff
@@ -193,20 +193,23 @@ def install_pypy(tmp: Path, url: str) -> Path:
     return installation_path / "bin" / "pypy3"
 
 
+def can_use_uv(python_configuration: PythonConfiguration) -> bool:
+    conditions = (Version(python_configuration.version) >= Version("3.8"),)
+    return all(conditions)
```
Czaki, line 198: Why not just 
```
return Version(python_configuration.version) >= Version("3.8")
```

Czaki (approved):
Does the Linux not need similar refactor?

cibuildwheel/macos.py
mayeut: Because I used the same layout as on Windows for consistency where using a conditions tuple is more readable.

mayeut:
> Does the Linux not need similar refactor?

Linux could if we want to push consistency even further but given it's used in only one place on Linux, the risk of modifying one place and forgetting the other does not exist there.

cibuildwheel/macos.py
Czaki: ok.

joerick (approved):

PR #2006, opened 2024-09-14
fix: file ownership of files copied into the container

### LINKED ISSUE #2004: File ownership not preserved when copying the project inside the container
### Description

File ownership not preserved when copying the project inside the container

Regression introduced by https://github.com/pypa/cibuildwheel/pull/1961

Originally reported by @cdce8p  in https://github.com/pypa/cibuildwheel/pull/1961#issuecomment-2349484355

FYI This change causes a build failure for mypy on `manylinux` and `musllinux` with Github actions and `ubuntu-latest`.
https://github.com/mypyc/mypy_mypyc-wheels/actions/runs/10852483095/job/30118641732#step:4:461

For the build, the mypy repo is first cloned in a previous step and cibuildwheel subsequently run with `package-dir: mypy`.

The error
```
fatal: detected dubious ownership in repository at '/project/mypy'
```

--
I was able to work around it by adding this to `linux.before-all`:
```
git config --global --add safe.directory {project}/mypy
```
The link to the PR: https://github.com/mypyc/mypy_mypyc-wheels/pull/82

### Build log

_No response_

### CI config

_No response_

mayeut:
fix file ownership of files copied into the container

Use `docker cp -a` to copy files into the container.
This requires to create the container with `--user` for the command not to fail.

fix #2004
based on #2005

### DIFF
--- a/cibuildwheel/macos.py
+++ b/cibuildwheel/macos.py
@@ -193,6 +193,11 @@ def install_pypy(tmp: Path, url: str) -> Path:
     return installation_path / "bin" / "pypy3"
 
 
+def can_use_uv(python_configuration: PythonConfiguration) -> bool:
+    conditions = (Version(python_configuration.version) >= Version("3.8"),)
+    return all(conditions)
+
+
 def setup_python(
     tmp: Path,
     python_configuration: PythonConfiguration,
@@ -200,13 +205,11 @@ def setup_python(
     environment: ParsedEnvironment,
     build_frontend: BuildFrontendName,
 ) -> tuple[Path, dict[str, str]]:
-    if build_frontend == "build[uv]" and Version(python_configuration.version) < Version("3.8"):
+    if build_frontend == "build[uv]" and not can_use_uv(python_configuration):
         build_frontend = "build"
 
     uv_path = find_uv()
-    use_uv = build_frontend == "build[uv]" and Version(python_configuration.version) >= Version(
-        "3.8"
-    )
+    use_uv = build_frontend == "build[uv]"
 
     tmp.mkdir()
     implementation_id = python_configuration.identifier.split("-")[0]
@@ -415,9 +418,7 @@ def build(options: Options, tmp_path: Path) -> None:
         for config in python_configurations:
             build_options = options.build_options(config.identifier)
             build_frontend = build_options.build_frontend or BuildFrontendConfig("pip")
-            use_uv = build_frontend.name == "build[uv]" and Version(config.version) >= Version(
-                "3.8"
-            )
+            use_uv = build_frontend.name == "build[uv]" and can_use_uv(config)
             uv_path = find_uv()
             if use_uv and uv_path is None:
                 msg = "uv not found"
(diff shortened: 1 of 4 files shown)


mayeut:
Related issues in docker:
https://github.com/moby/moby/issues/45793
https://github.com/moby/moby/issues/41727

There's also the fact that when using an image not specifying a user, it does not fallback to root as it should (hence the added `--user`).

For the `--user` arg added, it could conflict with the default user in custom images or an engine config option set by the cibuildwheel user so, it would in fact probably require:
- checking wether or not the engine config sets this
- if not, checking wether or not the image sets this
- if not, add `--user`

Even if doing this, docker might fix the reversed behavior of `-a` which would break again cibuildwheel depending on docker version.

Given all that I'm not sure whether or not we should just reverse to using `tar` ?

joerick:
Thanks for pulling these options together. I meant to check the permissions thing with the `docker cp` PR and I confess I forgot.

It's crazy to me that `docker cp` would retain UIDs by default. I suppose most users run their containers as root and therefore don't see the issue.

Regarding adding `--user` as an option into `oci_container.py`, the only place I've seen this mentioned is [this discussion thread](https://github.com/pypa/cibuildwheel/discussions/1808), where a user wanted to run their container as a different user, because there was some script in their build complaining about running as root. That issue _wasn't_ solved using `--user 1001` because oci_container.py writes to some locations like `/project` and `/constraints.txt` that wouldn't be writable by any user other than root. So I think that it's very unlikely we have any instances of users running their containers with the primary user as anything other than root.

> Even if doing this, docker might fix the reversed behavior of `-a` which would break again cibuildwheel depending on docker version.

I'm very confused by this... according to the documentation [documented](https://docs.docker.com/reference/cli/docker/container/cp/) by Docker:

> The cp command behaves like the Unix cp -a command in that directories are copied recursively with permissions preserved if possible. Ownership is set to the user and primary group at the destination. For example, files copied to a container are created with UID:GID of the root user. Files copied to the local machine are created with the UID:GID of the user which invoked the docker cp command. However, if you specify the -a option, docker cp sets the ownership to the user and primary group at the source.

By my reading, this paragraph contradicts itself "copied recursively with permissions preserved if possible" versus "Ownership is set to the user and primary group at the destination".

So yes, this does seem like it's not set in stone. Having said that, I can't imagine they would change the _de facto_ behaviour because it would break so many workflows. More likely, they'll fix the documentation.

joerick:
I definitely think, regardless of how we get there, that files in the container should be owned by the primary user, and once copied out of the container they should be owned by the user running cibuildwheel. 

I'm slightly torn on these options, right now, given the strangeness of the `docker cp`, regretfully I think I am more inclined to go back to the tar pipes setup we had before. Where is your preference @mayeut ?

mayeut:
> I'm slightly torn on these options, right now, given the strangeness of the docker cp, regretfully I think I am more inclined to go back to the tar pipes setup we had before. Where is your preference mayeut ?

I also think that going back to the tar pipes setup we had before is what we should do for now.
Just as a side note, I did not restore the tar pipe in `copy_out` as `docker cp` works properly in this case (with or without `-a`) in #2007. 
I will close this PR in favor of #2007

PR #2007, opened 2024-09-15
fix: file ownership of files copied into the container

### LINKED ISSUE #2004: File ownership not preserved when copying the project inside the container

mayeut:
Revert to using `cat`/`tar` to copy files/folders into the container.

Alternative to #2006 
Fix #2004

Based on #2005


joerick:
i'll rebase this on main, now that #2005 is in

henryiii:
Why still draft? Is there something left?

joerick:
I don't think so. I suspect it was set to draft while we were deciding between this and #2006 .

henryiii:
Ah, okay

joerick:
Thanks for the RtD fix - looks like my rebase missed it by seconds! nvm, we can ignore that failure.

henryiii (approved):

PR #2010, opened 2024-09-16
ci: modernize update workflow

henryiii:
* Use official action
* Sign commits (v7 feature)


mayeut (approved):
Nice, I did not read the change log of peter-evans/create-pull-request for v7 until now. I'll have to report this in a bunch of other repos then.

PR #2011, opened 2024-09-16
docs: bump pinned versions

henryiii:
Trying to fix ci; ~~might need to add `pathspec` if this isn't enough~~ (this did fix it).


joerick (approved):
i've had a good look through the PR build and it doesn't look like the theme css changes have broken anything, which is lucky!

PR #2012, opened 2024-09-16
ci: update to astral-sh/setup-uv

henryiii:
I think we need a fix here due to uv not handling pre-releases if `VIRTUAL_ENV` is used (I think this a bug with https://github.com/astral-sh/uv/pull/7300). Edit: change in behavior requires `--python`, which we were missing on Linux. Pushed fix to #2008.


henryiii:
Actually, since I included this in #2008, let's just merge it there.

PR #2017, opened 2024-09-22
feat: add musllinux armv7l

### LINKED ISSUE #1421: arm32 / armv7l wheels
### Description

Is there a way to build for arm32v7?

I looked at the documentation, but I could only see references to arm64. 

Specifically, I'm looking to build for arm32 on alpine, with Python 3.11.

Thanks in advance!

mayeut:
close #1421 

Only musllinux was asked in the description of the linked issue but the title is broader so we may want to rework the title or not close the issue.

Now that the multi-arch images support has landed, workarounds to build armv7l wheels can't be used anymore so getting musllinux armv7l also allows to get workarounds working again to build manylinux armv7l if need be.


cibuildwheel/oci_container.py
```diff
@@ -226,13 +227,14 @@ def __enter__(self) -> Self:
...
-                ).strip()
-            simulate_32_bit = container_machine != "i686"
+                if self.oci_platform == OCIPlatform.i386:
+                    # The image might have been built with amd64 architecture
+                    # Let's try that
+                    platform_args = self._get_platform_args(oci_platform=OCIPlatform.AMD64)
+                    container_machine = call(
+                        *run_cmd, *platform_args, self.image, *ctr_cmd, capture_stdout=True
+                    ).strip()
+            simulate_32_bit = container_machine not in {"i686", "armv7l", "armv8l"}
```
joerick, line 237: What's `armv8l` doing here? Remnant from testing or something required?

joerick (approved):

cibuildwheel/oci_container.py
mayeut: this comes from the fact that a 32 bit kernel will report `armv7l` but a 64 bit kernel running linux32 emulation in an armv7 image will report `armv8l`.

mayeut:
I added another commit as this was failing for local armv7l images

cibuildwheel/oci_container.py
```diff
@@ -182,13 +183,16 @@ def _get_platform_args(self, *, oci_platform: OCIPlatform | None = None) -> tupl
...
+            assert components in {2, 3}
+            variant = "/{{.Variant}}" if components == 3 else ""
             image_platform = call(
                 self.engine.name,
                 "image",
                 "inspect",
                 self.image,
                 "--format",
-                "{{.Os}}/{{.Architecture}}",
+                "{{.Os}}/{{.Architecture}}" + variant,
```
joerick, line 195:
```suggestion
                (
                    "{{.Os}}/{{.Architecture}}/{{.Variant}}"
                    if len(oci_platform.value.split("/")) == 3
                    else "{{.Os}}/{{.Architecture}}"
                ),
```

cibuildwheel/oci_container.py
```diff
@@ -182,13 +183,16 @@ def _get_platform_args(self, *, oci_platform: OCIPlatform | None = None) -> tupl
         # c.f. https://github.com/moby/moby/issues/48197#issuecomment-2282802313
         pull = "always"
         try:
+            components = len(oci_platform.value.split("/"))
+            assert components in {2, 3}
+            variant = "/{{.Variant}}" if components == 3 else ""
```
joerick, line 188: think this can be tidied up a little

```suggestion
```

2bndy5:
> this comes from the fact that a 32 bit kernel will report `armv7l` but a 64 bit kernel running linux32 emulation in an armv7 image will report `armv8l` so it's required to build armv7l on aarch64 runners.

So, to use this, I would need qemu setup for aarch64?
```yml
      - name: Set up QEMU
        uses: docker/setup-qemu-action@v3
        with:
          platforms: aarch64

      - name: Build wheels with cibuildwheels
        uses: pypa/cibuildwheel@v2.21.2
        env:
          CIBW_ARCHS_LINUX: aarch64 native armv7l
```
It would be nice to have an example for armv7l in the docs somewhere, just to be more explicit than an implicit mention in the changelog.

PR #2027, opened 2024-10-01
chore: remove a couple of things pylint is unhappy about

henryiii:
We are getting annotations, removing the issues.


joerick:
Nice, thanks!

PR #2037, opened 2024-10-08
fix: use bash instead of sh on musllinux

### LINKED ISSUE #2036: Inconsistent treatment of curly braces in `test-command`
### Description

I have the following `pyproject.toml` setup
```toml
[tool.cibuildwheel]
test-requires = "pytest"
test-command = "pytest {project}/tests/python/test_{hogpp,repr}.py"
```
which works in `manylinux` builds with the following (relevant) output:
```console
Successfully installed exceptiongroup-1.2.2 iniconfig-2.0.0 packaging-24.1 pluggy-1.5.0 pytest-8.3.3 tomli-2.0.2
    + mkdir -p /tmp/tmp.U5BgBhGzbO/test_cwd
    + sh -c 'pytest /project/tests/python/test_{hogpp,repr}.py'
============================= test session starts ==============================
platform linux -- Python 3.9.20, pytest-8.3.3, pluggy-1.5.0
rootdir: /project
configfile: pytest.ini
collected 1300 items

../../../project/tests/python/test_hogpp.py ..............xxxxxxxxxxxxxx [  2%]
xxx..................................................................... [  7%]
.....................xxxxxx............................................. [ 13%]
.xxxxxx................................................................. [ 18%]
........................xxxxxxxxxxxxxxxxxxxxxxxxxxxx.................... [ 24%]
..........xxx........................................................... [ 29%]
........................................................................ [ 35%]
........................................................................ [ 40%]
........................................................................ [ 46%]
........................................................................ [ 52%]
........................................................................ [ 57%]
........................................................................ [ 63%]
........................................................................ [ 68%]
........................................................................ [ 74%]
........................................................................ [ 79%]
........................................................................ [ 85%]
........................................................................ [ 90%]
........................................................................ [ 96%]
........................................                                 [ 99%]
../../../project/tests/python/test_repr.py ........                      [100%]

======================= 1240 passed, 60 xfailed in 2.76s =======================
```
 but fails in a `musllinux` build (specifically cp39-musllinux_x86_64) as follows:
```console
Successfully installed exceptiongroup-1.2.2 iniconfig-2.0.0 packaging-24.1 pluggy-1.5.0 pytest-8.3.3 tomli-2.0.2
    + mkdir -p /tmp/tmp.AbCKIN/test_cwd
    + sh -c 'pytest /project/tests/python/test_{hogpp,repr}.py'
ERROR: file or directory not found: /project/tests/python/test_{hogpp,repr}.py

============================= test session starts ==============================
platform linux -- Python 3.9.20, pytest-8.3.3, pluggy-1.5.0
rootdir: /tmp/tmp.AbCKIN/test_cwd
collected 0 items

============================ no tests ran in 0.00s =============================

                                                              ✕ 9.73s
Error: Command ['sh', '-c', 'pytest /project/tests/python/test_{hogpp,repr}.py'] failed with code 4.
```
I understand that the expansion of curly braces might be ambiguous here. However, it would be nice to have a consistent behavior across different shells. While the [docs mention](https://cibuildwheel.pypa.io/en/stable/options/#build-skip) the expansion with respect to options and the use of `bracex` (albeit, in a slightly different context), it is not clear what the behavior should actually be.

### Build log

_No response_

### CI config

_No response_

henryiii:
Fix #2036. As an alternate, manylinux could set the default shell on the musllinux images to bash, which would allow us to respect the default shell if someone provided a customize linux. And it would be more consistent across both sets of images. It's present there, but not default. @mayeut, thoughts?

### DIFF
--- a/cibuildwheel/linux.py
+++ b/cibuildwheel/linux.py
@@ -192,7 +192,7 @@ def build_in_container(
             project=container_project_path,
             package=container_package_dir,
         )
-        container.call(["sh", "-c", before_all_prepared], env=env)
+        container.call(["bash", "-c", before_all_prepared], env=env)
 
     built_wheels: list[PurePosixPath] = []
 
@@ -258,7 +258,7 @@ def build_in_container(
                     project=container_project_path,
                     package=container_package_dir,
                 )
-                container.call(["sh", "-c", before_build_prepared], env=env)
+                container.call(["bash", "-c", before_build_prepared], env=env)
 
             log.step("Building wheel...")
 
@@ -320,7 +320,7 @@ def build_in_container(
                 repair_command_prepared = prepare_command(
                     build_options.repair_command, wheel=built_wheel, dest_dir=repaired_wheel_dir
                 )
-                container.call(["sh", "-c", repair_command_prepared], env=env)
+                container.call(["bash", "-c", repair_command_prepared], env=env)
             else:
                 container.call(["mv", built_wheel, repaired_wheel_dir])
 
@@ -365,7 +365,7 @@ def build_in_container(
                     project=container_project_path,
                     package=container_package_dir,
                 )
-                container.call(["sh", "-c", before_test_prepared], env=virtualenv_env)
+                container.call(["bash", "-c", before_test_prepared], env=virtualenv_env)
 
             # Install the wheel we just built
             # Note: If auditwheel produced two wheels, it's because the earlier produced wheel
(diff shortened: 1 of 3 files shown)


sergiud:
Thank you for looking into this issue.

As a possibly less intrusive approach I can only suggest explicitly setting the default shell in the Docker image (e.g., by symlinking `/bin/sh` to `/bin/bash`). However, I don't know whether this approach is better maintainable than explicitly invoking the desired shell.

mayeut:
Changing the default shell in musllinux might be the correct option to get consistency between images.

That being said, cibuildwheel is already requiring bash to be present. I don't know if we're using specific bash syntax somewhere in the code but images are always started with `/bin/bash` so respecting the default shell will require other changes in cibuildwheel:  https://github.com/pypa/cibuildwheel/blob/7940a4c0e76eb2030e473a5f864f291f63ee879b/cibuildwheel/oci_container.py#L263-L281

henryiii:
Since we already use bash in some places, is this a good idea for now, then?

henryiii:
Okay, sounds good. Should we change the existing `/bin/bash` one too?

PR #2043, opened 2024-10-15
linux: use Python 3.9 in before_all

mayeut:
3.8 is EOL, move to 3.9

Also use Python 3.9 for utility python in oci_container.

As a side note, 3.9 is the last version available across all images being used in cibuildwheel tests.


Czaki (approved):

henryiii:
>  3.9 is the last version available across all images being used in cibuildwheel tests.

Interesting. So we might be faced with some choices next year.

Issue #2047, opened 2024-10-19
Next release versioning proposal


henryiii:
### Description

I'd like to make a proposal for the versioning strategy and contents of the next few cibuildwheel versions. Here's the suggestion:

The next release (2.22) has the following features:

* #2045 - But the default set of versions is the same if you don't set `enable`. We can warn if prerelease or free-threading standalone options are set.
* Warning for Python < 3.11 as host.

The following release of cibuildwheel I propose we call "3.0", with the following features:

* More `enable` groups, and removing the prerelease and free-threading standalone options.
* #1912

After this point, we can shift our numbering up one, with major releases for adding/removing Pythons, minor release for features, and patch releases for bug fixes. This would allow the much-requested `vX` GHA tag to be something we can provide.

Thoughts?

joerick:
> The following release of cibuildwheel I propose we call "3.0", with the following features (not a complete list):
> 
> * More `enable` groups, and removing the prerelease and free-threading standalone options.
> * [feat: use SPEC 0 schedule for cibuildwheel #1912](https://github.com/pypa/cibuildwheel/pull/1912)

This sounds good. It especially works well with the introduction of the `enable` option - we can make some braver decisions there, such as dropping pypy from defaults, or changing the policy on building non-eol pythons by default if we like.

> * New version of [Split file utils into dedicated submodule #1873](https://github.com/pypa/cibuildwheel/pull/1873)

Ah yes. I'm happy to take a pass at this if you haven't started already?

> After this point, we can shift our numbering up one, with major releases for adding/removing Pythons, minor release for features, and patch releases for bug fixes. This would allow the much-requested `vX` GHA tag to be something we can provide.

Here, I'm less sure. Personally, I see the *compatibility* part of our versioning system to be related to the package and configuration, not so much the Python versions that we build. That's my logic behind not being too worried about bumping major when dropping an old version. I see the specific versions of wheels produced more of an side-effect of the time that version of cibuildwheel was released, not part of the API contract.

Having said all that, we don't follow Semantic versioning anyway, so maybe the specific definition of compatibility is moot.

The bigger point is - do users need to care? When I see a package I use bump major, I'll assume there might be some work I need to do to update - read the release notes, maybe update the config files as a result. But I don't think that applies for Python addition/removal updates. The Python additions require no work (assuming you've already made the package compatible during the beta phase), and the removals are probably so old as to not be of interest (we're still building 3.6!).

[...]

PR #2050, opened 2024-10-20
chore: add warning when running cibuildwheel with python<3.11

mayeut:
towards #2047

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
 What does it do?
 ----------------
 
+While cibuildwheel itself requires a recent enough python version to run, it can target the following versions to build wheels:
+
 |                | macOS Intel | macOS Apple Silicon | Windows 64bit | Windows 32bit | Windows Arm64 | manylinux<br/>musllinux x86_64 | manylinux<br/>musllinux i686 | manylinux<br/>musllinux aarch64 | manylinux<br/>musllinux ppc64le | manylinux<br/>musllinux s390x | musllinux armv7l | Pyodide |
 |----------------|----|-----|-----|-----|-----|----|-----|----|-----|-----|---|-----|
 | CPython 3.6    | ✅ | N/A | ✅  | ✅  | N/A | ✅  | ✅  | ✅ | ✅  | ✅  | ✅ | N/A |
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
 What does it do?
 ----------------
 
-While cibuildwheel itself requires a recent enough python version to run, it can target the following versions to build wheels:
+While cibuildwheel itself requires a recent Python version to run (we support the last three releases), it can target the following versions to build wheels:
 
 |                | macOS Intel | macOS Apple Silicon | Windows 64bit | Windows 32bit | Windows Arm64 | manylinux<br/>musllinux x86_64 | manylinux<br/>musllinux i686 | manylinux<br/>musllinux aarch64 | manylinux<br/>musllinux ppc64le | manylinux<br/>musllinux s390x | musllinux armv7l | Pyodide |
 |----------------|----|-----|-----|-----|-----|----|-----|----|-----|-----|---|-----|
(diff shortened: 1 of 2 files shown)


cibuildwheel/__main__.py
```diff
@@ -402,6 +402,15 @@ def get_build_identifiers(
...
     warnings = []
 
+    python_version_deprecation = ((3, 11), 3)
+    if sys.version_info[:2] < python_version_deprecation[0]:
+        python_version = ".".join(map(str, python_version_deprecation[0]))
+        msg = (
+            f"cibuildwheel {python_version_deprecation[1]}+ will require Python {python_version}+, "
+            "please upgrade your python version."
+        )
+        warnings.append(msg)
```
mayeut, line 412: warnings are just printed (i.e. not using `log.warning`).
This might end up with warnings not being seen easily (mostly on GHA where warnings are reported outside of the logs themselves).
Should `print_preamble` be updated to use `log.warning` ? (or add the correct prefix for GHA) ?

mayeut: Done.
It used to look like:
```
...
 test_requires: 
    -r
    requirements-test.txt

Cache folder: /Users/Matt/Library/Caches/cibuildwheel

Warnings:
  cibuildwheel 3+ will require Python 3.11+, please upgrade your python version.
  test_command: '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.
  before_build: '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.

Here we go!
```

It now looks like:
```
...
  test_requires: 
    -r
    requirements-test.txt

Cache folder: /Users/Matt/Library/Caches/cibuildwheel

Warning: cibuildwheel 3+ will require Python 3.11+, please upgrade your python version.

Warning: test_command: '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.

Warning: before_build: '{python}' and '{pip}' are no longer needed, and will be removed in a future release. Simply use 'python' or 'pip' instead.

Here we go!
```

henryiii (approved):

cibuildwheel/__main__.py
```diff
@@ -402,6 +401,15 @@ def get_build_identifiers(
 def detect_warnings(*, options: Options, identifiers: Iterable[str]) -> list[str]:
     warnings = []
 
+    python_version_deprecation = ((3, 11), 3)
+    if sys.version_info[:2] < python_version_deprecation[0]:
+        python_version = ".".join(map(str, python_version_deprecation[0]))
+        msg = (
+            f"cibuildwheel {python_version_deprecation[1]}+ will require Python {python_version}+, "
+            "please upgrade your python version."
```
henryiii, line 409:
```suggestion
            "please upgrade your Python version used to run cibuildwheel."
```

cibuildwheel/__main__.py
henryiii: If you want to change it while touching this, we could say it will be removed in version 3. In version 3, we can detect this and throw an error instead, perhaps?

cibuildwheel/__main__.py
Czaki: looks good.

cibuildwheel/__main__.py
mayeut: > In version 3, we can detect this and throw an error instead, perhaps?

`requires-python` should prevent such a thing ?
Maybe it could  happen in some edge cases but given it's checked after all option parsing, I guess it would fail before the version check.

henryiii: Oh, sorry, I meant the `'{python}' and '{pip}' are no longer needed` part.

mayeut: It makes much more sense now.
Yes we can update those to say it'll be remove in cibuildwheel 3 while we're at it. It's probably a good thing to error out in 3.x for a while at least before removing the check (if we remove it).

cibuildwheel/__main__.py
mayeut: I'm not completely sure the commit I just added is the best way to do that but that's the best I could think of.
Having the full matrix of targets just below the sentence "While cibuildwheel itself requires a recent enough python version to run, it can target the following versions to build wheels", I don't think there can be any confusion left.

README.md
```diff
@@ -22,6 +22,8 @@ Python wheels are great. Building them across **Mac, Linux, Windows**, on **mult
 What does it do?
 ----------------
 
+While cibuildwheel itself requires a recent enough python version to run, it can target the following versions to build wheels:
```
henryiii, line 25:
```suggestion
While cibuildwheel itself requires a recent Python version to run (we support the last three releases), it can target the following versions to build wheels:
```

henryiii (approved):

PR #2048, opened 2024-10-19
feat: adding enable

henryiii:
Close https://github.com/pypa/cibuildwheel/issues/2045. Adds `enable`, deprecates the old way of specifying these (remove in 3.0).

PR #2053, opened 2024-10-21
chore: stricter mypy

henryiii:
I was looking at updating #1873 (and I realized it would be better to wait till after #1912) and I noticed we weren't very sensitive to updating things like `util.call` -> `utils.cmd.call` in our tests; I'd expect mypy to be able to detect all these without wasting time to run things. I also noticed this in #2048, where the tests weren't noticing the change to `enable`. These improvements should get proper mypy "coverage" of our tests. We still don't have 100% typing for arguments, but things like not allowing untyped function calls should require helper functions fully typed.

- **chore: improve mypy** (setuptools, pytest added, simpler strict setting, dropped unreachable since it depends on platform)
- **chore(types): type functions in tests**
- **chore(types): No partial types in tests**


.pre-commit-config.yaml
```diff
@@ -32,6 +32,7 @@ repos:
       - nox
       - packaging
       - pygithub
+      - pytest
```
Czaki, line 35: Does having so many non pinned dependencies do not make pre-commit mypy fragile?

Czaki (approved):
Looks good

.pre-commit-config.yaml
henryiii: A little (though pytest isn't too bad). Though generally if something changes we just adapt to the change.

Pre-commit doesn't have a way to pin these. Maybe eventually uv will.

Czaki: > Pre-commit doesn't have a way to pin these. Maybe eventually uv will.

What did you mean by `uv` will? pre-commit.ci do not allow to pass any constraints. 

The only possible way that I see is to have own CI job to bump constraints.

henryiii: I mean if uv adds multiple independent locked environments (and tasks), we could use that to move this check to uv.

Czaki: in napari (https://github.com/napari/napari) we use uv to compile a constraints file and call it from tox. 

Maybe the same approach could be used here?

henryiii: If this becomes too much of a problem, we could probably do something (setting it up with nox would be easy). But for now it's fine, and long term we might end up moving from mypy to red-knot anyway.

PR #2064, opened 2024-10-30
chore: use dependency-groups

henryiii:
Using dependency groups for cibuildwheel itself.


.github/workflows/test.yml
```diff
@@ -137,12 +137,13 @@ jobs:
...
         python-version: "3.x"
+    - uses: astral-sh/setup-uv@v3
     - name: Install dependencies
-      run: python -m pip install ".[test]"
+      run: uv sync
     - name: Get qemu emulated architectures
       id: archs
       run: |
-        OUTPUT=$(python -c "from json import dumps; from test.utils import EMULATED_ARCHS; print(dumps(EMULATED_ARCHS))")
+        OUTPUT=$(.venv/bin/python -c "from json import dumps; from test.utils import EMULATED_ARCHS; print(dumps(EMULATED_ARCHS))")
```
joerick, line 146: Does `uv run python` work here perhaps?

```suggestion
        OUTPUT=$(uv run python -c "from json import dumps; from test.utils import EMULATED_ARCHS; print(dumps(EMULATED_ARCHS))")
```

henryiii: Screenshot 2024-10-31 at 8 41 52 AM

🤦

henryiii:
Cirrus: Failed to start an instance: FAILED_PRECONDITION: Monthly compute limit exceeded!

.github/workflows/test.yml
```diff
@@ -67,11 +67,11 @@ jobs:
 
     - name: Install dependencies
       run: |
-        uv pip install --system ".[test]"
+        uv sync
```
henryiii, line 70:
```suggestion
        uv sync --no-dev --group test
```

.github/workflows/test.yml
```diff
@@ -137,12 +137,13 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: "3.x"
+    - uses: astral-sh/setup-uv@v3
     - name: Install dependencies
-      run: python -m pip install ".[test]"
+      run: uv sync
```
henryiii, line 142:
```suggestion
      run: uv sync --no-dev --group test
```

.github/workflows/test.yml
```diff
@@ -159,14 +160,15 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: "3.x"
+    - uses: astral-sh/setup-uv@v3
     - name: Install dependencies
-      run: python -m pip install ".[test,uv]"
+      run: uv sync
```
henryiii, line 165:
```suggestion
      run: uv sync --no-dev --group test
```

.github/workflows/test.yml
```diff
@@ -179,14 +181,14 @@ jobs:
       name: Install Python 3.12
       with:
         python-version: '3.12'
+    - uses: astral-sh/setup-uv@v3
 
     - name: Install dependencies
-      run: |
-        python -m pip install ".[test]"
+      run: uv sync
```
henryiii, line 187:
```suggestion
      run: uv sync --no-dev --group test
```

requirements-dev.txt
henryiii, line 1: We can generate this file with `uvx dependency-groups dev --output requirements-dev.txt`, FYI.

mayeut (changes requested):
This will require to wait for a new nox release as modifications are also needed in the noxfile that require non yet released features of nox.

henryiii:
Nox is just Python, and TOML support was added some time ago. So it's not required to wait for a nox release. I've updated the noxfile.

henryiii:
Okay to go in?

.cirrus.yml
```diff
@@ -2,7 +2,8 @@ only_if: changesInclude('.cirrus.yml') || ($CIRRUS_BRANCH == "main" && !changesI
 
 run_tests: &RUN_TESTS
   install_cibuildwheel_script:
-    - python -m pip install -e ".[dev]" pytest-custom-exit-code
+    - python -m pip install dependency-groups
+    - python -m dependency_groups dev | xargs python -m pip install -e. pytest-custom-exit-code
```
mayeut, line 6:
```suggestion
    - python -m dependency_groups test | xargs python -m pip install -e.
```
We should probably align everything to use `test` rather than `dev`.
Is `pytest-custom-exit-code` really needed on some CI ? The usage section on PyPI seems to indicate that it requires to call `pytest` with specific args to get it to do something but we're not (and nothing in the conf either unless I missed something ?).

.travis.yml
```diff
@@ -57,8 +57,8 @@ jobs:
 
 install:
 - if [ "${TRAVIS_OS_NAME}" == "linux" ]; then docker run --rm --privileged docker.io/tonistiigi/binfmt:latest --install all; fi
-- $PYTHON -m pip install -U pip
-- $PYTHON -m pip install -e ".[test]" pytest-custom-exit-code
+- $PYTHON -m pip install -U pip dependency-groups
+- $PYTHON -m dependency_groups test | xargs $PYTHON -m pip install -e. pytest-custom-exit-code
```
mayeut, line 61:
```suggestion
- $PYTHON -m dependency_groups test | xargs $PYTHON -m pip install -e.
```

azure-pipelines.yml
```diff
@@ -14,7 +14,8 @@ jobs:
         versionSpec: '3.8'
     - bash: |
         docker run --rm --privileged docker.io/tonistiigi/binfmt:latest --install all
-        python -m pip install -e ".[dev]"
+        python -m pip install dependency-groups
+        python -m dependency_groups dev | xargs python -m pip install -e.
```
mayeut, line 18:
```suggestion
        python -m dependency_groups test | xargs python -m pip install -e.
```

azure-pipelines.yml
```diff
@@ -24,7 +25,8 @@ jobs:
       inputs:
         versionSpec: '3.8'
     - bash: |
-        python -m pip install -e ".[dev]"
+        python -m pip install dependency-groups
+        python -m dependency_groups dev | xargs python -m pip install -e.
```
mayeut, line 29:
```suggestion
        python -m dependency_groups test | xargs python -m pip install -e.
```

azure-pipelines.yml
```diff
@@ -35,5 +37,6 @@ jobs:
...
```
mayeut, line 41:
```suggestion
        python -m dependency_groups test | xargs python -m pip install -e.
```

mayeut (changes requested):
Should we wait for pip builtin support to keep workarounds to a minimum ?

henryiii:
Using a tool to read the dependency groups doesn't seem too bad to me; pip won't support this until 25.1 at the earliest. But we can wait, this doesn't affect users.

joerick:
Whats the proposed way to setup a dev environment with this? I still use `pip install -e '.[dev]'`. I'm not using uv habitually yet. Maybe the answer is `uv`, but it would be nice to have that documented somewhere. I know I can run tests with nox, but I still like to have a live venv I can point mypy/pyright to from my editor.

(I guess if we waited for pip to have support, that would take away this question)

henryiii:
`uv` doesn't support this via `pip install *` yet. The uv way to set this up would be to run `uv sync`. That will make a `.venv` environment that has an editable install of the project and the `dev` group in it, and vscode, etc. will find it automatically. Also all `uv run` commands will do this for you too. `uv run cibuildwheel`, for example, will set up a venv as needed, install dev dependencies as needed, and then run the requested command inside that venv. I've really liked using that everywhere, as it's fast, doesn't require setup, and doesn't require much if any config.

The non-uv way to do this is to pip/pipx install dependency-groups, then use that to generate the list of items you want to install. It also provides a pip wrapper that installs dependency groups, though I like `dependency-groups --group dev | xargs pip install -e.` better. `uvx dependency-groups --group dev | xargs pip install -e.` is a one liner if you have uv. Or `pipx run dependency-groups --group dev | xargs pip install -e.` is a one liner if you have pipx.

joerick:
Thanks @henryiii, I've added that info into contributing.md.

joerick (approved):
I'm okay with this today, but I'm expecting the pip `--group` support to clean this up a but!

henryiii:
Okay, waiting for @mayeut to update review, then.

mayeut (approved):

mayeut:
Thanks @henryiii, all good now.

mayeut:
we missed the gitlab update which only runs on branches in the repo.

PR #2070, opened 2024-11-04
Docs: clarify how multiple values work for various options

akx:
I spent a _long_ while wondering why comma-separated values wouldn't work for `CIBW_SKIP`, since this information was cleverly hidden inside the Examples box:

Screenshot 2024-11-04 at 17 35 45

This PR unifies that for all of the multiple-values-supported options; this was already documented in the prose for the environment variable options, so the extra repetition of the same information is now gone from the example box's footer.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -284,7 +284,11 @@ List of builds to build and skip. Each build has an identifier like `cp38-manyli
 
 When both options are specified, both conditions are applied and only builds with a tag that matches `CIBW_BUILD` and does not match `CIBW_SKIP` will be built.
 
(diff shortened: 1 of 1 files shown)


joerick:
I believe we put them in the tabbed box because that rule only applies to environment variables. In TOML, the options are supplied as a list of strings.

henryiii:
Maybe we could improve the error message?

akx:
> I believe we put them in the tabbed box because that rule only applies to environment variables. 

I could edit this to mention that in the prose.

> Maybe we could improve the error message?

There is no error message – in the case of a malformed `SKIP`, cibuildwheel just tries to build everything since a pattern like `pp*,foo` matches nothing (and in the case of my project, failed on PyPy because I use CPython-specific APIs). 

It would maybe better to "Do What I Mean", i.e. allow e.g. a `[,\s]+` regexp as a separator?

henryiii:
Let's hold off until we move the default to TOML. That might help shape this.

joerick:
> Maybe we could improve the error message?

Hmm. Yeah. We could raise a warning if a BUILD/SKIP item doesn't match any identifier that cibuildwheel knows about.

PR #2071, opened 2024-11-04
tests: fix CIBW_CONFIG_SETTIGNS typo in test

akx:
Caught with https://github.com/crate-ci/typos.

This would probably only be an issue if there was an ambient `CIBW_CONFIG_SETTINGS` envvar that'd conflict with the test.


henryiii:
Thanks!

PR #2002, opened 2024-09-11
Pyodide improvements: version setting, standalone environments

agriyakhetarpal:
## Description

This PR updates the Pyodide build procedure (see #1456) that is enabled with `CIBW_PLATFORM: "pyodide"` (or with the `--platform pyodide` CLI equivalent) post the changes in https://github.com/pyodide/pyodide/pull/4882, where [`pyodide/pyodide-build`](https://github.com/pyodide/pyodide-build) was unvendored from the main Pyodide repository to accommodate faster updates and fixes.

This means that the Pyodide version and `pyodide-build` are not going to be in sync going forward, and that the Pyodide xbuildenv to install must be inferred by the versions available to install by `pyodide-build` through a recently added `pyodide xbuildenv search` command, which prints out this table:

<details>
<summary>Tap to expand table</summary>

<br>

```
┌────────────┬────────────┬────────────┬───────────────────────────┬────────────┐
│ Version    │ Python     │ Emscripten │ pyodide-build             │ Compatible │
├────────────┼────────────┼────────────┼───────────────────────────┼────────────┤
│ 0.27.0a2   │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.4     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.3     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.2     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.1     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.0     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
└────────────┴────────────┴────────────┴───────────────────────────┴────────────┘
```

Alternatively, one may use `pyodide xbuildenv search --all` to return both compatible and non-compatible versions. This would, however, be better received as JSON (please see https://github.com/pyodide/pyodide-build/pull/28).

</details>

<hr>

[...]

PR #2090, opened 2024-11-21
Updates for Pyodide builds after `pyodide-build` was unvendored

agriyakhetarpal:
## Description

This PR has been split off from #2002, where it has been attempted to allow the use of a build-time flag to build and test a Pyodide version that is not available with cibuildwheel, since the cibuildwheel version couples the Pyodide version available within it.

This PR is, rather, more limited in its approach. The key changes here are:
- Post https://github.com/pyodide/pyodide/pull/4882, [`pyodide-build` is now hosted in its own repository](https://github.com/pyodide/pyodide-build), and updates to it are not made in sync with updates to the Pyodide version, allowing `pyodide-build` to move faster with changes to it. Hence, the `pyodide-build` version (currently at 0.29.0) is now what is used to fetch the Pyodide xbuildenv for Pyodide builds.
- Updated to the recently released [Pyodide version `0.26.4`](https://github.com/pyodide/pyodide/releases/tag/0.26.4), which now uses Python 3.12.7 (but we reflect just 3.12 and don't reflect the patch version). The Emscripten version stays the same at 3.1.58 and there is no ABI break.

### DIFF
--- a/cibuildwheel/pyodide.py
+++ b/cibuildwheel/pyodide.py
@@ -66,6 +66,7 @@ def install_emscripten(tmp: Path, version: str) -> Path:
 
 
 def install_xbuildenv(env: dict[str, str], pyodide_version: str) -> str:
+    """Install a particular Pyodide xbuildenv version and set a path to the Pyodide root."""
     pyodide_root = (
         CIBW_CACHE_PATH
         / f".pyodide-xbuildenv-{pyodide_version}/{pyodide_version}/xbuildenv/pyodide-root"
--- a/cibuildwheel/pyodide.py
+++ b/cibuildwheel/pyodide.py
@@ -41,6 +41,7 @@ class PythonConfiguration:
     version: str
     identifier: str
     pyodide_version: str
+    pyodide_build_version: str
     emscripten_version: str
     node_version: str
 
@@ -65,11 +66,15 @@ def install_emscripten(tmp: Path, version: str) -> Path:
     return emcc_path
 
 
-def install_xbuildenv(env: dict[str, str], pyodide_version: str) -> str:
+def install_xbuildenv(env: dict[str, str], pyodide_build_version: str, pyodide_version: str) -> str:
     """Install a particular Pyodide xbuildenv version and set a path to the Pyodide root."""
+    # Since pyodide-build was unvendored from Pyodide v0.27.0, the versions of pyodide-build are
+    # not guaranteed to match the versions of Pyodide or be in sync with them. Hence, we shall
+    # specify the pyodide-build version in the root path, which will set up the xbuildenv for
+    # the requested Pyodide version.
     pyodide_root = (
         CIBW_CACHE_PATH
-        / f".pyodide-xbuildenv-{pyodide_version}/{pyodide_version}/xbuildenv/pyodide-root"
+        / f".pyodide-xbuildenv-{pyodide_build_version}/{pyodide_version}/xbuildenv/pyodide-root"
     )
     with FileLock(CIBW_CACHE_PATH / "xbuildenv.lock"):
         if pyodide_root.exists():
@@ -172,7 +177,9 @@ def setup_python(
     env["PATH"] = os.pathsep.join([str(emcc_path.parent), env["PATH"]])
 
     log.step(f"Installing Pyodide xbuildenv version: {python_configuration.pyodide_version} ...")
-    env["PYODIDE_ROOT"] = install_xbuildenv(env, python_configuration.pyodide_version)
+    env["PYODIDE_ROOT"] = install_xbuildenv(
+        env, python_configuration.pyodide_build_version, python_configuration.pyodide_version
+    )
 
     return env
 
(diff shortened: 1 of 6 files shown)


cibuildwheel/resources/build-platforms.toml
```diff
@@ -190,5 +190,5 @@ python_configurations = [
 
 [pyodide]
 python_configurations = [
-  { identifier = "cp312-pyodide_wasm32", version = "3.12.1", pyodide_version = "0.26.1", emscripten_version = "3.1.58", node_version = "v20" },
+{ identifier = "cp312-pyodide_wasm32", version = "3.12.7", pyodide_version = "0.26.4", pyodide_build_version = "0.29.0", emscripten_version = "3.1.58", node_version = "v20" },
```
henryiii, line 193:
```suggestion
{ identifier = "cp312-pyodide_wasm32", version = "3.12", pyodide_version = "0.26.4", pyodide_build_version = "0.29.0", emscripten_version = "3.1.58", node_version = "v20" },
```

cibuildwheel/resources/build-platforms.toml
```diff
@@ -190,5 +190,5 @@ python_configurations = [
 
 [pyodide]
 python_configurations = [
-  { identifier = "cp312-pyodide_wasm32", version = "3.12.1", pyodide_version = "0.26.1", emscripten_version = "3.1.58", node_version = "v20" },
+{ identifier = "cp312-pyodide_wasm32", version = "3.12", pyodide_version = "0.26.4", pyodide_build_version = "0.29.0", emscripten_version = "3.1.58", node_version = "v20" },
```
henryiii, line 193:
```suggestion
  { identifier = "cp312-pyodide_wasm32", version = "3.12", pyodide_version = "0.26.4", pyodide_build_version = "0.29.0", emscripten_version = "3.1.58", node_version = "v20" },
```

cibuildwheel/pyodide.py
```diff
@@ -65,10 +66,15 @@ def install_emscripten(tmp: Path, version: str) -> Path:
...
+def install_xbuildenv(env: dict[str, str], pyodide_build_version: str, pyodide_version: str) -> str:
+    """Install a particular Pyodide xbuildenv version and set a path to the Pyodide root."""
+    # Since pyodide-build was unvendored from Pyodide v0.27.0, the versions of pyodide-build are
+    # not guaranteed to match the versions of Pyodide or be in sync with them. Hence, we shall
+    # specify the pyodide-build version in the root path, which will set up the xbuildenv for
+    # the requested Pyodide version.
     pyodide_root = (
         CIBW_CACHE_PATH
-        / f".pyodide-xbuildenv-{pyodide_version}/{pyodide_version}/xbuildenv/pyodide-root"
+        / f".pyodide-xbuildenv-{pyodide_build_version}/{pyodide_version}/xbuildenv/pyodide-root"
```
henryiii, line 77: This will not respect overrides, but will always be the one hard-coded. Was that intentional?

agriyakhetarpal: Yes, this is intentional, since we don't have any ways to override here (no environment variables or the like, etc.). We should be fine with the hardcoded version. Users can still override `pyodide-build` using their constraint files if they like, but that won't be of much use. I plan to rebase #2002 after this gets reviewed/merged.

agriyakhetarpal: Ah, just saw your edit. How do we ensure that we use the version of `pyodide-build` that users install themselves, e.g., in `before-build`?

henryiii: Constraint flags apply when it is setting up the environment. If a user installs it themselves, that will happen after the constrained version has already set up a new environment. Do we need the pyodide-build version in the env path?

agriyakhetarpal: I am not sure if I understood that correctly 😅 If I got it right,

> the constrained version has already set up a new environment

Is "environment" here the cibuildwheel venv that will be used for the build? If these changes constrain the `pyodide-build` version in this environment and constraints files will have no effect, I'd definitely like to try out a different method. Could you give me a pointer on how I could proceed?

Also, a side note – after thinking a bit more about this, I think it would be great not to hardcode the Pyodide version, too (not in this PR, though). In case we release a new version of `pyodide-build`, say, X, that drops compatibility with the Pyodide version Y listed in `build-platforms.toml`, overriding the version here with constraints to use X would make it unable to work with Pyodide version Y.

henryiii: I mean the xbuildenv gets set up with the constrained versions, but before-build runs after the xbuildenv is setup, so it can't affect the version used to setup the xbuildenv. After that I believe you can change it, so it's only the pyodide-build version used for the xbuildenv setup that's a little harder to control.

henryiii (approved):

agriyakhetarpal:
Thanks for the review and merge! I'm sorry I didn't see this sooner, and more notifications got in the way later. I think we can think of more solutions as part of #2002 in case users want more flexible constraints. But most of the discussion seems to point to the fact that `cibuildwheel` is still going to remain a bit rigid and https://github.com/pyodide/pyodide-actions/issues/12 could be the way forward if users want more flexibility.

PR #2092, opened 2024-11-22
Bump Ruff to 0.8.0

AlexWaygood:
Hi! I just released Ruff 0.8.0. This release removes `PT004`, which had been deprecated for a few releases, so Ruff 0.8.0 would have started emitting warnings when linting `cibuildwheel` due to the fact that you have `PT004` explicitly ignored in your Ruff config.

This PR gets rid of `PT004` from your Ruff config. There's also some additional changes made here due to the stabilisation of [`unsorted-dunder-all`](https://docs.astral.sh/ruff/rules/unsorted-dunder-all/) (`RUF022`) in this release -- let me know if these are unwelcome, and I can add them to the `tool.ruff.lint.ignore` list.

### DIFF
--- a/cibuildwheel/typing.py
+++ b/cibuildwheel/typing.py
@@ -9,7 +9,6 @@ __all__ = (
     "PLATFORMS",
     "PathOrStr",
     "PlatformName",
-    "PLATFORMS",
     "PopenBytes",
 )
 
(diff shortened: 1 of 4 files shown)


cibuildwheel/typing.py
```diff
@@ -9,7 +9,6 @@
     "PLATFORMS",
     "PathOrStr",
     "PlatformName",
-    "PLATFORMS",
```
AlexWaygood, line 12: Ruff would have just moved this higher up, but I noticed that it actually appears twice in `__all__` here, so I removed the duplicate entry!

henryiii (approved):

PR #2093, opened 2024-11-23
fix: cibuildwheel.schema.json

mayeut:
We missed a final `nox -s generate_schema` before merging  #2048


henryiii:
We should add this in CI (thought it was there, actually)

henryiii (approved):

PR #2095, opened 2024-11-23
chore: drop deprecated options related to `CIBW_ENABLE`

mayeut:
Drop deprecated options related to `CIBW_ENABLE`.

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -30,6 +30,7 @@ from cibuildwheel.util import (
     CIBW_CACHE_PATH,
     BuildSelector,
     CIProvider,
+    EnableGroup,
     Unbuffered,
     detect_ci_provider,
     fix_ansi_codes_for_github_actions,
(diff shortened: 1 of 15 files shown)


cibuildwheel/__main__.py
```diff
@@ -157,12 +156,6 @@ def main_inner(global_options: GlobalOptions) -> None:
         help="Do not report an error code if the build does not match any wheels.",
     )
 
-    parser.add_argument(
-        "--prerelease-pythons",
-        action="store_true",
-        help="Enable pre-release Python versions if available.",
-    )
-
```
mayeut, line 165: Should we be keeping the command line option but turn it in an enable in `options.py` ?

henryiii: Should we instead support `--enable=<opt> --enable=<opt>`? We also now have `--only`, which you can use to build a pre-release wheel, so maybe we don't need a CLI option for this anymore? (It also is supported statically now)

mayeut: Only a few projects are [using this in their CI](https://sourcegraph.com/search?q=%22--prerelease-pythons%22&patternType=keyword&sm=0&__cc=1) and those could probably move to the static definition (despite our recommendation not to deliver wheels in pre-ABI stability, for now at least).

I liked `--prerelease-pythons` better than `--only` because it allows to check musl/glibc, gil/no-gil and multiple architectures in just one pass.

For UNIX users debugging locally, `CIBW_ENABLE=cpython-prerelease cibuildwheel ...` works quite easily so probably not worth adding an `--enable` flag. It might be a bit more painful on Windows if you don't want to leave the environment variable after execution (don't know if there's a one liner there) but we can still probably live without a new option as well.

So, maybe just remove the option for now ?

joerick: I'd be in favour of a `--enable=` command line option, because the ergonomics make sense to me - you could do one cibuildwheel execution for the PyPI artifacts, then another with experimental stuff enabled, that might go to a different place.

joerick: I pushed a commit to add the enable option.

joerick (approved):
Thanks for putting this together!

PR #2098, opened 2024-11-25
feat: add manylinux_2_34

mayeut:
Add support for manylinux_2_34 images

xref: https://github.com/pypa/manylinux/issues/1585


henryiii (approved):

joerick (approved):

Czaki (approved):

henryiii:
Should we get a release out? I secretly just want the docs typo fix, but this is a more valid reason.

joerick:
IMO it's nice to have a bit of 'heft' to a major release, this would contribute to that which would be nice. But if we don't want to wait that long we can put it out now.

henryiii:
I think this is pretty minor (most people won't jump on this for a while) and we have some nice things lined up for 3. And 3 will take a little while, if nothing more than to give people time to upgrade and see the warnings in 2.22 and react.

PR #2099, opened 2024-11-25
ci: warm-up docker images

mayeut:
Let's see if we can speed-up linux CI by warming-up docker images.
This helps locally but I have a very slow Internet connection so it might not apply here.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -86,6 +86,9 @@ def docker_warmup_fixture(
     # if we're in CI testing linux, let's warm-up docker images
     if detect_ci_provider() is None or platform != "linux":
         return None
+    if request.config.getoption("--run-emulation", default=None) is not None:
+        # emulation tests only run one test in CI, caching the image only slows down the test
+        return None
 
     if worker_id == "master":
         # not executing with multiple workers
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -53,15 +53,12 @@ def docker_warmup(request: pytest.FixtureRequest) -> None:
     ]
     for image in images:
         try:
+            command = (
+                "manylinux-interpreters ensure $(manylinux-interpreters list 2>/dev/null | grep -v graalpy) &&"
+                "cpython3.13 -m pip download -d /tmp setuptools wheel pytest"
+            )
             container_id = subprocess.run(
-                [
-                    "docker",
-                    "create",
-                    image,
-                    "bash",
-                    "-c",
-                    "manylinux-interpreters ensure-all && cpython3.13 -m pip download -d /tmp setuptools wheel pytest",
-                ],
+                ["docker", "create", image, "bash", "-c", command],
                 text=True,
                 check=True,
                 stdout=subprocess.PIPE,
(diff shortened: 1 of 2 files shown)


henryiii:
How much slower? Seems like an overall win?

mayeut:
It seems like an overall win, I will do a proper sum-up this week-end once a couple things are fixed like the emulation run on GHA.
TravisCI has ran now and it's 25 minutes less on x64 / 10 minutes less on aarch64

test/conftest.py
```diff
@@ -28,6 +31,77 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+                stdout=subprocess.PIPE,
+            ).stdout.strip()
+            subprocess.run(["docker", "start", container_id], check=True, stdout=subprocess.DEVNULL)
+            exit_code = subprocess.run(
+                ["docker", "wait", container_id], text=True, check=True, stdout=subprocess.PIPE
+            ).stdout.strip()
+            if exit_code == "0":
+                subprocess.run(
+                    ["docker", "commit", container_id, image], check=True, stdout=subprocess.DEVNULL
+                )
```
joerick, line 73: Should we print an error or fail if the command doesn't succeed?

test/conftest.py
```diff
@@ -28,6 +31,77 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    if detect_ci_provider() is None or platform != "linux":
+        return None
+    if request.config.getoption("--run-emulation", default=None) is not None:
+        # emulation tests only run one test in CI, caching the image only slows down the test
+        return None
+
+    if worker_id == "master":
+        # not executing with multiple workers
+        return docker_warmup(request)
+
```
joerick, line 93: Is this early return needed? I can't see the downside of the locking in the single-worker case?

test/conftest.py
```diff
@@ -28,6 +31,77 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+            ).stdout.strip()
+            subprocess.run(["docker", "start", container_id], check=True, stdout=subprocess.DEVNULL)
+            exit_code = subprocess.run(
+                ["docker", "wait", container_id], text=True, check=True, stdout=subprocess.PIPE
+            ).stdout.strip()
+            if exit_code == "0":
+                subprocess.run(
+                    ["docker", "commit", container_id, image], check=True, stdout=subprocess.DEVNULL
+                )
+            subprocess.run(["docker", "rm", container_id], check=True, stdout=subprocess.DEVNULL)
```
joerick, line 74: Might be polite to run this in a `finally:` block

joerick (commented):
A decent improvement! Thanks! Anything we can do to speed up the tests is good for productivity.

test/conftest.py
mayeut: This was taken from https://pytest-xdist.readthedocs.io/en/stable/how-to.html#making-session-scoped-fixtures-execute-only-once

I asked myself the same question but did not add a comment.
I can't see the downside of the locking but I can see one in writing to `tmp_path_factory.getbasetemp().parent` which might not be so temporary anymore in single worker mode.

test/conftest.py
mayeut: Good point. should we print or fail ?
Maybe we should just fail here. I will add a commit doing that (& revert if we decide not to fail but print).

test/conftest.py
mayeut: done in the commit that enforce the warm-up does not fail

joerick (approved):

PR #2108, opened 2024-12-07
download cpython installer to cache dir on macos

trim21:
When download cpython on macos, download it to `CIBW_CACHE_PATH/python/` instead of temp path.

So if user cache CIBW_CACHE_PATH between ci runs, they do not need to download cpython every times.

We have done similar things for pypy, but not cpython.

for example: 

```text
Building cp313t-macosx_x86_64 wheel
CPython 3.13t macOS x86_64
Installing Python cp313t...
  + pkgutil --pkgs
  + Download https://www.python.org/ftp/python/3.13.0/python-3.13.0-macos11.pkg to /Users/runner/Library/Caches/cibuildwheel/python/python-3.13.0-macos11.pkg
  + sudo installer -pkg /Users/runner/Library/Caches/cibuildwheel/python/python-3.13.0-macos11.pkg -applyChoiceChangesXML /Users/runner/work/_temp/cibw/lib/python3.12/site-packages/cibuildwheel/resources/free-threaded-enable-313.xml -target /
  installer: Package name is Python
  installer: choices changes file '/Users/runner/work/_temp/cibw/lib/python3.12/site-packages/cibuildwheel/resources/free-threaded-enable-313.xml' applied
```

close https://github.com/pypa/cibuildwheel/issues/2081

travis ci failed on unrelated os

image


trim21:
@trim21-bot squash

mayeut:
Unless the download of the installer is very slow itself, I would guess that it wouldn't change much (just moving the timings from cibuildwheel download to cache download+extract). We'd still need to install which is probably where most of the time is spent.

Given all those are just guesses, seeing the actual benefits might help reason about this. Do you have runs on one of your projects showing the improvement ?

trim21:
> Unless the download of the installer is very slow itself, I would guess that it wouldn't change much (just moving the timings from cibuildwheel download to cache download+extract). We'd still need to install which is probably where most of the time is spent.
> 
> Given all those are just guesses, seeing the actual benefits might help reason about this. Do you have runs on one of your projects showing the improvement ?
> 
> EDIT: I could confirm my guesses by running with and without installer cache https://github.com/mayeut/cibuildwheel/actions/runs/12221740312/usage

Thanks for your explanation.

I agree, on GitHub CI it has roughly same speeding cache and python pkg, it will only benefit users with low network speed with www.python.org, and saidly I'm one of them. 

for example:

```text
$ curl https://www.python.org/ftp/python/3.10.11/python-3.10.11-macos11.pkg -O
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 10 39.1M   10 4086k    0     0  45543      0  0:15:00  0:01:31  0:13:29 48863
```

But cache can be really fast in internal network with local S3, sorry I can't share a public url for this.

trim21:
I updated the PR title to make it more proper

cibuildwheel/macos.py
```diff
@@ -161,10 +161,10 @@ def install_cpython(tmp: Path, version: str, url: str, free_threading: bool) ->
                     """
                 )
                 raise errors.FatalError(msg)
-            pkg_path = tmp / "Python.pkg"
-            # download the pkg
-            download(url, pkg_path)
-            # install
+            python_filename = url.split("/")[-1]
+            pkg_path = CIBW_CACHE_PATH.joinpath("python", python_filename)
```
joerick, line 165:
```suggestion
            pkg_path = CIBW_CACHE_PATH / "cpython-installer" / python_filename
```

joerick (approved):
Useful to a small target audience, but I don't see a downside, so why not?

PR #2122, opened 2025-01-05
Inline dependency-versions syntax

joerick:
To help with #2002, I propose a way to specify dependency version constraints inline, without the extra file. This would eliminate the need for an additional CIBW_PYODIDE_BUILD_VERSION option, while being consistent with our existing version management system.

This would allow setting the pyodide-build version like:

```yaml
CIBW_DEPENDENCY_VERSIONS_PYODIDE: "packages: pyodide-build==0.29.1"
```

~So far, no implementation yet, just docs for discussion.~


joerick:
Having reviewed #2002 again, I think this would be a useful feature. Specifically, there's a check for compatible versions between pyodide and pyodide-build - if we provide an option for CIBW_PYODIDE_VERSION, and we don't have this, changing pyodide version becomes a bit hassle, because you need to make a constraints file as well.

joerick:
This should be ready to go, pending review.

henryiii:
I've managed to make some conflicts due to #2276. I can rebase if you prefer, sorry!

joerick:
> * `packages = []` would avoid pinning?

Good point. I think that should be equivalent to 'latest', as in it pins nothing. I'll make sure that's covered with a test case.

> * How does this interact with `inherit`? Can you inherit from a file and add a package?

It doesn't inherit. I considered including it but because of the backwards-compatibility requirements[^1] it's awkward to implement (the syntax is different from the other table-style options). Also, I couldn't think of a situation it might be useful. What do you think?


[^1]: the existing string-form (just a path) means that if we start using the table-form, characters like ` ;:` will be parsed differently. I'm sure it could be worked around somehow, but it might require some messy code, didn't seem worth the effort/complexity.

joerick:
This is ready for another review, the empty packages case is now covered.

henryiii (approved):

PR #2135, opened 2025-01-18
ci(gha): add tests on ubuntu-24.04-arm

mayeut:
Run all tests on GitHub Actions Ubuntu 24.04 aarch64 runner.

The emulation tests are kept for now.

We could also run armv7l tests on those same runners but it's not part of `auto` arch because:
1: not all aarch64 systems can run aarch32 binaries (we need EL0 aarch32 support - mac Mx SoC or  Cavium ThunderX do not provide this for exemple).
2: part of the armv7l images are still experimental

If we want to run armv7l, we can probably amend the tests on GHA to request both aarch64 & armv7l architectures.


joerick:
---

> **Examples test run**
> 
> Branch: [example-config-test---gha-arm-2025-01-20T09-04-55](https://github.com/pypa/cibuildwheel/tree/example-config-test---gha-arm-2025-01-20T09-04-55)
> 
> |---|---|---|
> | github | `examples/github-minimal.yml` | [Build](https://github.com/pypa/cibuildwheel/actions) |
> 

---
---

> **Examples test run**
> 
> Branch: [example-config-test---gha-arm-2025-01-20T09-05-21](https://github.com/pypa/cibuildwheel/tree/example-config-test---gha-arm-2025-01-20T09-05-21)
> 
> |---|---|---|
> | github | `examples/github-with-qemu.yml` | [Build](https://github.com/pypa/cibuildwheel/actions) |
> 

---

examples/github-with-qemu.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         env:
           # configure cibuildwheel to build native archs ('auto'), and some
           # emulated ones
```
joerick, line 27:
```suggestion
          # configure cibuildwheel on Linux to build native archs ('auto'),
          # and to split the remaining architectures between the x86_64 and
          # ARM runners
```

I don't love this way of doing ternary expressions, but it's all we have in GHA workflows, so I think adding a comment is nice.

joerick (commented):
Assuming the above goes green, this looks good to me!

examples/github-with-qemu.yml
mayeut: I included your proposal and extended a bit on the `auto armv7l` peculiarity. That might be a bit too verbose but that's strange enough to deserve a also comment (it's "native" in some circumstances hence not in `auto`).

henryiii (approved):

joerick (approved):

henryiii:
Not sure this is a feature. Current cibuildwheel works fine with native ARM, already using it in several projects. :)

examples/github-with-qemu.yml
joerick: This change looks good to me. Question for another PR, but it makes me wonder if we could detect armv7l support and include that in `auto` on machines where that's possible.

mayeut: > if we could detect armv7l support and include that in auto on machines where that's possible.

That might be possible:
- don't allow on platform other that Linux
- try to run a small statically linked Linux exe & check if it works.

README.md
```diff
@@ -59,7 +59,7 @@ Usage
 
 |                 | Linux | macOS | Windows | Linux ARM | macOS ARM | Windows ARM |
 |-----------------|-------|-------|---------|-----------|-----------|-------------|
-| GitHub Actions  | ✅    | ✅    | ✅       | ✅¹       | ✅        | ✅²         |
+| GitHub Actions  | ✅    | ✅    | ✅       | ✅        | ✅        | ✅²         |
```
webknjaz, line 62: Should the actual ¹ text be dropped below as well?

joerick: That footnote still applies to Gitlab, who don't yet offer ARM runners.

pllim:
Will this be released soon? Thanks!

henryiii:
The current version already supports Linux ARM, I’ve moved over several packages like boost-histogram and awkward. This just adds testing and examples.

PR #2259, opened 2025-02-02
feat: add armv7l in auto_archs when running on aarch64

mayeut:
This depends on aarch32 EL0 support and thus is done conditionally.

This uses a `linux32` call to check for support.

Things weren't working well on Travis CI for some reason. Given all examples are using LXD, I updated tests to use that as well and running armv7l  does not seem supported on that infrastructure.

The feature was first requested in https://github.com/pypa/cibuildwheel/pull/2135#discussion_r1924179661

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: focal
+dist: jammy
 language: python
 
 branches:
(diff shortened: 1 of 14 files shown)


mayeut:
Moving to draft while debugging Travis CI

henryiii (approved):

PR #2274, opened 2025-02-10
chore: use nox 2025.2.9 features

henryiii:
Using new features in nox 2025. This can now be run with `uv run`, `pipx run`, `hatch run`, etc.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,45 +1,36 @@
+#!/usr/bin/env -S uv run
+
+# /// script
+# dependencies = ["nox>=2025.2.9"]
+# ///
+
+"""
+cibuildwheel's nox support
+
+Tags:
+
+    lint: All linting jobs
+    update: All update jobs
+
+See sessions with `nox -l`
+"""
+
 from __future__ import annotations
 
 import os
 import shutil
 import sys
 from pathlib import Path
-from typing import Any
 
 import nox
 
-nox.needs_version = ">=2024.4.15"
-nox.options.sessions = ["lint", "pylint", "check_manifest", "tests"]
+nox.needs_version = ">=2025.2.9"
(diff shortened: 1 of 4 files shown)


noxfile.py
```diff
@@ -1,45 +1,36 @@
+#!/usr/bin/env uv run
```
joerick, line 1: I think maybe `-S` is required on some OSs?

https://simonwillison.net/2024/Aug/21/usrbinenv-uv-run/

```suggestion
#!/usr/bin/env -S uv run
```

noxfile.py
```diff
@@ -1,45 +1,36 @@
...
+
+# /// script
+# dependencies = ["nox>=2024.2.9"]
```
burgholzer, line 4:
```suggestion
# dependencies = ["nox>=2025.2.9"]
```
Potentially a typo 👆🏻

noxfile.py
```diff
@@ -59,16 +50,21 @@ def pylint(session: nox.Session) -> None:
...
 @nox.session
-def check_manifest(session: nox.Session) -> None:
+def tests(session: nox.Session) -> None:
     """
-    Ensure all needed files are included in the manifest.
+    Run the unit and regular tests.
     """
-
-    session.install("check-manifest")
-    session.run("check-manifest", *session.posargs)
```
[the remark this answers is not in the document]
henryiii: Intentional, since we don't have a MANIFEST anymore. I wrote `check-sdist` as a replacement that works with any backend, not just setuptools, if we want to use that. It also is faster.

mayeut:
IMHO, the linter changes should not be part of this PR but belongs more with #2276 unless there are some changes related to nox that I'm not seeing.

joerick (approved):

PR #2276, opened 2025-02-12
chore: cleanup for 3.11

henryiii:
Some cleanup, focused on some things 3.11 provides.

- **chore: cleanup for 3.11+**
- **chore: use StrEnum from Python 3.11**
- **chore: some cleanup from ruff --preview checks**


cibuildwheel/options.py
```diff
@@ -265,10 +265,12 @@ class EnvironmentFormat(OptionFormat):
...
     """
 
-    def format_table(self, table: SettingTable) -> str:
+    @staticmethod
+    def format_table(table: SettingTable) -> str:
         return " ".join(f'{k}="{v}"' for k, v in table.items())
 
-    def merge_values(self, before: str, after: str) -> str:
+    @staticmethod
+    def merge_values(before: str, after: str) -> str:
```
mayeut, line 273: These 2 changes shall probably be reverted, they do not respect the prototype inherited from OptionFormat

cibuildwheel/__main__.py
```diff
@@ -160,7 +160,7 @@ def main_inner(global_options: GlobalOptions) -> None:
     parser.add_argument(
         "package_dir",
         metavar="PACKAGE",
-        default=Path("."),
+        default=Path(),
```
joerick, line 163: `Path()` is less self-documenting, could add a comment or revert?

```suggestion
        default=Path(),  # working directory
```

```suggestion
        default=Path("."),
```

mayeut: or `Path.cwd()`

joerick: I wondered about that, but `Path.cwd()` would be an absolute path, which might make some logging more verbose unnecessarily.

joerick: Not a big deal fyi @henryiii, I can learn to like it I guess, was just curious if anybody agreed with me.

henryiii: The Ruff recommendation is to use the default constructor `Path()` instead of `Path(".")` or `Path("")`. (Yes, `Path.cwd()` is different). I'm happy with either as long as it's consistent, and there's a check to enforce `Path()`. IMO, it's pretty clear, not sure what else `Path()` could be. (And, in pathlib, you can completely avoid `"."` and `".."`)

cibuildwheel/options.py
henryiii: Yes, they do (in fact, mypy would complain if they didn't!). A `staticmethod` or `classmethod` is a perfectly valid implementation of an instance method. A `staticmethod` can also be an implementation of a `classmethod`. It's the other way around that's not allowed. I can't make the `OptionFormat` prototype a `staticmethod`, because then I can't ever make it anything else.

henryiii: Sadly, pylint is tripping up on this, because it's doing simple argument counting, and not taking into account that the `staticmethod` decorator adds the missing argument for you.

henryiii: I reverted it to make the pylint warning go away.  Though I think W0221 is incorrect and should take into account `@staticmethod`.

cibuildwheel/__main__.py
joerick: If you're happy with either, and we're looking for a decision, my preference for readability, would be `Path(".")`.

> not sure what else `Path()` could be

Off the top of my head, it could be working dir, `Path(__file__)`, `Path(__file__).parent`, the filesystem root, the user home dir. It does seem weird to me that `Path() / 'bin'` and `Path('.') / 'bin'` are the same but `Path() / 'bin'` and `Path('/bin')` are different.

henryiii: I don't have a way to enforce that it's always `Path(".")`, but I do have a way to enforce that it's always `Path()` (https://docs.astral.sh/ruff/rules/path-constructor-current-directory), and I value consistency over either form. I guess I could manually do a search for the characters, since we enforce consistent spacing and quoting.

Almost all the forms you mention are absolute paths (`__file__` is always absolute starting in 3.9, 3.4-3.8 it could be relative for scripts), which are not the same as a relative path. (FWIW, Plumbum's path system didn't allow relative paths - making one was a string, which IMO was simpler)

https://docs.python.org/3/library/pathlib.html#pure-paths states "When pathsegments is empty, the current directory is assumed", so it's basically `"."` is default. They've designed it so you never have to use "." and "..", though you can. `Path() / "bin"` and `Path("bin")` are the same.

henryiii: FYI, there's also a rule for `Path(".").resolve()` -> `Path.cwd()`, I can apply that. That's almost every usage of `Path(".")` that we have - there are only two that are actually relative paths.

```
cibuildwheel/__main__.py:163:        default=Path("."),
cibuildwheel/macos.py:722:                        project=Path(".").resolve(),
cibuildwheel/macos.py:737:                        test_cwd = Path(".").resolve()
cibuildwheel/options.py:76:            package_dir=Path("."),
cibuildwheel/pyodide.py:295:            if str(Path(".").resolve()).startswith("/tmp"):
cibuildwheel/pyodide.py:296:                extra_mounts.append(str(Path(".").resolve()))
cibuildwheel/pyodide.py:416:                    project=Path(".").resolve(),
cibuildwheel/pyodide.py:430:                    test_cwd = Path(".").resolve()
cibuildwheel/windows.py:552:                    project=Path(".").resolve(),
cibuildwheel/windows.py:566:                    test_cwd = Path(".").resolve()
```

Also, `cibuildwheel/pyodide.py:295: if str(Path(".").resolve()).startswith("/tmp"):` will trigger on `/tmp*`, I wonder if that was intentional?

joerick: Okay, let's go with `Path()`.

> FYI, there's also a rule for `Path().resolve()` -> `Path.cwd()`, I can apply that.

Sounds reasonable.

> Also, `cibuildwheel/pyodide.py:295: if str(Path(".").resolve()).startswith("/tmp"):` will trigger on `/tmp*`, I wonder if that was intentional?

Unlikely, based on context. I think that could be changed to `if str(Path.cwd()).startswith("/tmp/"):`

joerick:
I had a thought - can we remove `from __future__ import annotations` now? I can't remember how all that shook out but I think annotations are deferred by default now? Maybe worth trying it and seeing what breaks!

henryiii:
I think you are referring to 3.14's making them lazy by default? However, we might not need the annotations because the main reason it's used is to enable invalid expressions that became valid in 3.9 and 3.10 (and stuff that is often better spelled as `Self`). So I could try and see.

joerick:
Ahh, yeah it's the expression thing. So we might still want it to use more modern type features than exist in 3.11. So I guess 3.14 is where it's gonna get simple again :)

cibuildwheel/__main__.py
joerick: Sounds good to me.

henryiii:
I removed `from __future__ import annotations`. A few places needed strings, lots of things I moved to `Self`, and I was able to remove the one place we were using `TYPE_CHECKING` since the thing it was protecting was valid on 3.10+.

joerick (approved):
Looks good to me.

cibuildwheel/architecture.py
```diff
@@ -93,7 +82,7 @@ def parse_config(config: str, platform: PlatformName) -> set[Architecture]:
         return result
 
     @staticmethod
-    def native_arch(platform: PlatformName) -> Architecture | None:
+    def native_arch(platform: PlatformName) -> "Architecture | None":
```
henryiii, line 85: FYI, I tried `Self` on these (with `@classmethod`), but typing didn't seem to understand that enum is `final` (even if you tell it it's `final`), so it sees `cls.wasm32` as `Architecture.wasm32` and therefore not `Self`. Wrapping it in an extra `cls()` does make it happy, but `cls(cls.wasm32)` just to make the type checker happy wasn't worth it. Just added strings here instead of `Self`. It's technically correct since enums are always `final`.

cibuildwheel/macos.py
```diff
@@ -738,7 +736,7 @@ def build(options: Options, tmp_path: Path) -> None:
                 moved_wheel = move_file(repaired_wheel, output_wheel)
                 if moved_wheel != output_wheel.resolve():
                     log.warning(
-                        "{repaired_wheel} was moved to {moved_wheel} instead of {output_wheel}"
+                        f"{repaired_wheel} was moved to {moved_wheel} instead of {output_wheel}"
```
henryiii, line 739: This is a fix, I believe.

PR #2280, opened 2025-02-23
General docs improvements for v3.0

joerick:
I think a bit of organisation / rethinking is due. The main change I've made here so far is to the navigation - it now looks like this:
image

I decided to go for the tree-based approach because it deals better with many sections/pages.

We can make the change to pyproject.toml-style configuration in this PR too, if you like @henryiii .


henryiii:
We don't have to swap the envvar / TOML config in this PR, it can be a followup, but it is on the planned items for 3.0.

Also, I know I was considering making `cibuildwheel.toml` a supported filename around the time we added the schema (since I had to set a filename for SchemaStore), maybe we could look into doing that for 3.0. Might help encourage TOML usage.

henryiii:
I can work on the toml/envvar swap once this is done.

joerick:
Okay, I'll try to tie up these changes soon.

> Also, I know I was considering making `cibuildwheel.toml` a supported filename around the time we added the schema (since I had to set a filename for SchemaStore), maybe we could look into doing that for 3.0. Might help encourage TOML usage.

You mean a file that's auto-detected at the project root? Personally I'm not a fan of extra project root files but I can see that some might prefer it. But, I do like the shorthand of  "pyproject.toml" in tab labels and such. Diluting that to "TOML" might make it less friendly to new users.

joerick:
This should be ready for a review. I've made quite a few changes, mostly to organisation, but where I inevitably spot things that are inaccurate or out-of-date I've fixed them too, so there's quite a bit in this diff. 

Probably the easiest thing is to view the [rendered version](https://cibuildwheel--2280.org.readthedocs.build/en/2280/).

image

Big things:

- Moved the platforms pages into a single page, and grouped together FAQ entries here. The platform-specific information is very useful, but having them as separate pages took a lot of space in the top-level of the navbar. Also, grouping platform specific information like this is much better than scattering it around the docs, so I've moved a few FAQ entries into this page, too.
- Split out the CI services setup into its own page. It always felt strange that the CI configs were pushed down below the running locally section, and, I think it's good to have a page in the navbar with 'CI services', as a lot of people are just looking for copy/paste examples.
- Split out the configuration methods into their own page, separate from the options docs. This lets the options page load into the TOC, which is a much better experience.
- A general tidy up of the 'Tips & tricks' page.

docs/faq.md
```diff
@@ -6,88 +6,6 @@ title: Tips and tricks
 
 ## Tips
```
joerick, line 8: This info moved to platforms.md.

docs/configuration.md
```diff
@@ -0,0 +1,238 @@
...
+cibuildwheel to run tests, add the following YAML to your CI config file:
+
+!!! tab "GitHub Actions"
+
+    > .github/workflows/*.yml ([docs](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)) (can be global, in job, or in step)
+
+    ```yaml
+    env:
+      CIBW_TEST_REQUIRES: pytest
+      CIBW_TEST_COMMAND: "pytest {project}/tests"
```
henryiii, line 20: Should we update the for 3.0, now that `{project}/` isn't needed, but test-sources might be?

docs/options.md
```diff
@@ -310,7 +71,7 @@ Windows arm64 platform support is experimental.
 For an experimental WebAssembly build with `--platform pyodide`,
 `cp312-pyodide_wasm32` is the only platform identifier.
 
-See the [cibuildwheel 1 documentation](https://cibuildwheel.pypa.io/en/1.x/) for past end-of-life versions of Python, and PyPy2.7.
+See the [cibuildwheel 2 documentation](https://cibuildwheel.pypa.io/en/2.x/) for past end-of-life versions of Python.
```
henryiii, line 74: Should we link 1.x too?

docs/setup.md
```diff
@@ -1,36 +1,28 @@
...
-Before getting to CI setup, it can be convenient to test cibuildwheel
-locally to quickly iterate and track down issues without even touching CI.
+Before getting to [CI setup](ci-services.md), it can be convenient to test cibuildwheel locally to quickly iterate and track down issues without having to commit each change, push, and then check CI logs.
 
 Install cibuildwheel and run a build like this:
 
 ```sh
-# using pipx (https://github.com/pypa/pipx)
+# run using uv
+uvx cibuildwheel
```
henryiii, line 13:
```suggestion
uvx cibuildwheel

```

henryiii (approved):

docs/options.md
henryiii: I guess it's discoverable, and really old at this point.

docs/configuration.md
joerick: Yeah, good idea on removing {project}. I think we don't need TEST_SOURCES though? It's more of a thing that can be discovered as/when it's needed?

joerick: I've made the change to all instances in docs of "pytest {project}/tests". I also updated the tests, as we should probably be testing what we're documenting!

henryiii: If someone isn't using `src` structure, they may run into it. Probably should at least add a warning and maybe an upgrade guide. Not now though.

mayeut (approved):

PR #2282, opened 2025-02-23
drop python 3.6 & 3.7 support

mayeut:
Per https://github.com/pypa/cibuildwheel/issues/2047#issuecomment-2676957111, dropping Python 3.6 & 3.7


joerick (approved):
Thank you @mayeut! Man that was fast :)

mayeut:
Note: I did not update the build selector unit tests that still reference 3.6 & 3.7, they're working fine as is.

PR #2284, opened 2025-02-24
Correct handling of test_sources on universal macOS builds.

### LINKED ISSUE #2283: macOS universal builds fail if `test-sources` contains a directory
#2062 added a `test-sources` configuration, allowing projects to specify a directory of source code that would be copied into a test environment.

However, on macOS, a project with a `test-sources` configuration will fail if a `universal` build is requested. A universal build is built once, but is tested *twice* - once for each architecture that the build should support.

As the code to set up `test-sources` is contained inside the loop testing architectures, the second test will fail if `test-sources` references a directory, as the target directory will already exist.

The failure will look something like:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/__main__.py", line 487, in <module>
    main()
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/__main__.py", line 67, in main
    main_inner(global_options)
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/__main__.py", line 207, in main_inner
    build_in_directory(args)
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/__main__.py", line 408, in build_in_directory
    platform_module.build(options, tmp_path)
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/macos.py", line 730, in build
    copy_test_sources(
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/util/file.py", line 137, in copy_test_sources
    copy_into(source, test_dir / test_path)
  File "/Users/rkm/projects/cibuildwheel/cibuildwheel/util/file.py", line 110, in copy_into_local
    shutil.copytree(src, dst)
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/shutil.py", line 600, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/shutil.py", line 498, in _copytree
    os.makedirs(dst, exist_ok=dirs_exist_ok)
  File "<frozen os>", line 225, in makedirs
FileExistsError: [Errno 17] File exists: '/private/var/folders/b_/khqk71xd45d049kxc_59ltp80000gn/T/cibw-run-5t9ju11c/cp310-macosx_universal2/test_cwd/tests'
```

freakboy3742:
Ensure that if a `test-sources` configuration is provided, it's only copied on the *first* test pass. 

If a macOS project provides a `test-sources` configuration, and that project is tested on `universal` architecture, there are 2 test passes. As currently configured, the second pass will fail beacuse the `test-sources` folder already exists.

Adds an explicit test of `test-sources` on macOS, in both universal and non-universal configurations.

Fixes #2283.


cibuildwheel/macos.py
```diff
@@ -726,12 +726,13 @@ def build(options: Options, tmp_path: Path) -> None:
 
                     if build_options.test_sources:
                         test_cwd = identifier_tmp_dir / "test_cwd"
-                        test_cwd.mkdir(exist_ok=True)
-                        copy_test_sources(
-                            build_options.test_sources,
-                            build_options.package_dir,
-                            test_cwd,
-                        )
+                        if not test_cwd.exists():
```
joerick, line 729: I think a comment here would be nice

```suggestion
                        # only create test_cwd if it doesn't already exist - it
                        # may have been created during a previous `testing_arch`
                        if not test_cwd.exists():
```

joerick (approved):
Thanks for catching this @freakboy3742 !

test/test_macos_archs.py
```diff
@@ -62,12 +81,12 @@ def test_cross_compiled_test(tmp_path, capfd, build_universal2):
 
     actual_wheels = utils.cibuildwheel_run(
         project_dir,
-        add_env={
-            "CIBW_BUILD": "cp310-*" if build_universal2 else "*p310-*",
-            "CIBW_TEST_COMMAND": '''python -c "import platform; print('running tests on ' + platform.machine())"''',
-            "CIBW_ARCHS": "universal2" if build_universal2 else "x86_64 arm64",
-            "CIBW_BUILD_VERBOSITY": "3",
-        },
+        add_env=dict(
```
henryiii, line 84: Nit: I prefer `{}`, it's a bit faster (native syntax) and works in all cases (`dict` requires valid identifier keys), and `**` still works.

freakboy3742: TIL - I didn't know `**` syntax was legal inside `{}`. I've always fallen back to `dict()` when keyword expansion was needed. I'll push a cleanup.

Issue #1960, opened 2024-08-08
Add support for Android and iOS


timrid:
Are there plans to support Android and iOS wheels as well?

For Python 3.13, iOS ([PEP730](https://peps.python.org/pep-0730/)) and Android ([PEP738](https://peps.python.org/pep-0738/)) will also be supported. It would be practical to be able to easily create wheels for these platforms.

henryiii:
I think that the BeeWare folks plans to help support this in cibuildwheel, I was told later this year at PyCon.

freakboy3742:
Confirming that yes, we (the BeeWare folks) are working on this. I've got an [initial WIP branch with iOS support](https://github.com/freakboy3742/cibuildwheel/tree/ios-support), which is sufficient to get a simple project building, but still needs a lot of work. #2062 is the first piece of that work to be upstreamed.

Jzhenli:
Hi, will Android ([PEP738](https://peps.python.org/pep-0738/))  be supported?

freakboy3742:
@Jzhenli Yes - it's on our (BeeWare's) todo list. We're trying to get iOS support added first, rather than trying to add 2 platforms at once.

henryiii:
iOS merged. Now just Android left. ;)

freakboy3742:
We're working on it!

PR #2286, opened 2025-02-24
Add support for building iOS wheels.

freakboy3742:
Adds support for building PEP 730 iOS wheels.

Refs #1960.

tl;dr - for simple packages, `CIBW_PLATFORM=ios cibuildwheel` will produce 3 iOS wheels suitable for use on:
* ARM64 devices
* ARM64 simulators on macOS
* x86_64 simulators on macOS

As proof, this branch has been use to [publish iOS wheels for pyspamsum](https://pypi.org/project/pyspamsum/1.0.7/#files). These wheels were produced using [this Github Actions CI configuration](fhttps://github.com/freakboy3742/pyspamsum/blob/main/.github/workflows/ci.yml#L33).

A [branch adding iOS support on Pillow](https://github.com/freakboy3742/Pillow/tree/ios-build) also exists; this hasn't been submitted for inclusion in Pillow because of the dependency on this (as yet unmerged, unreleased) PR. However, it's a demonstration of a non-trivial Python package building iOS wheels.

## The details:

Edited 6 Mar 2025, to reflect most recent changes in `CIBW_PLATFORM` and `CIBW_ARCHS`.

### `CIBW_PLATFORM` values

iOS is effectively 2 platforms - physical devices, and the simulator. While the API for these two platforms are identical, the ABI is *not* compatible, even when dealing with a device and simulator with the same CPU architecture. For this reason, iOS support a `CIBW_PLATFORM` value of `ios`, and 3 `CIBW_ARCHS` values - `arm64_iphoneos` (for ARM64 devices), `arm64_iphonesimulator` (for simulators running on ARM64 Macs) and `x86_64_iphonesimulator` (for simulators running on Intel simulators). This also allows for configurations to be specified at the `ios` level, and specialised at the `iphoneos` or `iphonesimulator` level using a `select` clause in a settings override.

### Binary distributions

The binaries used to support iOS builds come from the BeeWare [Python Apple Support project](https://github.com/beeware/Python-apple-support). It is hoped that in the 3.14 release timeframe, these artefacts will be produced by Python.org itself - this is an topic I'm currently working on. In the meantime, the use of 'unofficial' support packages puts iOS in a similar boat as the Emscripten backend using Pyodide build artefacts.

### Cross platform builds

iOS builds are *always* cross platform builds, as it not possible to run compilers and other build tools "on device". The BeeWare iOS support package includes tooling that can convert any virtual environment into a cross platform virtual environment - that is, an environment that can run binaries on the build machine (macOS), but, if asked, will respond as if it is an iOS machine. This allows `pip`, `build`, and other build tools to perform iOS-appropriate behaviour. As a result, an iOS build environment also means downloading and installing CIBW's macOS tools.

### Build frontends

iOS builds support both the `pip` and `build` build frontends. In principle, support for `build[uv]` should be possible, but `uv` [doesn't currently have support for cross-platform builds](https://github.com/astral-sh/uv/issues/7957), and [doesn't have support for iOS (or Android) tags](https://github.com/astral-sh/uv/issues/8029).

### The build environment 
The environment used to run builds does *not* inherit the full user environment - in particular, PATH is deliberately re-written. This is because UNIX C tooling doesn't do a great job differentiating between "macOS ARM64" and "iOS ARM64" binaries. If (for example) Homebrew is on the path when compilation commands are invoked, it's easy for a macOS version of a library to be linked into the iOS binary, rendering it useless. To prevent this, iOS builds always force PATH to a "known minimal" path, that includes only the bare system utilities, plus cargo (to facilitate Rust builds).

### iOS test suites

Running iOS test suites also requires special attention. The iOS test environment can't support running shell scripts, so the entry point must be specified as a completion to `python -m ...`. In addition, the test itself must be run "on device", so the local project directory cannot be used to run tests. As a result, the project *must* specify `test_sources` as a minimum subset of files that should be copied to the test environment.

The test process uses the same testbed used by CPython itself to run the CPython test suite. It is an Xcode project that has been configured to have a single Xcode "XCUnit" test - the result of which is whatever the output of `python -m <CIBW_TEST_COMMAND>` is. This testbed is also included in the BeeWare support package.


freakboy3742:
I've clearly got some CI edge cases to clean up - moving to draft status.

examples/github-minimal.yml
joerick, line 1: My feeling is that adding Pyodide and iOS to this 'minimal' config is changing it to no longer be 'minimal'. My philosophy with the minimal config is to show the user the least amount of config required to get something up and running - something they can build upon to get to something that works for them.

I'm now thinking there's a place for a new 'platforms' example config, which is what you have here, and then reducing the minimal config back down. 

(Specifically, the minimal config probably shouldn't be requiring any messing with `matrix.include`. Even for me this is a head-scratcher of an option!)

timrid:
First of all, nice work what you are doing with the python support for mobile platforms. I am following most of your steps towards this goal :)

I am currently trying to get cibuildwheel running with `pybind11` and `cmake` using [this example project](https://github.com/timrid/pybind11_cmake_example) and have the following problem:

The project has `cmake` as a build dependency. Under macOS there is also a ready-made binary wheel for CMake [cmake-3.31.4-py3-none-macosx_10_10_universal2.whl](https://pypi.org/project/cmake/#cmake-3.31.4-py3-none-macosx_10_10_universal2.whl), which is used when building the wheel for macos. However, when I build the iOS wheel, `cmake-3.31.4.tar.gz` is downloaded and an attempt is made to build `cmake` from source, which ends in an error ([see here](https://github.com/timrid/pybind11_cmake_example/actions/runs/13495332539/job/37701229928)).

Since a cross platform build is made for iOS, I would assume that `cmake-3.31.4-py3-none-macosx_10_10_universal2.whl` should be used for ios instead of the `.tar.gz`. Is this possibly a bug in [make_cross_venv.py](https://github.com/beeware/Python-Apple-support/blob/main/patch/Python/make_cross_venv.py)?

examples/github-minimal.yml
freakboy3742: Thats fair. I can see how there's a balance to be had between "minimal for 'most' uses", and "minimal spanning set" - and it makes sense that the first introductory example is the former. I'll simplify/restore this example.

freakboy3742:
> Since a cross platform build is made for iOS, I would assume that `cmake-3.31.4-py3-none-macosx_10_10_universal2.whl` should be used for ios instead of the `.tar.gz`. Is this possibly a bug in [make_cross_venv.py](https://github.com/beeware/Python-Apple-support/blob/main/patch/Python/make_cross_venv.py)?

Hrm - that's an interesting one. It's not a clear bug in `make_cross_venv` to my mind... I'd argue its almost a "missing feature".

The problem is that cmake is in a weird situation as a dependency. Cython would be in a similar situation. On the one hand, it needs to run as part of the *build* - so we need to install the macOS wheel, so a macOS binary exists. The virtual environment will advertise itself as a `sys.platform == "ios"` environment, so it will be looking for iOS-compatible wheels. So, when it doesn't find an iOS-compatible binary wheel, it tries to compile one... and fails. 

However, an iOS-compatible cmake wheel is a bit of a contradiction... because you won't be able to run cmake *on iOS*. So it would appear as an iOS-compatible wheel that would break things if you actually added it to an app. If you put a macOS binary into an iOS wheel, it would obviously break.

The alternative approach would be to consider build dependencies to be "non cross", and Install build dependencies using the build platform's wheel tag. That means installing the macOS wheels for cmake, so that cmake can run in the build environment. I'm not 100% sure if the PEP 517 interface is literally the boundary that we need to care about here, or if we need an additional setting or control mechanism to allow the installation of build dependencies independent of package/testing dependencies... some more investigation may be required.

freakboy3742:
> The text you have in the PR description - we should add that to the docs somewhere. Perhaps a new page about other platforms... that could include some info about Pyodide too.

That's a good call - I'll add in a section to the docs to that effect.

freakboy3742:
> > I am currently trying to get cibuildwheel running with `pybind11` and `cmake` using [this example project](https://github.com/timrid/pybind11_cmake_example)
> 
> That example is old, you should be using scikit-build-core and https://github.com/pybind/scikit_build_example as the base.

That may be true for that specific example, but there are other projects (PyTorch is one notable example) that have a dependency on cmake, cython, or similar "binary tools that must be runnable in the build environment" requirement.

However, I'll make a note to add https://github.com/pybind/scikit_build_example as a test case for iOS/Android build tooling.

> > The alternative approach would be to consider build dependencies to be "non cross", and Install build dependencies using the build platform's wheel tag
> 
> This is tricky; a good counter example is numpy. If you want to run `numpy.get_include`, you would want the host wheel. If you want to link to numpy, you'd want the target wheel. I'm not sure this is properly possible unless PEP 517 were expanded to include cross-compilation. Maybe we need some sort of per-package override?

Pandas looks like a good test case here - it needs both cython (which needs to be a macOS version) and numpy (which needs to be the iOS version). 

However, I agree that this is fundamentally a gap in PEP 517 itself, and we need an extension of that PEP to cover the cross-compilation case. My off-the-cuff proposal would be to have `build-system.requires` be "host native" wheels, but a new "build-requires" setting defines wheels that must be "build native". For standard macOS/Linux/Windows builds, the two are the same; but for iOS/Android/other cross-platform builds, the difference would exist. I need to do some more experimentation here.

This would also give an opportunity to formalize what a "cross platform venv" actually means, which what the `make_cross_venv` script is doing, and pyodide is doing with its own tooling. I was planning to kick off a discuss.python.org thread about cross-venv stuff in the near future, with the intention of that turning into a PEP; adding the PEP517 interface issues to that PEP seems like it might be appropriate.

I guess the bigger question here is whether that discussion needs to precede a PR like this one being merged - i.e., is "can compile numpy and pandas" (which are projects with complex build systems) a prerequisite for adding *any* iOS/Android support to cibuildwheel. This would admittedly be a big gap, but I question whether "the perfect is the enemy of the good" in this instance.

freakboy3742:
@joerick I've restored most of the old examples of GitHub actions configuration, only including the "full" example for the last "deployment" case. I've also split out the platform details into standalone pages. That required a bit of surgery to the MKDocs configuration; I'm not sure if I've landed on the best layout for that new content, but the new content is *there* - I'm open to suggestions on other ways to structure that content.

I've fixed the issue with Linux and Windows CI - it required changing the "test" that is run to do something other than invoking `system()`, because `system()` isn't available on iOS. 

The tests are still failing macOS CI - that appears to be due to [this bug in the testbed](https://github.com/python/cpython/pull/130388), which I've got a fix in flight. As soon as that PR is merged, I can publish updated support packages that will include the patch.

mhsmith:
> If you want to run `numpy.get_include`, you would want the host wheel. If you want to link to numpy, you'd want the target wheel. I'm not sure this is properly possible unless PEP 517 were expanded to include cross-compilation.

NumPy itself may also need to change to support this, because even if we did install the iOS wheel, we wouldn't be able to import it to call `get_include`. 

In our Android recipes we worked around this by patching everything that used NumPy to set [`__NUMPY_SETUP__`](https://github.com/search?q=repo%3Achaquo%2Fchaquopy%20__numpy_setup__&type=code). But I see this doesn't appear in the [iOS recipes](https://github.com/beeware/mobile-forge), so maybe it's not necessary anymore.

freakboy3742:
> But I see this doesn't appear in the [iOS recipes](https://github.com/beeware/mobile-forge), so maybe it's not necessary anymore.

IIRC, this cleanup was possible because of the switch to Meson as a build system. The iOS patch is still on the 1.X branch, not the 2.X branch, but building with Meson was (unsurprisingly) a lot cleaner than the older setuptools build.

However - even then, I'm not sure it would be a problem with the way this PR is structured. The cross-env handling means that you can invoke Python code - and it runs as if `sys.platform == "ios"`. As a result, having a `setup.py` that references platform-specific behavior isn't a problem - in fact, it's desirable. The [presumptive patch for Pillow that I have developed does this](https://github.com/freakboy3742/Pillow/blob/ios-build/setup.py#L693).

freakboy3742:
> > But I see this doesn't appear in the [iOS recipes](https://github.com/beeware/mobile-forge), so maybe it's not necessary anymore.
> 
> IIRC, this cleanup was possible because of the switch to Meson as a build system. The iOS patch is still on the 1.X branch, not the 2.X branch, but building with Meson was (unsurprisingly) a lot cleaner than the older setuptools build.
> 
> However - even then, I'm not sure it would be a problem with the way this PR is structured. The cross-env handling means that you can invoke Python code - and it runs as if `sys.platform == "ios"`. As a result, having a `setup.py` that references platform-specific behavior isn't a problem - in fact, it's desirable. The [presumptive patch for Pillow that I have developed does this](https://github.com/freakboy3742/Pillow/blob/ios-build/setup.py#L693).

EDIT: ... as long as the code being executed is pure Python. I'm not sure if this is the case with `numpy.get_include()`

docs/contributing.md
```diff
@@ -36,7 +36,7 @@ So, if we can, I'd like to improve the experience on errors as well. In [this](h
 
 ### Running the tests
 
-When making a change to the codebase, you can run tests locally for quicker feedback than the CI runs on a PR. You can [run them directly](#making-a-venv), but the easiest way to run tests is using [nox](https://nox.thea.codes/).
+When making a change to the codebase, you can run tests locally for quicker feedback than the CI runs on a PR. You can run them directly, but the easiest way to run tests is using [nox](https://nox.thea.codes/).
```
freakboy3742, line 39: The `making-a-venv` anchor doesn't exist; I can't work out where it *was* pointing, so I've removed it to make the docs build run without warnings.

bin/run_tests.py
```diff
@@ -42,12 +42,15 @@
         sys.executable,
         "-m",
         "pytest",
+        "--dist",
+        "loadgroup",
```
freakboy3742, line 46: This allows the build to run all the iOS tests on the same machine; see the iOS tests for the reason.

bin/run_tests.py
```diff
@@ -42,12 +42,15 @@
...
         f"--numprocesses={args.num_processes}",
         "-x",
         "--durations",
         "0",
         "--timeout=2400",
         "test",
+        "-vv",
```
freakboy3742, line 53: This generates a lot more output - but when a test fails, it helps to have as much context as possible. It was difficult to diagnose some of the iOS failures with the low verbosity output.

freakboy3742:
This is now passing CI, and the documentation issues referenced in the earlier pre-review have been addressed, so I think this is ready for a full review.

test/test_projects/c.py
```diff
@@ -10,15 +10,16 @@
...
+    const char *content;
     int sts;
 
-    if (!PyArg_ParseTuple(args, "s", &command))
+    if (!PyArg_ParseTuple(args, "s", &content))
         return NULL;
 
-    sts = system(command);
+    // Spam should not be allowed through the filter.
+    sts = strcmp(content, "spam");
```
freakboy3742, line 22: `system()` is a prohibited call on iOS (there's essentially no access to spawn subprocesses). However, this didn't seem to be what was actually being tested, so I've converted it into a simpler "string check" with the name net result (a 0/1 integer result that is controlled by input)

cibuildwheel/ios.py
```diff
@@ -0,0 +1,635 @@
...
+
+
+def cross_virtualenv(
+    py_version: str,
+    host_python: Path,
+    multiarch: str,
+    build_python: Path,
+    venv_path: Path,
+    dependency_constraint_flags: Sequence[PathOrStr],
+) -> dict[str, str]:
```
freakboy3742, line 157: Not sure what I can do here... these are all required arguments, and factoring out a data structure for exactly one invocation seems excessive.

mhsmith:
> as long as the code being executed is pure Python. I'm not sure if this is the case with `numpy.get_include()`

`get_include` itself is pure-Python, but at least in the older versions I've built for Android, `import numpy` would immediately try to import its binary modules. If this has changed with Meson, then that's great news.

henryiii:
I'm curious, would it make sense to use the architecture instead of the platform for simulator? So basically split native and simulator architectures? Then "auto" could build all of them?

freakboy3742:
> I'm curious, would it make sense to use the architecture instead of the platform for simulator? So basically split native and simulator architectures? Then "auto" could build all of them? 

Unless I'm missing something, I don't think so. 

The current PR setup uses 2 configurations:

    CIBW_PLATFORM=iphoneos
    CIBW_ARCHS=arm64

and:

    CIBW_PLATFORM=iphonesimulator
    CIBW_ARCHS=arm64 x86_64

with `CIBW_PLATFORM=ios` existing purely as a shorthand to build both of those configurations.

If I'm understanding you correctly, what you're proposing is that there would be a single iOS platform, and 3 "architectures" - the long form representation would be something like:

    CIBW_PLATFORM=ios 
    CIBW_ARCHS=arm64_iphoneos arm64_iphonesimulator x86_64_iphonesimulator

with that explicit `CIBW_ARCHS` definition being the interpretation of `CIBW_ARCHS=auto`.

For simple wheels, that would indeed work, building all 3 binaries - just as the current PR does with `CIBW_PLATFORM=ios`. However, it becomes a problem with more complex builds, because `iphoneos` and `iphonesimulator` have different ABIs, and these different ABIs sometimes require different compilation and linking flags and paths.

Pillow is an illustrative example here - my [WIP branch for iOS support](https://github.com/freakboy3742/Pillow/tree/ios-build) contains (in part) the following:
```
[tool.cibuildwheel.iphoneos.environment]
PATH = "$(pwd)/build/deps/iphoneos/bin:..."

[tool.cibuildwheel.iphonesimulator.environment]
PATH = "$(pwd)/build/deps/iphonesimulator/bin:..."
```
That is - you need to be able to slice the *build configuration* on the ABI axis. Unless I'm missing something, it's not possible to do that level of slicing at the level of an architecture in cibuildwheel. The flags you pass to `iphonesimulator` builds are (in my experience) identical, except for the actual CPU architecture - and even that generally isn't needed, because Apple binaries are generally distributed in "fat binary" format that includes both x86_64 and arm64 slices, so you only need to link "the iPhone simulator library". There's a separate library for iPhone devices, and the flags used for linking are often *not*  the same as the flags for `iphoneos` builds.

You can't build a fat binary that includes arm64 iphoneos and arm64 iphonesimulator slices either, because Apple's "fat" binary mechanism doesn't allow different ABIs to co-exist, just different architectures. This is why Apple introduced the `XCframework` structure - it provides a way to have a single "meta-framework" that internally contains a slice for each ABI, each of which itself can contain slices for each CPU architecture supported by that ABI.

henryiii:
You can split those with overrides and `select = "*iphoneos"`, which is the modern replacement for the old `environment` mechanism. In fact, you could even use `select = "*simulator"` to get both.

freakboy3742:
> You can split those with overrides and `select = "*iphoneos"`, which is the modern replacement for the old `environment` mechanism. 

Interesting... I think this could work - and it would have some notable benefits, too. As it currently stands, there's sometimes need for duplication of settings between `iphoneos` and `iphonesimulator` configurations; Based on my reading, it looks like this would also allow for "common" configurations for all iOS builds, with a minimal number of overrides when the platform *actually* needs a different value. The Pillow configuration, for example, uses the same value for `test_sources` on both `iphoneos` and `iphonesimulator` - which means duplicating the setting. Presumably we could define a set of "common" overrides for `select = "*-ios_*", and the specific ones for  `select = "*_iphoneos"` and `select = "*_iphonesimulator"`? 

It also look like this would also potentially allow `select = "*_arm64_iphonesimulator"` and `select = "*_x86_64_iphonesimulator"` if you needed a configuration that was architecture specific inside the simulator ABI. I don't know that I have seen a use case for this, but it's nice to know it would be possible should the need arise. 

I'm also guessing that there *isn't* an equivalent for this syntax with `CIBW_*` environment variables (and this is by design)? I presume there's no analog for "overrides", and the intention is that "best practice" is/should be that environment variables are only used for *selecting* builds (with `CIBW_PLATFORM`, `CIBW_ARCHS`, `CIBW_BUILD` and `CIBW_SKIP`)?
 
> In fact, you could even use `select = "*simulator"` to get both.

I presume this is a typo, and you meant `select = "*iphone*"` or `select = "*-ios_*"` or similar? Or have I misunderstood something?

> (Though those names are long, I'd probably do something a bit shorter. Also, isn't iPads and such included even if it's called "iPhone"? `ios` for platform and `arm64_native arm64_sim x86_64_sim`, perhaps?)

`iphoneos` and `iphonesimulator` are the names used in the PEP 730 specification for `sys.implementation._multiarch` and in the wheel specifier; it seems to be appropriate to consistent, rather than invent new terminology

`iphoneos` and `iphonesimulator` is also how Apple itself internally identifies the ABIs in calls to tools like `xcrun --show-sdk-path --sdk iphoneos` (except when they use `iOS`, or `ios` and `ios-simulator`... there are some truly remarkable inconsistencies in Apple internal naming sometimes...)

And yes - an iPad identifies as iOS... except when it doesn't, and it identifies as iPadOS - but that never impacts on compilation. There isn't a separate API/framework for iPad support, even though Apple goes to extraordinary lengths to refer to iPadOS as a separate entity in documentation.

freakboy3742:
> Sticking with official names is fine, it's just longer than any of the others.

👍 

> There's no way to set overrides with envvars, but you can always just do multiple runs with different env vars manually set if you really want to. The ability to control how values inherit from base / other overrides makes these much more powerful than the classic platform-only variables.

Yeah - based on the documentation, I can see all sorts of possibilities here. I'll work on simplifying the patch early next week.

> `select = "*simulator"` would select every identifier that ends in `simulator`, which is just these two, and you could add a custom flag just for simulators that way.

Got it. I'm not sure there will be any other "simulator" frameworks - Android doesn't have a different ABI for emulators, so there's no analog there. But point taken about what is possible in theory.

henryiii:
Device/simulator as part of the arch was just an idea. I think it makes sense, but curious what others think (@joerick for example). I think the weirdest part of making it part of the platform is having the `ios` shortcut, which we don't have anything like for the other platforms, while we already have `auto` shortcuts for arch.

mkdocs.yml
joerick, line 1: Thanks for making these docs changes! I'll work on the organisation further in #2280, once this is merged, because I have some other changes I want to make, so I don't think it's worth discussing too much here.

I do like the breaking out the platform information into it's own section/page though. :)

test/test_ios.py
```diff
@@ -0,0 +1,106 @@
...
+        ["xcodebuild", "-version"],
+        text=True,
+        check=True,
+        stdout=subprocess.PIPE,
+    ).stdout
+    lines = output.splitlines()
+    _, version_str = lines[0].split()
+
+    version_parts = version_str.split(".")
+    return (int(version_parts[0]), int(version_parts[1]))
```
joerick, line 37: This also exists in `test/test_macos_archs.py`, so it can be moved into `test/utils.py` and simply referenced here.

test/test_projects/c.py
joerick: Yeah, that's fine. I think really all we care about is that it can call a C library function here.

joerick (commented):
Another preliminary review as it sounds like there are still some changes to come.

Glad to hear that the architecture refactor might condense things down into one platform.

As an aside, the fact that Russell didn't know about the override technique is a good reminder/motivation that we probably want to document pyproject.toml configuration and overrides primarily in #2280 - rather than primarily pointing people to the env vars.

joerick:
> Device/simulator as part of the arch was just an idea. I think it makes sense, but curious what others think, especially before you start changing code (@joerick for example).

Yeah, I like that idea. I'm imagining arch values like `arm64-iphoneos`, `arm64-iphonesimulator` and `x86_64-iphonesimulator`, as I think they're also used as values of `sys.implementation_.multiarch`? Granted, it's a bit of a deviation of the meaning of "arch", but I think it's neater than the many-platforms version we have here.

henryiii:
By the way, not in this PR, but shortly afterwards I'd like to move all the platform files into a directory, so we have `platform.windows` and such. We are getting quite a few of them at this point. :)

joerick:
> I'd like to move all the platform files into a directory

Spooky. I'm just now working on a PR where I made that change 20 minutes ago :)

henryiii:
I've managed to make a couple of conflicts with #2276. I can rebase if you prefer.

freakboy3742:
As discussed, I've updated the implementation to use a single platform, but a "multiarch" interpretation of architecture. As expected, using the `overrides` approach [significantly simplified the config for Pillow](https://github.com/freakboy3742/Pillow/blob/ios-build/pyproject.toml#L114), which was the major use case I had for the split platform config.

I've also merged with main, and the other comments that have been made to date; I've also been able to simplify the comments about restrictions on parallel test instances, as I've been able to fix the underlying bug upstream in CPython.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,603 @@
...
+                        # Build a sdist of the project
+                        call(
+                            "python",
+                            "-m",
+                            "build",
+                            build_options.package_dir,
+                            "--sdist",
+                            f"--outdir={identifier_tmp_dir}",
+                            capture_stdout=True,
+                        )
```
mayeut, line 533: This fails with `python` not found in PATH. If the build frontend is build (or build[uv]), it could succeed adding `env=build_env,` but will fail with `pip` frontend.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,603 @@
...
+                if built_wheel.name.endswith("none-any.whl"):
+                    raise errors.NonPlatformWheelError()
+
+                log.step_end()
+
+            if build_options.test_command and build_options.test_selector(config.identifier):
+                if not config.is_simulator:
+                    log.step("Skipping tests on non-simulator SDK")
+                elif config.arch != os.uname().machine:
+                    log.step("Skipping tests on non-native simulator architecture")
```
mayeut, line 492: would it work with Rosetta 2 ? (x86_64 on arm64 mac0S)

cibuildwheel/ios.py
```diff
@@ -0,0 +1,603 @@
...
+            # Cargo's binary directory (to allow for Rust compilation)
+            str(Path.home() / ".cargo" / "bin"),
+            # The bare minimum Apple system paths.
+            "/usr/bin",
+            "/bin",
+            "/usr/sbin",
+            "/sbin",
+            "/Library/Apple/usr/bin",
+        ]
+    )
```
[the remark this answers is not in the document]
freakboy3742: `pkg-config` is one vector that causes leakage. Another is when you have a build tool that uses the location of its own binary as an implicit root for libraries to find. Yet another - although one that only occurs on Intel Macs - is that `/usr/local/lib` is on the default `DYLD_LIBRARY_PATH` - and that's the location where Homebrew puts its binaries. Admittedly, this last one is becoming less of an issue over time as Intel Macs become less common, but at least for now, it's still a configuration we need to accomodate.

Just removing `/opt/homebrew` isn't enough in my experience. You have no idea what else the end-user has in their path, and *anything* that adds a library could be a problem. 

In terms of workarounds, though - if you need `cmake` (or any other tool), you can set `environment.PATH` in your `pyproject.toml` configuration. I do this in the pillow configuration to ensure that local build dependencies are included. You can't just add `/opt/homebrew/bin` without re-introducing the original problem; but you *can* use `/opt/homebrew/Cellar/cmake/3.31.2/bin` (or `$(dirname $(realpath /opt/homebrew/bin/cmake))` if you want to avoid the specific version number).

cibuildwheel/ios.py
freakboy3742: Good catch. I've defined `test-sources` in all my practical test cases; so this branch isn't getting tested.

Adding `env=build_env` is an obvious fix for the `build` case; but I can't see an obvious fix for the `pip` case. A literal "cp -r *" would *work* - but in my experience its a real problem in practice, because that will copy `.git` and `.venv` folders, which (1) take a long time to copy, and (2) can cause problems with building if there are any local binaries. 

I'm wonder if it might be better to just raise an error on this branch of the code. `test-sources` really is the best option for iOS projects, as you copy over *just* the test code into the test app; this branch is really a "fallback to make things easy as a default". But if the default won't actually be that easy.

There's a middle ground where we raise an error for the `pip` build case, but allow an sdist build for the `build` case... but I'm not sure if it's worth carrying that extra code just to make a subset of one case a little easier.

cibuildwheel/ios.py
freakboy3742: I honestly hadn't even considered that as an option. I guess the simulator is "just a macOS binary", so it might be possible. The complication is that we're not starting the simulator directly; we're using xcodebuild to start the simulator. 

From a quick search, it looks like once upon a time, it might have been possible to specify the simulator as `iPhone SE (3rd Generation),arch=x86_64`; but in my testing, that option doesn't work any more (or, at least, the option works, but the x86_64 simulator binaries are no longer provided by default, so it fails to start a simulator).

Given that it appears that the default CI configurations won't have x86_64 binaries available, it seems like skipping cross-platform tests, and encouraging the use of x86-64 build machines in CI will be the more reliable approach.

cibuildwheel/ios.py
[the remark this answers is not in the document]
mayeut: Should `DYLD_LIBRARY_PATH` be popped from the environment ?
Should we add some known safe tools - or add an option for them ?

We don't always know the path to some tools or it varies between CI & local tests. I could probably work around that with environment variable before invoking cibuildwheel but that seems overly verbose for something like that. How about giving - config - a list of executable names that should remain in PATH if found ? These could be symlinked in a temp directory added to the sanitized PATH ?

cibuildwheel/ios.py
mayeut: If overall we want to always use test-sources (even if only an override), then we're probably better off raising a configuration error in this case.

cibuildwheel/ios.py
mayeut: let's leave at is then. thanks for the detailed answered.

cibuildwheel/ios.py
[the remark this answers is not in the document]
freakboy3742: Removing `DYLD_LIBRARY_PATH` from the environment definitely makes sense.

As for a list of "known safe tools" - My concern is finding a way to make that work that can't backfire. For example - if an end user manually compiled and installed cmake into `/usr/local`, any mechanism that is based on "find the executable" would put `/usr/local/bin` into the path, and we'd be back to the same situation, in a way that can't be resolved without telling the user to restructure their entire system.

Homebrew is perhaps a special case - it would be possible to special case individual tools in the Homebrew cellar; but that effectively means making 2 moderately significant policy decisions:
* Explicitly blessing Homebrew, and not other mechanisms for installing common tools
* Explicitly blessing cmake, but not `<insert-pet-build-tool-here>`, as a "known safe tool".

Both of these seem like intensely political arguments where we're not going to be able to keep everyone happy.

My inclination here is that the better approach is to leave it to the end-user to configure with `CIBW_ENVIRONMENT`. A CI config file can set a `PATH` that is appropriate for the CI environment's specific requirements; end users can specify a `PATH` that reflects their local conditions (with documented warnings about blanket inclusions of Homebrew etc).

cibuildwheel/ios.py
freakboy3742: To clarify -  are you suggesting a configuration error for *any* "`tests-sources` is undefined" case, or only for the "`test-sources` is undefined and build backend is pip" case?

cibuildwheel/ios.py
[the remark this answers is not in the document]
mayeut: > any mechanism that is based on "find the executable" would put /usr/local/bin into the path, and we'd be back to the same situation, in a way that can't be resolved without telling the user to restructure their entire system.

I was thinking something along the line of (pseudo-code):
```
tools_bin = mkdtemp()
for tool in safe_tools:
    tool_path = which(tool)
    if tool_path:
        ln -s tool_path tools_bin / tool_path.name
PATH += tools_bin 
```

cibuildwheel/ios.py
mayeut: configuration error for any "tests-sources is undefined" case

cibuildwheel/ios.py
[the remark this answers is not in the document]
freakboy3742: I can see how that would work for a tool like cmake that is a standalone binary. It would fall apart if the binary then has other files that it needs (e.g., `pkg-config`) - but then I guess there's no real way to diagnose whether those data files are "cross platform appropriate" (e.g., in the case of `pkg-config`, they wouldn't be, almost by definition). 

I'm guessing this setting would only be meaningful on iOS (and potentially other cross-platform builds that do an "environment cleanse" as part of the build - possibly Android)? 

Is this something we'd need to add to this PR, or could it be tackled in a follow up PR?

docs/platforms/macos.md
```diff
@@ -0,0 +1,32 @@
...
+Because the builds are happening without full isolation, there might be some differences compared to CI builds (Xcode version, OS version, local files, ...) that might prevent you from finding an issue only seen in CI.
+
+In order to speed-up builds, cibuildwheel will cache the tools it needs to be reused for future builds. The folder used for caching is system/user dependent and is reported in the printed preamble of each run (e.g. `Cache folder: /Users/Matt/Library/Caches/cibuildwheel`).
+
+You can override the cache folder using the `CIBW_CACHE_PATH` environment variable.
+
+!!! warning
+    cibuildwheel uses official python.org macOS installers for CPython but those can only be installed globally.
+
+    In order not to mess with your system, cibuildwheel won't install those if they are missing. Instead, it will error out with a message to let you install th missing CPython:
```
mhsmith, line 20:
```suggestion
    In order not to mess with your system, cibuildwheel won't install those if they are missing. Instead, it will error out with a message to let you install the missing CPython:
```

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+title: 'iOS'
+---
+
+# iOS builds
+
+## Pre-requisites
+
+You must be building on a macOS machine, with Xcode installed. The Xcode installation must have an iOS SDK available, with all license agreements agreed to by the user. To check if an iOS SDK is available, open the Xcode settings panel, and check the Platforms tab. This will also ensure that license agreements have been acknowledged.
+
+Building iOS wheel also requires a working macOS Python configuration. See the notes on [macOS builds](./macos.md) for details about configuration of the macOS environment.
```
mhsmith, line 11:
```suggestion
Building iOS wheels also requires a working macOS Python installation. See the notes on [macOS builds](./macos.md) for details about configuration of the macOS environment.
```

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Specifying an iOS build
+
+iOS is effectively 2 platforms - physical devices, and simulators. While the API for these two platforms are identical, the ABI is not compatible, even when dealing with a device and simulator with the same CPU architecture. For this reason, the architecture specification for iOS builds includes *both* the CPU architecture *and* the ABI that is being targeted. There are three possible values for architecture on iOS; the values match those used by `sys.implementation._multiarch` when running on iOS:
+
+* `arm64-iphoneos` (for physical iOS devices);
+* `arm64-iphonesimulator` (for iOS simulators running on Apple Silicon macOS machines); and
+* `x64_64-iphonesimulator` (for iOS simulators running on Intel macOS machines).
+
+By default, cibuildwheel will build wheels for all three of these targets.
```
mhsmith, line 21: This may change depending on the conversation above.

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+
+If you need to specify different compilation flags or other properties on a per-ABI or per-CPU basis, you can use [configuration overrides](..#overrides) with a `select` clause that targets the specific ABI or architecture. For example, consider the following example:
```
mhsmith, line 23: The overrides link is broken.

docs/options.md
```diff
@@ -435,6 +436,7 @@ Options:
 - macOS: `x86_64` `arm64` `universal2`
 - Windows: `AMD64` `x86` `ARM64`
 - Pyodide: `wasm32`
+- iOS: `arm64-iphoneos` `arm64-iphonesimulator` `x86_64-iphonesimulator`
```
mhsmith, line 439: The mixed underscores and hyphens are hard to remember. Suggest using underscores everywhere, including in `CIBW_BUILD`. This is consistent both with the existing `x86_64`, and with wheel tags, which build identifiers are obviously supposed to resemble.

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+
+Building iOS wheel also requires a working macOS Python configuration. See the notes on [macOS builds](./macos.md) for details about configuration of the macOS environment.
+
+## Specifying an iOS build
+
+iOS is effectively 2 platforms - physical devices, and simulators. While the API for these two platforms are identical, the ABI is not compatible, even when dealing with a device and simulator with the same CPU architecture. For this reason, the architecture specification for iOS builds includes *both* the CPU architecture *and* the ABI that is being targeted. There are three possible values for architecture on iOS; the values match those used by `sys.implementation._multiarch` when running on iOS:
+
+* `arm64-iphoneos` (for physical iOS devices);
+* `arm64-iphonesimulator` (for iOS simulators running on Apple Silicon macOS machines); and
+* `x64_64-iphonesimulator` (for iOS simulators running on Intel macOS machines).
```
mhsmith, line 19: Punctuation may change depending on the conversation above.

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+select = "*-arm64-*"
+inherit.test-requires = "append"
+test-requires = ["arm64-testing-helper"]
+```
+
+This configuration would:
+
+ * Specify a `test-sources` and `test-requires` for all iOS targets;
+ * Add a `PATH` setting that will be used on physical iOS devices; and
+ * Add `arm64-testing-helper` to the test environment for all ARM64 devices (whether simulator or device).
```
mhsmith, line 44: The only reason this doesn't match other ARM64 plaforms is because Windows and macOS don't  have a hyphen after `arm64`, and Linux calls it `aarch64`. But this is a bit too fragile to recommend. Suggest explicitly limiting the pattern to iOS.

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+
+iOS builds will honor the `IPHONEOS_DEPLOYMENT_TARGET` environment variable to set the minimum supported API version for generated wheels. This will default to `13.0` if the environment variable isn't set.
+
+## Cross platform builds
+
+iOS builds are *cross platform builds*, as it not possible to run compilers and other build tools "on device". The pre-compiled iOS binaries used to support iOS builds include tooling that can convert any virtual environment into a cross platform virtual environment - that is, an environment that can run binaries on the build machine (macOS), but, if asked, will respond as if it is an iOS machine. This allows pip, build, and other build tools to perform iOS-appropriate behaviour.
+
+## Build frontend support
+
+iOS builds support both the pip and build build frontends. In principle, support for uv with the `build[uv]` frontend should be possible, but uv [doesn't currently have support for cross-platform builds](https://github.com/astral-sh/uv/issues/7957), and [doesn't have support for iOS (or Android) tags](https://github.com/astral-sh/uv/issues/8029).
```
mhsmith, line 56: Use formatting to indicate that "build" is a command.

```suggestion
iOS builds are *cross platform builds*, as it not possible to run compilers and other build tools "on device". The pre-compiled iOS binaries used to support iOS builds include tooling that can convert any virtual environment into a cross platform virtual environment - that is, an environment that can run binaries on the build machine (macOS), but, if asked, will respond as if it is an iOS machine. This allows `pip`, `build`, and other build tools to perform iOS-appropriate behaviour.

## Build frontend support

iOS builds support both the `pip` and `build` build frontends. In principle, support for `uv` with the `build[uv]` frontend should be possible, but `uv` [doesn't currently have support for cross-platform builds](https://github.com/astral-sh/uv/issues/7957), and [doesn't have support for iOS (or Android) tags](https://github.com/astral-sh/uv/issues/8029).
```

docs/platforms/ios.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Tests
+
+If tests have been configured, the test suite will be executed on the simulator matching the architecture of the build machine - that is, if you're building on an ARM64 macOS machine, the ARM64 wheel will be tested on an ARM64 simulator. It is not possible to use cibuildwheel to test wheels on other simulators, or on physical devices.
+
+The iOS test environment can't support running shell scripts, so the `CIBW_TEST_COMMAND` value must be specified as if it were a command line being passed to `python -m ...`. In addition, the test itself must be run "on device", so the local project directory cannot be used to run tests. Instead, the entire project sources must be copied onto the test device; or the project must use `CIBW_TEST_SOURCES` to specify the minimum subset of files that should be copied to the test environment.
```
mhsmith, line 66: It's unclear whether this means the cibuildwheel user must *do* something to copy the entire sources (in which case it should explain how), or that the default behavior is to copy the entire sources (in which case it should say so, both here and in the `CIBW_TEST_SOURCES` documentation).

Also, what exactly is "the entire project sources"? The directory containing the pypyroject.toml file?

mhsmith: There should also be an explanation of how this relates to the PYTHONPATH used for running `python -m`.

Finally, the `CIBW_TEST_SOURCES` documentation says the "temporary folder will be used as the working directory for running the test suite" – is this also true on iOS?

docs/options.md
```diff
@@ -452,14 +454,16 @@ Default: `auto`
 | Windows / ARM64 | `ARM64` | `ARM64` | `ARM64` | |
 | macOS / Intel | `x86_64` | `x86_64` | `x86_64` |  |
 | macOS / Apple Silicon | `arm64` | `arm64` | `arm64` |  |
+| iOS on macOS / Intel | `x86_64-iphonesimulator` | `arm64-iphoneos` `arm64-iphonesimulator` `x86_64-iphonesimulator` | `arm64-iphoneos` `arm64-iphonesimulator` `x86_64-iphonesimulator` |  |
+| iOS on macOS / Apple Silicon | `arm64-iphonesimulator` | `arm64-iphoneos` `arm64-iphonesimulator` `x86_64-iphonesimulator` | `arm64-iphoneos` `arm64-iphonesimulator` `x86_64-iphonesimulator` |  |
 
 If not listed above, `auto` is the same as `native`.
```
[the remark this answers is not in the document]
freakboy3742: > On the existing platforms, `auto` seems to mean "all architectures which could be run on the build machine". The equivalent on iOS would be the current simulator architecture only, i.e. the same as `native`.

I'm not sure that existing platforms are a good point of comparison here. My read is that `auto` should be "the most sensible default setting" - which for iOS, is all platforms that the compiler can support. That differs from most other platforms, because while cross-compilation might be theoretically possible, I can't think of any occasion where cross-compilation would be a reasonable default. 

The interpretation of `native` is a separate issue. I could see an argument that `native` on ARM64 should be arm64 device *and* simulator, rather than just simulator as it's currently defined; happy to take feedback on this one.
 
> Also, the "if not listed above" sentence is confusing, because it is in fact always listed above.

True - I'll clean that up.

examples/github-deploy.yml
```diff
@@ -12,18 +12,47 @@ on:
...
+        include:
+          - archs: auto
+            platform: auto
+          - os: linux-intel
+            runs-on: ubuntu-latest
+          - os: linux-arm
+            runs-on: ubuntu-24.04-arm
+          - os: windows
+            runs-on: windows-latest
+          - os: macOS-intel
```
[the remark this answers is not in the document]
freakboy3742: The capitalisation makes me twitch... but I can't disagree with your analysis :-)

docs/options.md
freakboy3742: I'm not sure I agree here. The use of a hyphen has existing precedent in cibw's build selectors - eg., `cp313-macosx_universal2`. There's also precedent in `sysconfigdata.get_platform()`, which *exclusively* uses hyphens to join (e.g., `macosx-10.13-x86_64`); `x86_64` is used because that's how the CPU identified. In the iOS case, specifically, the value being used  is the value of `sys.implementation._multiarch`, rather than the raw architecture; and that value uses a hyphen. 

Admittedly, *wheel* specifiers are all normalised to underscores; but I'd argue that's more of a quirk of the wheel format.

I could also make a (admittedly weak) argument that the `-iphoneos` / `-iphonesimulator` suffix is so odd it warrants being "separated" from the `ios_arm64` part.

Ultimately, it's a mostly cosmetic change though - I'm happy to take direction from project maintainers if they'd rather normalise to underscores.

docs/platforms/ios.md
freakboy3742: This section has changed somewhat because of the change to make `test-sources` a mandatory setting on iOS; but I've clarified some of the other points you've made here (such as clarifying the "temporary folder" on iOS.

docs/options.md
```diff
@@ -288,18 +289,18 @@ When setting the options, you can use shell-style globbing syntax, as per [fnmat
...
+| PyPy3.11 v7.3 | pp311-macosx_x86_64<br/>pp311-macosx_arm64                             | pp311-win_amd64                                     | pp311-manylinux_x86_64<br/>pp311-manylinux_i686                                                     | pp311-manylinux_aarch64                                                                                                                                                                                               |                                                                                                   |
 
 The list of supported and currently selected build identifiers can also be retrieved by passing the `--print-build-identifiers` flag to cibuildwheel.
 The format is `python_tag-platform_tag`, with tags similar to those in [PEP 425](https://www.python.org/dev/peps/pep-0425/#details).
```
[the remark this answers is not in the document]
mhsmith: Resolved.

docs/options.md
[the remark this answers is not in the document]
mhsmith: I see your point about cross compilation, but you could argue that building for a simulator isn't really cross compilation at all, because it's just macOS with different system libraries. 

And in a CI situation, building x86 on ARM or vice versa doesn't seem like a sensible default, because you won't be able to test the result. To test both architectures, the build needs to be split across two runners, just as you've done in your Pillow branch.

docs/options.md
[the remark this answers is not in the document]
mhsmith: A minor nitpick, but everywhere else in the docs lists iOS before Pyodide.

docs/options.md
```diff
@@ -1443,10 +1448,14 @@ Alternatively, you can use the [`CIBW_TEST_SOURCES`](#test-sources) setting to
 create a temporary folder populated with a specific subset of project files to
 run your test suite.
 
-The command is run in a shell, so you can write things like `cmd1 && cmd2`.
+On all platforms other than iOS, the command is run in a shell, so you can write
+things like `cmd1 && cmd2`.
+
+On iOS, the value of the `CIBW_TEST_COMMAND` setting is interpreted as the arguments
+to pass to `python -m`. Shell commands cannot be used.
```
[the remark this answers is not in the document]
mhsmith: Resolved.

docs/platforms/macos.md
mhsmith: Resolved.

docs/platforms/ios.md
mhsmith: It's still broken in the RTD preview.

docs/platforms/ios.md
mhsmith: Resolved.

docs/platforms/ios.md
mhsmith: Resolved.

docs/options.md
mhsmith: > The use of a hyphen has existing precedent in cibw's build selectors - eg., `cp313-macosx_universal2`

That also follows the [wheel tag format](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/), which separates the Python tag from the platform tag with a hyphen, while not allowing a hyphen within either part. This makes it easier to parse, and makes the structure visually obvious.

What do the cibuildwheel maintainers think?

docs/platforms/ios.md
mhsmith: And so are the other links to `/options` in ios.md.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,589 @@
...
+            "cibuildwheel: pip available on PATH doesn't match our installed instance. "
+            "If you have modified PATH, ensure that you don't overwrite cibuildwheel's "
+            "entry or insert pip above it."
+        )
+        raise errors.FatalError(msg)
+    call("pip", "--version", env=env)
+
+    # Ensure that IPHONEOS_DEPLOYMENT_TARGET is set in the environment
+    ios_deployment_target = os.getenv("IPHONEOS_DEPLOYMENT_TARGET", "13.0")
+    env["IPHONEOS_DEPLOYMENT_TARGET"] = ios_deployment_target
```
joerick, line 339: I think this would be more idiomatic, and would prevent overwriting values set by CIBW_ENVIRONMENT. unless there's a reason to specifically use `os.getenv` that I'm missing?

```suggestion
    env.setdefault("IPHONEOS_DEPLOYMENT_TARGET", "13.0")
```

cibuildwheel/ios.py
```diff
@@ -0,0 +1,589 @@
...
+    else:
+        assert_never(build_frontend)
+
+    return host_install_path, env
+
+
+def build(options: Options, tmp_path: Path) -> None:
+    if sys.platform != "darwin":
+        msg = "iOS binaries can only be built on macOS"
+        raise ValueError(msg)
```
joerick, line 363: This will give a better formatted error message.

```suggestion
        raise errors.FatalError(msg)
```
(import from `from cibuildwheel import errors`)

cibuildwheel/ios.py
```diff
@@ -0,0 +1,589 @@
...
+
+        built_wheels: list[Path] = []
+
+        for config in python_configurations:
+            build_options = options.build_options(config.identifier)
+            build_frontend = build_options.build_frontend or BuildFrontendConfig("pip")
+            # uv doesn't support iOS
+            if build_frontend.name == "build[uv]":
+                msg = "uv doesn't support iOS"
+                raise ValueError(msg)
```
joerick, line 396:
```suggestion
                raise errors.FatalError(msg)
```

cibuildwheel/architecture.py
```diff
@@ -63,6 +64,12 @@ class Architecture(StrEnum):
     # WebAssembly
     wasm32 = auto()
 
+    # iOS "multiarch" architectures that include both
+    # the CPU architecture and the ABI.
+    arm64_device = "arm64-iphoneos"
+    arm64_simulator = "arm64-iphonesimulator"
+    x86_64_simulator = "x86_64-iphonesimulator"
```
joerick, line 71: Clearly Apple is lugging around some technical debt here, but I think it's probably on my brain if we don't have two different names for these archs.

```suggestion
    arm64_iphoneos = "arm64-iphoneos"
    arm64_iphonesimulator = "arm64-iphonesimulator"
    x86_64_iphonesimulator = "x86_64-iphonesimulator"
```

docs/options.md
joerick: Ahh, thanks for raising this @mhsmith. I think I originally came up with the thought to mirror `sys.implementation._multiarch`, but I hadn't yet seen how [iOS wheel tags look](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#ios).

from that doc:

> The combination of `arch_sdk` is referred to as the “multiarch”. There are three possible values for multiarch:
> 
> - `arm64_iphoneos`, for physical iPhone/iPad devices. This includes every iOS device manufactured since ~2015;
> 
> - `arm64_iphonesimulator`, for simulators running on Apple Silicon macOS hardware; and
> 
> - `x86_64_iphonesimulator`, for simulators running on x86_64 hardware.

I do think the wheel tag is a better reference for cibuildwheel. And I'd agree with the delimiter arguments made by @mhsmith. So I'd say that underscores would suit better. Apologies for the bum steer @freakboy3742 !

docs/options.md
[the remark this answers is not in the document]
joerick: I'd agree with the current definition of `native`. 

I'm actually somewhat torn on the definition of `auto`. I tend to think that the normal way to set up iOS builds would be to have a separate arm64 runner for that, so the default of building all three multiarchs makes sense to me. So I guess I'm +0 for this current approach.

docs/options.md
```diff
@@ -1443,10 +1436,12 @@ Alternatively, you can use the [`CIBW_TEST_SOURCES`](#test-sources) setting to
 create a temporary folder populated with a specific subset of project files to
 run your test suite.
 
-The command is run in a shell, so you can write things like `cmd1 && cmd2`.
+On all platforms other than iOS, the command is run in a shell, so you can write things like `cmd1 && cmd2`.
+
+On iOS, the value of the `CIBW_TEST_COMMAND` setting is interpreted as the arguments to pass to `python -m` - that is, a Python module name, followed by arguments that will be assigned to `sys.argv`. Shell commands cannot be used.
```
joerick, line 1441: Ha! Clever way to make `pytest arg1 arg2` work transparently :)

test/test_ios.py
```diff
@@ -0,0 +1,101 @@
...
+
+basic_project = test_projects.new_c_project()
+basic_project.files["tests/__init__.py"] = ""
+basic_project.files["tests/__main__.py"] = r"""
+import platform
+import time
+
+# Workaround for CPython#130294
+for i in range(0, 10):
+    time.sleep(1)
```
joerick, line 19: Ah, that's a bummer. Hope this can be fixed upstream soon!

joerick (commented):
Very exciting stuff. Thank you for working on this!

cibuildwheel/ios.py
```diff
@@ -0,0 +1,603 @@
...
+    venv_path: Path,
+    dependency_constraint_flags: Sequence[PathOrStr],
+) -> dict[str, str]:
+    """Create a cross-compilation virtual environment.
+
+    In a cross-compilation environment, the *host* is the platform you're
+    targeting the *build* is the platform where you're running the compilation.
+    When building iOS wheels, iOS is the host machine and macOS is the build
+    machine.
+
```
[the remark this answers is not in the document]
freakboy3742: So - "host" and "build" are very much established terms for cross compilation. When you invoke `configure`, you use `--host=<gcc triple> --build=<gcc triple>` to define compilation conditions. Meson also [uses build/host terminology](https://mesonbuild.com/Cross-compilation.html).

That said - the terms aren't universal; CMake uses ["build host" and "target"](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html); Cargo refers to [target](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html), but doesn't really have the analogous concept of "build" (because rust, by virtue of how it's distributed, can rely on the fact that that if you have a toolchain at all, it will work locally).

I'm not seeing any existing uses of `target` in the existing codebase being used for this purpose. There's plenty of `target` objects - but they all refer to a "target version" or the destination location for copying files. The only "cross platform" target I can see is in the Windows backend, where it relates to cargo configuration. Even in the context of iOS, the term is entirely internal; there's no new external API that uses "host"/"build" as a distinction.

Ultimately, I'll defer to project maintainers here - but I'd caution against being too "innovative" in terminology.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,589 @@
...
+    def multiarch(self) -> str:
+        return self.identifier.split("-ios_")[1]
+
+    @property
+    def is_simulator(self) -> bool:
+        return self.identifier.endswith("-iphonesimulator")
+
+    @property
+    def slice(self) -> str:
+        return "ios-arm64_x86_64-simulator" if self.is_simulator else "ios-arm64"
```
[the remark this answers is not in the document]
freakboy3742: Why not both :-)

docs/options.md
freakboy3742: No problems - fix incoming shortly!

cibuildwheel/ios.py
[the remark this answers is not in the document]
freakboy3742: To clarify - do you mean:
1. You'd expect a hard coded, "known defaults" version for this PR, with a followup PR making it fully configurable; or
2. It's OK to leave this entire feature for a follow up PR?

If it's (1) - any requests other than `cmake` for the hard-coded list? Maybe `cython` and `cythonize`?

docs/options.md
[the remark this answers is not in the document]
freakboy3742: I agree that normal CI usage would likely be to split across 2 runners (if you're planning to produce x86 wheels *at all*... they're becoming an increasingly niche use case).

I will say it has been *really* convenient to be able to build all three wheels on one target locally just to confirm that I haven't got x86_64 headers wrong when I'm on an ARM laptop. However, I guess `CIBW_ARCHS=all` also exists, so maybe that's a better way to satisfy the "local convenience" use case while preserving useful default behavior for CI. On that basis, I'll make the change to "matching CPU architecture" for auto.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,589 @@
...
+    # load the macos configurations, determine the macOS configuration that
+    # matches the platform we're building, and embed that URL in the parsed iOS
+    # configuration.
+    macos_python_configs = resources.read_python_configs("macos")
+
+    def build_url(item: dict[str, str]) -> str:
+        # The iOS item will be something like cp313-ios_arm64_iphoneos. Drop
+        # the iphoneos suffix, then replace ios with macosx to yield
+        # cp313-macosx_arm64, which will be a macOS configuration item.
+        parts = item["identifier"].split("-")
```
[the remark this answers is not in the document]
freakboy3742: There's a couple of other cleanups I can do here as a result of other changes; but I've incorporated all this into the latest pushed version.

cibuildwheel/ios.py
freakboy3742: No reason other than oversight on my part. Corrected in latest version.

test/test_ios.py
freakboy3742: It's on my list. I've got some ideas that should be sufficient for at least a workaround.

cibuildwheel/ios.py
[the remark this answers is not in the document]
joerick: I meant (2), but I'm okay with either. I wracked my brains to think of other tools that might be useful here, but I could only think of cmake. But the scientific computing crew might have a few. I don't think that the cython stuff should be necessary, as that should be listed in the `build-system.requires` and therefore installed by the build frontend.

docs/options.md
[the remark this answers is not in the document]
mhsmith: Resolved.

docs/options.md
mhsmith: Resolved.

cibuildwheel/ios.py
```diff
@@ -0,0 +1,591 @@
...
+
+    # An iOS environment requires 2 python installs - one for the build machine
+    # (macOS), and one for the host (iOS). We'll only ever interact with the
+    # *host* python, but the build Python needs to exist to act as the base
+    # for a cross venv.
+    tmp.mkdir()
+    implementation_id = python_configuration.identifier.split("-")[0]
+    log.step(f"Installing Build Python {implementation_id}...")
+    if implementation_id.startswith("cp"):
+        free_threading = "t-iphone" in python_configuration.identifier
```
mhsmith, line 249: A couple of problems with this:

* The first word of the platform tag is "ios", not "iphone".
* If additional Python ABI flags are added in the future, "t" won't necessarily appear immediately before the hyphen.

cibuildwheel/ios.py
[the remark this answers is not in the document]
mhsmith: If we were starting from scratch then I would agree with your preferred terms. However, there are many existing uses of "host" used to mean "the build machine", and not just in the context of Docker.

cibuildwheel/ios.py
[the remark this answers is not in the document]
freakboy3742: Ok - on that basis, I'll leave it for now, and come back once this has landed with a fix for the cmake et al issue.

cibuildwheel/ios.py
freakboy3742: Good catch on the `t-ios` prefix; I've corrected that.

As for handling additional ABI flags - I agree, but (1) this is a pattern copied directly from the macOS and Windows backends; and (2) new ABIs are a sufficiently infrequent experience that I think we can treat that as a future problem.

docs/platforms/ios.md
mhsmith: Resolved.

cibuildwheel/ios.py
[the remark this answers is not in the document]
joerick: I think that the fact that `configure` uses 'host' to mean the opposite of what `cmake` means with 'build host' reinforces my argument that 'host' should probably be avoided.[^1]

> However, there are many existing uses of "host" used to mean "the build machine", and not just in the context of Docker, so I think that's best avoided.

Right - [quite a few references to 'host'](https://github.com/search?q=repo%3Apypa%2Fcibuildwheel%20host%20machine&type=code), referring to the Docker definition, which is spiritually closer to 'build' than 'target'. We also sometimes refer to the 'host python', meaning the python used to run cibuildwheel.

[^1]: I guess in the olden days 'host' just meant 'computer that runs the program', but that usage has faded, making this term confusing - to me it's closer to 'computer that provides a service', when we're talking about compiling programs for other computers, you can see where the confusion arises :)

> I'm not seeing any existing uses of `target` in the existing codebase being used for this purpose.

You're right, maybe that bit is mostly in my head! I still think it's the best word, given that cmake and cargo are using it.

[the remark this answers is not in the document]
freakboy3742: Fair enough. At the end of the day, this isn't a user-facing feature, so it doesn't matter *that* much (beyond maintenance overhead); I'll modify the usage, with a comment highlighting that there's at least *some* ambiguity in how "host/target" are referenced.

test/test_ios.py
freakboy3742: Actually... I think I've got a better solution that won't require any workaround at all. Fix incoming.

freakboy3742: (I'll still have to fix the issue upstream; it just won't be a source of a problem for this test)

cibuildwheel/ios.py
```diff
@@ -0,0 +1,593 @@
...
+                        testbed_path,
+                        env=build_env,
+                    )
+
+                    if not build_options.test_sources:
+                        # iOS requires an explicit test-sources, as the project directory
+                        # isn't visible on the simulator.
+
+                        msg = "Testing on iOS requires a definition of test-sources."
+                        raise ValueError(msg)
```
mayeut, line 524:
```suggestion
                        raise errors.FatalError(msg)
```

cibuildwheel/ios.py
```diff
@@ -0,0 +1,593 @@
...
+    # Skip builds as required by BUILD/SKIP
+    python_configurations = [c for c in python_configurations if build_selector(c.identifier)]
+
+    return python_configurations
+
+
+def install_target_cpython(tmp: Path, config: PythonConfiguration, free_threading: bool) -> Path:
+    if free_threading:
+        msg = "Free threading builds aren't available for iOS (yet)"
+        raise ValueError(msg)
```
mayeut, line 124:
```suggestion
        raise errors.FatalError(msg)
```

cibuildwheel/ios.py
```diff
@@ -0,0 +1,593 @@
...
+def setup_python(
+    tmp: Path,
+    python_configuration: PythonConfiguration,
+    dependency_constraint_flags: Sequence[PathOrStr],
+    environment: ParsedEnvironment,
+    build_frontend: BuildFrontendName,
+) -> tuple[Path, dict[str, str]]:
+    if build_frontend == "build[uv]":
+        msg = "uv doesn't support iOS"
+        raise ValueError(msg)
```
mayeut, line 241:
```suggestion
        raise errors.FatalError(msg)
```

cibuildwheel/ios.py
```diff
@@ -0,0 +1,593 @@
...
+        free_threading = "t-ios" in python_configuration.identifier
+        build_python = install_build_cpython(
+            tmp,
+            python_configuration.version,
+            python_configuration.build_url,
+            free_threading,
+        )
+    else:
+        msg = "Unknown Python implementation"
+        raise ValueError(msg)
```
mayeut, line 260:
```suggestion
        msg = f"Unknown Python implementation: {implementation_id}"
        raise errors.FatalError(msg)
```
or we shouldn't be here at all by design in which case an `assert implementation_id.startswith("cp")` might be more appropriate ?

cibuildwheel/ios.py
```diff
@@ -0,0 +1,593 @@
...
+        target_python = (
+            target_install_path
+            / "Python.xcframework"
+            / python_configuration.xcframework_slice
+            / "bin"
+            / f"python{python_configuration.version}"
+        )
+    else:
+        msg = "Unknown Python implementation"
+        raise ValueError(msg)
```
mayeut, line 278: This one is definitely dead code given the check for the `install_build_cpython` a dozen lines above

cibuildwheel/resources/build-platforms.toml
```diff
@@ -163,3 +163,10 @@ python_configurations = [
 python_configurations = [
   { identifier = "cp312-pyodide_wasm32", version = "3.12", pyodide_version = "0.27.0", pyodide_build_version = "0.29.2", emscripten_version = "3.1.58", node_version = "v20" },
 ]
+
+[ios]
+python_configurations = [
+  { identifier = "cp313-ios_arm64_iphoneos", version = "3.13", url = "https://github.com/beeware/Python-Apple-support/releases/download/3.13-b5/Python-3.13-iOS-support.b5.tar.gz" },
+  { identifier = "cp313-ios_x86_64_iphonesimulator", version = "3.13", url = "https://github.com/beeware/Python-Apple-support/releases/download/3.13-b5/Python-3.13-iOS-support.b5.tar.gz" },
+  { identifier = "cp313-ios_arm64_iphonesimulator", version = "3.13", url = "https://github.com/beeware/Python-Apple-support/releases/download/3.13-b5/Python-3.13-iOS-support.b5.tar.gz" },
+]
```
mayeut, line 172: Could `bin/update_pythons.py` be updated to update those as well ?

freakboy3742: Sure - that should be possible. I'll add it in.

freakboy3742: Done - most recent commit has modifications that will update the download URLs for iOS builds.

README.md
```diff
@@ -55,18 +55,19 @@ Usage
...
-| Cirrus CI       | ✅    | ✅    | ✅      | ✅        | ✅        |             |
+|                 | Linux | macOS | Windows | Linux ARM | macOS ARM | Windows ARM | iOS |
+|-----------------|-------|-------|---------|-----------|-----------|-------------|-----|
+| GitHub Actions  | ✅    | ✅    | ✅       | ✅        | ✅        | ✅²         | ✅³  |
+| Azure Pipelines | ✅    | ✅    | ✅       |           | ✅        | ✅²         |     |
+| Travis CI       | ✅    |       | ✅      | ✅        |           |             |     |
+| AppVeyor        | ✅    | ✅    | ✅      |           | ✅        | ✅²         |     |
+| CircleCI        | ✅    | ✅    |         | ✅        | ✅        |             |     |
+| Gitlab CI       | ✅    | ✅    | ✅      | ✅¹       | ✅        |             |     |
+| Cirrus CI       | ✅    | ✅    | ✅      | ✅        | ✅        |             |     |
```
mayeut, line 66: This only refers to GitHub Actions supporting building wheels for iOS.
There's probably little reason for other providers not to support this ?
It's probably already at least partially tested i this PR.

docs/options.md
```diff
@@ -452,14 +442,14 @@ Default: `auto`
 | Windows / ARM64 | `ARM64` | `ARM64` | `ARM64` | |
 | macOS / Intel | `x86_64` | `x86_64` | `x86_64` |  |
 | macOS / Apple Silicon | `arm64` | `arm64` | `arm64` |  |
-
-If not listed above, `auto` is the same as `native`.
```
mayeut, line 456: This should be kept below the table ?

cibuildwheel/ios.py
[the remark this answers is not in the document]
mayeut: Just a note to say that `CIBW_BEFORE_BUILD=pip install cmake` wouldn't work in this instance (the build venv is saying it's iOS, not macOS, there are no wheels).
I've tried `CIBW_BEFORE_BUILD="_PYTHON_HOST_PLATFORM=macosx-10.10-universal2 pip install cmake"` but the patched venv does not respect the - undocumented - `_PYTHON_HOST_PLATFORM` and always patches `sysconfig.get_platform()`

If it had been (1), I'd say `cmake`, `ninja` & `swig` for a start.

just FYI, here's the diff I apply to test the PR on a pet project of mine:
```diff
diff --git a/cibuildwheel/ios.py b/cibuildwheel/ios.py
index d18a370b..8080efae 100644
--- a/cibuildwheel/ios.py
+++ b/cibuildwheel/ios.py
@@ -206,6 +206,14 @@ def cross_virtualenv(
     #
     # To prevent problems, set the PATH to isolate the build environment from
     # sources that could introduce incompatible binaries.
+    tools_path = venv_path / "cibw_tools"
+    tools_path.mkdir()
+    allowed_tools = ("cmake", "ninja", "swig")
+    for tool in allowed_tools:
+        tool_path = shutil.which(tool)
+        if tool_path is not None:
+            tools_path.joinpath(tool).symlink_to(tool_path)
+
     env["PATH"] = os.pathsep.join(
         [
             # The target python's binary directory
@@ -214,6 +222,8 @@ def cross_virtualenv(
             str(venv_path / "bin"),
             # Cargo's binary directory (to allow for Rust compilation)
             str(Path.home() / ".cargo" / "bin"),
+            # Allowed tools
+            str(tools_path),
             # The bare minimum Apple system paths.
             "/usr/bin",
             "/bin",
```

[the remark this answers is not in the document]
joerick: Thanks for the analysis @mayeut! That looks like a great starting point for the followup PR.

docs/options.md
mhsmith: I'm not sure what the purpose of this sentence is, because it actually is listed above for all supported platforms.

mayeut: Linux s390x/ppc64le/... for example are not listed in the table.

README.md
freakboy3742: Good point - I only added Github because that was the only one where I had run it personally; but it should work anywhere that there's a macOS runner that has iOS tooling, and we're validating most of those in CI. I'll update the table.

docs/options.md
freakboy3742: Understood - I'll restore the line.

cibuildwheel/ios.py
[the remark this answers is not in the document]
freakboy3742: It's also worth noting that fixing `_PYTHON_HOST_PLATFORM` handling is also possible - although it would require an updated support package to roll out that support. 
 
I'm hoping to kick start a bigger discussion (possibly leading to a PEP) around cross-platform build environments; I'm expecting the handling of `_PYTHON_HOST_PLATFORM` to be part of that discussion.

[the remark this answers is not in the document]
freakboy3742: Also - reading between the lines... this implies you've been able to build (and test?) iOS wheels for your "pet project" - can I ask the name of your pet 😄 ?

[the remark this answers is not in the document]
mayeut: > Also - reading between the lines... this implies you've been able to build (and test?) iOS wheels for your "pet project" - can I ask the name of your pet 😄 ?

It's https://github.com/mayeut/pybase64, built & tested locally with minimal changes.
The output of the tests is a little bit verbose compared to other tests (but I set verbosity to 1 for the build which is used here for tests) and the output of pytest gets a little bit messed up compared to other platforms. Nothing major to report for this PR.

joerick (approved):
From my PoV, this is looking good! Just a few small notes.

cibuildwheel/architecture.py
```diff
@@ -152,12 +152,17 @@ def auto_archs(platform: PlatformName) -> "set[Architecture]":
...
+            # as the native simulator architecture
+            if native_arch == Architecture.x86_64_iphonesimulator:
+                result = {
+                    Architecture.x86_64_iphonesimulator,
+                }
+            else:
+                result = {
+                    Architecture.arm64_iphonesimulator,
+                    Architecture.arm64_iphoneos,
+                }
```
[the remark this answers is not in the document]
freakboy3742: You stare at something for too long, and the obvious stops standing out :-) Thanks for the cleanup.

freakboy3742:
Not sure what's going on with the new failures... they seem unrelated to the change that was just made. Investigating...

freakboy3742:
Ok... that took longer than expected. For some reason, testing was raising errors that indicated I wasn't using the updated dependency API from #2122... but I was getting that error *before* I merged the #2122 changes into this branch. 

Regardless, I've merged with main, and updated the API in question; It seems to have taken a couple of passes to get all the caches and the like to align, but it looks like it's passing now.

mayeut (approved):

joerick:
> For some reason, testing was raising errors that indicated I wasn't using the updated dependency API from #2122... but I was getting that error before I merged the #2122 changes into this branch.

Most CI services test PRs against a fictional commit, as if it was merged now. Quite handy, it catches errors like the above. It's also the reason that some CIs  refuse to run when there are merge conflicts. Maybe that explains some of the weirdness you were seeing!

henryiii (approved):

joerick:
Many thanks and congrats on landing this @freakboy3742 !

freakboy3742:
> Many thanks and congrats on landing this @freakboy3742 !

Thanks for the reviews and feedback! 

I'm hoping to find some time to work on the "safe tools" addition in a couple of days. Watch this space :-)

henryiii:
Any idea why the iOS tests are timing out a lot?

freakboy3742:
> Any idea why the iOS tests are timing out a lot?

I hadn't noticed that - where are you seeing the timeouts? Main, or branch CI?

henryiii:
I’ve been rebuilding them, but on most PRs it seems to happen 50% of the time. https://github.com/pypa/cibuildwheel/pull/2322 for example.

freakboy3742:
Hrm... that's concerning. It's not a problem I've seen in my own testing. It's a timeout that could happen if there are multiple iOS tests running at the same time - I wonder if it's less prevalent for me because of my timezone. I'll need to dig into this some more.

PR #2287, opened 2025-02-24
fix: avoid races in filesystem ops

### LINKED ISSUE #2279: race condition in output dir creation
### Description

While invoking cibuildwheel in parallel (multiple processes, each building one wheel), I encountered an error like:

```python
  File "cibuildwheel/__main__.py", line 403, in build_in_directory
    output_dir.mkdir(parents=True)
  File "/mnt/sw/nix/store/71ksmx7k6xy3v9ksfkv5mp5kxxp64pd6-python-3.10.13-view/lib/python3.10/pathlib.py", line 1175, in mkdir
    self._accessor.mkdir(self, mode)
FileExistsError: [Errno 17] File exists: 'wheelhouse'
```

I'm embarrassed to say that the logs on the CI server with the exact error message have been purged, so the above is my recreation from memory. Take it with a grain of salt! However, I think the TOCTOU race is clear just by reading the code:

https://github.com/pypa/cibuildwheel/blob/c93d51ec540da7537ae66107a32c60dccd705102/cibuildwheel/__main__.py#L402-L403

These two lines should probably be this one line: `output_dir.mkdir(parents=True, exist_ok=True)`.

I would submit a PR, except a quick [search](https://github.com/search?q=repo%3Apypa%2Fcibuildwheel%20exists&type=code) of `exists` suggests there may be similar issues throughout the code. Or perhaps not, but I wanted to ask the experts. First I should ask, though: is it safe/supported to run cibuildwheel in parallel this way (shared filesystem, shared output dir)?

### Build log

_No response_

### CI config

_No response_

lgarrison:
Use `exist_ok` and `missing_ok` in pathlib operations where we can.

I didn't do an exhaustive audit of pathlib usage, but I did look at places where the code was using `path.exists()`. There's still a couple of places with this pattern:

```python
if path.exists():
    shutil.rmtree(path)
```

I don't think there's an elegant solution that will work across Python versions currently, although in 3.13 I think one can do:

```python
try:
    shutil.rmtree(path)
except FileNotFoundError:
    pass
```

Because: "*Changed in version 3.13*: `rmtree()` now ignores `FileNotFoundError` exceptions for all but the top-level path." (https://docs.python.org/3.13/library/shutil.html#shutil.rmtree)

Fixes #2279.

### DIFF
--- a/test/test_custom_repair_wheel.py
+++ b/test/test_custom_repair_wheel.py
@@ -21,8 +21,7 @@ platform = wheel.stem.split("-")[-1]
 name = f"spam-0.1.0-py2-none-{platform}.whl"
 dest = dest_dir / name
 dest_dir.mkdir(parents=True, exist_ok=True)
-if dest.exists():
-    dest.unlink()
+dest.unlink(missing_ok=True)
 shutil.copy(wheel, dest)
 """
 
--- a/bin/update_how_it_works_image.py
+++ b/bin/update_how_it_works_image.py
@@ -42,8 +42,7 @@ def main() -> None:
     )
 
     dest_path = Path("docs/data/how-it-works.png")
-    if dest_path.exists():
-        dest_path.unlink()
+    dest_path.unlink(missing_ok=True)
 
     Path(screenshot).rename(dest_path)
 
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -399,8 +399,7 @@ def build_in_directory(args: CommandLineArguments) -> None:
 
     output_dir = options.globals.output_dir
 
-    if not output_dir.exists():
-        output_dir.mkdir(parents=True)
+    output_dir.mkdir(parents=True, exist_ok=True)
 
     tmp_path = Path(mkdtemp(prefix="cibw-run-")).resolve(strict=True)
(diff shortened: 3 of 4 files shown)


henryiii:
I think this is tripping up a test because we were catching an exception somewhere that has now changed type? (from a quick glance)

test/test_custom_repair_wheel.py
```diff
@@ -21,8 +21,7 @@
 name = f"spam-0.1.0-py2-none-{platform}.whl"
 dest = dest_dir / name
 dest_dir.mkdir(parents=True, exist_ok=True)
-if dest.exists():
-    dest.unlink()
+dest.unlink(missing_ok=True)
```
joerick, line 24: This one won't work, because it was added in Python 3.8, and the repair script runs in the build, so runs on 3.6 and 3.7.

henryiii: Though it would be fine after #2282?

lgarrison: I rebased on main and added this back in. Let's see if CI passes!

lgarrison:
CI is green!

henryiii (approved):

Issue #2277, opened 2025-02-13
No module named pip on self-hosted runner


michprev:
### Description

Hello,
I am observing the following error when using `pypa/cibuildwheel@v2.22.0` in my action:

```
  /home/username/actions-runner/_work/_tool/Python/3.12.9/x64/bin/python: No module named pip
  Traceback (most recent call last):
    File "<stdin>", line 28, in <module>
    File "/home/username/actions-runner/_work/_tool/Python/3.12.9/x64/lib/python3.12/venv/__init__.py", line 80, in create
      self.post_setup(context)
    File "<stdin>", line 20, in post_setup
    File "/home/username/actions-runner/_work/_tool/Python/3.12.9/x64/lib/python3.12/subprocess.py", line 573, in run
      raise CalledProcessError(retcode, process.args,
  subprocess.CalledProcessError: Command '['/home/username/actions-runner/_work/_tool/Python/3.12.9/x64/bin/python', '-m', 'pip', '--python', '/home/username/actions-runner/_work/_temp/cibw/bin/python', 'install', '/home/username/actions-runner/_work/_actions/pypa/cibuildwheel/v2.22.0']' returned non-zero exit status 1.
```

The error happens at the `Install cibuildwheel` step (https://github.com/pypa/cibuildwheel/blob/c93d51ec540da7537ae66107a32c60dccd705102/action.yml#L36).

The machine is Ubuntu 23 with `python3` and `python3-pip` installed through `apt` (externally managed environment).

henryiii:
You need to have setup-python to use our action. If you cannot use setup-python, you should not use our action either. Just use `cibuildwheel` directly with your externally managed pip.

henryiii:
Our action uses setup-python, which itself is not supposed to break like this - the links above might be the root bug.

joerick:
Closing due to inactivity.

PR #2289, opened 2025-02-25
fix: no module named pip in externally managed envs

michprev:
Fixes `No module named pip` on self-hosted runners with externally managed Python environments. See https://github.com/pypa/cibuildwheel/issues/2277

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -29,7 +29,7 @@ runs:
       id: python
       with:
         python-version: "3.12"
-        update-environment: false
+        update-environment: true
 
     - id: cibw
       run: |


joerick:
FYI the `update-environment: false` is deliberate, so as to avoid polluting the action context for subsequent actions - effectively, avoiding our implementation from leaking out. 

I don't understand your solution. This inline script is run with the installed python from `actions/setup-python`, i'm guessing it always has pip installed? Is there some other way that update-environment changes the behaviour of `setup-python`?

henryiii:
If you can't use setup-python, I don't think you should be able to use cibuildwheel's action either. Just use your external pip directly if that's what you need.

michprev:
I don't understand the fix either. My self-hosted runner comes with python3 and pip pre-installed (apt `python3-pip` package). `python3 -m pip` works without issues. Also, as far as I understand, setup-python installs pip as well. But, for some reason, the inline script fails with the error described in #2277, and setting `update-environment: true` fixes the error.

Czaki:
> I don't understand the fix either

If you do not understand it, it is the next argument to not do this. 

In general, using pip on external managed python (in your case, python is managed by apt) is bad practice and leads to unexpected and hard to debug errors.  

If you do not want to enable `setup-python` action, then please use at least virtualenv.  
As if you create virtualenv and activate it for action, then it will not be external managed.

mayeut:
> Is there some other way that update-environment changes the behaviour of setup-python?

Unfortunately, the most likely explanation is an issue with `setup-python` so the answer would be yes (it shouldn't).
I commented with the 2 relevant issues in #2277.

Closing this as we don't want to leak our implementation.

PR #2291, opened 2025-02-28
feat: stricter selector parsing, refactor to `platforms` module

joerick:
Implementation of an idea from https://github.com/pypa/cibuildwheel/pull/2070#issuecomment-2676888930

cibuildwheel will error out if the `build` config contains something that it doesn't recognise. For `skip` and `test-skip`, it just prints a warning.

This is to highlight misconfiguration as seen in #2070. Notably, the set of identifiers that cibuildwheel will consider is not dependent on platform, so users can still do a global CIBW_BUILD that's shared across platforms.

Also in this PR, I wanted to iterate through platform modules, and so decided it was time to factor them out into their own package.

I'll keep this draft to merge #2286 first, since that adds a new platform.

### DIFF
--- a/cibuildwheel/platforms/__init__.py
+++ b/cibuildwheel/platforms/__init__.py
@@ -14,6 +14,8 @@ from cibuildwheel.typing import GenericPythonConfiguration, PlatformName
 class PlatformModule(Protocol):
     # note that as per PEP544, the self argument is ignored when the protocol
     # is applied to a module
+    def all_python_configurations(self) -> Sequence[GenericPythonConfiguration]: ...
+
     def get_python_configurations(
         self, build_selector: BuildSelector, architectures: set[Architecture]
     ) -> Sequence[GenericPythonConfiguration]: ...
--- a/cibuildwheel/platforms/__init__.py
+++ b/cibuildwheel/platforms/__init__.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Final, Protocol, assert_never
+
+from cibuildwheel.architecture import Architecture
+from cibuildwheel.options import Options
+from cibuildwheel.platforms import ios, linux, macos, pyodide, windows
+from cibuildwheel.selector import BuildSelector
+from cibuildwheel.typing import GenericPythonConfiguration, PlatformName
+
+
+class PlatformModule(Protocol):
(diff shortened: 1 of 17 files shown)


joerick:
This change made the unit tests quite a bit slower (26s), so I broke out the profiler. In doing so I spotted a few places things could be sped up. I'll split out the unrelated changes to a separate PR.

henryiii:
Do you know why this change made it slower, though?

joerick:
Yeah, it was [these lines](https://github.com/pypa/cibuildwheel/pull/2291/commits/b1037d40e71015d23365c27bcd6856e7990db628#diff-cdf254321017f9a32499135cace65adc9085efa0eb28e54902f12e9968c7dd54R478-R498), it ended up doing a lot of TOML loading and running of the BuildSelector.

joerick:
I've kept only the optimisation relevant to this change, the rest are in #2293.

cibuildwheel/platforms/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+def get_platform_module(platform: PlatformName) -> PlatformModule:
+    if platform == "linux":
+        return linux
+    if platform == "windows":
+        return windows
+    if platform == "macos":
+        return macos
+    if platform == "pyodide":
+        return pyodide
+    assert_never(platform)
```
mayeut, line 35: This function is probably not needed anymore given it's just a look-up in the ALL_PLATFORM_MODULES dict now ?

henryiii:
Can be rebased now.

cibuildwheel/platforms/__init__.py
joerick: Done!

cibuildwheel/platforms/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Final, Protocol
+
+from cibuildwheel.architecture import Architecture
+from cibuildwheel.options import Options
+from cibuildwheel.platforms import ios, linux, macos, pyodide, windows
+from cibuildwheel.selector import BuildSelector
+from cibuildwheel.typing import GenericPythonConfiguration, PlatformName
+
+
+class PlatformModule(Protocol):
```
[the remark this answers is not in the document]
joerick: BTW, this module type is working well - mypy caught that the ios module was missing the new `all_python_configurations` function after the rebase.

PR #2293, opened 2025-02-28
chore: optimise unit tests

joerick:
A few changes to speed up the unit tests. Shouldn't change any functionality or test coverage.


cibuildwheel/options.py
```diff
@@ -593,6 +593,10 @@ def __init__(
         except FileNotFoundError:
             self.pyproject_toml = None
 
+        # cache the build options method so repeated calls don't need to
+        # resolve the options again
+        self.build_options = functools.lru_cache()(self.build_options)  # type: ignore[method-assign]
```
henryiii, line 598: This looks ugly. Can't you just use this as a decorator? [Seems like that's in the docs](https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls). There's also the manual way, with `self._build_options = {}` then compute only if `self._build_options` doesn't contain it.

unit_test/validate_schema_test.py
```diff
@@ -12,15 +12,22 @@
 DIR = Path(__file__).parent.resolve()
 
 
-def test_validate_default_schema():
+@pytest.fixture(scope="session")
+def validator() -> validate_pyproject.api.Validator:
+    """
+    Reuse the validator for all tests, to keep unit tests fast.
+    """
+    return validate_pyproject.api.Validator()
```
henryiii, line 20: Historical note: this used to accidentally mutate itself, so a year or two ago this might not have worked. :) (it was injecting defaults, so it probably would still work and I just was really unlucky that I was affected by defaults being injected elsewhere)

cibuildwheel/util/resources.py
```diff
@@ -18,6 +19,8 @@
 CIBUILDWHEEL_SCHEMA: Final[Path] = PATH / "cibuildwheel.schema.json"
 
 
+# this value is cached because it's used a lot in unit tests
+@functools.lru_cache
```
mayeut, line 23:
```suggestion
@functools.cache
```

c.f. https://docs.python.org/3/library/functools.html#functools.cache

> Because it never needs to evict old values, this is smaller and faster than [lru_cache()](https://docs.python.org/3/library/functools.html#functools.lru_cache) with a size limit.

cibuildwheel/options.py
joerick: You can put it as a method decorator, but then the cache is stored on the class itself - i.e. globally, which isn't really what I had intended. I researched it after you mentioned and there are a few posts online which also point out that it's a memory leak if applied at the class level, because all the instances are referenced from the cache.

cibuildwheel/selector.py
```diff
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import fnmatch
 import itertools
 from dataclasses import dataclass
 from enum import Enum
+from fnmatch import fnmatch
```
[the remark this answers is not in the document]
joerick: I got more consistent line-breaks from the autoformatter in the group filtering section with the imports like this. I tend to prefer top-level imports too but where the name is repeated it does feel a bit redundant to me.

unit_test/validate_schema_test.py
joerick: Ah, good to know :)

cibuildwheel/options.py
mayeut: Given how `lru_cache` works on instance method and `cached_property` can't be used, this leaves the manual method which does not require too much to be implemented:
```diff
diff --git a/cibuildwheel/options.py b/cibuildwheel/options.py
index f0999096..bd88da29 100644
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -577,6 +577,11 @@ def __init__(
         self.env = env
         self._defaults = defaults
 
+        # build options cache for different identifiers
+        # lru_cache is not well suited for instance methods
+        # see https://docs.python.org/3/faq/programming.html#faq-cache-method-calls
+        self._build_options: dict[str | None, BuildOptions] = {}
+
         self.reader = OptionsReader(
             None if defaults else self.config_file_path,
             platform=platform,
@@ -591,10 +596,6 @@ def __init__(
         except FileNotFoundError:
             self.pyproject_toml = None
 
-        # cache the build options method so repeated calls don't need to
-        # resolve the options again
-        self.build_options = functools.cache(self.build_options)  # type: ignore[method-assign]
-
     @functools.cached_property
     def config_file_path(self) -> Path | None:
         args = self.command_line_arguments
@@ -671,8 +672,7 @@ def globals(self) -> GlobalOptions:
             allow_empty=allow_empty,
         )
 
-    # pylint: disable-next=method-hidden
-    def build_options(self, identifier: str | None) -> BuildOptions:
+    def _compute_build_options(self, identifier: str | None) -> BuildOptions:
         """
         Compute BuildOptions for a single run configuration.
         """
@@ -826,6 +826,14 @@ def build_options(self, identifier: str | None) -> BuildOptions:
                 container_engine=container_engine,
             )
 
+    def build_options(self, identifier: str | None) -> BuildOptions:
+        """
+        Get BuildOptions for a single run configuration.
+        """
+        if identifier not in self._build_options:
+            self._build_options[identifier] = self._compute_build_options(identifier)
+        return self._build_options[identifier]
+
     def check_for_invalid_configuration(self, identifiers: Iterable[str]) -> None:
         if self.platform in {"macos", "windows"}:
             before_all_values = {self.build_options(i).before_all for i in identifiers}
```

joerick: Yeah, that works too. If the ugliness[^1] issue is linter complains about the method assignment, we could try...

```diff
@@ -671,8 +672,7 @@ def globals(self) -> GlobalOptions:
             allow_empty=allow_empty,
         )
 
-    def build_options(self, identifier: str | None) -> BuildOptions:
+    def _compute_build_options(self, identifier: str | None) -> BuildOptions:
         """
         Compute BuildOptions for a single run configuration.
         """

@@ -577,6 +577,11 @@ def __init__(
         self.env = env
         self._defaults = defaults
 
+        self.build_options = functools.cache(self._compute_build_options)
+
         self.reader = OptionsReader(
             None if defaults else self.config_file_path,
             platform=platform,
```

[^1]: Ugliness is in the eye of the beholder! I don't find the original pattern too strange, but I'm quite used to doing similar things in Typescript, where this pattern is quite normal `this.doThing = this.doThing.bind(this)`, although, admittedly, this is line working around a JS language deficiency

henryiii: We could try that, not replacing the original method already makes it look nicer. I think mypy would still keep the signature. If not, the manual method is fine too.

cibuildwheel/selector.py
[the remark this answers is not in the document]
henryiii: Okay, not a big deal.

cibuildwheel/options.py
joerick: Mypy and pylance both seem happy enough with it, so that works for me.

mayeut (approved):

PR #2304, opened 2025-03-04
fix(docs/faq): update rust toolchain intructions

Ravencentric:
Fixes https://github.com/pypa/cibuildwheel/issues/2303

This change did fix my CI, as you can see over here: https://github.com/Ravencentric/rnzb/pull/10/files.

I basically grabbed the PowerShell commands from the blog post: https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html#whats-new-in-rustup-1280, and then added equivalent commands for Linux and macOS. I've also added a `pyproject.toml` for the same. I'm not really sure if this is the best way to handle it, so any feedback or suggestions on whether this is the right approach would be really welcome!

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -288,11 +288,34 @@ If you build Rust wheels, you need to download the Rust compilers in manylinux.
 If you support 32-bit Windows, you need to add this as a potential target. You
 can do this on GitHub Actions, for example, with:
 
-```yaml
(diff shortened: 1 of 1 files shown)


Ravencentric:
Rustup 1.28.1 did fix this, so this PR is no longer needed.

PR #2312, opened 2025-03-10
fix: warn on deprecated images being set

henryiii:
See #2047.

### DIFF
--- a/cibuildwheel/logger.py
+++ b/cibuildwheel/logger.py
@@ -1,4 +1,5 @@
 import codecs
+import dataclasses
 import os
 import re
 import sys
@@ -64,16 +65,18 @@ class Symbols:
         self.error = "✕" if unicode else "failed"
 
 
+@dataclasses.dataclass
 class Logger:
-    fold_mode: str
-    colors_enabled: bool
-    unicode_enabled: bool
+    fold_mode: str = "disabled"
+    colors_enabled: bool = False
+    unicode_enabled: bool = False
     active_build_identifier: str | None = None
     build_start_time: float | None = None
     step_start_time: float | None = None
     active_fold_group_name: str | None = None
+    dedupiclate: set[str] = dataclasses.field(default_factory=set)
 
-    def __init__(self) -> None:
+    def __post_init__(self) -> None:
         if sys.platform == "win32" and hasattr(sys.stdout, "reconfigure"):
             # the encoding on Windows can be a 1-byte charmap, but all CIs
             # support utf8, so we hardcode that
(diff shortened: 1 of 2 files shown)


cibuildwheel/options.py
```diff
@@ -775,6 +775,14 @@ def _compute_build_options(self, identifier: str | None) -> BuildOptions:
                         # default to manylinux2014
                         image = pinned_images["manylinux2014"]
                     elif config_value in pinned_images:
+                        if config_value in {
+                            "manylinux1",
+                            "manylinux2010",
+                            "manylinux_2_24",
+                            "musllinux_1_1",
+                        }:
+                            msg = f"Unmaintained image, either upgrade to a supported image or pin directly to {pinned_images[config_value]!r}"
```
joerick, line 784: Adding a bit more detail...

```suggestion
                            msg = f"Deprecated image {config_value!r}. This value will not work in a future version of cibuildwheel. Either upgrade to a supported image or continue using the deprecated image by pinning directly to {pinned_images[config_value]!r}."
```

joerick (commented):
I think we'd want to check that the warning doesn't fire loads of times? I think that `build_options()` is called rather a lot... at least, once per identifier. Perhaps we'd need an instance variable tracking which deprecated images have already been seen.

cibuildwheel/logger.py
```diff
@@ -164,7 +165,11 @@ def notice(self, message: str) -> None:
             c = self.colors
             print(f"cibuildwheel: {c.bold}note{c.end}: {message}\n", file=sys.stderr)
 
-    def warning(self, message: str) -> None:
+    def warning(self, message: str, *, deduplicate: bool = False) -> None:
```
henryiii, line 168: Maybe we could just always do this for warnings?

joerick: Think I'd prefer to keep it explicit, it's a bit surprising at the call site otherwise. Might also make some unit tests buggy as this is shared global state.

joerick: ^ actually the global state thing might be a reason to do it on Options rather than here.

joerick: > I guess if it was an instance variable it still wouldn't over trigger?

I don't think so. There's only one instance of `Options` in a normal running process. The only reason I hesitate to do it globally is that I can imagine a unit test expecting to see a warning but depending on test execution order, it sometimes gets triggered somewhere else and the test becomes flaky/hard to debug.

That said it's a hypothetical issue at this point, we can leave it here if you wish.

joerick (approved):
Looks good. Thanks for putting it together. I'll roll this into a 2.x patch release.

PR #2313, opened 2025-03-11
chore: add a selection of PERF rules

DimitriPapadopoulos:


henryiii:
I think we can just make expected_wheels a generator, which would simply a bit. I can try that.

pyproject.toml
```diff
@@ -186,6 +186,7 @@ extend-select = [
   "ICN",         # flake8-import-conventions
   "ISC",         # flake8-implicit-str-concat
   "G",           # flake8-logging-format
+  "PERF",        # Perflint
```
joerick, line 189: The changes in this PR are nice, but conceptually, I'm pretty dubious about the idea of static performance linting. Performance problems are context-dependent. Priority one is readability and maintainability. Performance changes should be motivated by profiling data, not rules.

```suggestion
```

henryiii: [Perflint](https://github.com/tonybaloney/perflint) is a tool from [Anthony Shaw](https://github.com/tonybaloney), and looks for common performance anti patterns. I think it's pretty good and mostly focus on detectable things, like calling `for item in list(a)` where `a` is already iterable. But cibuildwheel will never be constrained by Python language performance, so I'd fine with not adding it to our checks. I'm also fine with adding it then dropping it if it ever becomes a problem by adding a rule that we don't want to follow. Some of the fancier rules that aren't in Ruff yet (I think) might not be ideal, like dotted-import-in-loop.

DimitriPapadopoulos: I agree, improving global performance should rely on profiling. The idea is not to spend precious developer time on non-issues.

However, these are “free” micro-optimisations. While in most cases they might not have measurable impact on performance, they are applied automatically, in a systematic way, without wasting developer time. I think the result is mostly consistent code — and sometimes (seldom?) measurable performance improvements.

joerick: Just based on skimming the [perflint readme](https://github.com/tonybaloney/perflint), (which does include the line "This project is an early beta. It will likely raise many false-positives in your code."), and looking at some of the loop-related rules listed there, it doesn't look like a good match for this project IMO. We're very unlikely to have performance issues related to Python language idioms - 99% of time in cibuildwheel is spent shelling out to child processes.

joerick (approved):
looks good to me!

PR #2314, opened 2025-03-12
fix: image deprecation warning

mayeut:
The `musllinux_1_1 ` warning was not triggered, rework a bit the check and add some unit tests.


joerick (approved):
Ah, the musl images are checked in a different location. Well done for noticing this!

PR #2316, opened 2025-03-13
Drop EOL manylinux images

mayeut:
Towards https://github.com/pypa/cibuildwheel/issues/2047#issuecomment-2711913040

EOL images have very little usage.
Drop support for those images, they can still be used by directly pinning their full registry name.

This mostly has an impact on documentation & tests.

The mechanism to check deprecation has been kept which leads to the check being a bit redundant regarding image names.


cibuildwheel/options.py
```diff
@@ -673,10 +673,17 @@ def globals(self) -> GlobalOptions:
...
+        warning_set = error_set  # shall include error_set
+
+        if value in warning_set and value not in self._image_warnings:
+            if value in error_set:
+                msg = (
+                    f"cibuildwheel 3.x does not support the image {value!r}. Either upgrade to a "
+                    "supported image or continue using the image by pinning it directly with"
+                    " its full OCI registry '<name>{:<tag>|@<digest>}'."
+                )
+                raise errors.DeprecationError(msg)
```
joerick, line 686: We can simplify a bit, I think, decouple errors and warnings...

```suggestion
        error_set = {"manylinux1", "manylinux2010", "manylinux_2_24", "musllinux_1_1"}
        warning_set = set()

        if value in error_set:
            msg = (
                f"cibuildwheel 3.x does not support the image {value!r}. Either upgrade to a "
                "supported image or continue using the image by pinning it directly with"
                " its full OCI registry '<name>{:<tag>|@<digest>}'."
            )
            raise errors.DeprecationError(msg)

        if value in warning_set and value not in self._image_warnings:
```

PR #2321, opened 2025-03-17
feat: use build by default

henryiii:
This is another thing we planned for 3.0, I think.


joerick (approved):
A satifyingly clean PR for this change! I had expected more test finagling to be required :)

PR #2325, opened 2025-03-21
examples/github-deploy.yml: add `print-hash: true` to log hashes to log

westurner:
- `print-hash: true` shows hashes of files to be uploaded:
  https://github.com/pypa/gh-action-pypi-publish#showing-hash-values-of-files-to-be-uploaded
- it is a good idea to log artifact hashes in the build logs
- in the docs as well:
  - [ ] https://github.com/pypa/cibuildwheel/blob/1826f7068262b8cb6b59fe9dd1b408a825d63fa6/docs/deliver-to-pypi.md?plain=1#L78-L84

### DIFF
--- a/examples/github-deploy.yml
+++ b/examples/github-deploy.yml
@@ -89,6 +89,7 @@ jobs:
           pattern: cibw-*
           path: dist
           merge-multiple: true
+          print-hash: true
 
       - uses: pypa/gh-action-pypi-publish@release/v1
         with:


joerick:
I'm -0 on such a change, it feels like something you could add if it's important to you, but most users wouldn't care to compare hashes anyway, and less config is better in my book.

henryiii:
Same here. Simpler is better. The PyPI action signs artifacts with sigstore already.

henryiii:
pypa/gh-action-pypi-publish prints out lots of info, including hashes. Logs get deleted after a time, though. If this actually matters to you, you should probably upload attestations to GitHub too, which looks like this:


```yaml
  permissions:
    attestations: write
  ...

    - name: Generate artifact attestation for sdist and wheel
      uses: actions/attest-build-provenance@v2.2.3
      with:
        subject-path: "dist/*"
```

You can see them, for example, at `<repo>/attestations`, for example, ours are here: https://github.com/pypa/cibuildwheel/attestations. But I don't think that's necessarily good for an example of how to use cibuildwheel, it's distracting from what we are trying to teach. It is mentioned in [the Scientific Python Development Guide](https://learn.scientific-python.org/development/), which covers more than just cibuildwheel.

westurner:
- https://learn.scientific-python.org/development/guides/gha-wheels/#publishing
- https://www.pypa.io/en/latest/search/?q=attestation
  - https://packaging.python.org/en/latest/search/?q=attestation
  - https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
  - https://cibuildwheel.pypa.io/en/stable/deliver-to-pypi/#github-actions

PR #2328, opened 2025-03-21
fix: always resolve --python argument

henryiii:
This will fix https://github.com/pypa/cibuildwheel/issues/2327. `uvx git+https://github.com/henryiii/cibuildwheel@henryiii/fix/uvx --only cp312-pyodide_wasm32` does seem to pass with this.

Virtualenv can break if the `--python` argument is a symlink.

### DIFF
--- a/cibuildwheel/venv.py
+++ b/cibuildwheel/venv.py
@@ -96,6 +96,10 @@ def virtualenv(
     dependency_constraint_flags are ignored since nothing is installed in the
     venv. Otherwise, pip is installed, and setuptools + wheel if Python < 3.12.
     """
+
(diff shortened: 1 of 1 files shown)


joerick:
I also hit the bug when using a `uv` installed version of python, which was somehow on my PATH, (maybe from `uv python install`, but I don't remember doing that). In my case, `UV` wouldn't have been set.

henryiii:
Looking into if we can fix the virtualenv creation.

henryiii:
PS: Also tagging @vineetbansal, who I originally found the issue with.

mayeut (approved):

PR #2331, opened 2025-03-22
[2.x] fix: always resolve --python argument (#2328)

henryiii:
Backport:

* fix: always resolve the --python argument (#2328)


mayeut (approved):

joerick:
Thanks! Would you like a release with this?

henryiii:
Not critical, but would be nice.

Issue #2335, opened 2025-03-25
Instability in iOS tests under CI conditions


freakboy3742:
Reported by @henryiii in [this comment](https://github.com/pypa/cibuildwheel/pull/2286#issuecomment-2744509095).

It appears that iOS builds are seeing unusually high failure rates on CI jobs - possibly as high as 50% of CI jobs. It's hard to get precise metrics out of the tools Github provides, but there's no shortage of failures on the [Test workflow](https://github.com/pypa/cibuildwheel/actions/workflows/test.yml) for macOS runs - usually macOS-14, but sometimes macOS-13 (x86_64).

Digging into the logs, the failures all seem to be timeouts:

* Hitting the 5 minute timeout for running a build on the simulator
* Hitting a 10 second timeout waiting for the return value of the simulator list tool.

I'm not seeing this failure rate locally - but local conditions will always be more favourable for test timing, as they can involve partially warmed simulator images, and won't usually be competing with other build tasks on the same physical hardware.

freakboy3742:
Anecdotally, there also seems to be highly variable performance - especially I/O performance - on GitHub Actions macOS runners. [Toga](https://github.com/beeware/toga) doesn't use cibuildwheel to run tests, but it does start an iOS simulator as part of tests. Sometimes the simulator can start almost immediately; and sometimes it can take a couple of minutes. My current hypothesis is that the 5 minute timeout is a little too aggressive for the "worst case" CI performance. 

Unfortunately, there's no real way to replicate this other than "start a CI jobs"... and the fix will likely require an updated support package. So - apologies in advance for the thrashing I'm about to give CI. I'll try to minimise the impact to just macOS jobs.

freakboy3742:
This also seems to be a problem that is specific to Github CI. As far as I can make out from CI logs, CircleCI and Azure Pipeline tests are passing consistently.

joerick:
I have never run the iOS tests locally, so I tried running them just now.

The first run took a pretty long time, pytest tells me

> 3 passed in 507.62s (0:08:27)

Activity monitor showed lots of CPU usage in that time from various processes, including `/Library/Developer/CoreSimulator/Volumes/iOS_21C62/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS\ 17.2.simruntime/Contents/Resources/update_dyld_sim_shared_cache`. I also got a crash report for MobileCal in that time, though it didn't appear to affect anything.

The second run of pytest took:

[...]

PR #2336, opened 2025-03-25
[DO NOT MERGE] Evaluate iOS CI reliability issues.

freakboy3742:
This is an exploratory PR to check behavior of the iOS test runner under CI conditions.

Refs #2335


freakboy3742:
With 7 successful passes, I'm going to call this experiment a success; see #2337 for the final form of the fix.

PR #2339, opened 2025-03-28
fix: support wider range of verbosity settings on other build backends

henryiii:
See https://github.com/pypa/cibuildwheel/issues/2338.

This makes the verbosity setting a bit more useful on other backends. This tries to map the current `pip` meanings to other backends. The key changes are:

* `<0` will support `uv` (#2322), `build` continues to produce a warning. This is a feature that's been requested before for build, btw.
* `1` continues to do nothing on non-`pip` backends
* 2+ now passes multiple `-v` to non-pip backends, since both build and uv support adding extra resolving info (similar to pip's `-vv`) if passing `-v`.

### DIFF
--- a/cibuildwheel/frontend.py
+++ b/cibuildwheel/frontend.py
@@ -45,8 +45,6 @@ def _get_verbosity_flags(level: int, frontend: BuildFrontendName) -> list[str]:
     elif level > 1:
         return ["-v"]
     elif level < 0:
-        if frontend not in {"build", "build[uv]"}:
-            return ["-q"]
         msg = f"build_verbosity {level} is not supported for {frontend} frontend. Ignoring."
         log.warning(msg)
     return []
--- a/cibuildwheel/frontend.py
+++ b/cibuildwheel/frontend.py
@@ -42,7 +42,11 @@ def _get_verbosity_flags(level: int, frontend: BuildFrontendName) -> list[str]:
             return ["-" + level * "v"]
         if level < 0:
             return ["-" + -level * "q"]
-    elif not 0 <= level < 2:
+    elif level > 1:
+        return ["-v"]
+    elif level < 0:
+        if frontend not in {"build", "build[uv]"}:
+            return ["-q"]
         msg = f"build_verbosity {level} is not supported for {frontend} frontend. Ignoring."
         log.warning(msg)
     return []
--- a/cibuildwheel/frontend.py
+++ b/cibuildwheel/frontend.py
@@ -37,16 +37,19 @@ class BuildFrontendConfig:
 
 
 def _get_verbosity_flags(level: int, frontend: BuildFrontendName) -> list[str]:
-    if frontend == "pip":
-        if level > 0:
-            return ["-" + level * "v"]
-        if level < 0:
+    if level < 0:
+        if frontend == "pip":
             return ["-" + -level * "q"]
-    elif level > 1:
-        return ["-v"]
-    elif level < 0:
+
         msg = f"build_verbosity {level} is not supported for {frontend} frontend. Ignoring."
         log.warning(msg)
+
+    if level > 0:
+        if frontend == "pip":
+            return ["-" + level * "v"]
+        if level > 1:
+            return ["-" + (level - 1) * "v"]
+
     return []
 
 
--- a/docs/options.md
+++ b/docs/options.md
@@ -1762,14 +1762,25 @@ export CIBW_DEBUG_TRACEBACK=TRUE
 ### `CIBW_BUILD_VERBOSITY` {: #build-verbosity}
 > Increase/decrease the output of the build
 
-The setting vary a bit between build frontends.
+This setting controls `-v`/`-q` flags to the build frontend. Since there is
+no communication between the build backend and the build frontend, build
+messages from the build backend will always be shown with `1`; higher levels
+will not produce more logging about the build itself. Other levels only affect
+the build frontend output, which is usually things like resolving and
+downloading dependencies. The settings are:
 
-* `-1`: Hide as much build output as possible; passes `-q` to the build frontend. Not supported by `build`/`build[uv]`.
-* `0`: The default. On pip, this hides the build output if the build succeeds, other build frontends produce output from the build backend.
-* `1`: Produces build backend output. On `pip`, this passes `-v`. Other frontends do this by default.
-* `2`: Produces extra output from resolving packages too. On `pip`, this passes `-vv`, other build frontends use `-v`.
-* `3`: Even more resolving output from pip with `-vvv`, other build frontends continue to just pass `-v`.
+|             | build | pip    | desc                             |
+|-------------|-------|--------|----------------------------------|
+| -2          | N/A   | `-qq`  | even more quiet, where supported |
+| -1          | N/A   | `-q`   | quiet mode, where supported      |
+| 0 (default) |       |        | default for build tool           |
+| 1           |       | `-v`   | print backend output             |
+| 2           | `-v`  | `-vv`  | print log messages e.g. resolving info |
+| 3           | `-vv` | `-vvv` | print even more debug info       |
 
+Settings that are not supported for a specific frontend will log a warning.
+The default build frontend is `build`, which does show build backend output by
+default.
 
 Platform-specific environment variables are also available:<br/>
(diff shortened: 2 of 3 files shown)


docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
 ```
 
 ### `CIBW_BUILD_VERBOSITY` {: #build-verbosity}
-> Increase/decrease the output of pip wheel
+> Increase/decrease the output of the build
+
+The setting vary a bit between build backends.
```
mayeut, line 1765:
```suggestion
The setting vary a bit between build frontends.
```

docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
...
+
+* `-1`: Hide as much build output as possible; passes `-q` to the build backend. Not supported by `build`/`build[uv]`.
```
mayeut, line 1767:
```suggestion
* `-1`: Hide as much build output as possible; passes `-q` to the build frontend. Not supported by `build`/`build[uv]`.
```

docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
...
+* `0`: The default. On pip, this hides the build output if the build succeeds, other build backends produce output from the build backend.
```
mayeut, line 1768:
```suggestion
* `0`: The default. On pip, this hides the build output if the build succeeds, other build frontends produce output from the build backend.
```

docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
...
+* `1`: Produces build backend output. On `pip`, this passes `-v`. Other backends do this by default.
```
mayeut, line 1769:
```suggestion
* `1`: Produces build backend output. On `pip`, this passes `-v`. Other frontends do this by default.
```

docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
...
+* `2`: Produces extra output from resolving packages too. On `pip`, this passes `-vv`, other build backends use `-v`.
```
mayeut, line 1770:
```suggestion
* `2`: Produces extra output from resolving packages too. On `pip`, this passes `-vv`, other build frontends use `-v`.
```

cibuildwheel/frontend.py
```diff
@@ -42,7 +42,11 @@ def _get_verbosity_flags(level: int, frontend: BuildFrontendName) -> list[str]:
             return ["-" + level * "v"]
         if level < 0:
             return ["-" + -level * "q"]
-    elif not 0 <= level < 2:
+    elif level > 1:
+        return ["-v"]
+    elif level < 0:
+        if frontend not in {"build", "build[uv]"}:
+            return ["-q"]
```
mayeut, line 49:
```suggestion
```

mayeut:
Should we be trying to get something consistent between frontends ? i.e. 0 maps to `-v` with `pip`

joerick:
> Should we be trying to get something consistent between frontends ? i.e. 0 maps to `-v` with `pip`

If we want to keep this option, I like this idea. Previously I was in favour of a light-touch approach to adding arguments to `pip wheel`. But printing the build backend output by default makes a lot of sense in a CI scenario, and with `build` the default now, we're just making the option consistent between frontends.

so that would be 

|  | build | pip | uv | desc |
|---|---|---|---|---|
| -2 | not supported | `-q` | not supported | quiet mode |
| -1 | not supported | ` ` | `-q` | reduce output, where supported |
| 0 (default) | ` ` | `-v` | ` ` | print backend output (default) |
| 1 | `-v` | `-vv` | `-v` | print log messages e.g. resolving info |
| 2 | `-vv` | `-vvv` | `-vv` | print even more debug info |

I think it would be nice to warn on unsupported settings i.e. (not supported, above) or outside `-2...2`.

The other approach, of course, is to avoid this mess and just refer users to `build-frontend` with the `args:` param. I'd be open to either, I think.

henryiii:
That is very similar to what I had, but with one difference: 0 is "native", which doesn't add any flags, keeping to the native behavior of the tool, and allows a user to use the `args:` if they prefer. If you set `verbosity=1`, that tells all backends to produce verbose output consistently.

|  | build | pip | uv | desc |
|---|---|---|---|---|
| -2 | not supported | `-q` | not supported | quiet mode |
| -1 | not supported | ` ` | `-q` | reduce output, where supported |
| 0 (default) | ` ` | ` ` | ` ` | default for build tool |
| 1 | ` ` | `-v` | ` ` | print backend output |
| 1 | `-v` | `-vv` | `-v` | print log messages e.g. resolving info |
| 2 | `-vv` | `-vvv` | `-vv` | print even more debug info |

I prefer the default not adding flags (especially since we have a mechanism for users to add flags themselves). I expect most users won't change the default (which now produces build output because it uses `build` by default), or they'll want build output so they'll set `verbose=1`, which will produce build output on all backends.

Since this is such an important feature, I like having the flag for it (probably not enough to have added it if we didn't already have it, but certainly happy to keep it). I have a `verbose` setting for scikit-build-core even though it's possible to manually pass through the verbosity flag go CMake, and it's been popular.

cibuildwheel/frontend.py
henryiii: This was there to support #2322, but I guess I can re-add it.

burgholzer:
Just fyi:
Since 0.6.11, uv also supports `-qq`. See https://github.com/astral-sh/uv/pull/12300

joerick:
Yeah, seeing it in a table, I agree with you @henryiii. It's better to have a default that matches the build-frontend's default.

> Plus this is backwards compatible, the values keep their meanings. Otherwise, `1` becomes too verbose, and a lot of projects have `1` set.

Also a good point.

docs/options.md
```diff
@@ -1760,9 +1760,16 @@ export CIBW_DEBUG_TRACEBACK=TRUE
...
-> Increase/decrease the output of pip wheel
+> Increase/decrease the output of the build
+
+The setting vary a bit between build frontends.
+
+* `-1`: Hide as much build output as possible; passes `-q` to the build frontend. Not supported by `build`/`build[uv]`.
+* `0`: The default. On pip, this hides the build output if the build succeeds, other build frontends produce output from the build backend.
+* `1`: Produces build backend output. On `pip`, this passes `-v`. Other frontends do this by default.
+* `2`: Produces extra output from resolving packages too. On `pip`, this passes `-vv`, other build frontends use `-v`.
+* `3`: Even more resolving output from pip with `-vvv`, other build frontends continue to just pass `-v`.
```
joerick, line 1771: I'd consider replacing this with the table from your comment - I think the table does a better job of conveying the meaning, especially with uv soon to be added to the mix.

```suggestion
|             | Description                            | build | pip   | uv   | 
|-------------|----------------------------------------|-------|-------|------|
| -2          | quiet mode                             | N/A   | `-q`  | N/A  | 
| -1          | reduce output, where supported         | N/A   | ` `   | `-q` |
| 0 (default) | default for build tool                 | ` `   | ` `   | ` `  |
| 1           | print backend output                   | ` `   | `-v`  | ` `  |
| 2           | print log messages e.g. resolving info | `-v`  | `-vv` | `-v` |
| 3           | print even more debug info             | `-vv` | `-vvv` | `-vv` |

```

cibuildwheel/frontend.py
```diff
@@ -42,7 +42,9 @@ def _get_verbosity_flags(level: int, frontend: BuildFrontendName) -> list[str]:
             return ["-" + level * "v"]
         if level < 0:
             return ["-" + -level * "q"]
-    elif not 0 <= level < 2:
+    elif level > 1:
+        return ["-v"]
+    elif level < 0:
         msg = f"build_verbosity {level} is not supported for {frontend} frontend. Ignoring."
         log.warning(msg)
     return []
```
joerick, line 50: I think there are a few changes required here to match the table in readme. e.g. -1 on pip doesn't do anything. It might be worth changing this to a map, with only a few supported settings per frontend, warning otherwise; up to you.

henryiii:
Okay, new version. Added the table, and modified it a tiny bit after observing `pip wheel . -q` and `uv build -q` look identical, with no output normally. So that's now -1 in the table.

henryiii:
Also, `-vv` works on build, even though I don't think it does anything different, I'm only warning and skipping if it is explicitly not supported.

joerick (approved):
Looks good to me!

mayeut (approved):

PR #2347, opened 2025-04-03
refactor: pull out dependency_constraint

henryiii:
Pulled out from #2322 to simplify that one. Passes the actual Path or None instead of passing the flags and then parsing the argument back out (!). Also switches to using the URI, but I could do that in 2322 instead if there are any issues (edit: seems fine!).

### DIFF
--- a/cibuildwheel/platforms/macos.py
+++ b/cibuildwheel/platforms/macos.py
@@ -7,7 +7,7 @@ import shutil
 import subprocess
 import sys
 import typing
-from collections.abc import Sequence, Set
+from collections.abc import Set
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Literal, assert_never
@@ -23,7 +23,6 @@ from ..frontend import BuildFrontendConfig, BuildFrontendName, get_build_fronten
 from ..logger import log
 from ..options import Options
 from ..selector import BuildSelector
-from ..typing import PathOrStr
 from ..util import resources
 from ..util.cmd import call, shell
 from ..util.file import (
@@ -34,7 +33,7 @@ from ..util.file import (
 )
 from ..util.helpers import prepare_command, unwrap
 from ..util.packaging import combine_constraints, find_compatible_wheel, get_pip_version
-from ..venv import find_uv, virtualenv
+from ..venv import constraint_flags, find_uv, virtualenv
 
 
 @functools.cache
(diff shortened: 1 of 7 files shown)


cibuildwheel/platforms/windows.py
```diff
@@ -248,7 +247,7 @@ def setup_python(
     if build_frontend == "build[uv]" and not can_use_uv(python_configuration):
         build_frontend = "build"
 
-    use_uv = build_frontend == "build[uv]"
+    use_uv = build_frontend in {"build[uv]", "uv"}
```
mayeut, line 250:
```suggestion
    use_uv = build_frontend == "build[uv]"
```

"uv" is not there yet, I would have expected an error from the linter saying something like "condition always false for `uv`"

cibuildwheel/platforms/windows.py
```diff
@@ -370,15 +371,15 @@ def build(options: Options, tmp_path: Path) -> None:
                 version=config.version,
                 tmp_dir=identifier_tmp_dir,
             )
-            dependency_constraint_flags: Sequence[PathOrStr] = (
-                ["-c", constraints_path] if constraints_path else []
+            dependency_constraint_flags = (
+                ["-c", constraints_path.as_uri()] if constraints_path else []
             )
```
mayeut, line 376:
```suggestion
            dependency_constraint_flags = constraint_flags(constraints_path)
```
but the same remark as macOS applies here so might not be worth merging the suggestion as-is

cibuildwheel/platforms/pyodide.py
```diff
@@ -269,8 +269,8 @@ def build(options: Options, tmp_path: Path) -> None:
             constraints_path = build_options.dependency_constraints.get_for_python_version(
                 version=config.version, variant="pyodide", tmp_dir=identifier_tmp_dir
             )
-            dependency_constraint_flags: Sequence[PathOrStr] = (
-                ["-c", constraints_path] if constraints_path else []
+            dependency_constraint_flags = (
+                ["-c", constraints_path.as_uri()] if constraints_path else []
             )
 
             env = setup_python(
```
mayeut, line 276: setup_python should likely be updated to take constraints_path as is done on other platforms.

cibuildwheel/platforms/macos.py
```diff
@@ -427,14 +428,12 @@ def build(options: Options, tmp_path: Path) -> None:
             constraints_path = build_options.dependency_constraints.get_for_python_version(
                 version=config.version, tmp_dir=identifier_tmp_dir
             )
-            dependency_constraint_flags: Sequence[PathOrStr] = (
-                ["-c", constraints_path] if constraints_path else []
-            )
+            dependency_constraint_flags = constraint_flags(constraints_path)
```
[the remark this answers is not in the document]
joerick: +1 - moving the flags generation to where they are used makes sense to me.

mayeut (approved):

PR #2357, opened 2025-04-13
Test enable groups as specified by PR labels

joerick:
Don't run all enable groups on PRs, but allow them to be enabled using PR labels.

This is mostly to keep the PR builds fast. As noted in https://github.com/pypa/cibuildwheel/pull/1538#issuecomment-2796457742 , extra enable groups affect the CI times a lot.

The only CI affected by the labels is Github Actions. The others just run the default set, which probably isn't ideal, but I can't think of another way to do it!


henryiii:
When I've added this sort of thing in the past, it wouldn't trigger the expanded tests on the first push (since there's no label there), and then you have to add the label then push something to get the expanded testing.

joerick:
I did set it to trigger on 'labelled', so hopefully that shouldn't be a problem? Once I get these tests to pass we can play with it.

joerick:
Cool, I added the PyPy tag and the tests reran. I'll update this table when it finishes:

| Test  durations                                                 | no labels |
| :-------------------------------------------------------------- | :-------- |
| Get qemu emulated architectures (pull_request)                  | 8s        |
| Linters (mypy, flake8, etc.) (pull_request)                     | 1m        |
| Test cibuildwheel building Pyodide wheels (pull_request)        | 5m        |
| Test Linux aarch64 using qemu (pull_request)                    | 17m       |
| Test Linux armv7l using qemu (pull_request)                     | 16m       |
| Test Linux ppc64le using qemu (pull_request)                    | 15m       |
| Test Linux s390x using qemu (pull_request)                      | 18m       |
| Test on macos-13 (3.13) (pull_request)                          | 23m       |
| Test on macos-15 (3.13) (pull_request)                          | 15m       |
| Test on ubuntu-24.04-arm (3.13) (pull_request)                  | 21m       |
| Test on ubuntu-latest (3.11) (pull_request)                     | 31m       |
| Test on ubuntu-latest (3.13) (pull_request)                     | 30m       |
| Test on windows-11-arm (3.13) (pull_request)                    | 15m       |
| Test on windows-latest (3.13) (pull_request)                    | 30m       |

joerick:
Looks like it's working based on the above table!

The remaining question is what to do with the other CIs. Perhaps the right approach is to run everything on `main`, but just the core set in PRs. That way we'll catch those incompatibilities eventually.

joerick:
That's done. This is ready to review.

henryiii:
Should we merge #1538 first, then add graalpy here?

Also, I'd like to remove the deprecated PyPy's or put then behind a separate enable group. The original reason to add an enable group was to avoid depreciated PyPy's, but since it enables both current and deprecated PyPy's, it's not helpful for the original use case.

joerick:
> Should we merge #1538 first, then add graalpy here?

Yeah, that would make sense to me.

> Also, I'd like to remove the deprecated PyPy's or put then behind a separate enable group.

Sounds good to me!

henryiii (approved):

Issue #2361, opened 2025-04-14
Windows ARM Runners support


johnnynunez:
### Description

windows-11-arm

https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

henryiii:
Does anyone know how to setup Rust for Windows ARM? It seems rustup doesn't support Windows ARM with binaries? There are no Windows ARM wheels for cryptography.

johnnynunez:
> Does anyone know how to setup Rust for Windows ARM? It seems rustup doesn't support Windows ARM with binaries? There are no Windows ARM wheels for cryptography.

https://doc.rust-lang.org/nightly/rustc/platform-support.html

henryiii:
Okay, I think the actions need to explicitly support it (https://github.com/actions-rust-lang/setup-rust-toolchain, https://github.com/moonrepo/setup-rust@v1). It hurts that rustup seems to be already present on the other runners, so this one likely requires a different code path unless GitHub adds rustup to the default image. We might see a method come out of https://github.com/pyca/cryptography/issues/11199. Anyway, we don't need cryptography in the tests (it was a mistake that we were syncing the dev dependencies), so it looks like the only thing needed was some minor updates to the test suite.

henryiii:
Adding rustup to the images is tracked at https://github.com/actions/partner-runner-images/issues/77 - that might make all the setup actions just start working (and you could avoid them with that).

johnnynunez:
> Adding rustup to the images is tracked at [actions/partner-runner-images#77](https://github.com/actions/partner-runner-images/issues/77) - that might make all the setup actions just start working (and you could avoid them with that).

yeah for sure, we have to wait

PR #2362, opened 2025-04-14
ci: add Windows ARM

henryiii:
See #2361.


mayeut:
should `uv run` use the `--no-sync` or another argument in order not to try and install dev dependency group ? or should we update some `UV_...` environment variables in the `install dependencies` step ?
All needed test dependencies should already be installed and cryptography is only a dependency for something in the dev group not needed for tests.

PR #2365, opened 2025-04-18
chore: type check install_certifi.py

hauntsaninja:


joerick (approved):
This file is vendored from cpython, but sure, why not.

henryiii:
Same thought here; that's why it wasn't typed originally. But this is a pretty trivial addition to get typing, and we aren't re-syncing this regularly or anything.

Issue #2263, opened 2025-02-06
riscv64 support


eshattow:
Request: Add riscv64 support

Ref: https://github.com/aio-libs/aiohttp/pull/10330 reverted because cibuildwheel is missing support

henryiii:
I think it has to be supported here first: https://github.com/pypa/manylinux. Is riscv64 a valid wheel tag accepted by PyPI? I don't see it listed in any PEPs. I think it has to be standardized and support added to PyPI.

mayeut:
> Is riscv64 a valid wheel tag accepted by PyPI?

It's not

> I think it has to be standardized and support added to PyPI.

It does

There's a [discussion](https://discuss.python.org/t/packaging-support-for-riscv64/58475) about riscv64 on discuss.python.org

eshattow:
More recently a talk presented on the topic: https://fosdem.org/2025/schedule/event/fosdem-2025-5659-towards-seamless-python-package-installation-on-riscv64/

mayeut:
This is now fully fixed in main.
3.1.1 still requires opt-in but is supported

CaselIT:
is seems that's not documented in the [readme](https://github.com/pypa/cibuildwheel?tab=readme-ov-file#what-does-it-do)

would be nice to add it also there

PR #2366, opened 2025-04-19
feat: initial support for riscv64

mayeut:
Towards #2263

While riscv64 would need a PEP in order to publish to PyPI, the community is already using a patch version of cibuildwheel to populate a repository hosted on GitLab: https://gitlab.com/riseproject/python/wheel_builder.

In order to get things moving forward, initial support could be added to cibuildwheel while waiting for full support once a PEP is available.

IMHO, the fact that those wheels can't be uploaded to PyPI isn't a showstopper here (in fact, the same is true of pyodide wheels). The only thing I'm concerned about with this PR is that building "all" platforms on Linux would now include something not upload able to PyPI so maybe it should be protected by an enable flag ?


joerick:
I expect only a few people have set `CIBW_ARCHS_LINUX: all` but [I did find a few](https://sourcegraph.com/search?q=context:global+CIBW_ARCHS_LINUX.*all&patternType=regexp&sm=0).

However, making this an enable group is a good idea IMO, since it's kind of experimental... In theory, something about these wheels might change before the PEP is approved. That way it matches the logic of the prerelease enable group. `cpython-experimental-riscv` would work for me as a name.

cibuildwheel/logger.py
```diff
@@ -22,12 +22,14 @@
...
     "manylinux_s390x": "manylinux s390x",
     "manylinux_armv7l": "manylinux armv7l",
+    "manylinux_riscv64": "manylinux riscv64",
     "musllinux_x86_64": "musllinux x86_64",
     "musllinux_i686": "musllinux i686",
     "musllinux_aarch64": "musllinux aarch64",
     "musllinux_ppc64le": "musllinux ppc64le",
     "musllinux_s390x": "musllinux s390x",
     "musllinux_armv7l": "musllinux armv7l",
+    "musllinux_riscv64": "manylinux riscv64",
```
joerick, line 32:
```suggestion
    "musllinux_riscv64": "musllinux riscv64",
```

joerick (commented):
Looks good! I think we'd want to document the option in the docs for CIBW_ENABLE, in options.md.

henryiii (approved):

PR #2372, opened 2025-04-22
Addition to allow Bitbucket Pipelines to work with cibuildwheel

t-s:


henryiii:
Is it possible to run pipelines on this repo?

joerick:
Hi there. It's hard for us to add this as official support as we can't test it in our CI because I believe it's only supported if the repo lives on Bitbucket.

Perhaps we can add 'unofficial support' - we could link from our CI setup page to a page somewhere on the web that you control that contains the documentation for Bitbucket?

Issue #2380, opened 2025-05-05
Virtualenv 20.31.0 breaks cibuildwheel testing


jeremyleung521:
### Description

I was in a unique position where my CI (https://github.com/westpa/westpa/actions/runs/14840584129/attempts/1) ran slightly before and after the release of `virtualenv` [20.31.0](https://github.com/pypa/virtualenv/pull/2879/files).

From my investigation, it seems like `virtualenv` v20.31.0 no longer bundles/supports `wheel` [directly](https://github.com/pypa/virtualenv/pull/2868/files#diff-9de4cd7912fc4aecf3fbfe3f6012f14b27a08a497ecbefb8be0eb79e2eb54eccL414) and that leads to ```virtualenv: error: unrecognized arguments: --wheel=embed``` during testing, where it would somehow pull the newest virtualenv version (?). This suggests that there needs to be some changes in cibuildwheel workflow.

EDIT: Seems like related to this: https://github.com/pypa/cibuildwheel/pull/2329

### Build log

https://github.com/westpa/westpa/actions/runs/14840584129/job/41665220582#step:4:288

### CI config

https://github.com/westpa/westpa/actions/runs/14840584129/workflow

henryiii:
I assume you hit this due to  `CIBW_DEPENDENCY_VERSIONS: latest`?

jeremyleung521:
> I assume (hope) you hit this due to `CIBW_DEPENDENCY_VERSIONS: latest`?

That seems to be the case:
https://github.com/jeremyleung521/westpa/actions/runs/14843565224

jeremyleung521:
> We do need to remove `--no-wheel`, but I don't see a `--wheel=embed`, do you know where that's coming from? Upstream discussion is at [pypa/virtualenv#2868](https://github.com/pypa/virtualenv/pull/2868). Maybe Poetry? I think that might get turned into a warning in a future version of virtualenv.

Yea, no clue where that's coming from. Maybe the `--no-wheel` got translated to `--wheel=embed` [here](https://github.com/pypa/cibuildwheel/blob/edbd234b820f271eca6dadad30c6a2d67b5cb685/cibuildwheel/platforms/linux.py#L356C101-L357)?  

Anyhow, `virtualenv` is in the process of reintroducing the `--wheel` flag, so that might eventually solve it.
https://github.com/pypa/virtualenv/pull/2884

henryiii:
[...]

PR #2382, opened 2025-05-06
fix: virtualenv 20.31 no-wheel

henryiii:
Fixes our side of #2380.

### DIFF
--- a/cibuildwheel/venv.py
+++ b/cibuildwheel/venv.py
@@ -20,7 +20,7 @@ _IS_WIN: Final[bool] = sys.platform.startswith("win")
 
 
 @functools.cache
-def _ensure_virtualenv(version: str) -> Path:
+def _ensure_virtualenv(version: str) -> tuple[Path, Version]:
     version_parts = version.split(".")
     key = f"py{version_parts[0]}{version_parts[1]}"
     with resources.VIRTUALENV.open("rb") as f:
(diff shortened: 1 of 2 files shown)


henryiii:
I can't get the output from the container to be non-blank, so I just remove the arg, and it should be fine once all the containers are updated. (If not, it just adds wheel when not needed).

joerick (approved):

cibuildwheel/venv.py
```diff
@@ -110,10 +110,12 @@ def virtualenv(
     if use_uv:
         call("uv", "venv", venv_path, "--python", python)
     else:
-        virtualenv_app = _ensure_virtualenv(version)
+        virtualenv_app, virtualenv_version = _ensure_virtualenv(version)
         if pip_version is None:
             pip_version = _parse_pip_constraint_for_virtualenv(dependency_constraint)
-        additional_flags = [f"--pip={pip_version}", "--no-setuptools", "--no-wheel"]
+        additional_flags = [f"--pip={pip_version}", "--no-setuptools"]
+        if virtualenv_version < Version("20.31") or Version(version) < Version("3.9"):
```
joerick, line 117: I haven't followed this bug closely, but would just 
```suggestion
        if virtualenv_version < Version("20.31"):
```
be sufficient?

henryiii: virtualenv 20.31 was released for 3.8 too. It still includes `wheel` for 3.8 and so therefore it is helpful to remove it with `--no-wheel`.

joerick: gotcha

PR #2389, opened 2025-05-11
docs: flip TOML and envvars order

henryiii:
To start, I've asked Copilot for a script to flip the order of the tabs. It gave me this one:

```python
import re

# filepath: /Users/henryschreiner/git/software/cibuildwheel/docs/options.md
file_path = "/Users/henryschreiner/git/software/cibuildwheel/docs/options.md"

# Regular expression to match the examples section
example_pattern = re.compile(
    r"(!!! tab examples \"Environment variables\".*?```.*?```.*?)"
    r"(!!! tab examples \"pyproject\.toml\".*?```.*?```.*?)",
    re.DOTALL
)

def flip_examples(file_path):
    with open(file_path, "r") as file:
        content = file.read()

    # Swap the order of the examples
    updated_content = example_pattern.sub(r"\2\n\n\1", content)

    with open(file_path, "w") as file:
        file.write(updated_content)

    print("Examples flipped successfully!")

# Run the function
flip_examples(file_path)
```

This commit is the result of that.

---

Next I'd like to update the headings so that both options are shown. I'd like to put one on the left, the other on the right, but not sure how to do that yet. Putting this here for now.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -76,6 +76,38 @@ See the [cibuildwheel 2 documentation](https://cibuildwheel.pypa.io/en/2.x/) for
 
 #### Examples
 
+!!! tab examples "pyproject.toml"
+
+    ```toml
+    [tool.cibuildwheel]
+    # Only build on CPython 3.8
+    build = "cp38-*"
+
+    # Skip building on CPython 3.8 on the Mac
+    skip = "cp38-macosx_x86_64"
+
+    # Skip building on CPython 3.8 on all platforms
+    skip = "cp38-*"
+
+    # Skip CPython 3.8 on Windows
+    skip = "cp38-win*"
+
+    # Skip CPython 3.8 on 32-bit Windows
+    skip = "cp38-win32"
+
+    # Skip CPython 3.8 and CPython 3.9
+    skip = ["cp38-*", "cp39-*"]
+
+    # Skip Python 3.8 on Linux
+    skip = "cp38-manylinux*"
+
+    # Skip 32-bit builds
+    skip = ["*-win32", "*-manylinux_i686"]
+
+    # Disable building PyPy wheels on all platforms
+    skip = "pp*"
+    ```
+
 !!! tab examples "Environment variables"
 
     ```yaml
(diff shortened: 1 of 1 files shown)


joerick:
I added some code to make little tags like this on each option

image

They work by adding attributes using markdown [attribute list](https://python-markdown.github.io/extensions/attr_list/) syntax, like so:

```md
### `archs` {: #archs cmd-line env-var toml }
```

joerick:
But, I did notice a problem with the result of the swapping code, it seems that some of the tabs got out-of-sync in the swap - here the examples for `project-requires-python` and `enable` got mixed up.


Screen Recording 2025-05-16 at 6 54 35 pm

henryiii:
I was secretly hoping you would jump in and fix this, though this is even better than what I was imagining. :)

joerick:
I've only flipped the titles and added attrs for the first few headings, but thought I'd push where I was up to.

henryiii:
I can fix this if you want, but at PyCon, so might not be soon. (Probably tonight)

henryiii:
Fixed a few small things too, like the old mention of numpy needing to be old (not true since some 2.x version), and the text on `enable`.

henryiii:
The table at the top seems to be picking up the last section and sections, though they aren't options. It's kind of handy, but noticed the current docs don't do that.

docs/options.md
```diff
@@ -1235,18 +1248,9 @@ This option is particularly useful for:
...
-
-!!! tab examples "Environment variables"
-
-    ```yaml
-    # Build Pyodide wheels using Pyodide version 0.27.6
-    CIBW_PYODIDE_VERSION: 0.27.6
-
-    # Build Pyodide wheels using a specific alpha release
-    CIBW_PYODIDE_VERSION: 0.28.0a2
     ```
```
mayeut, line 1254:
```suggestion
```

docs/options.md
```diff
@@ -1261,17 +1265,27 @@ The available Pyodide versions are determined by the version of `pyodide-build`
...
 
+!!! tab examples "Environment variables"
+
+    ```yaml
+    # Build Pyodide wheels using Pyodide version 0.27.6
+    CIBW_PYODIDE_VERSION: 0.27.6
+
+    # Build Pyodide wheels using a specific alpha release
+    CIBW_PYODIDE_VERSION: 0.28.0a2
+
```
mayeut, line 1276:
```suggestion
    ```

```

joerick:
> By the way, the badges align a little strangely when the window is narrow. Not terrible.

Yeah, I noticed that too. I can fix it when I get half an hour to play in detail.

joerick:
A few things I noticed when looking though just now. Most are styling-related, which I can deal with.

image

- [ ] the two ALL_CAPS options jump out like a sore thumb, I think we should just normalise the header naming to kebab-case, the badge handles the specifics
- [ ] The sidebar TOC is missing a few sections

image

- [ ] This formatting of 'command-line' section in the top-of-page TOC is a bit janky

image

- [ ] this formatting of badges is odd due to the use of `float`. I'll switch to a flexbox container.

image

- [ ] the env var badge for this entry is missing a `CIBW_`

(while i'm at it)

image

- [ ] this table (build+skip) formatting is terrible with the line-breaks inside each identifier

PR #2407, opened 2025-05-22
docs: color output for docs

henryiii:
Close https://github.com/pypa/cibuildwheel/issues/2399. A little hacky for now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,7 +87,8 @@ repos:
     - id: check-github-actions
     - id: check-github-workflows
     - id: check-gitlab-ci
-    - id: check-readthedocs
+    # Disabled due to schema issue for now: 
+    # - id: check-readthedocs
     - id: check-travis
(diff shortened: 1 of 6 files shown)


henryiii:
Waiting on https://github.com/astral-sh/python-build-standalone/pull/610.

joerick (commented):
Can we tell nox that the docs need python 3.14 as well?

henryiii:
Schema validation fix in https://github.com/readthedocs/readthedocs.org/pull/12222

henryiii:
If we want to put this is, we can; I want to show off the color 3.14 output in our new docs. :) On <3.14, it should still work, it just doesn't add color.

If we want to wait, that's fine too. 3.14.0b1 is being worked on at https://github.com/astral-sh/python-build-standalone/pull/610.

joerick (approved):

joerick:
Yeah, we can put it into v3.0 I think!

PR #2411, opened 2025-05-25
fix: rename `gp242` identifier to `gp311_242`

### LINKED ISSUE #2410: Using `requires_python` skips GraalPy identifiers
### Description

Found while testing 3.0.0b1 on pybase64 https://github.com/mayeut/pybase64/actions/runs/15225239809/job/42826448795

Adding graalpy to `enable` in `pyproject.toml` had no effect on the build because it's filtered by `requires_python`:
https://github.com/pypa/cibuildwheel/blob/327f5c327dcf37610148cf8e1e93a9a086c0da2c/cibuildwheel/selector.py#L75-L84

The `requires_python` filter expects build identifier to be in the form `f{impl}{pyver}t?` but for GraalPy, the version used is not the one of Python & it's being filtered out.

### Build log

https://github.com/mayeut/pybase64/actions/runs/15225239809/job/42826448795

### CI config

https://github.com/mayeut/pybase64/blob/graalpy/.github/workflows/build-upload.yml

mayeut:
fixes #2410

This normalises identifiers to always start with `{impl}{pyver}`. It's one way of fixing #2410
The rationale behind this solution is while there's only one python version supported by GraalPy for a given version, if PyPy starts to ship a new ABI, we'll probably end-up using `pp311_{ABI_VERSION}` and identifiers will be consistent in between PyPy & GraalPy even if a bit more verbose than needed for GraalPy.


mayeut:
cc @timfel

bin/update_pythons.py
```diff
@@ -147,7 +147,11 @@ def update_version(self, identifier: str, spec: Specifier) -> AnyConfig:
             msg = f"{identifier} not supported yet on GraalPy"
             raise RuntimeError(msg)
 
+        gpspec_str = identifier.split("-")[0].split("_")[1]
+        gpspec = Specifier("==24.2.*") if gpspec_str == "242" else Specifier(f"=={gpspec_str}.*")
```
mayeut, line 151: this is likely to need an update once a new GraalPy version is released

timfel:
Looks good to me as well. I took a note to update the update_pythons.py script when we move to 25

joerick (commented):
Just missing the update to the table at https://cibuildwheel.pypa.io/en/latest/options/#build-skip

(i added the row in a parallel PR)

henryiii:
Yes, I noticed that too. Should we merge this and then update in the parallel PR?

henryiii (approved):

PR #2414, opened 2025-05-25
Debug Cirrus CI testing

joerick:
Fix the build on `main`. 

## Disables iOS testing on Cirrus.

Cirrus has a problem running the simulator, see a recent build on main, for example:

https://cirrus-ci.com/task/6394806932340736?logs=run_cibuildwheel_tests#L1832

I've removed the support tick from cirrus in the readme too, until/unless we can figure this out. 

cc @freakboy3742

## Fixes a bug with the GraalPy build

Sometimes vsdevcmd can print an error to console while it works. This messed up the environment JSON blob, so it's saved to a file instead.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,45 +5,45 @@ run_tests: &RUN_TESTS
     - python -m pip install dependency-groups
     - python -m dependency_groups test | xargs python -m pip install -e.
   run_cibuildwheel_tests_script:
-    - python ./bin/run_tests.py
+    - pytest test -k test_abi3
 
-linux_x86_task:
-  timeout_in: 120m
-  compute_engine_instance:
-    image_project: cirrus-images
-    image: family/docker-builder
-    platform: linux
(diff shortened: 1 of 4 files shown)


freakboy3742:
AFAICT, this is a manifestation of python/cpython#133131 - Xcode removed the iPhone SE simulator from the default set when they released Xcode 16.3, in favor of the newly released "iPhone 16e". As a result, the default simulator picked by the iOS testbed runner no longer exists. 

This has been fixed upstream, and the patch will be rolled into the next release of the iOS binary support package. I was planning to cut an update once 3.14 beta 2 is released (which is due in a day or two).

henryiii:
Beta 2 is out now! Also waiting on @mayut to trigger manylinux.

joerick:
This is in fact, ready, the Cirrus CI failures are due to lack of credits.

henryiii (approved):

Issue #2406, opened 2025-05-20
Test CWD issues discovered in beta


joerick:
It seems that in the beta we're seeing a couple instances of the test cwd change causing significant backwards incompatibility-

- https://github.com/numpy/numpy/pull/29007
- https://github.com/astropy/astropy/pull/18159

Even with PYTHONSAFEPATH applied, it seems that there's some behaviour (could be pytest's `prepend` mode?) that is causing issues with import paths in test commands. I haven't looked into it in any depth, but I wanted to make an issue as somewhere to discuss.

For what it's worth, I suggested making the change to the test CWD to clean up options specification, as I thought the temp dir wasn't providing us much benefit anyway. It looks like that might not have been 100% right!

I think we still have the option of keeping the temp CWD behaviour when test-sources is unset. One thing to consider is if it's worth just returning to that behaviour. Of course the best thing would be for projects to switch to `/src` for their code, but we know that a cibuildwheel upgrade is unlikely to be enough of a reason to make that (huge) change.

henryiii:
My thought was to check for usage of `{project}`. The object we pass in can detect if it's used. But it might be simpler to just base it on `test-sources`. You can't tell empty from non-defined (at least if you want to be nice to users), but `test-sources=["*"]` is pretty close to the current default. It would be nice to offer "normal" builds without copies for `src` structured projects, so I think `{project}` detection might still be a good idea. There are a couple of things still to understand or fix: #2405 and I want to understand what numpy is doing to try to import `numpy/conftest.py`. It's not in the installed wheel, so this is a really weird mishmash of local and installed files, I think. I'll try to look at it tonight or tomorrow.

henryiii:
Here's what I was thinking:

```python
            if build_options.test_sources:
                test_cwd = testing_temp_dir / "test_cwd"
                container.call(["mkdir", "-p", test_cwd])
                copy_test_sources(
                    build_options.test_sources,
                    build_options.package_dir,
                    test_cwd,
                    copy_into=container.copy_into,
                )
            elif test_command_prepared != prepare_command(
                build_options.test_command,
                wheel=wheel_to_test,
            ):
                # Back-compat mode for cibuildwheel < 3
                # Run the tests from a different directory
                log.warning(
                    "Using {project} and {package} in test_command is deprecated; triggering back-compat mode for cibuildwheel < 3."
                )
                test_cwd = testing_temp_dir / "test_cwd"
                container.call(["mkdir", "-p", test_cwd])
[...]

PR #2420, opened 2025-05-27
Revert to running tests from from a temp dir when test-sources is unset

joerick:
Fix https://github.com/pypa/cibuildwheel/issues/2406.

Alternative to #2419.

This approach restores the v2.x behaviour of running tests from a temp directory. That resolves the issues noted in #2406.

This also removes the ambiguity from https://github.com/pypa/cibuildwheel/pull/2363#issuecomment-2851205233 , so I've removed the restriction that iOS requires test-sources to be set - in this world, we can use the inclusion of the `{project}` placeholder in `test-command` as a signal that users haven't configured things correctly for iOS. Failing that, the `test_fail.py` script that a test runner (pytest) will find can prompt the user to use `test-sources`.

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -559,19 +559,19 @@ def build(options: Options, tmp_path: Path) -> None:
                         env=test_env,
                     )
 
-                    if not build_options.test_sources:
-                        # iOS requires an explicit test-sources, as the project directory
-                        # isn't visible on the simulator.
-
-                        msg = "Testing on iOS requires a definition of test-sources."
-                        raise errors.FatalError(msg)
+                    testbed_app_path = testbed_path / "iOSTestbed" / "app"
 
                     # Copy the test sources to the testbed app
-                    copy_test_sources(
-                        build_options.test_sources,
-                        build_options.package_dir,
-                        testbed_path / "iOSTestbed" / "app",
-                    )
+                    if build_options.test_sources:
+                        copy_test_sources(
+                            build_options.test_sources,
+                            build_options.package_dir,
+                            testbed_app_path,
+                        )
+                    else:
+                        (testbed_app_path / "test_fail.py").write_text(
+                            resources.TEST_FAIL_CWD_FILE.read_text()
+                        )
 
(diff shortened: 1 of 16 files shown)


cibuildwheel/platforms/ios.py
```diff
@@ -598,6 +598,26 @@ def build(options: Options, tmp_path: Path) -> None:
...
 
+                    # iOS doesn't support placeholders in the test command,
+                    # because the source dir isn't visible on the simulator.
+                    if (
+                        "{project}" in build_options.test_command
+                        or "{package}" in build_options.test_command
+                    ):
+                        msg = unwrap_preserving_paragraphs(
+                            f"""
+                            iOS tests configured with a test command that uses the {{"project"}} or
```
henryiii, line 609: Are the quotes supposed to be on the inside? `{"project"}` vs. `"{project}"`?

docs/options.md
```diff
@@ -1284,23 +1284,23 @@ Shell command to run tests after the build. The wheel will be installed
...
+- You can use the `{package}` or `{project}` placeholders in your
+  `test-command` to refer to the package being built or the project root,
+  respectively.
+
+    - `{package}` is the path to the package being built - the `package_dir`
+      argument supplied to cibuildwheel on the command line.
+    - `{project}` is an absolute path to the project root - the working
+      directory where cibuildwheel was called.
 
 On all platforms other than iOS, the command is run in a shell, so you can write things like `cmd1 && cmd2`.
```
henryiii, line 1305: Unrelated, shouldn't we support running multiple commands on iOS with `&&` (list in TOML)? We could split and run one at a time?

cibuildwheel/platforms/ios.py
joerick: Good catch!

docs/options.md
joerick: I created #2423

joerick:
Thanks for sorting the tests!

joerick:
I'll cut a new beta release now.

joerick:
...after updating dependencies to get v3.14.0b2!

PR #2427, opened 2025-05-28
Update options table in README and keep up-to-date with commit hook

joerick:
I noticed this was out-of-date and it often is, so I wrote a script to keep it fresh. It'll run via pre-commit whenever the options.md changes.


henryiii (approved):
Nevermind, I'll try `cog` as a followup with all the readme injections.

PR #2429, opened 2025-05-28
fix: drop PYTHONSAFEPATH

henryiii:
Not really needed now. This doesn't remove the test-environment, just PYTHONSAFEPATH.


joerick (approved):
Thanks!

PR #2431, opened 2025-05-29
Add pyodide 0.28

joerick:
Following on from #2002, this adds pyodide 0.28, which corresponds to cpython 3.13.

This can be held to after v3.0. If we were to ship this in a beta, would we put it behind the `python-prerelease` flag? On the one hand, it is prerelease, but on the other, these wheels are all sorta experimental from a PyPI perspective anyway, so maybe it doesn't need to hide.


yeicor:
I understand this can wait, but I’d like to kindly request making this available in the v3.0 release if possible—mainly out of convenience to avoid migrating from my fork in the future.

FWIW, [my project builds successfully](https://github.com/yeicor/OCP-wasm/actions/runs/15387191266/job/43294401848) with this version and requires it as a minimum supported version.

henryiii:
The problem is that 0.28 isn't final, and if you build binaries for 0.28, they might not work in the actual 0.28; it's quite possible there will be one last Emscripten update, or adjustments to the flags until the platform is locked in.

Behind a flag would be fine, but we'd probably want a new one, so that enabling 3.14 doesn't suddenly add new Pyodide wheels you don't want to distribution. I'm hoping we get PyPI support soon, [PEP 783 discussion](https://discuss.python.org/t/pep-783-emscripten-packaging/86862/64) seems to be stuck.

PR #2432, opened 2025-05-29
feat: support multiple commands on iOS

### LINKED ISSUE #2423: Support running multiple test commands on iOS
> Unrelated, shouldn't we support running multiple commands on iOS with `&&` (list in TOML)? We could split and run one at a time?

_Originally posted by @henryiii in https://github.com/pypa/cibuildwheel/pull/2420#discussion_r2110049868_

henryiii:
This should add support for multiple commands on iOS.

Remember to select "hide whitespace" when viewing the diff. :)

Fix #2423.

CC @freakboy3742

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -7,7 +7,7 @@ import shutil
 import subprocess
 import sys
 import textwrap
-from collections.abc import Iterator, Sequence, Set
+from collections.abc import Sequence, Set
 from pathlib import Path
 from typing import assert_never
 
@@ -25,7 +25,7 @@ from ..logger import log
 from ..options import Options
 from ..selector import BuildSelector
 from ..util import resources
-from ..util.cmd import call, shell
+from ..util.cmd import call, shell, split_command
 from ..util.file import (
     CIBW_CACHE_PATH,
     copy_test_sources,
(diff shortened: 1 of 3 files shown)


cibuildwheel/platforms/ios.py
```diff
@@ -636,43 +648,39 @@ def build(options: Options, tmp_path: Path) -> None:
...
-                            "--",
-                            *test_command_parts,
-                            env=test_env,
-                        )
-                        failed = False
-                    except subprocess.CalledProcessError:
-                        failed = True
+                            failed = False
+                        except subprocess.CalledProcessError:
+                            failed = True
```
joerick, line 677: I guess we should break out of the loop here. Or the CalledProcessError exception should be caught outside of the loop - that would negate the need for the `failed` flag I think.

joerick: A test that checks the failure short-circuits the rest of the command would be good, too.

joerick: Yeah, I get where you're coming from, but I don't think it's cibuildwheel's place to redefine what `&&` means. As idioms go, it's pretty widespread. 

(we can't really change the delimiter either, because we've always accepted string-style options with `&&` as the delimiter.)

cibuildwheel/platforms/ios.py
```diff
@@ -618,14 +629,15 @@ def build(options: Options, tmp_path: Path) -> None:
...
-                            log.warning(
-                                unwrap_preserving_paragraphs(f"""
+                    test_command_list = shlex.split(build_options.test_command)
+                    for test_command_parts in split_command(test_command_list):
+                        match test_command_parts:
+                            case ["python", "-m", *rest]:
+                                final_command = rest
+                            case ["pytest", *rest]:
+                                # pytest works exactly the same as a module, so we
+                                # can just run it as a module.
```
joerick, line 639: Sidenote, but I do really like the use of match case here :)

PR #2433, opened 2025-05-29
fix: use standard schema line

henryiii:
This was a workaround for a bug in validate-pyproject, but that bug was fixed some time ago, so we can use the standard line here now. 3.0 seems like a good time to fix this.


joerick (approved):
I don't follow tbh, but I'll trust you :)

PR #2437, opened 2025-05-30
fix: test-sources should use project dir

### LINKED ISSUE #2436: test-sources issue
### Description

If you use a subdirectory, test-sources doesn't seem to work correctly. For pybind11, there's a `tests/pyproject.toml` that is the subdirectory being tested. This is what I would have expected to work:

```diff
 [tool.cibuildwheel]
-test-command = "pytest -o timeout=0 -p no:cacheprovider {project}/tests/test_*.py"
+test-command = "python -m pytest -o timeout=0 -p no:cacheprovider tests/test_*.py"
+test-sources = ["tests", "pyproject.toml"]
```

However, it seems to be running inside the subdirectory instead.

### Build log

https://github.com/pybind/pybind11/actions/runs/15350529674/job/43197316393?pr=5705

### CI config

https://github.com/pybind/pybind11/pull/5705

henryiii:
Critical fix before 3.0: test-sources should be relative to the project root, not the package directory. The package directory is always a subdirectory of the project directory, so this is more flexible.

By the way, on iOS, it might make sense to expand `{project}` to the relative path? While you normally do know it as well, I'm not sure what all the details are if you run directly on an SDist. (So test-sources work, `{project}`/`cwd()` should probably be the root of the SDist, would be good to verify)

No documentation change since this is how it's described in the docs already. :)

Fix #2436.

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -565,7 +565,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     if build_options.test_sources:
                         copy_test_sources(
                             build_options.test_sources,
-                            build_options.package_dir,
+                            Path.cwd(),
                             testbed_app_path,
                         )
                     else:
--- a/cibuildwheel/platforms/linux.py
+++ b/cibuildwheel/platforms/linux.py
@@ -403,7 +403,7 @@ def build_in_container(
             if build_options.test_sources:
                 copy_test_sources(
                     build_options.test_sources,
-                    build_options.package_dir,
+                    Path.cwd(),
                     test_cwd,
                     copy_into=container.copy_into,
                 )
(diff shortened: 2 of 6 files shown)


joerick:
> By the way, on iOS, it might make sense to expand {project} to the relative path? While you normally do know it as well, I'm not sure what all the details are if you run directly on an SDist. (So test-sources work, {project}/cwd() should probably be the root of the SDist, would be good to verify)

Outside of test-command, I think we should can support {project} and {package} on iOS. 

But I'm okay with leaving placeholder support out of test-sources because the current implementation relies on it being a relative path. (Technically we could support relative paths for test-sources as you say, but I think it does confuse things as the same paths don't work in test-command, so I think it would be a bad idea to encourage this, and I'm not sure why we'd bother implementing it. )

In the sdist case, all of package, project and cwd are the same path, which is the root of the sdist.

henryiii:
Looks like there is an issue copying files on iOS in https://github.com/pybind/pybind11/pull/5705. That or "*" doesn't work on iOS, which is possible.

henryiii:
https://github.com/pybind/pybind11/pull/5705 is passing!

joerick (approved):
Copilot's low-confidence comments are worth looking at, but otherwise this looks good to me.

PR #2456, opened 2025-06-09
chore: order python-standalone releases for stability

henryiii:
Let's order these to make sure we don't have random reordering as seen in #2455.

By the way, we could mark these (and other generated files) as linguist-generated in our `.gitattributes`. Since we isolate these changes to PRs that are just making these changes, I'm thinking it's fine as is, but wanted to throw that out there in case others have an opinion otherwise. I think this is more for uninteresting changes (like conda-forge's plethora of support files).

### DIFF
--- a/bin/update_python_build_standalone.py
+++ b/bin/update_python_build_standalone.py
@@ -25,25 +25,25 @@ def main() -> None:
         f"repos/astral-sh/python-build-standalone/releases/tags/{latest_tag}"
     )["assets"]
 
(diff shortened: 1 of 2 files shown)


joerick:
> By the way, we could mark these (and other generated files) as linguist-generated

I think it's okay to keep the diffs visible. I will skim them generally speaking. And, in theory a supply-chain hack could be hidden in one of these files in an unrelated PR, keeping the diff visible would increase the chances of catching that.

PR #2462, opened 2025-06-10
ci: use uv python for docs (binary b1)

henryiii:
There's now a binary (of b1, not b2, but that works), which downloads in a few seconds instead of building.


joerick (approved):

henryiii:
The docs build takes 10 seconds. :) It did need to be retriggered once, network issue downloading. Not sure how common/rare that is.

PR #2475, opened 2025-06-24
Added Python 3.14 preview for iOS

radarhere:
cc @freakboy3742


freakboy3742 (approved):
Looks good to me!

mayeut:
It seems to be missing tests as all iOS tests are forcing cp313 only for now. `test_ios_platforms` should probably not enforce cp313 but rely on defaults and thus allow cp314 to be tested.

radarhere:
Ok, I've pushed a commit for that.

joerick (approved):

henryiii:
Can't check at the moment, but isn't there a table somewhere this should be added to?

radarhere:
The table at https://github.com/pypa/cibuildwheel?tab=readme-ov-file#what-does-it-do already lists it - I think https://github.com/pypa/cibuildwheel/pull/2390 just presumed it would be automatically added.

joerick:
You're right @henryiii - the options.md table was missing. I've added that.

PR #2469, opened 2025-06-13
chore: add summary to Action

henryiii:
This prints out [a summary](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/) in the GitHub action step.

Some logging commands were missing for Pyodide.

You can see the general idea, but I plan to flesh it out soon.

PR #2478, opened 2025-06-24
fix: support only one output wheel from repair

henryiii:
This removes the ability to pick up multiple output wheels from the linux repair step, brining it into line with the other operating systems. Auditwheel has, as far as I can tell, switched some time ago to just using multiple tags, not outputting multiple wheels. This will simplify #2469. If it's not true, then I can add multiple wheel outputs to that PR instead, but I think this is the cleaner solution.

### DIFF
--- a/cibuildwheel/errors.py
+++ b/cibuildwheel/errors.py
@@ -85,3 +85,21 @@ class RepairStepProducedNoWheelError(FatalError):
         )
         super().__init__(message)
         self.return_code = 8
+
+
+class RepairStepProducedMultipleWheelsError(FatalError):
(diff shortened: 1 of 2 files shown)


mayeut (approved):
> Auditwheel has, as far as I can tell, switched some time ago to just using multiple tags, not outputting multiple wheels

That's been the case for a bit more than 4 years (auditwheel 4.0.0) so LGTM

joerick (approved):

henryiii:
Should we consider this fine for a patch release, or should we hold it for a minor release?

joerick:
Hmm... maybe should wait for a minor. It might break somebody somewhere.

PR #2487, opened 2025-07-07
Updates for Pyodide 0.28 release

### LINKED ISSUE #2485: Pyodide 0.28.0 is now stable
### Description

I believe Pyodide version `0.28.0a3` should be replaced with `0.28.0` and tests updated accordingly.

https://github.com/pyodide/pyodide/releases

### Build log

_No response_

### CI config

_No response_

agriyakhetarpal:
In account of the new Pyodide release, this PR adds a build identifier for Pyodide 0.28.0, updates the constraints, and updates the docs and tests.

Closes #2485

cc: @hoodmane @ryanking13

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -233,21 +233,6 @@ jobs:
       env:
         CIBW_PLATFORM: pyodide
 
-    - name: Run a sample build (GitHub Action) for an overridden Pyodide version
-      uses: ./
-      with:
-        package-dir: sample_proj
-        output-dir: wheelhouse
-        # In case this breaks at any point in time, switch to using the latest version
-        # available or any other version that is not the same as the default one set
-        # in cibuildwheel/resources/build-platforms.toml.
-      env:
-        CIBW_PLATFORM: pyodide
-        CIBW_BUILD: "cp313*"
-        # We don't have an alpha version for 0.29.0 yet, so we use the latest released
-        # alpha version that was available for 0.28.0.
-        CIBW_PYODIDE_VERSION: "0.28.0a3"
-
     - name: Run tests with 'CIBW_PLATFORM' set to 'pyodide'
       run: |
         uv run --no-sync ./bin/run_tests.py


cibuildwheel/selector.py
```diff
@@ -99,10 +99,17 @@ def __call__(self, build_id: str) -> bool:
...
+        # Python 3.13 support became stable in Pyodide 0.28.0, so it no longer needs a prerelease
+        # flag.
+        # Also update Pyodide tests in unit_test/build_selector_test.py accordingly.
+        # When re-enabling, update the pattern to match the experimental Python version in case
+        # it is bumped to Python 3.14 (likely cp314-pyodide_* but could remain as 3.13 as well).
+        # This depends on the CPython version being used in the Pyodide runtime at the time.
+        # if EnableGroup.PyodidePrerelease not in self.enable and fnmatch(
+        #     build_id, "cp313-pyodide_*"
+        # ):
+        #     return False
```
agriyakhetarpal, line 112: I'm not sure what the best way to handle this would be. We don't have a Pyodide 0.29.0 pre-release yet because it's too early to have one. We usually stay one year behind CPython version compatibility, allowing our packages to complete their upgrades. The `cp313-pyodide_wasm32` identifier might stay for Pyodide 0.29, or it might not. The CPython version could even change between Pyodide 0.29 alphas (or 0.30 alphas, for that matter).

We currently say that "If there are pre-releases available for a newer Pyodide version, the `pyodide-prerelease` can be used to include pre-release versions." in the documentation.

We could document there that `cibuildwheel` can disable the `pyodide-prerelease` option when there are no pre-releases available?

cibuildwheel/selector.py
```diff
@@ -99,10 +99,17 @@ def __call__(self, build_id: str) -> bool:
...
```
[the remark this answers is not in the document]
agriyakhetarpal: The comment exists _because_ Python 3.13 is now stable in a released version of Pyodide, and it is not stale. So I'd mark this suggestion as incorrect. I'm happy to remove the comment, as the check would be available in the VCS history anyway; however, my above review comment about it requires a resolution first.

.github/workflows/test.yml
```diff
@@ -243,8 +243,10 @@ jobs:
         # in cibuildwheel/resources/build-platforms.toml.
       env:
         CIBW_PLATFORM: pyodide
-        CIBW_BUILD: "cp312*"
-        CIBW_PYODIDE_VERSION: "0.27.6"
+        CIBW_BUILD: "cp313*"
```
hoodmane: Why not just leave it as is?

joerick: I was gonna say, we could just remove this whole build step now? I think everything is covered in integration tests anyway now.

cibuildwheel/selector.py
joerick: I'm personally okay with the commenting approach above.

As for the other points, yeah, I can see that it might get a little awkward around alphas, but that's probably okay, that's what alphas are for :) @henryiii were discussing some ideas in #2448 about the possibility of being able to add new identifiers at runtime, which might help? Having said that, I think we're gonna continue to want the python version in the identifier to support the functionality of the [requires-python feature](https://github.com/pypa/cibuildwheel/blob/a6ec930f15689c27998ef75440c06258ae143a33/cibuildwheel/selector.py#L78).

joerick (approved):
Thank you :)

.github/workflows/test.yml
agriyakhetarpal: Sounds good – I can drop it in that case!

agriyakhetarpal: Dropped in b19e9dd94eb0136f6af7491e291512b343c295e3

joerick:
This test failure has me stumped-

> ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
>     numpy>=1.26.0 from https://cdn.jsdelivr.net/pyodide/v0.28.0/full/numpy-2.2.5-cp313-cp313-pyodide_2025_0_wasm32.whl#sha256=ebf03d021b8b89cbb2e2a8a31655e8051ee2580e2bc0729f17484066ecdc4ec7 (from spam):
>         Expected sha256 ebf03d021b8b89cbb2e2a8a31655e8051ee2580e2bc0729f17484066ecdc4ec7
>              Got        467a88e8c79f9e10c8e7052bcc32dd12d44d51098beb3b8f970e1cfc965c6757
> 

The confusing bit is that numpy isn't referenced by this test.

agriyakhetarpal:
I think the lockfile for 0.28.0 might be wrong... perhaps @hoodmane or @ryanking13 might know more? (Sorry, I'm crawling my way back to relevance after not having worked on Pyodide for a week).

hoodmane:
Yes we're aware =)

hoodmane:
We'll have to make a 0.28.1 with the fix. It's fixed by pyodide/pyodide#5749

ryanking13:
I patched the xbuildenv file in the GitHub releases manually. I think it will work now. Could you retrigger the CI? If the xbuilenv is cached, purging cache might be needed.

agriyakhetarpal:
Thanks, @ryanking13!

PR #2489, opened 2025-07-07
feat: add a `--clean-cache` command to clean up locations specified at `CIBW_CACHE_PATH`

### LINKED ISSUE #2397: One follow-up that I think I could take on here is to add a utility command around this the cache directory – it would be nice to have something like `cibuildwheel cache clean` (similar to `uv cache clean`) or `cibuildwheel cache purge` (similar to `pip cache purge`) – if y'all agree to this, of course. Currently, `emsdk` installed through takes 1.52 GiB on my machine and the `pyodide venv` is another 200 MiB, and as these are buried within the `platformdirs` cache location, I have to go there and manually delete these files. It doesn't bother me too much: while cibuildwheel can be executed locally, the "ci" in the name does mean that not everyone will run into this locally. :) Would you be open to this idea?
One follow-up that I think I could take on here is to add a utility command around this the cache directory – it would be nice to have something like `cibuildwheel cache clean` (similar to `uv cache clean`) or `cibuildwheel cache purge` (similar to `pip cache purge`) – if y'all agree to this, of course. Currently, `emsdk` installed through takes 1.52 GiB on my machine and the `pyodide venv` is another 200 MiB, and as these are buried within the `platformdirs` cache location, I have to go there and manually delete these files. It doesn't bother me too much: while cibuildwheel can be executed locally, the "ci" in the name does mean that not everyone will run into this locally. :) Would you be open to this idea?

P.S. It would be pretty useful if  `platformdirs` were to have a cache-clear command of its own, too, but I won't open an issue in their tracker just yet :)

_Originally posted by @agriyakhetarpal in https://github.com/pypa/cibuildwheel/issues/2002#issuecomment-2881739071_

agriyakhetarpal:
As described in #2397, this PR adds a `--clean-cache` command to clean `CIBW_CACHE_PATH`. It does not implement granular cache cleans at the moment (it does not clean per supported platform), and `shutil.rmtree()`s the entire folder. I added three tests that account for possible code paths: if cache files exist, if they don't, or if the cache files are somehow in a non-readable/writable location (though I think that aspect should also be validated at the creation time of the cache paths).

This is especially useful for Pyodide builds, as the Emscripten SDK can occupy up to 1.5 GiB of space; however, it is also beneficial for builds on all other platforms.

Closes #2397


unit_test/main_tests/main_options_test.py
```diff
@@ -421,6 +424,71 @@ def test_debug_traceback(monkeypatch, method, capfd):
         assert "Traceback (most recent call last)" in err
 
 
+def test_clean_cache_when_cache_exists(tmp_path, monkeypatch, capfd):
+    monkeypatch.undo()
```
agriyakhetarpal: I had to add `monkeypatch.undo()` here and in the tests below because `fake_package_dir_autouse` is autoused and applied to all tests, and it breaks these tests because I'm monkeypatching `CIBW_CACHE_PATH`.

joerick: Hmm. could we move these tests to a new module `main_commands_test.py` (`--print_build_identifiers` tests can come too, if they exist), and then scope the autouse fixture to only apply to tests in main_options_test.py, main_platform_test.py, and main_requires_python_test.py?

henryiii: @agriyakhetarpal Did you want to do this? Or should I add it to my list of things I'll do if there's a lull in the SciPy sprints? :)

agriyakhetarpal: Thanks for the reminder, @henryiii – I am onboard with @joerick's idea here. I think this is indeed something a new contributor/you can take at the sprints, so I am okay with it :D

henryiii:
The one thought I had was could we check for something specific to make sure this is in fact a cibuildwheel cache dir? Just to protect for a mistake like setting the cache dir to something like $HOME. :)

agriyakhetarpal:
> The one thought I had was could we check for something specific to make sure this is in fact a cibuildwheel cache dir? Just to protect for a mistake like setting the cache dir to something like $HOME. :)

Do you mean we should check if `CIBW_CACHE_PATH` contains `cibuildwheel` or if it matches `platformdirs.user_cache_path()`?

Realistically, I think we'd have no defense against this unless we make the cache path non-configurable and drop the ability to set it via the `DEFAULT_CIBW_CACHE_PATH` environment variable. This is unless I've misunderstood you here :)

henryiii:
I was thinking looking for a specific file or folder that would exist if it was cibuildwheel's cache.

agriyakhetarpal:
Right, we could include a sentinel `.cibuildwheel_cached` file inside the cache folder; we do similar things with Pyodide's xbuildenv cache.

joerick:
> Right, we could include a sentinel `.cibuildwheel_cache` file inside the cache folder; we do similar things with Pyodide's xbuildenv cache.

That's a nice idea. Might save somebody's day.

cibuildwheel/util/file.py
```diff
@@ -20,6 +20,20 @@
...
+def ensure_cache_sentinel(cache_path: Path) -> None:
+    """Create a sentinel file to mark a cibuildwheel cache directory.
+
+    We help prevent accidental deletion of non-cache directories with this.
+
+    Args:
+        cache_path: The cache directory path where the sentinel should be created
+    """
+    if cache_path.exists():
+        sentinel_file = cache_path / ".cibuildwheel_cached"
```
henryiii: I guess this is the standard tense? I would have expected `.cibuildwheel_cache`, I think, based on [CACHEDIR.TAG](https://bford.info/cachedir/) (which is what uv uses). I [asked ChatGPT](https://chatgpt.com/share/68a346c4-ac28-8012-be37-cb78b5051305), and it guessed `.cibuildwheel_cache` too, though it also recommended `CACHEDIR.TAG`, and I think that would be fine, actually. If you did point it at a .venv or something else with a `CACHEDIR.TAG`, I think deleting it would be fine.

```suggestion
        sentinel_file = cache_path / "CACHEDIR.TAG"
```

cibuildwheel/util/file.py
```diff
@@ -20,6 +20,20 @@
...
+        if not sentinel_file.exists():
+            sentinel_file.write_text("# Created by cibuildwheel automatically\n")
```
henryiii:
```suggestion
            sentinel_file.write_text("Signature: 8a477f597d28d172789f06886806bc55\n"
            "# This file is a cache directory tag created by cibuildwheel.\n")
```

unit_test/main_commands_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_clean_cache_when_cache_exists(tmp_path, monkeypatch, capfd):
+    fake_cache_dir = (tmp_path / "cibw_cache").resolve()
+    monkeypatch.setattr(main_module, "CIBW_CACHE_PATH", fake_cache_dir)
+
+    fake_cache_dir.mkdir(parents=True, exist_ok=True)
+    assert fake_cache_dir.exists()
+
+    cibw_sentinel = fake_cache_dir / ".cibuildwheel_cached"
+    cibw_sentinel.write_text("# Created by cibuildwheel automatically", encoding="utf-8")
```
henryiii:
```suggestion
    cibw_sentinel = fake_cache_dir / "CACHEDIR.TAG"
    cibw_sentinel.write_text("Signature: 8a477f597d28d172789f06886806bc55\n"
    "# This file is a cache directory tag created by cibuildwheel.\n", encoding="utf-8")
```

unit_test/main_commands_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_clean_cache_with_error(tmp_path, monkeypatch, capfd):
+    fake_cache_dir = (tmp_path / "cibw_cache").resolve()
+    monkeypatch.setattr(main_module, "CIBW_CACHE_PATH", fake_cache_dir)
+
+    fake_cache_dir.mkdir(parents=True, exist_ok=True)
+    assert fake_cache_dir.exists()
+
+    cibw_sentinel = fake_cache_dir / ".cibuildwheel_cached"
+    cibw_sentinel.write_text("# Created by cibuildwheel automatically\n")
```
henryiii:
```suggestion
    cibw_sentinel = fake_cache_dir / "CACHEDIR.TAG"
    cibw_sentinel.write_text("Signature: 8a477f597d28d172789f06886806bc55\n"
    "# This file is a cache directory tag created by cibuildwheel.\n")
```

cibuildwheel/__main__.py
```diff
@@ -196,6 +202,31 @@ def main_inner(global_options: GlobalOptions) -> None:
...
+            sentinel_file = CIBW_CACHE_PATH / ".cibuildwheel_cached"
+            if not sentinel_file.exists():
+                print(
+                    f"Error: {CIBW_CACHE_PATH} does not appear to be a cibuildwheel cache directory.",
+                    file=sys.stderr,
+                )
+                print(
+                    "Only directories with a .cibuildwheel_cached sentinel file can be cleaned.",
+                    file=sys.stderr,
+                )
```
henryiii:
```suggestion
                print(
                    f"Error: {CIBW_CACHE_PATH} does not appear to be a cibuildwheel cache directory.",
                    "Only directories with a CACHEDIR.TAG sentinel file can be cleaned.",
                    sep="\n",
                    file=sys.stderr,
                )
```

cibuildwheel/__main__.py
```diff
@@ -196,6 +202,31 @@ def main_inner(global_options: GlobalOptions) -> None:
 
     global_options.print_traceback_on_error = args.debug_traceback
 
+    if args.clean_cache:
+        if CIBW_CACHE_PATH.exists():
+            sentinel_file = CIBW_CACHE_PATH / ".cibuildwheel_cached"
+            if not sentinel_file.exists():
```
henryiii: All CACHEDIR.TAG file start with the exact same signature string, so that could be checked here, but not really needed, probably, safe enough. We could also add a comment stating it's a cibuildwheel cache dir (added below).

cibuildwheel/__main__.py
```diff
@@ -196,6 +202,29 @@ def main_inner(global_options: GlobalOptions) -> None:
 
     global_options.print_traceback_on_error = args.debug_traceback
 
+    if args.clean_cache:
+        if CIBW_CACHE_PATH.exists():
+            sentinel_file = CIBW_CACHE_PATH / ".cibuildwheel_cached"
```
agriyakhetarpal:
```suggestion
            sentinel_file = CIBW_CACHE_PATH / "CACHEDIR.TAG"
```

agriyakhetarpal:
```suggestion
            sentinel_file = CIBW_CACHE_PATH / "CACHEDIR.TAG"
```

agriyakhetarpal:
`CACHEDIR.TAG` makes a lot of sense. Linking the specification here for future reference: https://bford.info/cachedir/

cibuildwheel/__main__.py
agriyakhetarpal: Makes sense to me! I think we can add it.

agriyakhetarpal: Addressed in 698c5f7a117bb2def96e36f50c545b0b137426ef.

agriyakhetarpal:
Should we also comply with https://arc.net/l/quote/bosrzvuf?

Edit: quoting here, it says the following:

>  For the benefit of anyone who happens to find and look at a cache directory tag directly, it is recommended that applications include in the file a comment referring back to this specification. For example:
>

	Signature: 8a477f597d28d172789f06886806bc55
	# This file is a cache directory tag created by (application name).
	# For information about cache directory tags, see:
	# http://www.brynosaurus.com/cachedir/
	
>
> The official "home" URL of this specification may change from the above at some point if/when this proposal becomes > more of a formal standard, but in that case an appropriate forwarding link will be left in the old location.

henryiii:
I'd be fine with that - I was tempted to add it but was a little hard in comments.

FWIW, uv does not add this.

henryiii:
@freakboy3742 I'm guessing that since Azure hasn't migrated "macOS-latest" to 15 yet, there's nothing we can do about the Azure Pipelines failures, since those are on 14?

freakboy3742:
> @freakboy3742 I'm guessing that since Azure hasn't migrated "macOS-latest" to 15 yet, there's nothing we can do about the Azure Pipelines failures, since those are on 14?

Azure started moving `macos-latest` to `macos-15` on August 4; the failed run was [running on macos-15](
https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=7980&view=logs&j=9129c50d-495a-566a-0c8b-b212c617768f&t=2a0afef3-d06d-4c6e-911a-47a0f9468785&l=15).

GitHub's response to the problem appears to be ["yeah, looks like macOS-15 is broken; try using macOS-14 for now, and maybe it will be fixed in the next release"](https://github.com/actions/runner-images/issues/12777#issuecomment-3200018075).

I've opened #2563 to switch the Azure image, and add some documentation about the problem. That documentation is hopefully temporary, but since I can't say for certain, it doesn't hurt to throw a paragraph in.

PR #2514, opened 2025-07-24
Add Cirrus macOS as a runner that supports building for Android but not testing

mhsmith:
This should prevent failures like [this one](https://cirrus-ci.com/task/6086399490785280).


mhsmith:
Everything's passing now except CircleCI, which doesn't appear to have anything to do with this PR.

henryiii (approved):

PR #2512, opened 2025-07-24
BUG: hotfix report defects in logs and HTML summaries

neutrinoceros:
close #2511

disclaimer: I have not attempted to run tests locally yet (I have very limited battery at the moment), but I did run typechecking via `pre-commit`, which correctly caught a mistake in my initial iteration, so I'm reasonably confident in this patch.

PR #2517, opened 2025-07-25
BUG: fix incorrect number of wheels in logs

neutrinoceros:
hotfix for my own hotfix #2512
Thanks @joerick for quickly finding my mistake !

### DIFF
--- a/unit_test/wheel_print_test.py
+++ b/unit_test/wheel_print_test.py
@@ -15,9 +15,11 @@ def test_printout_wheels(capsys):
     log.colors_enabled = False
 
     with log.print_summary(options=OPTIONS_DEFAULTS):
+        # the number of BuildInfo with & without filename shall be different for this test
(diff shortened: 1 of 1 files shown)


mayeut:
thanks for the quick fix @neutrinoceros 
I modified the test to be able to check the fix (fails without the fix, passes with the fix).

mayeut (approved):

neutrinoceros:
Thanks for completing and merging the PR @mayeut !
out of curiosity, are there plans to release the fix soon, or any reason to wait for it ? cibuildwheel is only *indirectly* pinned in some re-usable workflows I help maintain, and I'm wondering if we should go straight from 3.0.0 to a hypothetical 3.1.2. Not a big deal if that's not planned. Thanks !

henryiii:
Personally, I'd like to get a 3.1.2 release out soon.

PR #2519, opened 2025-07-25
ci: rework tests to break out ios/android tests

henryiii:
Fail fast disabled, and trying to split out the android/ios runs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,18 +47,44 @@ jobs:
       run: pipx run --python "${{ steps.python.outputs.python-path }}" nox -s pylint -- --output-format=github
 
   test:
-    name: Test on ${{ matrix.os }} (${{ matrix.python_version }})
+    name: Test on ${{ matrix.os }} (${{ matrix.python_version }}) ${{ matrix.platform }}
     needs: lint
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-13, macos-15]
-        python_version: ['3.13']
         include:
+        # Min Python
         - os: ubuntu-latest
           python_version: '3.11'
+        # Max Python
         - os: ubuntu-latest
           python_version: '3.14'
+        - os: ubuntu-latest
+          python_version: '3.13'
+          platform: android
+        - os: ubuntu-24.04-arm
+          python_version: '3.13'
+        - os: windows-latest
+          python_version: '3.13'
+        - os: windows-11-arm
+          python_version: '3.13'
+        - os: macos-13
+          python_version: '3.13'
+        - os: macos-15
+          python_version: '3.13'
+        - os: macos-13
+          python_version: '3.13'
+          platform: ios
+        - os: macos-15
+          python_version: '3.13'
+          platform: ios
+        - os: macos-13
+          python_version: '3.13'
+          platform: android
+        - os: macos-15
+          python_version: '3.13'
+          platform: android
     timeout-minutes: 180
     steps:
     - uses: actions/checkout@v4
@@ -97,6 +123,7 @@ jobs:
         uv sync --no-dev --group test
 
     - uses: joerick/pr-labels-action@v1.0.9
+
     - name: Set CIBW_ENABLE
       shell: bash
       run: |
@@ -128,9 +155,11 @@ jobs:
       env:
         CIBW_ARCHS_MACOS: x86_64 universal2 arm64
         CIBW_BUILD_FRONTEND: 'build[uv]'
+        CIBW_PLATFORM: ${{ matrix.platform }}
 
     - name: Run a sample build (GitHub Action, only)
       uses: ./
+      if: matrix.platform == ''
       with:
         package-dir: sample_proj
         output-dir: wheelhouse_only
@@ -150,6 +179,8 @@ jobs:
 
     - name: Run a sample build (GitHub Action, config-file)
       uses: ./
+      env:
+        CIBW_PLATFORM: ${{ matrix.platform }}
       with:
         package-dir: sample_proj
         output-dir: wheelhouse_config_file
@@ -168,6 +199,8 @@ jobs:
         path: wheelhouse/*.whl
 
     - name: Test cibuildwheel
+      env:
+        CIBW_PLATFORM: ${{ matrix.platform }}
       run: |
         uv run --no-sync bin/run_tests.py ${{ (runner.os == 'Linux' && runner.arch == 'X64') && '--run-podman' || '' }}
 
--- a/bin/run_tests.py
+++ b/bin/run_tests.py
@@ -62,6 +62,14 @@ if __name__ == "__main__":
         "test",
         "-vv",
     ]
+    match os.environ.get("CIBW_PLATFORM", "native"):
+        case "":
+            serial_integration_test_args += ["-m not pyodide", "-m not android", "-m not ios"]
+        case "native":
+            pass
+        case platform:
+            serial_integration_test_args += ["-m {platform}"]
+
     print(
         "\n\n=========================== SERIAL INTEGRATION TESTS ===========================",
         flush=True,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,9 @@ filterwarnings = ["error"]
 log_cli_level = "INFO"
 markers = [
     "serial: tests that must *not* be run in parallel (deselect with '-m \"not serial\"')",
+    "ios: tests that run on iOS",
+    "android: tests that run on Android",
+    "pyodide: tests that run on Pyodide",
 ]
 
 [tool.mypy]
--- a/test/test_android.py
+++ b/test/test_android.py
@@ -12,6 +12,9 @@ import pytest
 from .test_projects import new_c_project
 from .utils import cibuildwheel_run, expected_wheels
 
+pytestmark = pytest.mark.pyodide
+
+
 CIBW_PLATFORM = os.environ.get("CIBW_PLATFORM", "android")
 if CIBW_PLATFORM != "android":
     pytest.skip(f"{CIBW_PLATFORM=}", allow_module_level=True)
--- a/test/test_ios.py
+++ b/test/test_ios.py
@@ -12,6 +12,12 @@ from cibuildwheel.ci import CIProvider, detect_ci_provider
 
 from . import test_projects, utils
 
+pytestmark = pytest.mark.ios
+
+CIBW_PLATFORM = os.environ.get("CIBW_PLATFORM", "ios")
+if CIBW_PLATFORM != "ios":
+    pytest.skip(f"{CIBW_PLATFORM=}", allow_module_level=True)
+
 basic_project_files = {
     "tests/test_platform.py": f"""
 import platform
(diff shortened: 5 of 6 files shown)


joerick:
What do you think about also introducing [flaky](https://pypi.org/project/flaky/) on the ios tests?

henryiii:
We could, or there's an action that reruns on failure; actually I see at least two:

* https://github.com/marketplace/actions/retry-step
* https://github.com/marketplace/actions/retry-action

I've used one of those before, not sure which one.

bin/run_tests.py
```diff
@@ -62,6 +62,14 @@
         "test",
         "-vv",
     ]
+    match os.environ.get("CIBW_PLATFORM", "native"):
+        case "":
+            serial_integration_test_args += ["-m not pyodide", "-m not android", "-m not ios"]
+        case "native":
+            pass
+        case platform:
+            serial_integration_test_args += ["-m {platform}"]
```
henryiii:
```suggestion
            serial_integration_test_args += [f"-m {platform}"]
```

.github/workflows/test.yml
```diff
@@ -128,9 +155,11 @@ jobs:
       env:
         CIBW_ARCHS_MACOS: x86_64 universal2 arm64
         CIBW_BUILD_FRONTEND: 'build[uv]'
```
henryiii:
```suggestion
        CIBW_BUILD_FRONTEND: ${{ matrix.platform && 'build' || 'build[uv]' }}
```

.github/workflows/test.yml
```diff
@@ -128,9 +155,11 @@ jobs:
...
+        CIBW_PLATFORM: ${{ matrix.platform }}
 
     - name: Run a sample build (GitHub Action, only)
       uses: ./
+      if: matrix.platform == null || matrix.platform == ''
```
henryiii:
```suggestion
      if: matrix.platform == ''
```

.github/workflows/test.yml
```diff
@@ -161,15 +190,21 @@ jobs:
       shell: bash
       run: |
         test $(find wheelhouse -name '*.whl' | wc -l) -ge 1
-        test $(find wheelhouse_only -name '*.whl' | wc -l) -eq 1
         test $(find wheelhouse_config_file -name '*.whl' | wc -l) -eq 1
 
+    - name: Check Action artifacts (native build only)
+      if: matrix.platform == ''
+      run: |
```
mayeut, line 198:
```suggestion
      shell: bash
      run: |
```

henryiii:
Weird, android is getting stuck in the unit_tests and hanging, while that should be identical to native. And macOS android is failing to download. Not sure what is causing this yet, though.

bin/run_tests.py
```diff
@@ -27,6 +27,12 @@
         default=default_cpu_count,
         help="number of processes to use for testing",
     )
+    parser.add_argument(
+        "--platform",
+        choices={"all", "native", "android", "ios", "pyodide"},
+        default="all",
+        help="Either 'native' or 'android'/'ios'/'pyodide'",
+    )
```
joerick: Given we already have sys.platform and CIBW_PLATFORM changing the behaviour in this script, having another thing called "platform" is confusing me! Maybe we call this `--test-select` instead?

joerick (commented):
The android unit test is probably failing due to something related to Docker - the docker tests are selected on that runner but binfmt isn't installed. I think my suggestion would be to not run the docker unit tests when testing android.

I haven't been able to figure out why the macOS android build is failing on Azure though.

joerick:
@mhsmith there's a test failure here that you might be interested in? On Azure pipelines, macOS x86_64, the android runner seems to fail while installing the android ndk. The error looks like this:

> `2025-08-11T09:49:21.8926460Z >       assert "Android requires the build frontend to be 'build'" in capfd.readouterr().err
2025-08-11T09:49:21.8942480Z E       assert "Android requires the build frontend to be 'build'" in 'Installing NDK - this may take several minutes\nWarning: An error occurred while preparing SDK package NDK (Side by side) 27.2.12479018: /Users/runner/Library/Android/sdk/.temp/PackageOperation01/unzip/android-ndk-r27c/sources/third_party/shaderc/third_party/spirv-tools/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp.\ncibuildwheel: \x1b[91merror\x1b[0m: Command [\'/private/var/folders/hg/d8rl25qs4nl00s45_rtwkjmr0000gn/T/cibw-run-ho0e8esi/cp313-android_x86_64/python/android.py\', \'env\'] failed with code 1. Loading package information...                                                  \nLoading local repository...                                                     \n[                                       ] 3% Loading local repository...        \n[                                       ] 3% Fetch remote repository...         \n[=                                      ] 3% Fetch remote repository...         \n[=                                      ] 4% Fetch remote repository...         \n[=                                      ] 5% Fetch remote repository...         \n[==                                     ] 5% Fetch remote repository...         \n[==                                     ] 6% Fetch remote repository...         \n[==                                     ] 7% Fetch remote repository...         \n[==                                     ] 7% Computing updates...               \n[===                                    ] 8% Computing updates...               \n[===                                    ] 10% Computing updates...              \n[===                                    ] 10% Installing NDK (Side by side) 27.2\n[===                                    ] 10% Verifying previous download...    \n[===                                    ] 10% Checking existing file...         \n[===                                    ] 10% Downloading android-ndk-r27c-darwi\n[====                                   ] 10% Downloading android-ndk-r27c-darwi\n[====                                   ] 11% Downloading android-ndk-r27c-darwi\n[====                                   ] 12% Downloading android-ndk-r27c-darwi\n[=====                                  ] 13% Downloading android-ndk-r27c-darwi\n[=====                                  ] 14% Downloading android-ndk-r27c-darwi\n[=====                                  ] 15% Downloading android-ndk-r27c-darwi\n[======                                 ] 15% Downloading android-ndk-r27c-darwi\n[======                                 ] 16% Downloading android-ndk-r27c-darwi\n[======                                 ] 17% Downloading android-ndk-r27c-darwi\n[=======                                ] 18% Downloading android-ndk-r27c-darwi\n[=======                                ] 19% Downloading android-ndk-r27c-darwi\n[=======                                ] 20% Downloading android-ndk-r27c-darwi\n[========                               ] 20% Downloading android-ndk-r27c-darwi\n[========                               ] 21% Downloading android-ndk-r27c-darwi\n[========                               ] 22% Downloading android-ndk-r27c-darwi\n[=========                              ] 23% Downloading android-ndk-r27c-darwi\n[=========                              ] 24% Downloading android-ndk-r27c-darwi\n[=========                              ] 25% Downloading android-ndk-r27c-darwi\n[==========                             ] 25% Downloading android-ndk-r27c-darwi\n[==========                             ] 26% Downloading android-ndk-r27c-darwi\n[==========                             ] 27% Downloading android-ndk-r27c-darwi\n[===========                            ] 28% Downloading android-ndk-r27c-darwi\n[===========                            ] 29% Downloading android-ndk-r27c-darwi\n[===========                            ] 30% Downloading android-ndk-r27c-darwi\n[============                           ] 30% Downloading android-ndk-r27c-darwi\n[============                           ] 31% Downloading android-ndk-r27c-darwi\n[============                           ] 32% Downloading android-ndk-r27c-darwi\n[============                           ] 33% Downloading android-ndk-r27c-darwi\n[============                           ] 33% Unzipping...                      \n[============                           ] 33% Unzipping... android-ndk-r27c/    \n[============                           ] 33% Unzipping... android-ndk-r27c/simp\n[=============                          ] 33% Unzipping... android-ndk-r27c/simp\n[=============                          ] 33% Unzipping... android-ndk-r27c/ndk-\n[=============                          ] 33% Unzipping... android-ndk-r27c/wrap\n[=============                          ] 33% Unzipping... android-ndk-r27c/meta\n[=============                          ] 33% Unzipping... android-ndk-r27c/pyth\n[=============                          ] 33% Unzipping... android-ndk-r27c/sour\n[=============                          ] 33% Unzipping... android-ndk-r27c/preb\n[=============                          ] 33% Unzipping... android-ndk-r27c/READ\n[=============                          ] 33% Unzipping... android-ndk-r27c/sour\n[=====================                  ] 55% Unzipping... android-ndk-r27c/sour\n                                                                                \n[=====================                  ] 55% Unzipping... android-ndk-r27c/sour\nCommand "set -eu; HOST=x86_64-linux-android; PREFIX=/private/var/folders/hg/d8rl25qs4nl00s45_rtwkjmr0000gn/T/cibw-run-ho0e8esi/cp313-android_x86_64/python/prefix; . /private/var/folders/hg/d8rl25qs4nl00s45_rtwkjmr0000gn/T/cibw-run-ho0e8esi/cp313-android_x86_64/python/android-env.sh; export" returned exit status 1\n\n\n'
`

It seems to just stop unzipping the archive and error out. [The output](https://gist.github.com/joerick/18c59d0c31ac71c41eb5aecd792739fb#file-gistfile1-txt-L209)'s a bit cluttered because pytest is capturing it, I'll try to make it easier to read...

mhsmith:
Could the runner be running out of disk space while unzipping? I hit that problem on GHA, though I think the symptoms were different. I worked around it here:

https://github.com/pypa/cibuildwheel/blob/3ebb6bd9429b35f6b00c98f5ef191fd4c391709d/.github/workflows/test.yml#L74-L75

joerick:
Ah, it looks like I might have accidentally worked around it! I moved the first android test into the serial phase of the tests and it disappeared. So I'm thinking that maybe the issue was that the `sdkmanager` call was being performed concurrently, and that was failing in a weird way. Maybe the `android-env.sh` script could lock around this, as it is system global state. I'm actually a bit surprised that sdkmanager doesn't do that itself.

bin/run_tests.py
joerick: I hope you don't mind, I went ahead and made this change.

mhsmith:
Ah, this might be because the NDK version specified in the android-env file is no longer pre-installed on the runners, so the android-env file now needs to install it. The NDK version will be updated in the next release of Python 3.13, but the versions on the runners change frequently, so the tests should be robust enough to cope with it not being present.

It should be a sufficient workaround to run at least one Android test in the serial stage that gets as far as checking the NDK installation. This wasn't previously happening on runners that don't support the emulator, because the serial tests all require the emulator and were therefore skipped.

Now that you've moved another test into the serial phase, it should be possible to remove the `flaky` decorator, unless it was there for some other reason.

azure-pipelines.yml
```diff
@@ -16,62 +16,61 @@ pr:
...
-        fi
-        python ./bin/run_tests.py
+    - bash: docker run --rm --privileged docker.io/tonistiigi/binfmt:latest --install all
+      condition: and(eq(variables['imageName'], 'ubuntu-latest'), eq(variables['testSelect'], 'native'))
+      displayName: 'Install binfmt on Linux'
+
+    - bash: python -m pip install -U pip && python -m pip install -e. --group test
+      displayName: 'Update pip and install cibuildwheel --group test'
+
+    - bash: echo "##vso[task.setvariable variable=CIBW_ENABLE;all]"
```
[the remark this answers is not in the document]
henryiii:
```suggestion
    - bash: echo "##vso[task.setvariable variable=CIBW_ENABLE;]all"
```
Looks like ti should be this, actually, from https://learn.microsoft.com/en-us/azure/devops/pipelines/process/set-variables-scripts?view=azure-devops&tabs=bash

joerick (approved):
Nice, I think this is a great improvement to our CI setup.

PR #2523, opened 2025-07-26
tests: reduce testing time on iOS

mayeut:
A number of tests only require to run with a single version of python.
This will reduce the time needed to run those tests.

Complements #2519


freakboy3742 (approved):
👍 Agreed this drops matrix tests where they're not adding any real test coverage.

PR #2524, opened 2025-07-27
tests: mark and add skips

henryiii:
Pulled from #2519 to simplify that one.


.github/workflows/test.yml
```diff
@@ -143,7 +145,7 @@ jobs:
         [tool.cibuildwheel]
         # Only build on CPython 3.12 on native arch
```
mayeut:
```suggestion
        # Only build on CPython 3.13 on native arch
```

PR #2525, opened 2025-07-27
break: drop unused riscv64 enable option

Xeonacid:
Fixes #2509

cc @mayeut


mayeut:
This will break some users, holding for a future major or minor release.

henryiii:
I think a future minor release should be fine, like 3.3, since it was experimental and probably doesn't have a huge number of users. Fine to wait for 4.0 if others think differently, though.

mayeut:
Something like 3.3 sounds good

mayeut (approved):
We're now after 3.3.0, I think it's fine to merge, even if the next version is 3.3.1

mayeut:
The 3.3 series is being handled as a branch (#2708), we can safely merge this now.

PR #2544, opened 2025-08-05
fix: apply base constraints to build env

henryiii:
This hopefully will help with failures like #2542 and main. Currently, we lose all constraints, even our base ones, when there's custom constraints. This adds any required constraints used to generate the constraint file to the constraints always.

### DIFF
--- a/cibuildwheel/util/packaging.py
+++ b/cibuildwheel/util/packaging.py
@@ -86,9 +86,7 @@ class DependencyConstraints:
                 if variant == "pyodide"
                 else resources.CONSTRAINTS_IN
             )
-            input_packages = (
-                s.strip() for s in file_name.read_text().splitlines()
-            )
+            input_packages = (s.strip() for s in file_name.read_text().splitlines())
             input_requirements = (
                 packaging.requirements.Requirement(s)
                 for s in input_packages


cibuildwheel/util/packaging.py
```diff
@@ -80,8 +81,23 @@ def get_for_python_version(
         self, *, version: str, variant: Literal["python", "pyodide"] = "python", tmp_dir: Path
     ) -> Path | None:
         if self.packages:
+            file_name = (
+                resources.PATH / f"pyodide{version}.in"
```
henryiii:
```suggestion
                resources.PATH / f"constraints-pyodide{version}.in"
```

joerick:
Hmmm. I don't follow this yet... why would adding more constraints fix the dependency resolve - it looks to me that pip [tries](https://github.com/pypa/cibuildwheel/actions/runs/16715390777/job/47307843871?pr=2542#step:8:824) to install a later version of pyodide-build but that doesn't satisfy the resolver for some reason...

joerick:
I think I found a neater solution here - https://github.com/pypa/cibuildwheel/pull/2542/commits/ddc285ccca564d898b030ce4423cac912119a7db

joerick:
#2548 fixes the current problem. I think unless there's a strong reason to do something like this, I'd be opposed to a solution like this, 

- because of the increased complexity
- because there's no way for a user to override constraints. I think this could become an issue down the line - say for example we have an upper bound in our base constraints, (like `click<8.2` due to some bug that is fixed after our release), but a future requirement needs something in click>8.3. There would be no way for a user to override our built-in constraints.

PR #2548, opened 2025-08-11
fix: pyodide build fails when dependency-versions is set

joerick:
Fix the pyodide failures on main.

auditwheel-emscripten is already a dep of pyodide-build, and cibuildwheel doesn't use it directly. But by listing it in this "pip install" command, it was causing the resolver to backtrack to 0.25.0 of pyodide-build, because more recent versions of pyodide-build pin auditwheel-emscripten.

Pyodide folks, does that sound right to you?

### DIFF
--- a/cibuildwheel/platforms/pyodide.py
+++ b/cibuildwheel/platforms/pyodide.py
@@ -284,9 +284,8 @@ def setup_python(
         "pip",
         "install",
         "--upgrade",
-        "auditwheel-emscripten",
-        "build[virtualenv]",
         "pyodide-build",
+        "build[virtualenv]",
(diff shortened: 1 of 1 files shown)


henryiii:
Ahh, so that's it. I still think something like #2544 would be good eventually, as our limits (like capping click, or selecting a pyodide version) aren't applying to unpinned environments.

agriyakhetarpal (approved):
This LGTM from my end – yes, `auditwheel-emscripten` is now at 0.2.0 after RPATH support was added. We used to previously pin it but there is no reason to do so now. Thank you!

hoodmane:
Thanks @joerick!

PR #2549, opened 2025-08-11
Improve docs on how to skip free-threaded builds

### LINKED ISSUE #2531: Provide clear documentation on not supporting free threaded Python
### Description

The latest cibuildwheel automatically enables free threaded Python 3.14 compilation.    It isn't clear from the documentation how to disable it.

[Doc search](https://cibuildwheel.pypa.io/en/stable/search.html?q=free+threaded)

The first result [shows how to enable](https://cibuildwheel.pypa.io/en/stable/options/#examples_3) but not disable.

The second result [shows enable option](https://cibuildwheel.pypa.io/en/stable/options/#enable) but there is no corresponding disable option.

It seems like enable uses `cpython-freethreading` while disabling requires skip using a glob pattern with `t` somewhere in the name, hoping that you have the exact right pattern and don't accidentally match an unintended `t`.  Hence why clear documentation is needed.

My C extension uses the GIL so it compiles fine, but loading the extension re-enables the GIL.  It also isn't clear if I should disable the building, or let it be published to pypi anyway.

### Build log

_No response_

### CI config

_No response_

joerick:
Fix #2531.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -74,7 +74,7 @@ Windows arm64 platform support is experimental.
 
 See the [cibuildwheel 2 documentation](https://cibuildwheel.pypa.io/en/2.x/) for past end-of-life versions of Python.
 
-#### Examples
+#### Examples {: #build-skip-examples }
 
 !!! tab examples "pyproject.toml"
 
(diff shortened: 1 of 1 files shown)


henryiii:
Would it make sense to insert a link to https://py-free-threading.github.io here (or maybe two)?

joerick:
Yeah, that's a nice reference, thanks.

PR #2552, opened 2025-08-12
fix: add patchelf requirement for android

henryiii:
This adds `"patchelf"` on platforms where Android wheels can be built, since it's required. It's not added on platforms where a binary isn't available, which shouldn't be building Android wheels either.

The alternative is to just require users to pre-install it, which isn't bad, but I think this is nicer?

See https://github.com/pybind/pybind11/pull/5714#discussion_r2265401851.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,8 @@ dependencies = [
     "filelock",
     "humanize",
     "packaging>=20.9",
+    # patchelf is used for Android
+    "patchelf; (sys_platform == 'linux' or sys_platform == 'darwin') and (platform_machine == 'x86_64' or platform_machine == 'arm64' or platform_machine == 'aarch64')",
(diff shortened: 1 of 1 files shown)


joerick (approved):

henryiii:
@mhsmith, it looks like we are getting `Error: cibuildwheel: Couldn't find 'patchelf' in PATH` when running on macOS (https://github.com/pybind/pybind11/actions/runs/17138877560/job/48621564021?pr=5809), maybe we should see if patchelf is installed alongside cibuildwheel and then use it from there if it was present?

mhsmith:
Yes, this is the same issue we previously had with `wheel`: the cibuldwheel GitHub action runs cibuildwheel without activating the environment, so the environment's `bin` directory isn't on the PATH. I'll deal with this in #2568.

PR #2568, opened 2025-08-21
Android updates

mhsmith:
* Add Python 3.14 support, including downloading official binaries from python.org.

* Simplify update_pythons.py:
  * Remove the redundant Windows `arch` setting.
  * Unify `Config` classes which are now identical.
  * Make the new config replace the old config rather than merge with it. This makes the behavior easier to reason about.

* Update to latest versions of Python 3.13 and 3.14:
  * CPython's android.py script now automatically configures KVM on GitHub Actions, so cibuildwheel users no longer need to do that manually.

* Don't assume that `patchelf` will be on the PATH ([#2552](https://github.com/pypa/cibuildwheel/pull/2552)).


cibuildwheel/platforms/android.py
```diff
@@ -508,17 +508,21 @@ def repair_default(
         new_soname = soname_with_hash(src_path)
         dst_path = libs_dir / new_soname
         shutil.copyfile(src_path, dst_path)
-        call("patchelf", "--set-soname", new_soname, dst_path)
+
+        # If cibuildwheel was called without activating its environment, the `bin`
+        # directory will not be on the PATH.
+        cibw_bin = Path(sys.executable).parent
```
henryiii: This is not guaranteed to be present. On conda-forge, patchelf comes from the feedstock, not our install - it's probably in the same bin dir, but I'd rather not force it. So I think this should revert (or use `shutil.which`) if `cibw_bin / "patchelf"` doesn't exist.

joerick: I did a little spelunking through pypa/installer and [found](https://github.com/pypa/installer/blob/3d1d1c416113a2c9e90c814d238fd21e28cf143f/src/installer/__main__.py#L74C34-L74C34) that it installs scripts into `sysconfig.get_path('scripts')` - that might be more reliable than what we have here. I suppose that still wouldn't address the conda issue though.

So, maybe the most reliable thing would be to do-

```python
patchelf_bin = shutil.which(
  'patchelf',
  path=sysconfig.get_path('scripts') + os.pathsep + os.environ['PATH']
)
```

henryiii: You could do:

```python
patchelf_bin = Path(sysconfig.get_path('scripts')) / "patchelf"
if not patchelf_bin.exists():
    find_patchelf = shutil.which("patchelf")
    if find_patchelf is None:
        # raise some error
    patchelf_bin = Path(find_patchelf)
```

Or the suggestion above, which maybe is a bit better, but it still could return None if not found, so that needs to be handled.

henryiii: Ahh, what you put looks fine, though I do think I like `sysconfig.get_path('scripts')` a bit better.

mhsmith: Thanks, I've gone with Joe's suggestion.

henryiii:
FYI, if it helps at all, currently we just read the groups out and pass them as requirements, but we can probably move to using the `--groups` option at some point now that pip and uv support it.

mhsmith:
There are a few more things I want to deal with, but they're all connected to testing wheels, so I think I'll put them in a separate PR.

henryiii:
Should we consider this the start of 3.2?

mhsmith:
I'm not exactly sure what the version number policy is. There's nothing backward incompatible about this PR, but it does add a new platform / Python version pair.

joerick:
> **Examples test run**
> 
> Branch: [example-config-test---android-2025-08-2025-08-30T18-48-50](https://github.com/pypa/cibuildwheel/tree/example-config-test---android-2025-08-2025-08-30T18-48-50)
> 
> |---|---|---|
> | github | `examples/github-deploy.yml` | [Build](https://github.com/pypa/cibuildwheel/actions?query=branch%3Aexample-config-test---android-2025-08-2025-08-30T18-48-50) |
> 

joerick (approved):
Looks good to me. I'll have a look at the other PRs regarding the versioning question...

joerick:
The current PR queue feels like we're heading to 3.2 at this point.

PR #2574, opened 2025-08-26
Add documentation for test dependency groups on mobile platforms.

freakboy3742:
The documentation for test groups doesn't list configuration options for iOS or Android. 

It turns out that because of the way `test-groups` is handled (by rolling dependencies in to `test-requires`), it works fine on both iOS and Android, and the settings files are already populated. I can only assume this was an oversight on my part when I wrote the iOS docs, which @mhsmith inherited by doing a "search for _IOS and add the same change for Android" when he added Android support.

[This pull request](https://github.com/freakboy3742/pyspamsum/pull/78) proves that `test-groups` works on iOS and Android with cibuildwheel 3.1.4.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -1576,7 +1576,7 @@ tests. This can be used to avoid having to redefine test dependencies in
 `test-requires` if they are already defined in `pyproject.toml`.
 
 Platform-specific environment variables are also available:<br/>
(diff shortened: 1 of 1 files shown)


henryiii:
There shouldn’t be any problem switching to `-—group` either, now that pip and uv support it?

freakboy3742:
> There shouldn’t be any problem switching to `-—group` either, now that pip and uv support it?

Do you mean switching to using `--group` explicitly, rather than the "roll test-depends into the test-requires" approach that is currently being used? If that's the case, then yes, it should be possible for pip on iOS. uv doesn't support iOS at this time, so it's a moot point there.

PR #2575, opened 2025-08-26
Android test updates

mhsmith:
* Use Android environment markers when installing the test environment – see [pybind11#5809](https://github.com/pybind/pybind11/pull/5809#issuecomment-3215219502).

* Pass the `build-verbosity` setting through to the test runner.

* I thought I'd need to add support for `test-groups`, but it turns out this already works ([#2574](https://github.com/pypa/cibuildwheel/pull/2574)).

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -578,7 +578,7 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
         site_packages_dir,
         f"{wheel}{state.options.test_extras}",
         *state.options.test_requires,
-        env=state.build_env,
+        env=state.android_env,
     )
 
     # Copy test-sources.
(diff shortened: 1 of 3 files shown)


mhsmith:
@freakboy3742: The iOS tests have failed several times in the last 2 days, but only on macos-14. I don't see any obvious explanation in the logs.

freakboy3742:
> @freakboy3742: The iOS tests have failed several times in the last 2 days, but only on macos-14. I don't see any obvious explanation in the logs.

As far as I can make out, it's a timeout launching the simulator. Launch times on macOS-14 have always significantly slower than macOS-15 - it's not clear why, but possibly due to a change/improvement in Xcode 16. 

The prevalence of that mode of failure will be exacerbated by the way we're launching the simulator (i.e., the 2-thread parallelised build approach). My hope is that when I push out new support packages that use the single-threaded approach, we can move back to macOS-15, which should make this sort of failure much less common.

joerick (approved):

PR #2586, opened 2025-09-10
feat: support uv with iOS

henryiii:
Now supported in the latest uv release.


.github/workflows/test.yml
```diff
@@ -119,6 +119,8 @@ jobs:
         allow-prereleases: true
 
     - uses: astral-sh/setup-uv@v7
+      with:
+        version: "0.9.9"
```
henryiii:
```suggestion
```

PR #2589, opened 2025-09-11
Make the `{project}` placeholder available to `repair-wheel-command`

arcondello:
Closes https://github.com/pypa/cibuildwheel/issues/2588
Closes https://github.com/pypa/cibuildwheel/issues/1931

Let me know if you'd prefer a different approach for testing, the change I made amounts to a smoke test.

### DIFF
--- a/test/test_custom_repair_wheel.py
+++ b/test/test_custom_repair_wheel.py
@@ -80,7 +80,7 @@ def test_repair_wheel_command_structure(tmp_path, repair_command):
         dest_dir = Path(sys.argv[2])
 
         dest_dir.mkdir(parents=True, exist_ok=True)
-        shutil.copy(wheel, dest_dir / "spam-repaired.whl")
+        shutil.copy(wheel, dest_dir / "spamrepaired-0.0.1-py-none-any.whl")
     """)
 
     # Combined test for repair wheel command formats (plain, {package}, {project})
@@ -90,8 +90,9 @@ def test_repair_wheel_command_structure(tmp_path, repair_command):
         project_dir,
         add_env={
             "CIBW_REPAIR_WHEEL_COMMAND": repair_command,
+            "CIBW_ARCHS": "native",
         },
         single_python=True,
     )
 
-    assert result == ["spam-repaired.whl"]
+    assert result == ["spamrepaired-0.0.1-py-none-any.whl"]


arcondello:
Oh wow, totally missed that one. Updated

henryiii:
I'm a little surprised we don't run the command from the project directory on Linux.

joerick:
> I'm a little surprised we don't run the command from the project directory on Linux.

This is what's confusing me also... I'm 90% sure that we do...! But I think this can still be useful so I'm happy to merge it.

On testing, it appears that the previous test did actually assert that the cwd was the project directory previously. Could we make sure that we don't lose that too? It might be best to revert the test change and add another test, perhaps on just a single python version, that runs the script with the `{project}` placeholder.

arcondello:
> On testing, it appears that the previous test did actually assert that the cwd was the project directory previously. Could we make sure that we don't lose that too? It might be best to revert the test change and add another test, perhaps on just a single python version, that runs the script with the {project} placeholder.

Sure, I can do that

arcondello:
As an alternative, if the intent is always to run the `repair-wheel-command` in the project directory, I can change the PR to update the documentation to specify that. IMO the explicit approach gives more flexibility in the future, but I am happy either way.

joerick:
> As an alternative, if the intent is always to run the `repair-wheel-command` in the project directory, I can change the PR to update the documentation to specify that.

Yes, that is the intention. I doubt that will change - the other commands like before-build etc. also run from the project dir (which is the cwd when cibuildwheel starts). Logic with relative paths gets pretty strange if we change this rule!

Happy to see this noted in the docs.

Also, I think it might be nice for consistency for `{package}` to be expanded in repair-wheel-command as well. If you have time :)

arcondello:
> Happy to see this noted in the docs.

Will do

> Also, I think it might be nice for consistency for {package} to be expanded in repair-wheel-command as well. If you have time :)

Sure, I can do that.

Likely will be able to update the PR tomorrow

arcondello:
Ok updated, sorry for the delay. Though I fear my pattern matching for the tests is falling short and that something more sophisticated would be better.

arcondello:
I dont think the readthedocs error is related to this PR? but if it is let me know

PR #2590, opened 2025-09-11
Update Android test-command handling

### LINKED ISSUE #2611: CMake cannot found Python Development.Module for Android with Python 3.14 (3.13 works fine)
### Description

I'm trying to build wheels in GitHub Action for Android with Python 3.14 (newly supported by `cibuildhwheel` 3.2). The CMake program fails to find Python modules with `cp314-android` while `cp313-android` works fine.

The CMake error log (see https://github.com/metaopt/optree/actions/runs/18182914758/job/51772310290?pr=242):

```txt
  CMake Error at /tmp/cibw-run-e6ejroha/cp314-android_arm64_v8a/venv/lib/python3.14/site-packages/cmake/data/share/cmake-4.1/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
    Could NOT find Python (missing: Development.Module) (found suitable version
    "3.14.0", minimum required is "3.8")
  
        Reason given by package: 
            Development: Cannot find the directory "/tmp/cibw-run-e6ejroha/cp314-android_arm64_v8a/venv/include/python3.14"

  Call Stack (most recent call first):
    /usr/local/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
    /usr/local/share/cmake-3.31/Modules/FindPython/Support.cmake:4002 (find_package_handle_standard_args)
    /usr/local/share/cmake-3.31/Modules/FindPython.cmake:631 (include)
    /tmp/cibw-run-bewfqrm_/cp314-android_arm64_v8a/venv/lib/python3.14/site-packages/pybind11/share/cmake/pybind11/pybind11NewTools.cmake:54 (find_package)
    /tmp/cibw-run-bewfqrm_/cp314-android_arm64_v8a/venv/lib/python3.14/site-packages/pybind11/share/cmake/pybind11/pybind11Common.cmake:204 (include)
    /tmp/cibw-run-bewfqrm_/cp314-android_arm64_v8a/venv/lib/python3.14/site-packages/pybind11/share/cmake/pybind11/pybind11Config.cmake:257 (include)
    CMakeLists.txt:261 (find_package)
```

Note that I have tried both CMake 3.31 in the build container and the latest CMake 4.x from PyPI. Either is failed with the above error.

### Build log

https://github.com/metaopt/optree/actions/runs/18182914758/job/51772310290?pr=242

### CI config

https://github.com/metaopt/optree/pull/242/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R210-R212

mhsmith:
In https://github.com/python/cpython/pull/138805 I'm updating the Android testbed so it can take all Python command line options, not just `-c` and `-m`. However, this will be slightly backward incompatible for cibuildwheel, because the `-c` or `-m` must be moved after the `--` separator on the testbed command line. So this PR will need to remain a draft until the next releases of Python 3.13 and 3.14, which are currently scheduled for October 7.

* Fixes #2611

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -619,7 +619,7 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
     test_args = shlex.split(test_command)
     if test_args[0] in ["python", "python3"] and any(arg in test_args for arg in ["-c", "-m"]):
         # Forward the args to the CPython testbed script. We require '-c' or '-m'
-        # to be in the command, because without those flags, the testbed script 
+        # to be in the command, because without those flags, the testbed script
         # will prepend '-m test', which will run Python's own test suite.
         del test_args[0]
     elif test_args[0] in ["pytest"]:


mhsmith:
This PR is ready for review.

cibuildwheel/platforms/android.py
```diff
@@ -617,8 +617,8 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
 
     # Parse test-command.
     test_args = shlex.split(test_command)
-    if test_args[:2] in [["python", "-c"], ["python", "-m"]]:
-        test_args[:3] = [test_args[1], test_args[2], "--"]
+    if test_args[0] == "python" and any(arg in test_args for arg in ["-c", "-m"]):
```
henryiii: Python also supports merged short args, like `python3 -Om calendar`. Maybe a leading `python` should be stripped regardless of what follows? Also, what about `python3`?

Also, I think this would read nicely with pattern matching:

```python
    in_test_args = shlex.split(test_command)
    match in_test_args:
        case ["python" | "python3", *test_args]:
            pass
        case ["pytest", *_]:
            # We transform some commands into the `python -m` form, but this is deprecated.
            msg = (
                f"Test command {test_command!r} is not supported on Android. "
                "cibuildwheel will try to execute it as if it started with 'python -m'. "
                "If this works, all you need to do is add that to your test command."
            )
            log.warning(msg)
            test_args = ["-m", *in_test_args]
        case _:
            msg = (
                f"Test command {test_command!r} is not supported on Android. "
                f"Command must begin with 'python' and contain '-m' or '-c'."
            )
            raise errors.FatalError(msg)
```

mhsmith: > Python also supports merged short args, like `python3 -Om calendar`.

That's a good point, but the arguments are processed by the [cpython test script](https://github.com/python/cpython/blob/v3.14.0/Android/android.py#L567-L568), which, if it doesn't find a literal `-m` or `-c`, will insert `-m test` at the start of the command. This would be very confusing for cibuildwheel users. 

The script could be improved to detect merged short args, but then this PR would be blocked for 2 months until the next Python release.

This PR also fixes a regression in cibuildwheel 3.2.1, which updated the Android section of build-platforms.toml to Python 3.14.0 (which has the new version of the test script), but didn't include the cibuildwheel side of the changes. This has [broken Android testing on Python 3.14](https://github.com/metaopt/optree/actions/runs/18610782532/job/53068566623?pr=247). 

To make sure any similar problems are caught in the future, I’ve updated `test_expected_wheels` so it not only builds on all Python versions, but also tests on them.

> Also, what about `python3`?

I've updated the code to accept that.

mhsmith:
To avoid merge conflicts, I've also included a fix for #2611. 

@freakboy3742: `_cross_venv_test_android.py` in this PR, which validates that fix, could eventually be migrated into a test suite for [xvenv](https://github.com/beeware/xbuild) (https://github.com/beeware/xbuild/issues/21).

cibuildwheel/platforms/android.py
```diff
@@ -617,8 +617,8 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
 
     # Parse test-command.
     test_args = shlex.split(test_command)
-    if test_args[:2] in [["python", "-c"], ["python", "-m"]]:
-        test_args[:3] = [test_args[1], test_args[2], "--"]
+    if test_args[0] in ["python", "python3"] and any(arg in test_args for arg in ["-c", "-m"]):
+        del test_args[0]
```
joerick, line 624: This code looks a little odd, let's comment that.
```suggestion
        # Pass the args to be parsed by the cpython testbed script. We only do this
        # when '-c' or '-m' is found in the command, because without those flags, the
        # testbed script will prepend '-m test', which isn't what the user wants.
        del test_args[0]
```

mhsmith: Thanks; I've added the comment and rephrased it slightly.

joerick: thanks!

PR #2597, opened 2025-09-17
Add GraalPy 25 and remove workarounds for that version.

timfel:


joerick:
Hm, I still see lots of workarounds...! I had expected them to go for some reason, maybe because of how the comments were worded. I'm wondering - did you expect to have a fix release out for GP24 or did I misunderstand?

joerick:
But yeah, otherwise LGTM!

timfel:
> Hm, I still see lots of workarounds...! I had expected them to go for some reason, maybe because of how the comments were worded. I'm wondering - did you expect to have a fix release out for GP24 or did I misunderstand?

I had hoped to have a fix for GP24, but we couldn't get it out and for reasons I have no control over there will not be another release off the GP24 branch, so the workarounds have to stay for that version. Once we decide to drop it, we can remove them entirely. I wouldn't mind dropping it before we release the next version in the spring of next year, i just didn't want to do it without any overlap time in case someone did use a build spec with gp24 in it.

timfel:
Hm, both the iOS as well as the windows tests seem flaky. They failed, they passed, they failed two times now... this PR shouldn't affect iOS tests, and the Windows tests seem to fail with network connectivity issues (the latest one running on CPython 3.8 [afaict](https://github.com/pypa/cibuildwheel/actions/runs/17914148505/job/50949894989?pr=2597#step:18:1197), so also not GraalPy related).

joerick:
Yeah, the iOS tests being flaky is a known issue. The windows failure is test_ssl again, which honestly, I have no idea why it's such a problem child. It does hit the network, but then again, every build hits PyPI, and that's the one that has the problem. 

I added a flaky marker, but maybe it needs a pause to get around temporary network issues.

joerick (approved):

PR #2601, opened 2025-09-19
Replace the `macos-13` images with the `macos-15-intel` images (GHA)

agriyakhetarpal:
See https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/; we can replace `macos-13` images with the new `macos-15-intel` runners, which will be phased out in 2027.

The `macos-14` images continue to be used and are not changed here, as `macos-15` continues to be broken for iOS wheel support.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,17 +69,17 @@ jobs:
           python_version: '3.13'
         - os: windows-11-arm
           python_version: '3.13'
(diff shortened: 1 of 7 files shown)


joerick (approved):
Thanks for picking this up. I'm pleased they're continuing to offer intel runners, should reduce the amount of cross-compiling people have to do.

PR #2607, opened 2025-09-29
Resolve issues with macOS-15 runners

freakboy3742:
Due to actions/runner-images#12777, the macOS-15 runner has been unstable for iOS test runs. This was reported to Python as python/cpython#137973; and corrected in python/cpython#138018.

This PR:
* Bumps the iOS support packages to versions that include this fix
* Restores the use of macOS-15 in CI environments
* Removes the documentation note about macOS 15 usage
* Forces the selection of Xcode 16.4 in macOS environments. This *should* be the default, but it appears explicitly selecting the Xcode version can have other effects that improve iOS simulator reliability.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,6 +110,12 @@ jobs:
       if: runner.os == 'Linux'
       uses: docker/setup-qemu-action@v3
 
+    - name: Set up Xcode
+      if: ${{ startsWith(matrix.os, 'macos-15') }}
+      run: |
+        # GitHub recommends explicitly selecting the desired Xcode version:
+        sudo xcode-select --switch /Applications/Xcode_16.4.app
+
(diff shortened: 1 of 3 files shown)


freakboy3742:
Not sure what is going on with the Android tests here - Any ideas @mhsmith?

mhsmith:
Looks like it's OK now, and there are no issues shown at https://status.maven.org/, except with the search function, which we're not using.

I thought it might be a GitHub-focused brown-out connected with [this statement](https://openssf.org/blog/2025/09/23/open-infrastructure-is-not-free-a-joint-statement-on-sustainable-stewardship/), but I can't find any announcement of that, nor any comments about other people having the same problems.

joerick:
Thanks for continuing to work on the stability issues @freakboy3742 !

PR #2617, opened 2025-10-08
doc: include free-threading builds in identifier list

cdce8p:
Update `options.build` in the documentation to include the free-threading identifiers.


joerick:
Thanks! Sidenote, but it's hard to stay on top of this, I'm thinking that this table should really be generated from `build-platforms.toml`.

PR #2629, opened 2025-10-18
Fix formatting of footnote in README.md

msyriac:


joerick:
thanks, i noticed that @mhsmith also fixed this in #2590 so i'd rather merge that

PR #2630, opened 2025-10-19
Allow installation of extras in GitHub action

ofek:
When setting the [build-frontend](https://cibuildwheel.pypa.io/en/stable/options/#build-frontend) to `build[uv]` one must manually make UV globally available on Windows and macOS. This leads to conditional execution as seen in this snippet of real CI configuration:

```yaml
    - name: Download source distribution
      uses: actions/download-artifact@v4
      with:
        name: artifact-sdist
        path: dist

    - name: Install UV
      if: runner.os != 'Linux'
      uses: astral-sh/setup-uv@v6

    - name: Build wheels
      uses: pypa/cibuildwheel@v3.2.1
      with:
        package-dir: dist/${{ needs.sdist.outputs.sdist-name }}
```

Such an option would allow for removing the middle step:

```yaml
    - name: Download source distribution
      uses: actions/download-artifact@v4
      with:
        name: artifact-sdist
        path: dist

    - name: Build wheels
      uses: pypa/cibuildwheel
      with:
        extras: uv
        package-dir: dist/${{ needs.sdist.outputs.sdist-name }}
```

Here's a [PR](https://github.com/ofek/dotslash-python/pull/4) to showcase this change being successfully used.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -17,6 +17,10 @@ inputs:
     description: 'Build a specific wheel only. No need for arch/platform if this is set'
     required: false
     default: ''
+  extras:
+    description: 'Comma-separated list of extras to install'
+    required: false
+    default: ''
 branding:
   icon: package
   color: yellow
@@ -42,6 +46,10 @@ runs:
         from pathlib import Path
         from subprocess import run
 
+        EXTRAS = set(e.strip() for e in "${{ inputs.extras }}".split(",") if e.strip())
+        if sys.platform == "linux":
+            EXTRAS.discard("uv")
+
 
         class EnvBuilder(venv.EnvBuilder):
             def __init__(self):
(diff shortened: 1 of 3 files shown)


action.yml
```diff
@@ -62,30 +73,47 @@ runs:
...
+                try:
+                    os.symlink(path, clean_bin_path / path.name)
+                except OSError:
+                    import shutil
+
+                    shutil.copy2(path, clean_bin_path / path.name)
+        full_path = f"{clean_bin_path}{os.pathsep}{os.environ['PATH']}"
         with open(os.environ["GITHUB_OUTPUT"], "at") as f:
-            f.write(f"cibw-path={cibw_path}\n")
+            f.write(f"updated-path={full_path}\n")
```
joerick, line 91: I wonder, if we're gonna modify PATH within the scope of the action, is there a reason to not simplify a little and add the venv bin dir to the PATH instead?

```python
        builder = EnvBuilder()
        builder.create(venv_path)
        full_path = f"{builder.bin_path}{os.pathsep}{os.environ['PATH']}"
        with open(os.environ["GITHUB_OUTPUT"], "at") as f:
            f.write(f"updated-path={full_path}\n")
```

I believe this would still achieve the objective of #1745 - not contaminating the rest of the action context.

ofek: Installation of `cibuildwheel` may bring in any number of executables via transitive dependencies. Exposing them would lead to 2 issues:

1. The build environment would no longer be hermetic, limiting the prospects of reproducibility.
2. We want `cibuildwheel` (and optionally `uv`) to come first on PATH but users may require specific versions of executables that happen to come with our transitive dependencies. Alternatively, if we append rather than prepend then we couldn't guarantee which of our required executables is selected.

joerick: Executables from transitive dependencies is a fair point. I'm not too concerned about those since any other method of invoking cibuildwheel is gonna have them on path (pipx, uvx etc) but I take the point that more isolated is better.

joerick:
thank you for putting this together @ofek ! A nice improvement for sure.

PR #2636, opened 2025-10-22
Add configuration option for test executor arguments

freakboy3742:
iOS and Android test suites are executed using a testbed app. This app needs to be started, and the test arguments (the value of `test_command`) is passed to that testbed. 

However, the testbed app itself has configuration arguments, and it can sometimes be useful to be able to use these arguments.

This issue has become mildly critical as a result of the recent changes to the `macos-15` test runner. At present, the simulator that is picked by default by the iOS test runner is entirely valid from the perspective of Xcode, but then refuses to launch. 

This is a [known issue with the macOS-15 runner that GitHub is investigating](https://github.com/actions/runner-images/issues/12777); but it's not a once-off issue. When Xcode 16.4 was released, it removed the "iPhone SE" simulator as a default image, and as there was no way to configure the simulator used at runtime, a full update of the iOS support packages and a cibuildwheel release was necessary to resolve the issue.

So - this PR adds a `CIBW_TEST_EXECUTION_ARGS` setting (with related platform-specific alternatives). These arguments are passed to the testbed runner on iOS and Android.

As Android requires explicitly specifying the simulator to use as part of the runner arguments, the value of `CIBW_TEST_EXECUTION_ARGS` is lightly processed. If there's a `--managed` or `--connected` value in `CIBW_TEST_EXECUTION_ARGS`, that value will be used by the testbed runner; if no such argument exists, `--managed maxVersion` will be used by the test runner.

I'm not entirely sure how best to test this; it's in a portion of the code that can't be easily tested without actually starting simulators, and... well... the problem that led to this PR is that we can't use other simulators :-) So - I've gone with testing this empirically through the Github Actions configuration, passing in the necessary arguments so that on GitHub, simulators are specified on iOS and Android x86 simulator. On other platforms, the default behavior will continue to operate.


cibuildwheel/platforms/android.py
```diff
@@ -635,17 +635,28 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
...
+    # By default, run on a testbed managed emulator running the newest supported
+    # Android version. However, if the user specifies a --managed or --connected
+    # test execution argument, that argument takes precedence.
+    if state.options.test_execution_args and (
+        "--managed" in state.options.test_execution_args
+        or "--connected" in state.options.test_execution_args
+    ):
+        simulator_args = []
+    else:
+        simulator_args = ["--managed", "maxVersion"]
```
mhsmith: Android uses the term "emulator", not "simulator".

cibuildwheel/platforms/android.py
```diff
@@ -635,17 +635,28 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
...
         "test",
-        "--managed",
-        "maxVersion",
         "--site-packages",
         site_packages_dir,
         "--cwd",
         cwd_dir,
+        *simulator_args,
         *(["-v"] if state.options.build_verbosity > 0 else []),
+        *(state.options.test_execution_args if state.options.test_execution_args else []),
```
```suggestion
        *(state.options.test_execution_args or []),
```

docs/options.md
```diff
@@ -1669,6 +1669,41 @@ Platform-specific environment variables are also available:<br/>
     CIBW_TEST_ENVIRONMENT: PYTHONSAFEPATH=1
     ```
 
+### `test-execution-args` {: #test-execution-args toml env-var }
+
+> Define additional arguments that will be passed to the command that runs the tests.
```
mhsmith: `test-execution-args` still sounds like they might be arguments to the tests themselves. Would `test-runner-args` be a clearer name?

cibuildwheel/platforms/ios.py
```diff
@@ -658,6 +658,11 @@ def build(options: Options, tmp_path: Path) -> None:
                                 testbed_path,
                                 "run",
                                 *(["--verbose"] if build_options.build_verbosity > 0 else []),
+                                *(
+                                    build_options.test_execution_args
+                                    if build_options.test_execution_args
+                                    else []
+                                ),
```
mhsmith:
```suggestion
                                *(build_options.test_execution_args or []),
```

.github/workflows/test.yml
```diff
@@ -79,9 +79,16 @@ jobs:
         - os: macos-15
           python_version: '3.13'
           test_select: ios
+          # Github Actions macOS-15 runner has disk performance issues; so it
+          # has problems using any simulator that isn't pre-warmed.
+          # Unfortunately, the simulator that it picks by default isn't
+          # pre-warmed. So - we need to explicitly select a simulator.
+          test_execution_args: '--simulator "iPhone 16e,OS=18.5"'
```
mhsmith: Could we make this the default, at least when running on GitHub Actions, so each project doesn't need to specify it? Experience so far has shown that every required bit of manual configuration will trip up a significant number of package developers, even if the documentation is clear enough.

docs/platforms.md
```diff
@@ -233,6 +233,8 @@ machine – for example, if you're building on an ARM64 machine, then you can te
 ARM64 wheel. Wheels of other architectures can still be built, but testing will
 automatically be skipped.
 
+Any arguments specified using [`test-execution-args`](options.md#test-execution-args) will be passed as arguments to the Python script that starts the testbed project. The most common arguments to use will be `--managed minVersion` or `--managed maxVersion`, specifying the use of a managed Android emulator with the minimum or maximum supported Android version; or `--connected <emulator ID>`, specifying the use of an existing booted Android emulator or device.
```
mhsmith: "Serial" is the term used in the Android tools, because for a physical device it is literally the serial number.

```suggestion
Any arguments specified using [`test-execution-args`](options.md#test-execution-args) will be passed as arguments to the Python script that starts the testbed project. The most common arguments to use will be `--managed minVersion` or `--managed maxVersion`, specifying the use of a managed Android emulator with the minimum or maximum supported Android version; or `--connected <serial>`, specifying the use of an existing booted Android emulator or device.
```

docs/platforms.md
```diff
@@ -321,3 +323,7 @@ If tests have been configured, the test suite will be executed on the simulator
 The iOS test environment can't support running shell scripts, so the [`test-command`](options.md#test-command) value must be specified as if it were a command line being passed to `python -m ...`.
 
 The test process uses the same testbed used by CPython itself to run the CPython test suite. It is an Xcode project that has been configured to have a single Xcode "XCUnit" test - the result of which reports the success or failure of running `python -m <test-command>`.
+Any arguments specified using [`test-execution-args`](options.md#test-execution-args) will be passed as arguments to the Python script that starts the testbed project. The most common argument to use will be `--simulator`, which allows the specification of a specific device or iOS version for the test simulator. For example, `test_execution_args = ["--simulator", "iPhone 16e,OS=18.5"]` would specify the use of an iPhone 16e simulator running OS 18.5.
```
mhsmith: This should state what the default arguments are on both Android and iOS.

.github/workflows/test.yml
freakboy3742: The problem is that the "working default" can (and based on recent evidence, does) change somewhat arbitrarily. If a value is hard coded into cibuildwheel, then it needs a cibuildwheel release to fix it. 

Even the fact that an override is needed *at all* is a temporary measure - once GitHub/Apple resolve this hypervisor performance issue, there presumably won't be a need for an override.

docs/options.md
freakboy3742: I picked `test-execution-args` based on [the discussion that lead to this PR](https://github.com/pypa/cibuildwheel/pull/2624#issuecomment-3418619041); but `test-runner-args` would also work. The exact naming isn't a hill I'll die on, and it's easy to change; would be interested in hearing other opinions.

docs/platforms.md
```diff
@@ -233,7 +233,7 @@ machine – for example, if you're building on an ARM64 machine, then you can te
...
 
-Any arguments specified using [`test-execution-args`](options.md#test-execution-args) will be passed as arguments to the Python script that starts the testbed project. The most common arguments to use will be `--managed minVersion` or `--managed maxVersion`, specifying the use of a managed Android emulator with the minimum or maximum supported Android version; or `--connected <emulator ID>`, specifying the use of an existing booted Android emulator or device.
+Any arguments specified using [`test-execution-args`](options.md#test-execution-args) will be passed as arguments to the Python script that starts the testbed project. cibuildwheel will automatically start the testbed project with `--site-packages` and `--cwd` arguments matching your test environment, as well as enabling verbose output with `-v` if [`build_verbosity`](options.md#build_verbosity) is enabled. The most common additional arguments to use will be `--managed minVersion` or `--managed maxVersion`, specifying the use of a managed Android emulator with the minimum or maximum supported Android version; or `--connected <serial>`, specifying the use of an existing booted Android emulator or device.
```
mhsmith: If we're going into this level of detail, I guess we should link to the [script documentation](https://github.com/python/cpython/blob/main/Android/README.md#testing) so people can find out what options are available, and similarly for iOS.

.github/workflows/test.yml
mhsmith: An alternative idea – though I guess this might need to be done on the CPython side – if we need to use a simulator which is warmed up, can we detect which simulators are warmed up and use them by default? That would give better performance even after the current issue is fixed,

Again, this could be a CI-specific behavior, but there is precedent for doing that to work around issues which are likely to keep happening, such as the [Android emulator permissions](https://github.com/python/cpython/blob/v3.14.0/Android/android.py#L284-L292).

docs/platforms.md
mhsmith: It should also say what the default emulator/simulator is, so the user can judge whether an override is necessary.

docs/options.md
joerick: I had considered `test-runner-...` too, but I didn't like it because to my mind 'pytest' is a test runner, and we're not talking about that. Ideally, the name would evoke more 'environment' or 'context' in its flavour, but nothing fit that didn't already have a meaning. So I landed on 'execution', which feels higher-level than runner(?).

docs/options.md
```diff
@@ -1669,6 +1669,41 @@ Platform-specific environment variables are also available:<br/>
...
+
+A list of arguments that will be used by the test runner when running tests. This is used by environments where the execution environment can be customized. For example, mobile platforms will use these arguments to control the execution of the testbed application that is used to run tests.
+
+This option will be ignored on platforms that do not have a separate test runner.
+
+Platform-specific environment variables are also available:<br/>
+`CIBW_TEST_EXECUTION_ARGS_MACOS` | `CIBW_TEST_EXECUTION_ARGS_WINDOWS` | `CIBW_TEST_EXECUTION_ARGS_LINUX` | `CIBW_TEST_EXECUTION_ARGS_ANDROID` |`CIBW_TEST_EXECUTION_ARGS_IOS` | `CIBW_TEST_EXECUTION_ARGS_PYODIDE`
```
```suggestion
`CIBW_TEST_EXECUTION_ARGS_ANDROID` |`CIBW_TEST_EXECUTION_ARGS_IOS`
```

cibuildwheel/platforms/android.py
```diff
@@ -635,17 +635,28 @@ def test_wheel(state: BuildState, wheel: Path) -> None:
...
+    # Android version. However, if the user specifies a --managed or --connected
+    # test execution argument, that argument takes precedence.
+    if state.options.test_execution_args and (
+        "--managed" in state.options.test_execution_args
+        or "--connected" in state.options.test_execution_args
+    ):
+        emulator_args = []
+    else:
+        emulator_args = ["--managed", "maxVersion"]
+
```
joerick, line 650: This is fine, but I'm just wondering if it would be simpler to merely document a default value for the testbed args on android, and let the user decide how they wanna roll there... what happens if `--managed` or `--connected` are not set?

.github/workflows/test.yml
joerick: I wouldn't be opposed to putting some CI-specific code into cibuildwheel to smooth over temporary issues like this, especially if there's a user-controllable escape hatch to override our probably-helpful defaults.

We'll need to adapt to CI images for the sake of our CI anyway, so might as well roll that into the product and let our users benefit.

cibuildwheel/options.py
```diff
@@ -761,6 +762,12 @@ def _compute_build_options(self, identifier: str | None) -> BuildOptions:
                 msg = f"Malformed environment option {test_environment_config!r}"
                 raise errors.ConfigurationError(msg) from e
 
+            test_execution_args = shlex.split(
+                self.reader.get(
+                    "test-execution-args", option_format=ListFormat(sep=" ", quote=shlex.quote)
+                )
+            )
+
```
joerick: May I suggest a change here - I'd rather this option was named something like `test-execution` or `test-execution-config` and was parsed with `parse_key_value_string()`. So an example config might be 

 
```toml
[tool.cibuildwheel.ios]
test-execution = { args = ["--simulator", "iPhone 16e,OS=18.5"] }
# or, if we wanna get fancy (we probably don't want to show this in docs)...
test-execution.args = ["--simulator", "iPhone 16e,OS=18.5"]
```
or
```bash
CIBW_TEST_EXECUTION_IOS="args: --simulator 'iPhone 16e,OS=18.5'"
```

Reason being - I think we might want to use such an option in future for other platform-specific purposes, like running tests in a separate Docker container on Linux.

[build-frontend](https://cibuildwheel.pypa.io/en/stable/options/#build-frontend) would be a good reference for [how to parse the option](https://github.com/mhsmith/cibuildwheel/blob/0495c1e4dbc90250734c1c6a784eed5e7604466a/cibuildwheel/options.py#L775-L788). Happy to help make this change, if you like!

docs/options.md
mhsmith: Here are some other options:
* "testbed" is the term we've used for both Android and iOS, but that wouldn't be appropriate for a Docker container, as mentioned in the linked discussion.
* "container" is what Docker would call it, but that implies a certain level of isolation which we don't necessarily provide.
* "launcher" seems like a fairly generic way of saying "the thing that starts (but doesn't "run") the test suite".

On the other hand, if this option could have additional sub-keys in the future as suggested [below](https://github.com/pypa/cibuildwheel/pull/2636#discussion_r2462835093), then maybe a general name like "execution" is OK.

cibuildwheel/platforms/android.py
mhsmith: Exactly one of them must always be set, or the testbed will give an error.

joerick: Ok cool, in that case this approach seems good to me.

cibuildwheel/options.py
freakboy3742: Seems like a reasonable suggestion (especially with potential Docker usage on the horizon); I'll take a look at this, and get back to you if I hit any problems.

.github/workflows/test.yml
freakboy3742: > can we detect which simulators are warmed up and use them by default?

The only suggestion I've seen to date involves a command that "succeeds on warm simulators, hangs on non-warm simulators".

> I wouldn't be opposed to putting some CI-specific code into cibuildwheel to smooth over temporary issues like this, especially if there's a user-controllable escape hatch to override our probably-helpful defaults.

Adding a (currently known to work) default value when on GitHub Actions + macOS 15 is certainly a possibility, and I agree that it would definitely be preferable for this to work out of the box without configuration. As I mentioned previously - my hesitation is that we don't know what Github will do to fix the problem, and on what timeframe; and once it's baked into cibuildwheel itself, it takes a full release cycle to "unbake" it. 

I guess taking a similar approach to Android (i.e., "here's a default; if you provide something manually we'll ignore that default") is an option, as it at least provides a window for an immediate workaround if it turns out the update *is* incompatible. I'll take a look to see if I can make that work.

freakboy3742:
@joerick I've just pushed an update that includes:

* Moving to `test-execution = { args = [...] }` as the syntax for configuration
* Adding a default simulator on iOS when running under Github Actions on a macos-15 runner.
* Links to the iOS and Android test runner documentation.

I've kept the explicit use of `TEST_EXECUTION` in the GitHub Actions CI configuration, as it's a viable "live test" of the new functionality.

docs/platforms.md
mhsmith: I think we're still missing this for Android.

freakboy3742: Added.

cibuildwheel/platforms/ios.py
```diff
@@ -653,11 +654,35 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                                and arch == "arm64"
+                                and not any(
+                                    arg.startswith("--simulator") for arg in test_execution_args
+                                )
+                            ):
+                                test_execution_args = [
+                                    "--simulator",
+                                    "iPhone 16e,OS=18.5",
+                                    *test_execution_args,
+                                ]
```
joerick: It would be nice if the user could pass `test-execution = "args: --simulator auto"` to activate the `select_simulator_device(platform)` function from the test script. But I suppose that's a CPython change. Not essential, this is okay, it's probably a temporary issues anyway.

docs/options.md
```diff
@@ -1672,6 +1672,41 @@ Platform-specific environment variables are also available:<br/>
     CIBW_TEST_ENVIRONMENT: PYTHONSAFEPATH=1
     ```
 
+### `test-execution` {: #test-execution toml env-var }
+
+> Controls how the tests will be executed.
+
+In simple desktop environments, the test suite will be executed using the command defined by [`test-command`][test-command]. However, on some platforms, the test is executed using a test runner; the `test-execution` setting is used to control that test execution environment. The value of the setting is a dictionary that can contain additional keys, depending on the platform being targeted.
+
+On mobile platforms (iOS and Android), a testbed project is used to run the tests. The `test-execution` setting can define an `args` key that defines additional arguments that will be used when starting the testbed project.
```
joerick: I think we can be a bit more direct here, the generalities aren't relevant to the end-user.

```suggestion
On desktop environments, the tests are executed on the same machine/container as the wheel was built. However on Android and iOS, the tests are run inside a virtual machine – a simulator or emulator – representing the target.

For these embedded platforms, a testbed project is used to run the tests. The `test-execution` setting can define an `args` key that defines additional arguments that will be used when starting the testbed project.
```

joerick (commented):
Thanks for these latest changes - just a couple things...

docs/options.md
joerick: Yeah, if I'm completely honest, I'm still not in love with 'test-execution'. Seeing it in the sidebar, it really gives no hints about what it does-

image

I'm now coming back around to `test-target`, `test-host` or `test-runtime`. Actually `test-target` looks quite sensible in the options sidebar-

image

I'm curious what resonates with you? bearing in mind I could also imagine this being used to spec a docker image, or set some pyodide/node options.

(apologies to go around the houses with this @freakboy3742 !)

cibuildwheel/platforms/ios.py
freakboy3742: Yes - adding `auto` would require an upstream CPython change; and "auto" is already the default behavior for the iOS test runner. 

This is only needed as a workaround for the current set of GitHub Actions weirdness - and the 2 month release cadence of CPython definitely doesn't lend itself to hard-coding defaults like this.

docs/options.md
freakboy3742: > I'm curious what resonates with you? bearing in mind I could also imagine this being used to spec a docker image, or set some pyodide/node options.

Of those options, `test-runtime` makes the most sense to me. 

`test-target` sounds like "where do I want to run the tests?" rather than "how will I run this test on a target?" (and similar for `test-host`, with the added benefit of the confusion of whether the "host" is the architecture running the build, or the architecture you're building for).`test-runtime` sounds like the one that is most directly tied to "what runtime will I use for the tests?" - which is essentially what we're configuring, whether that's iOS, Android or Docker.

> (apologies to go around the houses with this @freakboy3742 !)

No problems at all - it's an easy set of changes to make, and I'd rather go round a few times now and get it right than have to walk back a change in a couple of weeks/months.

docs/options.md
```diff
@@ -1672,6 +1672,41 @@ Platform-specific environment variables are also available:<br/>
...
+    ```
+
+!!! tab examples "Environment variables"
+
+    ```yaml
+    # Run the tests on an iPhone 16e simulator running iOS 18.5.
+    CIBW_EXECUTION_IOS: "args: --simulator='iPhone 16e,OS=18.5'"
+
+    # Run the Android tests on the minimum supported Android version.
+    CIBW_EXECUTION_ANDROID: "args: --managed minVersion"
```
mhsmith:
```suggestion
    CIBW_TEST_EXECUTION_IOS: "args: --simulator='iPhone 16e,OS=18.5'"

    # Run the Android tests on the minimum supported Android version.
    CIBW_TEST_EXECUTION_ANDROID: "args: --managed minVersion"
```

docs/options.md
mhsmith: `test-runtime` seems like a reasonable name to me as well.

joerick: Thanks both for your input! Okay, that feels like consensus, let's run with `test-runtime` :)

mhsmith (approved):

PR #2650, opened 2025-11-05
feat(android): add Rust cross-compilation setup for Android environment

ririv:
Like platform `windows`

https://github.com/pypa/cibuildwheel/blob/c53e541c2de29c8bb0a8f6e3ff13d96046e8000c/cibuildwheel/platforms/windows.py#L206-L232

Add Rust cross-compilation setup for Android environment, it can fix the cross-compilation problem for the rust project built with setuptools-rust


ririv:
For Rust projects built with Maturin, it currently does not work. Maturin has some cross-compilation issues on Android that still need to be fixed. 

For details, please see https://github.com/PyO3/maturin/issues/2810

joerick:
This looks nice and neat! Is there an example of it working in a real project that you've been testing it with?

ririv:
I want to test， but I did not know how to run this without the github action，if you know how to do, you can tell me，and I will test it.

This commit is a modification I made based on the actual missing environment configuration built on Android using cbuildwheel.

Of course, I manually configured it externally in a non-isolated environment, and after the configuration, I built the wheel correctly and it can run on Android

joerick:
You can test a project using something like this in a github workflow-

```yaml
      - name: Build wheels
        uses: ririv/cibuildwheel@main
```

ririv:
Thanks, I have tested it, and made some new commits. Now, it works.
- The `CARGO_TARGET_<triple>_LINKER` must be set. Otherwise, it will use the host's linker `cc`, which will result in an error.
- The `PYO3_CROSS_LIB_DIR` must be set to link against `libpython3.x.so` explicitly. 
   See: https://peps.python.org/pep-0738/#linkage and https://pyo3.rs/v0.27.1/building-and-distribution.html#cross-compiling

The `PYO3_CROSS_LIB_DIR` environment variable should be set here rather than requiring users to specify it manually, for the following reasons:
- PYO3 has become the de facto standard for Python extension modules built with Rust. Both setuptools-rust and maturin are tools developed by the PYO3 team specifically to serve PYO3.
- The value set for `PYO3_CROSS_LIB_DIR` requires the directory path of target_python. This path is already configured within cbuildwheel's `android.py` but cannot be directly accessed externally. External configuration would necessitate manually constructing the path.
- PYO3's documentation for Android cross-compilation is insufficiently detailed. If users attempt to set this variable manually, they are likely to become confused and unsure how to proceed.

ririv:
https://github.com/ririv/android-wheels/blob/main/.github/workflows/cbuildweel-repair-test.yml
This is the workflow file. It performed some conversions.
For the maturin project, it was converted to a setuptools-rust project. This is because maturin has some bugs in Android cross-compilation. I've submitted a PR to fix it, but the maturin maintainer hasn't responded yet.

Here are some logs:
- pydantic-core | maturin convert to setuptools-rust
   https://github.com/ririv/android-wheels/actions/runs/19223626338/job/54946144941
- tiktoken | based on setuptools-rust
   https://github.com/ririv/android-wheels/actions/runs/19223649349/job/54946212372

joerick:
Looks good to me! @mhsmith what do you think?

mhsmith:
Sorry for the delay, I'll look at this as soon as I can.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+        if env["CARGO_BUILD_TARGET"] != cargo_target:
+            log.notice("Not overriding CARGO_BUILD_TARGET as it has already been set")
+        # No message if it was set to what we were planning to set it to
+    elif cargo_target:
+        cargo_target_linker_env_name = (
+            f"CARGO_TARGET_{cargo_target.upper().replace('-', '_')}_LINKER"
+        )
+        log.notice(
+            f"Setting CARGO_BUILD_TARGET={cargo_target} and {cargo_target_linker_env_name} for cross-compilation"
+        )
```
mhsmith: This logging is unnecessary, as the entire environment is already logged at line 398.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+) -> None:
+    # All Python extension modules must therefore be explicitly linked against libpython3.x.so when building for Android.
+    # See: https://peps.python.org/pep-0738/#linkage
+    # For projects using PyO3, this requires setting PYO3_CROSS_LIB_DIR to the directory containing libpython3.x.so.
+    # See: https://pyo3.rs/v0.27.1/building-and-distribution.html#cross-compiling
+    if env.get("PYO3_CROSS_LIB_DIR"):
+        log.notice("Not overriding PYO3_CROSS_LIB_DIR as it has already been set")
+    else:
+        env["PYO3_CROSS_LIB_DIR"] = str(python_dir / "prefix" / "lib")
+        log.notice("Setting PYO3_CROSS_LIB_DIR for PyO3 cross-compilation")
```
mhsmith: This logging is unnecessary.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+    python_configuration: PythonConfiguration,
+    env: MutableMapping[str, str],
+) -> None:
+    cargo_target = android_triplet(python_configuration.identifier)
+    call("rustup", "target", "add", cargo_target)
+
+    # CARGO_BUILD_TARGET is the variable used by Cargo and setuptools_rust
+    if env.get("CARGO_BUILD_TARGET"):
+        if env["CARGO_BUILD_TARGET"] != cargo_target:
+            log.notice("Not overriding CARGO_BUILD_TARGET as it has already been set")
```
mhsmith: I can't think of a plausible reason for the user to override any of these variables, so let's set them all unconditionally. This will make both functions so much shorter, that they might as well be combined into a single `setup_rust_cross_compile` function.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+        env[f"{cargo_target_linker_env_name}"] = env["CC"]
+    else:
+        log.warning(f"Unable to configure Rust cross-compilation for architecture {cargo_target}")
+
+
+def setup_PYO3_cross_compile(
+    python_dir: Path,
+    env: MutableMapping[str, str],
+) -> None:
+    # All Python extension modules must therefore be explicitly linked against libpython3.x.so when building for Android.
```
```suggestion
    # All Python extension modules must be explicitly linked against libpython3.x.so when building for Android.
```

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
     return android_env
 
 
+def setup_rust_cross_compile(
+    python_configuration: PythonConfiguration,
+    env: MutableMapping[str, str],
+) -> None:
+    cargo_target = android_triplet(python_configuration.identifier)
+    call("rustup", "target", "add", cargo_target)
```
mhsmith: This will fail if `rustup` isn't installed, so that must be handled. 

Even if it is installed, it would be a waste of time to download Android Rust files for packages that don't actually use Rust. How long will this take? If it's more more than a couple of seconds, I don't think that's acceptable.

Is the `rustup target add` command still necessary even if CARGO_BUILD_TARGET is set? If so, Is there some way we could trigger the command only for packages that use Rust? 

For example, if calling `rustup` is part of the build process (I don't know whether it is), then we could add a `rustup` script to the environment's `bin` directory, and have it run `rustup target add` before forwarding the command to the real `rustup`. I was thinking of doing something similar to install the [Android Fortran compiler](https://github.com/mzakharo/android-gfortran/releases) on demand for those packages that attempt to use it.

cibuildwheel/platforms/android.py
mhsmith: Or even just `setup_rust`, since everything in this file is cross compiling by definition.

cibuildwheel/platforms/android.py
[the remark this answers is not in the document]
mhsmith: Rather than checking for the environment variables directly, it would be better to actually build and run a simple Rust project, just as we currently do for C++ in `test_android.py::test_libcxx`.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+
```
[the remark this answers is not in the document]
mhsmith: I'll merge this into the previous thread about this issue.

cibuildwheel/platforms/android.py
mhsmith: Copilot adds:

> the Windows implementation (lines 206-232 in windows.py) does not call rustup at all - it only sets environment variables. This inconsistency should be evaluated to determine if rustup installation is actually necessary

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +403,47 @@ def setup_android_env(
...
+    elif cargo_target:
+        cargo_target_linker_env_name = (
+            f"CARGO_TARGET_{cargo_target.upper().replace('-', '_')}_LINKER"
+        )
+        log.notice(
+            f"Setting CARGO_BUILD_TARGET={cargo_target} and {cargo_target_linker_env_name} for cross-compilation"
+        )
+        env["CARGO_BUILD_TARGET"] = cargo_target
+        # CC has already been set by calling android.py (it calls android-env.sh)
+        env[f"{cargo_target_linker_env_name}"] = env["CC"]
```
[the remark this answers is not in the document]
mhsmith: Unnecessary: this is part of the basic contract of android-env.sh.

ririv:
Thanks for the suggestions. I've updated the PR to address the feedback:

*   **Lazy Rust Setup**: I implemented the shim/wrapper approach for `cargo` and `rustup` as suggested. This ensures zero overhead for non-Rust projects, as the Android target is only installed when a Rust command is actually triggered. I used absolute paths in the shims to avoid recursion.
*   **Platform Alignment**: Cleaned up the redundant `log.notice` calls and removed the defensive `CC` check to follow the `android-env.sh` contract.
*   **New Integration Test**: Added `test_setuptools_rust` to `test_android.py`. It builds a minimal Rust extension and runs it on the emulator to verify the whole flow.

ririv:
Since PyO3/maturin#2825 has been merged into main, we can now use cibuildwheel to test Rust projects built with maturin. However, as a new version hasn't been released yet, we have to build maturin from source for now.

This is the GItHub Action script:

https://github.com/ririv/android-wheels/blob/main/.github/workflows/test-cbuildwheel-repair-maturin-source.yml

This is the logs:

- zstandard |  non-Rust project
   https://github.com/ririv/android-wheels/actions/runs/20427696018
- tiktoken  | based on setuptools-rust
   https://github.com/ririv/android-wheels/actions/runs/20427678027

---

Sorry, there are still some issues with the maturin integration: it's building for Linux instead of Android. This might be a problem with the `build` module; I will investigate and fix it.

- pydantic_core | based on maturin
   https://github.com/ririv/android-wheels/actions/runs/20427674411
   
However, we can still use a conversion tool to transform maturin projects into setuptools-rust projects and successfully build Android wheels.

- pydantic_core | maturin convert to setuptools-rust
   https://github.com/ririv/android-wheels/actions/runs/20429779899

ririv:
I have investigated the issue with the wheel generated by maturin. It actually produced the correct wheel, but the tag was incorrect.
This occurred because maturin does not yet support Android platform tags, causing it to mistakenly identify Android as Linux and generate an incorrect tag. 
I have submitted a PR to maturin to fix this. Details can be found at https://github.com/PyO3/maturin/pull/2900

This issue is unrelated to the current PR. So, no modification is necessary.
Please review, thanks. @mhsmith

mhsmith:
Thanks; I'm on vacation at the moment, but I'll look at this in the new year.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +401,90 @@ def setup_android_env(
...
+    # See: https://pyo3.rs/v0.27.1/building-and-distribution.html#cross-compiling
+    env["PYO3_CROSS_LIB_DIR"] = str(python_dir / "prefix" / "lib")
+
+
+def create_rust_shims(env: MutableMapping[str, str], cargo_target: str, rustup_path: str) -> None:
+    venv_bin = Path(env["VIRTUAL_ENV"]) / "bin"
+
+    shim_script = dedent(f"""\
+           #!/usr/bin/env python3
+           import os
```
mhsmith: Suggest moving this script to a separate file under `resources`, using the same mechanism as `_cross_venv`. This would have several advantages:
* More readable – would allow `create_rust_shims` to be inlined into `setup_rust`
* Better interaction with IDE features
* Easier to test on its own
* Allows it to be checked by pre-commit.

Of the placeholders in the existing script:

* `cargo_target` is already available as the `CIBW_HOST_TRIPLET` environment variable.
* `rustup_path` appears to be unnecessary, since the script already contains logic for searching the PATH to find the real `rustup`.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +401,90 @@ def setup_android_env(
     return android_env
 
 
+def setup_rust(
+    python_configuration: PythonConfiguration,
```
mhsmith: Other variables of this type are simply called `config`.

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +401,90 @@ def setup_android_env(
...
+               try:
+                   subprocess.run(
+                       [REAL_RUSTUP, "target", "add", TARGET],
+                       check=True,
+                       stdout=subprocess.DEVNULL,
+                       stderr=subprocess.PIPE,
+                       text=True
+                   )
+               except subprocess.CalledProcessError as e:
+                   print(f"cibuildwheel: Warning: Failed to install Rust target {{TARGET}}: {{e.stderr}}", file=sys.stderr)
```
mhsmith: Shouldn't this be a fatal error?

cibuildwheel/platforms/android.py
```diff
@@ -398,6 +401,90 @@ def setup_android_env(
...
+    python_configuration: PythonConfiguration,
+    python_dir: Path,
+    env: MutableMapping[str, str],
+) -> None:
+    cargo_target = android_triplet(python_configuration.identifier)
+
+    # Check if rustup is available. If not, we can't install targets, so we skip shimming.
+    rustup_path = shutil.which("rustup")
+    if not rustup_path:
+        log.warning(f"rustup not found; skipping Rust target configuration for {cargo_target}")
```
mhsmith: This warning isn't relevant if the package doesn't use Rust. Suggest installing the shims and setting the environment variables unconditionally, and making the shim fail if it can't find `rustup`.

ririv:
Key Changes:

1. Resource-based Shim Script:
    * Moved the inline shim script from `android.py` to a standalone static resource file: `cibuildwheel/resources/_rust_shim.py`.
    * The shim logic is now consistent with `_cross_venv.py`. It dynamically detects the real `rustup/cargo` in `$PATH` and reads the target triplet from the `CIBW_HOST_TRIPLET` environment variable at runtime.

2. Simplified `android.py` Logic:
    * Removed the conditional check for rustup existence in setup_rust.
    * The setup_rust function now unconditionally installs the shim scripts into the virtual environment's bin directory and sets the necessary environment variables (`CARGO_BUILD_TARGET`, `PYO3_CROSS_LIB_DIR`, linkers, etc.).
    * This ensures a lazy setup: rustup target add is only triggered if and when the build process actually calls cargo.
    * Rename: `python_configuration` -> `config`

3. Expanded Test Coverage:
    * Added `test_maturin` integration test: Since the relevant PR has been merged into maturin (now requiring maturin >= 1.11.0), I've added a new test in test/test_android.py to verify this path.

Ready for re-review. Thanks! @mhsmith

ririv:
Here is how the Rust shim script integrates into the build process:

1.  **Deployment (The "Trap")**:
    During environment setup (`setup_android_env`), `cibuildwheel` copies `_rust_shim.py` into the virtual environment's binary directory (`venv/bin/`), renaming it to `cargo` and `rustup`. Since `venv/bin/` is prepended to the system `PATH`, these shims effectively shadow the system's real Rust binaries.

2.  **Interception & Setup (Lazy Execution)**:
    When the build frontend (e.g., `pip` or `build`) invokes `cargo`, the OS executes our shim script first (via the `#!/usr/bin/env python3` shebang). The script reads the target architecture from the `CIBW_HOST_TRIPLET` environment variable and, if the required Android target is missing, silently executes `rustup target add <target>` to install it on demand.

3.  **Handoff (Transparency)**:
    Once the environment is ready, the shim locates the **real** `cargo` executable in the remaining `PATH`. It then replaces the current process with the real `cargo` (using `os.execv`), passing along all original CLI arguments. This ensures the build continues seamlessly with the correct cross-compilation environment now in place.

PR #2654, opened 2025-11-07
chore: enable more Ruff checks

henryiii:
Adding more checks. Used sp-ruff-checks from sp-repo-review. Mostly improved typing.

After:
Screenshot 2025-11-07 at 1 46 07 PM

### DIFF
--- a/bin/bump_version.py
+++ b/bin/bump_version.py
@@ -68,8 +68,8 @@ def bump_version() -> None:
         sys.exit(1)
 
     # fmt: off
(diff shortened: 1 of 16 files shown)


joerick (approved):
Yeah, looks good to me!  I'm a little dubious about TC006 but I can run with it if you prefer it.

henryiii:
I thought TC006 looks a little strange at first, but I’ve gotten used to it; casting should have no effect at run time, and using a cast shouldn’t force an import or declaration.

Issue #2663, opened 2025-11-15
meson-python link.exe issue since 3.3.0 release


jeandet:
### Description

I tried to update my CI scripts to use the latest (3.3.0) cibuildwheel release, but somehow it seems to mess with the way meson finds msvc link.exe. I tried several cibuildwheel versions:
- 3.3.0 ❌
- 3.2.1  ✅
- 3.1.1 ✅
- 2.23.3 ✅

I tried look quickly in the release diff between 3.3.0 and 3.2.1 I didn't see anything obvious.
I believe I also have the correct [Meson setup](https://github.com/SciQLop/SciQLopPlots/blob/main/pyproject.toml#L40-L51).

### Build log

https://github.com/SciQLop/SciQLopPlots/actions/runs/19391928744/job/55486608055

henryiii:
https://github.com/pypa/cibuildwheel/compare/v3.2.1...v3.3.0

I see #2630; That's now modifying the path, rather than just pointing at cibuildwheel. If there was anything else in that location, that would get exposed too. However, it seems to be pretty carefully exposing only the binaries. I don't think that change is actually needed, since we use `uv.find_uv_bin()` to find uv if it's installed alongside us, but I don't see how that would have broken anything.

The windows.py file didn't really change in a way that could affect this.

Build dep pinning changes shouldn't affect meson-python.

Other than that, just normal dependency updates, which I wouldn't think would affect a Windows build. I'd be curious to see a build from just before / just after #2630.

henryiii:
Better yet, could you try #2673 on your workflow? Just use `@henryiii/fix/simpleraction` instead of `@3.2.1`.

jeandet:
@henryiii thanks a lot for your help, I started a build [here](https://github.com/SciQLop/SciQLopPlots/actions/runs/19645453249/job/56259468807), it seems to work so far.

Starbuck5:
@henryiii I ran into the same problem as jeandet on my project.

Using your custom branch fixes the problem. Tested here: https://github.com/pygame-community/pygame-ce/pull/3655

jborean93:
I think I'm coming across a similar problem here where 3.3.x somehow changes the env so that setuptools is no longer able to detect the Windows SDK paths and thus won't include them in the `-I` include dirs for `cl.exe`.

+ Working run with `v3.2.1` - https://github.com/pythongssapi/python-gssapi/actions/runs/21275731526/job/61234998635

[...]

PR #2673, opened 2025-11-24
fix(action): no need to modify path for extras

henryiii:
Try simpler action. See #2663.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -73,47 +73,30 @@ runs:
             shutil.rmtree(venv_path)
         builder = EnvBuilder()
         builder.create(venv_path)
-        exposed_binaries = {"cibuildwheel"}
-        if "uv" in EXTRAS:
-            exposed_binaries.add("uv")
-        clean_bin_path = builder.bin_path.parent / f"{builder.bin_path.name}.clean"
-        clean_bin_path.mkdir()
-        for path in list(builder.bin_path.iterdir()):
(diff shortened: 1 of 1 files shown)


mayeut:
I'm not sure why this had an impact as you mentioned in https://github.com/pypa/cibuildwheel/issues/2663#issuecomment-3571618005, maybe we should add a backend test (meson-python, scikit-build-core, maturin) ?

`uv` is still used directly in the `virtualenv` function and `pip` on Windows
https://github.com/pypa/cibuildwheel/blob/3c2d5961de159ac990d4225e499c4f554ee3e0e8/cibuildwheel/venv.py#L111
https://github.com/pypa/cibuildwheel/blob/3c2d5961de159ac990d4225e499c4f554ee3e0e8/cibuildwheel/platforms/windows.py#L578

henryiii:
Should we fix the direct usage, then?

joerick:
Thanks for this henryiii! I think both are a good idea - removing the direct `uv` usage and adding a test for an alternate backend.

Do we understand why the action change causes an issue with link.exe in meson-python? I don't see how that relates... regardless this solution seems good to me.

joerick:
Since I'd like to keep the `extras: uv` functionality, I've raised an alternative PR - https://github.com/pypa/cibuildwheel/pull/2723 - that keeps that working while avoiding the PATH changes that are causing the issue seen in #2663.

PR #2675, opened 2025-11-30
Added more infos about muslinux support of uv in options.md

Marco-Sulla:
uv doesn't support also ppc64le and riscv64 for muslinux

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -478,8 +478,6 @@ pre-installing it, or installing cibuildwheel with the `uv` extra, which is
 possible by manually passing `cibuildwheel[uv]` to installers or by using the
 `extras` option in the [cibuildwheel action](ci-services.md#github-actions).
 uv currently does not support iOS or musllinux on s390x, ppc64le and riscv64.
-Legacy dependencies like setuptools on Python < 3.12 and pip are not
-installed if using uv.
 
 On Android and Pyodide, the "pip" frontend is not supported.
 


docs/options.md
```diff
@@ -477,7 +477,7 @@ Make sure you have an external uv on Windows and macOS, either by
 pre-installing it, or installing cibuildwheel with the `uv` extra, which is
 possible by manually passing `cibuildwheel[uv]` to installers or by using the
 `extras` option in the [cibuildwheel action](ci-services.md#github-actions).
-uv currently does not support iOS or musllinux on s390x. Legacy dependencies
+uv currently does not support iOS or musllinux on s390x, ppc64le and riscv64. Legacy dependencies
 like setuptools on Python < 3.12 and pip are not installed if using uv.
```
```suggestion
uv currently does not support iOS or musllinux on s390x, ppc64le and riscv64.
Legacy dependencies like setuptools on Python < 3.12 and pip are not
installed if using uv.
```

docs/options.md
```diff
@@ -477,8 +477,9 @@ Make sure you have an external uv on Windows and macOS, either by
 pre-installing it, or installing cibuildwheel with the `uv` extra, which is
 possible by manually passing `cibuildwheel[uv]` to installers or by using the
 `extras` option in the [cibuildwheel action](ci-services.md#github-actions).
-uv currently does not support iOS or musllinux on s390x. Legacy dependencies
-like setuptools on Python < 3.12 and pip are not installed if using uv.
+uv currently does not support iOS or musllinux on s390x, ppc64le and riscv64.
+Legacy dependencies like setuptools on Python < 3.12 and pip are not
+installed if using uv.
```
```suggestion
```

This is now true for all backends as of 3.0, so let's clean it up while we are touching this.

mayeut (approved):

PR #2687, opened 2025-12-07
Update Android Python versions

mhsmith:
These Python versions contain the test script disk space fixes from https://github.com/python/cpython/pull/142289, which should fix https://github.com/pybind/pybind11/pull/5914.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,6 +100,14 @@ jobs:
 
     - uses: astral-sh/setup-uv@v7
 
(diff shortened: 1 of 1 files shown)


mhsmith:
It looks like the native Linux jobs also have trouble with disk space, so I've left the "free up disk space" step in place for those jobs, with the following changes:

* Removed `docker system prune`, since it was never actually freeing any space.
* Delete all Android NDK versions, since these jobs don't run any Android tests.

PR #2690, opened 2025-12-14
fix: always use `find_uv`

mayeut:
This allows to use the `uv` extra without `uv` being added to `PATH` however it requires to install `build` with the `uv` extra. This was already done on macOS.
I'm not sure what's the correct way forward given this results in multiple `uv` installations.

towards #2673

### DIFF
--- a/cibuildwheel/platforms/windows.py
+++ b/cibuildwheel/platforms/windows.py
@@ -329,7 +329,7 @@ def setup_python(
                 "pip",
                 "install",
                 "--upgrade",
-                "build[virtualenv]",
+                "build[virtualenv, uv]",
                 *constraint_flags(dependency_constraint),
                 env=env,
             )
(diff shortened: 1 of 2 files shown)


cibuildwheel/platforms/windows.py
```diff
@@ -329,7 +329,7 @@ def setup_python(
                 "pip",
                 "install",
                 "--upgrade",
-                "build[virtualenv]",
+                "build[virtualenv, uv]",
```
henryiii, line 332: This will install another copy of there is already an external one. I think we try to avoid that everywhere else.

mayeut: > This will install another copy of there is already an external one.

Yes indeed, hope it's clear enough in the PR description.

>  I think we try to avoid that everywhere else.

macOS is already doing that.

I'm wondering if the `uv` extra is not just plain broken when the virtual environment is not activated.
I'll try something else in a following commit.

mayeut: #2691 opened as an alternative

mayeut:
This won't work as explained in the other 2 PRs

PR #2691, opened 2025-12-14
fix: always expose `uv` in `PATH`

mayeut:
This PR changes the way `uv` is used in cibuildwheel for direct invocations.
At the moment, there's a mixed usage of direct invocations (assumed in `PATH`) & `find_uv()` path invocations.
#2690 has been opened to always use `find_uv()` but this is not really compatible with `build` unless it's installed with the `uv` extra.
In order not to install `build`'s `uv` extra, `find_uv` is updated to symlink/copy `uv` in a temporary directory. This temporary directory is added to build/test virtual environments `PATH` which removes the need to install `build`'s `uv` extra when using `cibuildwheel`'s `uv` extra.

close #2690
close #2673

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -71,47 +71,30 @@ runs:
             shutil.rmtree(venv_path)
         builder = EnvBuilder()
         builder.create(venv_path)
-        exposed_binaries = {"cibuildwheel"}
-        if "uv" in EXTRAS:
-            exposed_binaries.add("uv")
-        clean_bin_path = builder.bin_path.parent / f"{builder.bin_path.name}.clean"
-        clean_bin_path.mkdir()
-        for path in list(builder.bin_path.iterdir()):
-            if path.stem in exposed_binaries:
-                try:
-                    os.symlink(path, clean_bin_path / path.name)
-                except OSError:
-                    import shutil
-
-                    shutil.copy2(path, clean_bin_path / path.name)
-        full_path = f"{clean_bin_path}{os.pathsep}{os.environ['PATH']}"
+        cibw_path = [path for path in builder.bin_path.glob("cibuildwheel*") if path.stem == "cibuildwheel"][0]
         with open(os.environ["GITHUB_OUTPUT"], "at") as f:
-            f.write(f"updated-path={full_path}\n")
+            f.write(f"cibw-path={cibw_path}\n")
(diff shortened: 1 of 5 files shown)


mhsmith (commented):
The Android change looks OK to me, but I don't know enough to comment on the larger question of this PR vs #2690.

henryiii:
Build should behave similarly, using a uv in the path if there's no `uv` package installed?

https://github.com/pypa/build/blob/ed9c379f51abca7c0c0ae20bee66aa548d597dc3/src/build/env.py#L305-L307

mayeut:
> Build should behave similarly, using a uv in the path if there's no uv package installed?

If cibuildwheel's `uv` extra is not exposed in the PATH (as proposed in #2673) then build won't find `uv` unless it was installed with its own `uv` extra. See https://github.com/mayeut/cibuildwheel/actions/runs/20258197705/job/58164557217 for a failure when #2673 is applied on Windows which does not install build's `uv` extra.

I see you opened an issue in build and that's probably a good thing which should be applied ecosystem wide and would allow for an easier resolution.

joerick:
Hmm. All this is starting to feel a little messy/hacky. I'm thinking there's another option here, where we say that if the build-frontend uses `uv`, we require the caller of cibuildwheel to ensure that uv is somewhere on PATH (and bail if it's not found). So, we don't even attempt to look in the python import path for it. That way we'll be sure that `build` can use it too.

In the case of the github action, we could just use astral/setup-uv conditionally, and keep the direct invocation there?

As I see it, the `cibuildwheel[uv]` extra would continue to work in this setup, assuming that users have the virtualenv active when they call cibuildwheel (this must be the most common setup, I'd expect).

mayeut:
closing for the same reasons as explained in https://github.com/pypa/cibuildwheel/pull/2696#issuecomment-3862569874
If the user wants to use `uv` directly from commands, it's their responsibility to make sure it's in PATH.
Alternatives I see to fixing this are:
- rewrite commands from the user as proposed in https://github.com/pypa/cibuildwheel/pull/2696#issuecomment-3862569874 but this will not cover the case of `uv` being called from a script used as a command.
- link or copy/expose `uv` in PATH when cibuildwheel starts (different from this PR)

PR #2695, opened 2025-12-16
Android improvements needed for building NumPy and related packages

mhsmith:
Fixes for several issues found while building NumPy and related packages for Android:

* Replace the limited built-in `repair-wheel-command` with full auditwheel support.
  * Requires https://github.com/pypa/auditwheel/pull/643

* Add a shim which installs a Fortran compiler on demand, using a similar approach as [we did for Rust](https://github.com/pypa/cibuildwheel/pull/2650).

* Set up pkgconf in the build environment, and enable PKG_CONFIG_RELOCATE_PATHS. This is necessary for Meson, which uses pkgconf to find Python.

* When a package builds against NumPy, replace platform-specific files with their Android equivalents, using a similar approach to pyodide-build.

This PR branch is temporarily used by several other PRs: see the cross-references below.


henryiii:
Is there anything you'd like to pull out and get in for the next version?

mhsmith:
Not at the moment, thanks.

mhsmith:
@mayeut: Since you're familiar with https://github.com/pypa/auditwheel/pull/643, it might make sense for you to review this PR as well. It won't be ready to merge until the auditwheel work has been released, but it's otherwise complete.

cibuildwheel/platforms/android.py
```diff
@@ -43,6 +38,15 @@
     "x86_64": "x86_64-linux-android",
 }
 
+CROSS_BUILD_FILES = {
```
mayeut, line 41: I think this shall go in the configuration in order to allow users to extend/override without the need for a new cibuildwheel version.

cibuildwheel/platforms/android.py
```diff
@@ -400,6 +413,15 @@ def setup_android_env(
     # Cargo target linker needs to be specified after CC is set
     setup_rust(config, python_dir, android_env)
 
+    # Create shims which install additional build tools on first use.
+    setup_fortran(android_env)
```
mayeut, line 417: Do we really want to maintain this ? Should it be an example instead ?
Fortran is not setup for Windows & macOS and it's up to projects using it to set it up, however they choose to do it.

mhsmith: If it was simple to install the Android Fortran compiler then I would agree, but as you can see from the shim, it involves a complex merge of files from several different sources. The code to do this should be centralized somewhere, and cibuildwheel seems like the obvious choice, since it's already responsible for installing the other Android compilers (done by the [`android.py env`](https://github.com/pypa/cibuildwheel/blob/v3.4.1/cibuildwheel/platforms/android.py#L369) command, which ultimately runs [this script](https://github.com/python/cpython/blob/v3.14.4/Android/android-env.sh#L27-L33)).

Fortran is a slow-moving language, so I don't expect this code will need to change very often. For example, previously I was using the Fortran compiler from [GCC 4.9](https://github.com/mzakharo/android-gfortran), which was released in 2014, but was still able to build [SciPy 1.8.1](https://github.com/chaquo/chaquopy/pull/970), which was released in 2022.

cibuildwheel/platforms/android.py
mhsmith: I've added an `xbuild-files` option, named that way for consistency with `xbuild-tools`.

pyproject.toml
```diff
@@ -44,8 +44,6 @@ dependencies = [
     "filelock",
     "humanize",
     "packaging>=20.9",
-    # patchelf is used for Android
-    "patchelf; (sys_platform == 'linux' or sys_platform == 'darwin') and (platform_machine == 'x86_64' or platform_machine == 'arm64' or platform_machine == 'aarch64')",
     "platformdirs",
     "pyelftools>=0.29",
     "wheel>=0.33.6",
```
mayeut:
```suggestion
```

These should not be needed anymore

pyproject.toml
mayeut, line 42: Is `build`  still needed here ?

mhsmith: Yes, it's imported in android.py.

pyproject.toml
mhsmith: FIxed.

cibuildwheel/platforms/android.py
```diff
@@ -240,14 +239,21 @@ def setup_env(
...
-    android_env = setup_android_env(config, python_dir, venv_dir, build_env)
-
     # Install build tools
-    if build_frontend not in {"build", "build[uv]", "uv"}:
-        msg = "Android requires the build frontend to be 'build' or 'uv'"
-        raise errors.FatalError(msg)
-    call(*pip, "install", "build", *constraint_flags(dependency_constraint), env=build_env)
+    # TODO: use an official auditwheel version once
+    # https://github.com/pypa/auditwheel/pull/643 has been released, and add it to the
+    # constraints files.
```
mayeut: auditwheel 6.7.0 has been released

henryiii:
Are we going to try to get this into 4.0? Still waiting on PyPy anyway.

docs/options.md
```diff
@@ -939,6 +967,7 @@ The following placeholders must be used inside the command and will be replaced
 - `{wheel}` for the absolute path to the built wheel
 - `{dest_dir}` for the absolute path of the directory where to create the repaired wheel
 - `{delocate_archs}` (macOS only) comma-separated list of architectures in the wheel.
+- `{ldpaths}` (Android only) colon-separated list of directories to search for external libraries. cibuildwheel will set this to include any necessary locations in the NDK. To add your own locations, use the `LD_LIBRARY_PATH` environment variable.
```
joerick: v. minor phrasing point- it strikes me that this looks like a line of docs for the auditwheel `--ldpaths` option, not the `{ldpaths}` placeholder.

```suggestion
- `{ldpaths}` (Android only) colon-separated list of directories to search for external libraries, set by cibuildwheel to include any necessary locations in the NDK. You can add more directories by appending your own directories with a colon separator after the placeholder, or by setting the `LD_LIBRARY_PATH` environment variable.
```

cibuildwheel/resources/android/fortran_shim.py
```diff
@@ -0,0 +1,110 @@
...
+from cibuildwheel.util.file import CIBW_CACHE_PATH, download
+
+# In the future we might pick a different Flang release depending on the NDK version,
+# but so far all Python versions use the same NDK version, so there's no need.
+RELEASE_URL = "https://github.com/termux/ndk-toolchain-clang-with-flang/releases/download"
+RELEASE_VERSION = "r27c"
+ARCHS = ["aarch64", "x86_64"]
+
+# The compiler is built for Linux x86_64, so we use Docker on macOS.
+DOCKER_IMAGE = "debian:trixie"
```
joerick: Would we get away with `debian:trixie-slim` here?

cibuildwheel/resources/android/fortran_shim.py
```diff
@@ -0,0 +1,110 @@
...
+    ]:
+        archive_path = tmp_dir / archive_name
+        download(f"{RELEASE_URL}/{RELEASE_VERSION}/{archive_name}", archive_path)
+        shutil.unpack_archive(archive_path, tmp_dir)
+        archive_path.unlink()
+
+    # Merge the extracted trees together, along with the necessary parts of the NDK. Based on
+    # https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/recipes/fortran/__init__.py)
+    flang_toolchain = tmp_dir / "toolchain"
+    (tmp_dir / "out/install/linux-x86/clang-dev").rename(flang_toolchain)
```
joerick, line 52: Does this assume the build machine is x86? Perhaps the build machine is (linux) aarch64.

cibuildwheel/resources/android/fortran_shim.py
```diff
@@ -0,0 +1,110 @@
...
+    if sys.platform == "linux":
+        args = flang_args
+    elif sys.platform == "darwin":
+        args = ["docker", "run", "--rm", "--platform", "linux/amd64"]
+        for path in ["/private", "/Users", "/tmp"]:
+            # Docker on macOS only allows certain directories to be mounted as volumes
+            # by default, but they include all the locations we're likely to need.
+            args += ["-v", f"{path}:{path}"]
+        args += ["--workdir", str(Path.cwd())]
+        args += ["--entrypoint", flang_args[0], DOCKER_IMAGE, *flang_args[1:]]
```
joerick, line 112: This is quite the hack! I am quite impressed, but I do wonder what the potential for unintended consequences is here? Is it possible that linux does something to the filesystem that mac doesn't understand, or vice-versa? do we need to mount `/private` and `/tmp`? `/private` in particular contains a lot of stuff that is very OS-specific.

It also implicitly requires Docker, which might not be available on a Mac. I suppose I just wonder if we should just declare that the Fortran compilation shim is a Linux-only thing and leave it at that. Are people doing a lot of Android compilation on Mac in CI?

cibuildwheel/platforms/android.py
joerick: While it's a bit strange compared to other platforms, since `android.py env` does indeed install all the compilers, this seems consistent.

test/test_android.py
```diff
@@ -486,34 +501,203 @@ def test_libcxx(tmp_path: Path, capfd: pytest.CaptureFixture[str]) -> None:
...
+        )
+    )
+
+    # Add Fortran code.
+    meson_build_path = tmp_path / "meson.build"
+    meson_build_path.write_text(
+        meson_build_path.read_text()
+        .replace("'c'", "'c', 'fortran'")
+        .replace("'spam.c'", "'spam.c', 'fortran.f90', link_language: 'fortran'")
+    )
```
joerick: It would be slightly better to add these customisations as parameters to `new_meson_project()`, populated into the `meson.build` file via jinja variables.

docs/options.md
```diff
@@ -916,15 +916,43 @@ Platform-specific environment variables are also available on platforms that use
     ```
 
 
+### `xbuild-files` {: #xbuild-files env-var toml}
+> Platform-specific files in the build environment
+
+When cross-compiling a package, any dependencies in its [`build-system.requires`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) are installed for the build platform. However, some dependencies contain platform-specific files such as headers and static libraries, which must correspond to the target platform.
```
joerick: so far, this option is only supported on Android.

```suggestion
When cross-compiling a package for Android, any dependencies in its [`build-system.requires`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) are installed for the build platform. However, some dependencies contain platform-specific files such as headers and static libraries, which must correspond to the target platform.
```

joerick (commented):
Really nice work on this. If we can get this into 4.0 that would be nice, but I suppose it probably shouldn't block for more than a few days. 

The other question would be - can/should xbuild-files also be supported on iOS and pyodide too?

cibuildwheel/platforms/android.py
mhsmith: I've updated the PR to use it.

henryiii:
I expect we are about 24 hours from release - PyPy just got released, so just waiting on it in manylinux.

cibuildwheel/resources/android/fortran_shim.py
mhsmith: Apparently yes, so I've changed it.

cibuildwheel/resources/android/fortran_shim.py
mhsmith: None of the other Android development tools are available for Linux aarch64, so that isn't a concern.

cibuildwheel/resources/android/fortran_shim.py
mhsmith: > do we need to mount `/private` and `/tmp`? `/private` in particular contains a lot of stuff that is very OS-specific.

I've reduced this to only the home directory and the temporary directory, and verified that this works with SciPy.

> Are people doing a lot of Android compilation on Mac in CI?

Not in CI, because none of the major CI services' Mac runners support running the Android emulator. But [this comment](https://github.com/ReactiveCircus/android-emulator-runner/issues/350#issuecomment-2318425242) suggests this might be improved in the future with the M3 processor. I hope that happens, because Mac is the only platform on which Android aarch64 wheels can be tested, since the emulator isn't available for Linux or Windows aarch64. 

Outside of CI though, I compile for Android on Mac all the time, because it's my main development machine.

test/test_android.py
mhsmith: Done.

mhsmith:
> can/should xbuild-files also be supported on iOS and pyodide too?

It will probably be needed for iOS, but I don't think anyone's tried to build a package that needs it yet. I encountered it when building SciPy, which links against `numpy/_core/lib/libnpymath.a`.

On Pyodide a similar feature is already [built into pyodide-build](https://github.com/pyodide/pyodide-build/blob/v0.30.2/pyodide_build/recipe/builder.py#L638), which is where I got the idea from. It might be useful for Pyodide to support the cibuildwheel option as well, but it looks like that would require changes to pyodide-build.

> If we can get this into 4.0 that would be nice, but I suppose it probably shouldn't block for more than a few days.

I think I've dealt with all of your comments, so let's get this into 4.0 if possible.

bin/generate_schema.py
```diff
@@ -374,14 +380,15 @@ def as_object(d: dict[str, Any]) -> dict[str, Any]:
...
-    **schema["properties"]["repair-wheel-command"],
-    "default": "auditwheel repair -w {dest_dir} {wheel}",
-}
-oses["macos"]["properties"]["repair-wheel-command"] = {
-    **schema["properties"]["repair-wheel-command"],
-    "default": "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}",
-}
+for os_name, command in [
+    ("linux", "auditwheel repair -w {dest_dir} {wheel}"),
+    ("macos", "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}"),
```
henryiii, line 385: I think we have windows now too

cibuildwheel/resources/cibuildwheel.schema.json
```diff
@@ -1233,11 +1279,27 @@
...
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          ],
+          "title": "CIBW_REPAIR_WHEEL_COMMAND",
+          "default": "auditwheel repair --ldpaths {ldpaths} -w {dest_dir} {wheel}"
```
henryiii, line 1302: Not sure what the point of the three way (should be four) setting was, only one make it into the schema?

henryiii (approved):
This looks great. I think we missed a few things for Windows devlewheel, but that's not related to this PR.

cibuildwheel/resources/cibuildwheel.schema.json
mhsmith: The schema has separate sections for each platform.

henryiii: I saw that when doing #2877 👍

PR #2696, opened 2025-12-19
Fix the PATH computation in GHA on windows, and require `uv` on PATH

joerick:
Don't get clever about it at all, instead make uv on PATH a hard requirement for the caller of the cibuildwheel CLI. See https://github.com/pypa/cibuildwheel/pull/2691#issuecomment-3675725746 for context.

In the github action, this approach does add the cibuildwheel install venv bin to PATH. This has the potential downside of exposing some other transitive deps of cibuildwheel, but I think that's a theoretical downside, I'm not aware of a way that could be an issue in practice. I think it's an okay tradeoff for more simplicity.

close https://github.com/pypa/cibuildwheel/pull/2690
close https://github.com/pypa/cibuildwheel/pull/2673
close #2691

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -75,32 +75,37 @@ runs:
 
         with open(os.environ["GITHUB_OUTPUT"], "at") as f:
             f.write(f"cibw-bin={cibw_bin}\n")
-            f.write(f"bin-path={builder.bin_path}\n")
+            f.write(f"prepend-path={builder.bin_path if 'uv' in EXTRAS else ''}\n")
+
         print("::endgroup::")
         EOF
       shell: bash
 
     # Redirecting stderr to stdout to fix interleaving issue in Actions.
-    - run: >
-        "${{ steps.cibw.outputs.cibw-bin }}"
-        "${{ inputs.package-dir }}"
-        ${{ inputs.output-dir != ''  && format('--output-dir "{0}"', inputs.output-dir) || ''}}
-        ${{ inputs.config-file != '' && format('--config-file "{0}"', inputs.config-file) || ''}}
-        ${{ inputs.only != ''        && format('--only "{0}"', inputs.only) || ''}}
-        2>&1
-      env:
-        PATH: ${{ contains(inputs.extras, 'uv') && format('{0}:{1}', steps.cibw.outputs.bin-path, env.PATH) || env.PATH }}
+    - run: |
+        prepend_path="${{ steps.cibw.outputs.prepend-path }}"
+        if [ -n "$prepend_path" ]; then
+            export PATH="$prepend_path:$PATH"
+        fi
+        "${{ steps.cibw.outputs.cibw-bin }}" \
+          "${{ inputs.package-dir }}" \
+          ${{ inputs.output-dir != ''  && format('--output-dir "{0}"', inputs.output-dir) || ''}} \
+          ${{ inputs.config-file != '' && format('--config-file "{0}"', inputs.config-file) || ''}} \
+          ${{ inputs.only != ''        && format('--only "{0}"', inputs.only) || ''}} \
+          2>&1
       shell: bash
       if: runner.os != 'Windows'
 
     # Windows needs powershell to interact nicely with Meson
-    - run: >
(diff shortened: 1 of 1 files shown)


henryiii:
I'm back now, I'll try to take a look soon, I would like for all of the packages I work on that can use uv (build, nox, cibuildwheel, check-sdist) to do exactly the same behavior (I think this is also true for tox and others). If the problem is that we are not setting up our fake virtual environment correctly, then we should fix that.

I can get a build release out soon to see if that helps, as well.

joerick:
Thanks. Yeah, just a thought- it's kinda hard for the `build` behaviour to always match cibuildwheel, because `build` doesn't expose uv to its target venv, but cibuildwheel does. 

Python packages don't inherit between venvs. But PATH entries do. So that's some of the logic behind this approach.

joerick:
By the way, I have a theory why #2630 broke some builders e.g. meson-python in https://github.com/pypa/cibuildwheel/issues/2663. 

In the version on `main`, the PATH variable is being extended in a bash step. 

https://github.com/pypa/cibuildwheel/blob/24bade1d153af61fcc9fc6865ba34fe4d9b274e8/action.yml#L89-L91

And the PATH from that bash step is written, and used in the powershell step. So we don't actually get the powershell-native PATH, we get whatever bash is giving (perhaps that includes MinGW tools).

joerick:
This PR avoids that by outputting a 'prepend-path' variable from the bash step, and updating PATH in the pwsh step.

henryiii:
I think I'd prefer the patch-up fixes #2673 and most of #2690; those restore the previous behavior and fix an oversight. Then longer term we can decide if we go this way to improve using "uv" _from_ cibuildwheel's commands (those fixes are related to cibuildwheel itself working correctly). If we do go this way, I want us to remove the `uv` extra since it doesn't work correctly if we go this way.

joerick:
Going with #2673 would stop [`extras: uv`](https://github.com/pypa/cibuildwheel/blob/f3ba8a83a5df58f2f646e12e21468f0cfd0f7a36/action.yml#L20-L23) from working, which would be a shame because I think it's a pretty reasonable syntax that should work.

I've done a little testing of [my theory](https://github.com/pypa/cibuildwheel/pull/2696#issuecomment-3708116126) that the way the PATH is extended in a bash step is the reason behind the #2663 meson link.exe issue. Test results are here - https://github.com/joerick/ping/actions/runs/20798304116/job/59737382262 - I must say that it's a little confusing, but certainly there are a few extra PATH entries and different link.exe in the bash shell.

Another option (that would be less disruptive) would be to keep the prepend approach as seen in this PR, but still resolve uv using find_uv. 

Having said that, my preference is still to ensure that `uv` is on PATH:

- `uv` can be [installed](https://docs.astral.sh/uv/getting-started/installation/) in many different ways - python venv, the curl|bash script, homebrew, the only commonality is PATH.
- we _can_ access `uv` that's installed in a Python environment and not on PATH, but that doesn't propagate that to child venvs (without hacks)
- `uv` itself creates venvs without installing itself into the venv - unlike venv, which installs pip into venvs. So that implies it should be on PATH, as we can't use python package resolution in sub-venvs.

henryiii:
I think the current form is one I'm pretty happy with: We don't require uv on the path for our use. If a user puts `uv <...>` into their commands, they are responsible to make sure uv is on the path (on Windows/macOS), much like any other tool they use.

I do like "fixing" this when possible (`nox`, for example, does transform any command that starts with the single string "uv"), but I don't think we can do that, since we are just passing though text. Many projects don't need to interact with uv manually, though.

joerick:
Yeah, I agree that if the user is using `uv` in their command lines/scripts it seems reasonable that it's their responsibility that it's on PATH.

I think there is still one outstanding issue though... if `build-frontend = "build[uv]"` and `uv` isn't on PATH, the calls that `build` makes to `uv` will fail. For that problem, if we're not requiring `uv` on PATH, we'd need something like @mayeut's #2691. Or we just say "uv is required on PATH if you use `build[uv]`".

PR #2708, opened 2026-01-04
[3.3.x] Update dependencies

joerick:
Updates deps for a 3.3.x patch release.

    nox --force-color -s update_constraints && nox --force-color -s update_pins

### DIFF
--- a/cibuildwheel/resources/pinned_docker_images.cfg
+++ b/cibuildwheel/resources/pinned_docker_images.cfg
@@ -1,54 +1,54 @@
 [x86_64]
-manylinux2014 = quay.io/pypa/manylinux2014_x86_64:2026.01.03-3
-manylinux_2_28 = quay.io/pypa/manylinux_2_28_x86_64:2026.01.03-3
-manylinux_2_34 = quay.io/pypa/manylinux_2_34_x86_64:2026.01.03-3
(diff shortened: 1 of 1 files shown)


mayeut:
@joerick, please hold a bit on this one.
auditwheel is buggy in the latest manylinux images.
auditwheel 6.6.0 has been released today and manylinux images are being rebuilt.

henryiii:
I can also try to get a build release out.

mayeut:
I ran `nox -s update_pins` again now that the new images are published

joerick:
I suppose a `build` update will need a container rebuild too. I'm inclined to release as-is, unless you have a particular reason to get an updated `build` in @henryiii ?

joerick:
Ah, I guess https://github.com/pypa/cibuildwheel/pull/2709#issuecomment-3709320754 refers to this.

henryiii:
Ah, yes, I meant to comment here. I've got a build update proposed in https://github.com/pypa/build/pull/972; the main things affecting cibuildwheel is `$UV` and `-q` support, but at least `-q` will require a (small) update in cibuildwheel to support it, so won't really have an effect in the patch release. I think `$UV` alone isn't enough to fix most of the issues; we'd also need support in cibw. #2673 is basically just a revert to older behavior, so I'd be fine with that going in, I think it would at least fix meson.

PR #2711, opened 2026-01-09
Add a small note about security implications of using `cibuildwheel` in CI runners

agriyakhetarpal:
Based on recent discussions, this PR adds a small note for our users about the security implications of using `cibuildwheel` in their CI configuration. Suggestions on improving the language are welcome.

I also bumped `actions/checkout` to `v6` everywhere, since they finally fixed the issue where it leaked credentials and previously required `persist-credentials: false` to avoid that; ~it no longer needs that now.~ However, it is possible that it may not be enough as people may not upgrade anyway if their Dependabot/Renovate/etc. is not configured or if their cadence is too low. I wonder if we should set that explicitly in our docs? Edit: I have set it explicitly indeed, as not enabling the option still makes it store credentials, just in a location that's a bit more resistant, and Zizmor dropped the severity of the rule for v6 and above and didn't remove the rule itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,9 +71,6 @@ Usage
 
 <!--intro-end-->
 
-> [!WARNING]
-> Building and testing wheels executes arbitrary code from your project and its dependencies. To maintain security standards: keep the job that builds distributions separate from the job that uploads them to PyPI, handle secrets and credentials with care and rotate them regularly, and follow the principle of least privilege when granting permissions. Do not store sensitive data on CI runners. It is a good idea to follow [the Python Packaging Authority's guides](https://packaging.python.org/en/latest/guides/).
-
 Example setup
 -------------
 
@@ -128,6 +125,11 @@ The following diagram summarises the steps that cibuildwheel takes on each platf
 ![](docs/data/how-it-works.png)
 
 <sup>Explore an interactive version of this diagram [in the docs](https://cibuildwheel.pypa.io/en/stable/#how-it-works).</sup>
+
+> [!WARNING]
+> Building and testing wheels executes arbitrary code from your project and its dependencies. Although cibuildwheel uses OCI containers and Pyodide for some builds, these provide no security guarantees - the code you're building and testing has full access to the environment that's invoking cibuildwheel.
+>
+> If you cannot trust all the code that's pulled in, maintain good security hygiene: keep the job that builds distributions separate from the job that uploads them to PyPI, handle secrets and credentials with care and rotate them regularly, and follow the principle of least privilege when granting permissions. Do not store sensitive data on CI runners.
 
 
 <!--[[[cog from readme_options_table import get_table; print(get_table()) ]]]-->
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,9 +10,6 @@ title: Home
    end="<!--intro-end-->"
 %}
 
-!!! warning "A note on security"
-      Building and testing wheels executes arbitrary code from your project and its dependencies. To maintain security standards: keep the job that builds distributions separate from the job that uploads them to PyPI, handle secrets and credentials with care and rotate them regularly, and follow the principle of least privilege when granting permissions. Do not store sensitive data on CI runners. It is a good idea to follow [the Python Packaging Authority's guides](https://packaging.python.org/en/latest/guides/).
-
 To get started, head over to the [setup guide](setup.md).
 
 How it works
@@ -25,3 +22,8 @@ This diagram summarises the steps that cibuildwheel takes on each platform to bu
 %}
 
 This isn't exhaustive, for a full list of the things cibuildwheel can do, check the [options](options.md) page.
+
+!!! warning "A note on security"
+      Building and testing wheels executes arbitrary code from your project and its dependencies. Although cibuildwheel uses OCI containers and Pyodide for some builds, these provide no security guarantees - the code you're building and testing has full access to the environment that's invoking cibuildwheel.
+
+      If you cannot trust all the code that's pulled in, maintain good security hygiene: keep the job that builds distributions separate from the job that uploads them to PyPI, handle secrets and credentials with care and rotate them regularly, and follow the principle of least privilege when granting permissions. Do not store sensitive data on CI runners.


henryiii:
See https://github.com/zizmorcore/zizmor/pull/1353 - `persist-credentials: false` still is better than not having it, it's just persisted in a different location.

Bumping versions is always recommended, our examples should stay on the latest versions.

agriyakhetarpal:
Ah, thanks for linking that one and clarifying – I had this in mind.

README.md
```diff
@@ -91,7 +94,7 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-15-intel, macos-latest]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 94:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

docs/deliver-to-pypi.md
```diff
@@ -29,7 +29,7 @@ GitHub actions has pipx in all the runners as a supported package manager, as we
     name: Make SDist
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0  # Optional, use if you use setuptools_scm
         submodules: true  # Optional, use if you have submodules
```
agriyakhetarpal, line 35:
```suggestion
    - uses: actions/checkout@v6
       with:
         fetch-depth: 0  # Optional, use if you use setuptools_scm
         submodules: true  # Optional, use if you have submodules
         persist-credentials: false # Highly recommended as a good security practice to not store credentials in disk, unless you need this for your specific use case
```

examples/github-deploy.yml
```diff
@@ -48,7 +48,7 @@ jobs:
             platform: pyodide
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 51:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

examples/github-deploy.yml
```diff
@@ -70,7 +70,7 @@ jobs:
     name: Build source distribution
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 75:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

examples/github-minimal.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-15-intel, macos-14]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 15:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

examples/github-pipx.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-15-intel, macos-14]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 15:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

examples/github-with-qemu.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-15-intel, macos-14]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
```
agriyakhetarpal, line 15:
```suggestion
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
```

docs/deliver-to-pypi.md
```diff
@@ -29,10 +29,11 @@ GitHub actions has pipx in all the runners as a supported package manager, as we
...
     steps:
-    - uses: actions/checkout@v5
-      with:
-        fetch-depth: 0  # Optional, use if you use setuptools_scm
-        submodules: true  # Optional, use if you have submodules
+    - uses: actions/checkout@v6
+       with:
+         fetch-depth: 0  # Optional, use if you use setuptools_scm
+         submodules: true  # Optional, use if you have submodules
+         persist-credentials: false # Highly recommended as a good security practice to not store credentials in disk, unless you need this for your specific use case
```
agriyakhetarpal:
```suggestion
    - uses: actions/checkout@v6
      with:
        fetch-depth: 0  # Optional, use if you use setuptools_scm
        submodules: true  # Optional, use if you have submodules
        persist-credentials: false # Highly recommended as a good security practice to not store credentials in disk, unless you need this for your specific use case
```

agriyakhetarpal:
Done, thanks!

agriyakhetarpal:
> Thank you @agriyakhetarpal for doing this!
> 
> I think the best place for this warning will be at the end of the How it Works section in both the README and the docs index - it relates to the mechanism of building and the diagram provides some context.
> 
> As for the text of the warning, I think it would be good to make sure there's information specific to cibuildwheel, as well as some general advice as well. Essentially, the information from a threat model perspective is that cibuildwheel offers no isolation from the project it's building - even on Docker or Pyodide.
> 
> Here's a draft from me:
> 
> > Building and testing wheels executes arbitrary code from your project and its dependencies. Although cibuildwheel uses OCI containers and Pyodide for some builds, these provide no security guarantees - the code you're building and testing has full access to the environment that's invoking cibuildwheel.
> > If you cannot trust all the code that's pulled in, maintain good security hygiene: keep the job that builds distributions separate from the job that uploads them to PyPI, handle secrets and credentials with care and rotate them regularly, and follow the principle of least privilege when granting permissions. Do not store sensitive data on CI runners.

Thanks a lot @joerick, your proposed text reads much better to me overall! I'll move its location to the end of the "How it works" section as well. For the README, I'll keep it above the build options table, since it's fairly long.

> Also, I removed the link to the PyPA guides - I couldn't see anything relevant to the security model of building in there.

Fair call; I was previously thinking that we should keep it because we have this page: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/, but I now notice that it itself indirectly links to our docs for advice on building platform-specific distributions.

joerick (approved):

PR #2712, opened 2026-01-11
ci: test action with `uv` extra

mayeut:
Whatever the fix ends-up being for `uv` extra - if keeping it - it needs to be tested.
This PR adds a minimal test of the action for the `uv` extra and update the action not to skip `uv` on Linux as it's required when building Android on Linux.
The sample generation is moved to its own job in order to allow for the minimal `uv` extra jobs.

towards #2696, #2691 or #2673

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -47,8 +47,6 @@ runs:
         from subprocess import run
 
         EXTRAS = set(e.strip() for e in "${{ inputs.extras }}".split(",") if e.strip())
-        if sys.platform == "linux":
-            EXTRAS.discard("uv")
 
 
         class EnvBuilder(venv.EnvBuilder):


action.yml
```diff
@@ -47,8 +47,6 @@ runs:
         from subprocess import run
 
         EXTRAS = set(e.strip() for e in "${{ inputs.extras }}".split(",") if e.strip())
-        if sys.platform == "linux":
-            EXTRAS.discard("uv")
```
joerick, line 51: Is this intentional? uv shouldn't be referenced on the local machine in the linux build - it's used in the container.

mayeut: This is intentional as noted in the description:
> update the action not to skip uv on Linux as it's required when building Android on Linux.

Although `uv` is not required when building Linux wheels on Linux, it's required when building Android wheels on Linux.

joerick: Apologies for missing that!

PR #2719, opened 2026-01-26
Bump to Pyodide v0.29.2

agriyakhetarpal:
and also bump some constraints as https://github.com/advisories/ghsa-8rrh-rw8j-w5fx/ was fixed in `wheel` recently; users would appreciate a clean security report.

### DIFF
--- a/test/test_pyodide.py
+++ b/test/test_pyodide.py
@@ -99,7 +99,7 @@ def test_pyodide_version_incompatible(tmp_path, capfd):
             tmp_path,
             add_args=["--platform", "pyodide"],
             add_env={
-                "CIBW_DEPENDENCY_VERSIONS": "packages: pyodide-build==0.29.3",
+                "CIBW_DEPENDENCY_VERSIONS": "packages: pyodide-build==0.31.2",
                 "CIBW_PYODIDE_VERSION": "0.26.0a6",
             },
         )


ryanking13:
It looks like the `wheel` package made [`wheel.cli` private](https://github.com/pypa/wheel/releases/tag/0.46.0). I guess we need to update our code not to rely on that.

```
Made the wheel.cli package private (no deprecation warning)
```

agriyakhetarpal:
Hmm, I'm not sure I follow... we have been using the wheel CLI instead of the API since https://github.com/pyodide/auditwheel-emscripten/pull/45, and that PR has been in a release since v0.1.0?

agriyakhetarpal:
I see; the https://github.com/pyodide/auditwheel-emscripten/releases/tag/0.2.1 tag was created on December 17th last year, but your release was posted five hours ago. But that still doesn't explain it. I'll update the constraints again and check.

ryanking13:
Oh I see. That's weird. For some reason, it looks like `auditwheel-emscripten<0.1.0` is used in CI

ryanking13:
The failing test `test_pyodide_version_incompatible` is installing pyodide-build==0.29.3 and that version is using older version of auditwheel-emscripten. That should be the reason.

https://github.com/pypa/cibuildwheel/blob/dc4fc9c3fe05c08e571fbbb4f816ab36454f88de/test/test_pyodide.py#L91-L105

agriyakhetarpal:
My bad, I should have checked the actual code too. 😄 Thanks!

PR #2686, opened 2025-12-07
break: drop Python 3.8

mayeut:
Close #2685 
There's quite a bit of specific code for 3.8 that can be removed, especially on macOS.

Will be rebased once #2684 goes in

Issue #2685, opened 2025-12-07
Drop Python 3.8


mayeut:
[manylinux currently plans](https://github.com/pypa/manylinux/issues/1882) to drop Python 3.8 builds on 2026-05-05

Python 3.8 reached EOL on 2024-10-07. At this time, it's still used to produce wheels for 35% of active projects (including py3 or abi3 wheels which would not be affected by the drop in the same way). Python 3.8 represents less than 5% of overall manylinux wheels downloads. All those numbers are dropping.
Python 3.8 is the version used on Ubuntu 20.04 (currently in extended support phase, the standard support ended on 2025-04-02) which represents a 3rd of downloads. Another large share of downloads come from Amazon Linux 2 which will reach EOL on 2026-06-30.

The date chosen by manylinux is both a 5 months notice to gather feedback early on and maps to CPython 3.15.0b1 release date.

This issue is mostly created for the [manylinux one](https://github.com/pypa/manylinux/issues/1882) to get more visibility as it will have a direct impact on cibuildwheel.

PR #2737, opened 2026-02-07
fix: support -q for build/uv

henryiii:
Build now supports this in 1.4.0, and uv does too.

### DIFF
--- a/cibuildwheel/frontend.py
+++ b/cibuildwheel/frontend.py
@@ -4,6 +4,7 @@ import typing
 from collections.abc import Sequence
 from typing import Literal, Self, get_args
 
+from .logger import log
 from .util.helpers import parse_key_value_string
 
 BuildFrontendName = Literal["pip", "build", "build[uv]", "uv"]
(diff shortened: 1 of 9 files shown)


mayeut:
We're still using `build==1.2.2.post1` for python 3.8 so we'll need to wait for #2685 or make this conditional somehow.
https://github.com/pypa/cibuildwheel/blob/0d2547bf6f4be44e39b082971605921e4ba16738/cibuildwheel/resources/constraints-python38.txt#L5

freakboy3742 (commented):
The changes for iOS make sense to me, 

More generally, the only question I'd have is whether we actually want to retain support for Python 3.8 *at all*...

henryiii:
I think we plan to drop it in May

* #2686, #2685

henryiii:
@mayeut the last commit here can be reverted when we drop Python 3.8!

mayeut:
I rebased the drop 3.8 PR & removed the workaround introduced here.

PR #2726, opened 2026-02-02
docs: fix broken TOML links in options.md

0xkems:
## Summary

Fixes broken TOML links in `docs/options.md` where `[TOML][]` (reference-style link with empty reference) was used instead of a proper link.

## Changes

- Replaced `[TOML][]` with `[TOML](https://toml.io)` on lines 621 and 682
- This matches the style used in `docs/configuration.md` (line 71)

## Testing

The fix is a simple markdown link correction. No code changes.

Fixes #2725

PR #2740, opened 2026-02-13
docs: fix TOML reference links

### LINKED ISSUE #2725: [docs][good first issue] Broken `TOML` links in the options doc
https://cibuildwheel.pypa.io/en/latest/options/ has a few places with `[TOML][]`. Looks like #684 introduced the mistake and `[]` just needs to be dropped provided that the detached definition is still in the doc.

LouisLau-art:
Fixes #2725.

The options docs referenced TOML using `[TOML][]`, but the reference definition was missing, so the site rendered broken links.

- Switch to `[TOML]` and add a `[TOML]: https://toml.io/en/` definition in `docs/options.md`.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -627,7 +627,7 @@ Platform-specific environment variables are also available:<br/>
     SAMPLE_TEXT = "sample text"
     ```
 
-    In configuration files, you can use a [TOML] table instead of a raw string as shown above.
+    In configuration files, you can use a [TOML][] table instead of a raw string as shown above.
 
 [TOML]: https://toml.io/en/
 
@@ -690,7 +690,7 @@ To specify more than one environment variable, separate the variable names by sp
     environment-pass = ["BUILD_TIME", "SAMPLE_TEXT"]
     ```
 
-    In configuration files, you can use a [TOML] list instead of a raw string as shown above.
+    In configuration files, you can use a [TOML][] list instead of a raw string as shown above.
 
 !!! tab examples "Environment variables"
 


docs/options.md
```diff
@@ -627,7 +627,9 @@ Platform-specific environment variables are also available:<br/>
     SAMPLE_TEXT = "sample text"
     ```
 
-    In configuration files, you can use a [TOML][] table instead of a raw string as shown above.
+    In configuration files, you can use a [TOML] table instead of a raw string as shown above.
```
```suggestion
    In configuration files, you can use a [TOML][] table instead of a raw string as shown above.
```

docs/options.md
```diff
@@ -688,7 +690,7 @@ To specify more than one environment variable, separate the variable names by sp
     environment-pass = ["BUILD_TIME", "SAMPLE_TEXT"]
     ```
 
-    In configuration files, you can use a [TOML][] list instead of a raw string as shown above.
+    In configuration files, you can use a [TOML] list instead of a raw string as shown above.
```
```suggestion
    In configuration files, you can use a [TOML][] list instead of a raw string as shown above.
```

henryiii (commented):
Ahh, I realize the description in the issue is slightly wrong. You don't want to remove the `[]`, since `[link][]` is a standard short form for `[link][link]`, but `[link]` as a shortcut for `[link][]` is not universal. It _does_ work for our renderer, so it's not wrong, but some syntax highlighters (including the one I use in vi) don't support it. And I think it's more obviously supposed to be a link, and easier to spot and fix if it's incorrect (like here).

I also slightly prefer putting the definitions at the bottom, but we don't have any there currently, so I guess putting it here is fine.

Thanks, this is basically the correct fix! Closes #2726 too.

webknjaz:
> [link][] is a standard short form for [link][link]

Wow, TIL!

PR #2743, opened 2026-02-16
Bump to Pyodide v0.29.3

agriyakhetarpal:


joerick:
Error is-

>   ERROR: Cannot install pyodide-build==0.32.0 because these package versions have conflicting dependencies.
>   
>   The conflict is caused by:
>       pyodide-build 0.32.0 depends on virtualenv
>       The user requested (constraint) virtualenv==20.37.0
>   
>   Additionally, some packages in these conflicts have no matching distributions available for your environment:
>       virtualenv
> 

Looks like that virtualenv release was deleted (not yanked), so it's failing there. The fix is in #2734.

agriyakhetarpal:
Thanks! I bumped them here too, just in case that helps.

PR #2752, opened 2026-02-26
Fix incorrect document regarding pyodide auditwheel

ryanking13:
Pyodide does not run `auditwheel` by default as we don't know where to locate the cross-compiled libraries, it needs to be done by the user.


agriyakhetarpal:
Note that the test is failing; should be a quick fix though.

ryanking13:
> Thanks, @ryanking13! Makes sense. Side note: any reason why we are at auditwheel-emscripten==0.2.1 in cibuildwheel and not v0.2.3?

No reason, I guess https://github.com/pypa/cibuildwheel/pull/2734 will pickup the latest version soon?

joerick (approved):

joerick:
Docs change looks good, thanks. But there's a test failure on the test that was added in this PR.

PR #2761, opened 2026-03-08
fix: call `repair-wheel-command` in iOS builds

### LINKED ISSUE #2760: iOS builds do not call `repair-wheel-command`
### Description

Currently trying to embed an SBOM as part of the repair step, the iOS build does not call the custom `repair-wheel-command` added in the configuration.

cc @freakboy3742

### Build log

https://github.com/mayeut/pybase64/actions/runs/22819482127/job/66190025018

### CI config

https://github.com/mayeut/pybase64/blob/sbom-template/pyproject.toml

mayeut:
fix #2760

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -450,6 +450,7 @@ def build(options: Options, tmp_path: Path) -> None:
             identifier_tmp_dir = tmp_path / config.identifier
             identifier_tmp_dir.mkdir()
             built_wheel_dir = identifier_tmp_dir / "built_wheel"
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
Looks good to me. I hadn't considered that `repair_command` could be used for SBOM-type tooling.

mayeut:
It can also be used for any checks that can run on the build system requiring the wheel itself  (versus checks in the test-command that would only run in the host system from the installed wheel). There's an example for this regarding `abi3audit` in the [docs](https://cibuildwheel.pypa.io/en/stable/options/#examples_11).

PR #2766, opened 2026-03-11
SEC: add a 7 days cooldown period to dependabot settings

neutrinoceros:


agriyakhetarpal:
@joerick should we consider adding a 7-day cooldown to our `update_constraints` dependency updates too?

henryiii:
This also keeps us from getting security updates for a week if vulnerabilities are found, so I'm neutral. For our constraints, we already have a delay in releasing updates, so I don't think so.

agriyakhetarpal:
Yeah, we have leaned on _not_ pinning actions up till now; we have had moving tags and thus wouldn't benefit much from cooldowns. The currently pinned action is `actions/setup-python`, which, in my opinion, would benefit from a cooldown because it prevents us from updating too quickly and releasing a broken release if setup-python's functionality is found to be broken in their new release. I am +0.5 on this, which is the reason for my approval. That said, I hope GitHub is taking the security of that action seriously enough.

henryiii:
The things we are mostly likely to be broken on with setup-python is python-versions and the runner, both of which we can't pin. Partially pinning on top of a moving environment isn't helpful, IMO, since you might need fixes for that moving environment, and you are not protected against breakages by just partially pinning.

joerick:
Just a gut feeling, but I reckon a 7 day cooldown on dependabot is probably net-net a slight positive.

joerick:
That is, the security advantages probably tip the scales to positive. I agree that the stability pros/cons is a mixed bag!

Issue #2770, opened 2026-03-11
SEC: several vulnerabilities detected by zizmor


neutrinoceros:
I would like to propose enabling continuous security scanning with zizmor given the critical place this repo sits in the ecosystem's suplly chain.
At the time of opening, most (possibly all) existing vulnerabilities can be fixed automatically or with minimal effort, and I've started issuing PRs in that direction.

Linked PRs
- #2766
- #2767
- #2744
- #2769 
- #2771 
- #2776

neutrinoceros:
There's an audit I cannot address with *just* a PR: `secrets-outside-env`

```
warning[secrets-outside-env]: secrets referenced without a dedicated environment
  --> ./.github/workflows/update-dependencies.yml:30:21
   |
18 |   update-dependencies:
   |   ------------------- this job
...
30 |         app-id: ${{ secrets.CIBUILDWHEEL_BOT_APP_ID }}
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ secret is accessed outside of a dedicated environment
   |
   = note: audit confidence → High

warning[secrets-outside-env]: secrets referenced without a dedicated environment
  --> ./.github/workflows/update-dependencies.yml:31:26
   |
18 |   update-dependencies:
   |   ------------------- this job
...
31 |         private-key: ${{ secrets.CIBUILDWHEEL_BOT_APP_PRIVATE_KEY }}
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ secret is accessed outside of a dedicated environment
   |
   = note: audit confidence → High
```
remediation is easy, but it does require setting up at least one env at the repo level, which only an admin can do.

henryiii:
By the way, the biggest security issue is probably one that the tool can't detect. We just directly download python, and don't check to see if the hash of the download matches our expectations. We should record to the hash and do hash checking. Of course, if Python does re-upload a modified binary, which they have done before, that would break us.

neutrinoceros:
You're right. The GHA ecosystem is actually pretty brittle as a whole and there are many attack vectors we cannot guard against. I don't think that's a good reason not to guard against the ones we can though.

agriyakhetarpal:
[...]

PR #2771, opened 2026-03-11
SEC: disable default gha permissions

neutrinoceros:
ref #2770


neutrinoceros:
I don't think the one failing job is related (looks like some random timeout to me), so I'll open this for review now.

PR #2778, opened 2026-03-14
Add GHA environment to update-dependencies workflow

joerick:
See #2770. I've moved the secrets into the environment so they can only be accessed from here.


joerick:
image

All this GitHub UI around 'deployments' is a bit confusing, but let's see if it works...

joerick:
> All this GitHub UI around 'deployments' is a bit confusing

It looks like that issue will be fixed in the next couple of weeks https://github.com/actions/runner/issues/2120#issuecomment-3863243378

joerick:
Cool, that's working, see #2780. I've removed the environment from runs that don't need it, so we avoid most of the deployment noise.

PR #2775, opened 2026-03-12
[2.x] Update virtualenv URLs to avoid blob URLs

joerick:
It looks like [a few](https://github.com/pypa/cibuildwheel/pull/2555#issuecomment-4049234080) people are hitting the HTTP 429 for old versions of cibuildwheel. These hosted versions of virtualenv should work, let's see. If it's easy to change to this, I'm willing to do a 2.x patch for those packaging for older versions of Python.

PR #2782, opened 2026-03-16
Drop explicit `build` and `virtualenv` installation from Pyodide build tools

agriyakhetarpal:
See [`0911b78` (#2775)](https://github.com/pypa/cibuildwheel/pull/2775/commits/0911b78ff7c94d38c01fdcbad2651a38969ec624) and https://github.com/pypa/cibuildwheel/pull/2775#issuecomment-4067444018: it is better that we remove `build[virtualenv]` from the top-level install command entirely. `pyodide-build` already declares `build~=1.2.0` and `virtualenv` as direct dependencies, so they'll come in transitively at the right versions. `cibuildwheel` shouldn't be separately requesting `build[virtualenv]` for the Pyodide platform.

Perhaps, at some later point, we should vendor `build` into `pyodide-build` and drop the pins. I briefly recall some conversations about this with @ryanking13 (not sure where that was?).

### DIFF
--- a/cibuildwheel/platforms/pyodide.py
+++ b/cibuildwheel/platforms/pyodide.py
@@ -285,7 +285,6 @@ def setup_python(
         "install",
         "--upgrade",
         "pyodide-build",
-        "build[virtualenv]",
         *constraint_flags(constraints_path),
         env=env,
     )


ryanking13:
> Perhaps, at some later point, we should vendor build into pyodide-build and drop the pins. I briefly recall some conversations about this with @ryanking13 (not sure where that was?).

Not sure if we want to vendor the whole `build` package to pyodide-build, but I think we can relax the pin to make it more flexible. We've removed all usage of the private methods, so I think it is quite stabler than before when we first started using it.

agriyakhetarpal:
> I think we can relax the pin to make it more flexible. We've removed all usage of the private methods, so I think it is quite stabler than before when we first started using it.

This is indeed what I was working on in the past hour; sounds good to me!

PR #2784, opened 2026-03-16
Fix zizmor "code injection via template expansion"

joerick:
- **Refactor action.yml to avoid template expansion, composing command line in Python**
- **Remove more template expansion**
- **Make a string quoting that's compatible with pwsh**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -156,8 +156,6 @@ jobs:
 
     - name: Set CIBW_ENABLE
       shell: bash
-      env:
-        GITHUB_REF_NAME: ${{ github.ref_name }}
       run: |
         if [[ "$GITHUB_REF_NAME" == "main" ]]; then
           CIBW_ENABLE=all
--- a/action.yml
+++ b/action.yml
@@ -60,7 +60,7 @@ runs:
             def post_setup(self, context):
                 super().post_setup(context)
                 self.bin_path = Path(context.env_exe).parent
-                install_spec = os.environ["ACTION_PATH"]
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -154,8 +156,10 @@ jobs:
 
     - name: Set CIBW_ENABLE
       shell: bash
+      env:
+        GITHUB_REF_NAME: ${{ github.ref_name }}
```
mayeut: This is a pre-defined variable. It's not necessary here.
```suggestion
```

action.yml
```diff
@@ -59,14 +60,14 @@ runs:
             def post_setup(self, context):
                 super().post_setup(context)
                 self.bin_path = Path(context.env_exe).parent
-                install_spec = r"${{ github.action_path }}"
+                install_spec = os.environ["ACTION_PATH"]
```
mayeut: There's a predefined variable for this one.
```suggestion
                install_spec = os.environ["GITHUB_ACTION_PATH"]
```

action.yml
```diff
@@ -88,34 +89,61 @@ runs:
...
+            f.write(f"cmd-bash={cmd_bash}\n")
+            f.write(f"cmd-pwsh={cmd_pwsh}\n")
 
         print("::endgroup::")
         EOF
       shell: bash
+      env:
+        PYTHON: ${{ steps.python.outputs.python-path }}
+        ACTION_PATH: ${{ github.action_path }}
+        RUNNER_TEMP: ${{ runner.temp }}
```
mayeut: Use the predefined variables for those 2
```suggestion
```

PR #2812, opened 2026-04-06
Pyodide updates: pyemscripten platform tag (PEP 783), 314.0a1, `pyodide-eol` flag, and maintenance updates

agriyakhetarpal:
As stated in the PR title. Through this PR:
- We now use `pyodide config get emscripten_dir` to get the Emscripten installation location (and always with `--force` to ensure a fresh, unmodified toolchain on every run)
- Since PEP 783 has been accepted, I have updated the wheel platform tag to `pyemscripten_year_number` for all supported versions – Pyodide 0.29.4 (the first release on the 0.29.x patch line with `pyemscripten` wheel support) introduces this for the 2025 ABI, and our new 314.0.0a1 prerelease (enabled under `pyodide-prerelease`) does so for the 2026 ABI
- We now have a `pyodide-eol enable group (analogous to `pypy-eol`) to keep cp312/Pyodide 0.27.7 available behind an opt-in flag
- I've introduced a short Pyodide maintainer guide encompassing the recurring tasks around Pyodide releases in cibuildwheel
- Lastly, new dependency updates via the constraints updater split out from #2810 and other recent bot PRs


henryiii:
It has been accepted. :D

henryiii (approved):

test/utils.py
```diff
@@ -414,14 +414,14 @@ def _expected_wheels(
 
         elif platform == "pyodide":
             platform_tags = {
-                "cp312-cp312": ["pyodide_2024_0_wasm32"],
-                "cp313-cp313": ["pyodide_2025_0_wasm32"],
+                "cp312-cp312": ["pyemscripten_2024_0_wasm32"],
```
henryiii: I assume this is an accepted wheel tag for the old pyodide? Did it need a patch?

agriyakhetarpal:
One question: should we rename Pyodide to Pyemscripten in most other places as well?

hoodmane:
We should keep calling it Pyodide, nobody knows what pyemscripten is.

henryiii:
If we wanted to match everything else, then the platform would be emscripten (like linux), the identifier would be pyemscripten (like manylinux) or maybe even pyodide. I don't think it's important enough to change now, though now would be the time to do it when this is about to make a fairly big change to the wheel name.

cibuildwheel/platforms/pyodide.py
```diff
@@ -94,29 +94,28 @@ def ensure_node(major_version: str) -> Path:
...
     with FileLock(CIBW_CACHE_PATH / "emscripten.lock"):
-        if emcc_path.exists():
-            return emcc_path
+        if emscripten_dir.exists():
+            return emscripten_dir
         call(
             "pyodide",
             "xbuildenv",
             "install-emscripten",
+            "--force",
```
ryanking13, line 109: Why do we need force?

cibuildwheel/platforms/pyodide.py
```diff
@@ -315,9 +314,9 @@ def setup_python(
     log.step(
         f"Installing Emscripten {emscripten_version} and applying Pyodide-specific patches ..."
     )
-    emcc_path = install_emscripten(env, emscripten_version, xbuildenv_cache_path, pyodide_version)
+    emscripten_dir = install_emscripten(env, emscripten_version, xbuildenv_cache_path)
 
-    env["PATH"] = os.pathsep.join([str(emcc_path.parent), env["PATH"]])
+    env["PATH"] = os.pathsep.join([str(emscripten_dir), env["PATH"]])
```
ryanking13, line 319: I wonder if we should put this path before or after the original PATH.

In pyodide-build, if a system emscripten is installed, it dominates. This is to ensure we don't shadow the user-installed emscripten. However, regarding the cibuildwheel would normally run in CI (I guess)? It is rare that people would want to install emscripten themselves, so this might make sense.

ryanking13 (approved):
> One question: should we rename Pyodide to Pyemscripten in most other places as well?

+1 for keeping it as Pyodide.

cibuildwheel/platforms/pyodide.py
agriyakhetarpal: We need it to ensure that we always have a fresh copy of the Emscripten toolchain installed and patched. One might argue that cibuildwheel runs in CI on fresh runners and therefore this wouldn't be necessary, but I would suggest there are two reasons:
1. Sequential builds: if, for whatever reason, someone builds multiple Pyodide wheels in CI using separate cibuildwheel CLI invocations or GitHub Actions within the same job in a workflow, then we will always have a self-contained build that does not rely on a stale Emscripten toolchain.
3. Security: although this may be somewhat unlikely, cibuildwheel does execute user code, which could potentially tamper with the Emscripten toolchain with the right payload when building the wheel. I am aware that this risk could apply to any file on a system too, but even if we confine ourselves to this scenario, it would be beneficial for the user to have confidence that their Emscripten installation is always the one we intended them to use. This also supports the use case for cibuildwheel outside of CI and ties into the staleness aspect in point 1. I tend to use cibuildwheel to build Pyodide wheels locally quite often.

cibuildwheel/platforms/pyodide.py
agriyakhetarpal: Good question. We have been doing `env["PATH"] = os.pathsep.join([str(emcc_path.parent), env["PATH"]])` even before #2800, i.e., in all our previous releases with Pyodide support. So, yes, I would expect a user of pyodide-build standalone to want a higher level of control over the build process and be able to their own Emscripten installation, whereas a user of cibuildwheel would trust what it does and not mess with the moving parts required to build Pyodide wheels. This means cibuildwheel should dominate with its Emscripten installation indeed, like it does right now.

That said, we can always change this behaviour and append to PATH if someone requests it at a later point in time, of course.

agriyakhetarpal:
Alright, I think Pyodide stays as the name, then. We can always create an alias and add both Pyodide and pyemscripten at a time when the latter gains popularity.

test/utils.py
hoodmane: The existing tags should stay the same right? There is no version of Pyodide that accepts a wheel tagged `pyemscripten_2024_0` so if we're going to make them we might as well just delete support for versions older than 314. But it'd be better not to change the existing behavior.

agriyakhetarpal: I believe this is a regression, then. @ryanking13, were we too early in changing to the `pyemscripten` platform tag in pyodide-build? Ideally the previous xbuildenvs (0.27.7, 0.29.3, etc.) should have set `USE_LEGACY_PLATFORM`, but then that's not possible to fix retroactively as we didn't know that at the time. So maybe we should yank pyodide-build==0.34.1, set `USE_LEGACY_PLATFORM` to `1` and create a new release with it (0.35.0 because semver), and then ensure that the upcoming 314.0 xbuildenv(s) will set `USE_LEGACY_PLATFORM` to `0` in `Makefile.envs`. I would like to support both 0.29.3 and 314.0 releases at the same time, to allow users extra time to switch ABIs.

ryanking13: Well, I am not sure to be honest. If we keep the `pyodide` tag, then the package will not be uploadable to PyPI, if we change it to `pyemscripten`, old versions of micropip will deny the package. So both choices have downsides. 

I am personally +1 for using the `pyemscripten` name for even older versions for consistency. Older Pyodide versions can still either 1) install newer micropip from PyPI, or 2) install it directly through pyodide.loadPackage. I think it is less confusing than outputting different platform names for differenet Python versions.

agriyakhetarpal: Hmm, given that users do have an option to update micropip, and now that the first alpha of 314.0 is out (thanks to you!), I will add it in this PR and drop Pyodide 0.27.7. 

Will that resolve your concerns, @hoodmane? Or, do you think that cibuildwheel should work around it now, and rename `2024_0` and `2025_0` ABI wheels that contain `pyemscripten` in the platform tag string back to `pyodide` for them?. We do have some time until this PR is merged and released, and it brings pyodide-build 0.34.1, so it's better that we do things right in the first go.

agriyakhetarpal:
I've added the new alpha release (prerelease) and corresponding code, and dropped Pyodide 0.27.7. Thanks!

cibuildwheel/resources/constraints-pyodide314.txt
agriyakhetarpal, line 1: Git/GitHub shows this file as renamed and can't identify this – more accurately, the 3.12 constraints were deleted and followed by the creation of a new file for the 3.14 constraints.

henryiii: You can split them into two commits, and then we can rebase and merge instead of squash.

agriyakhetarpal:
Having noticed several moving parts in the Pyodide test suite that I didn't fully address while running it locally, I took the liberty of writing a short maintainer-specific guide for Pyodide maintainers in 6092434647a4ac731f44f1e661a7309c4b2befb7. The guide outlines what to do on upcoming occasions when we have Pyodide releases in relation to the cibuildwheel code. It is not intended to be overly detailed, and I intentionally omitted test-related information because tests drift over time. cc: @ryanking13 and @hoodmane, if you'd like to read it.

This guide could be expanded to include documentation for additional nitty-gritties or decisions about our code that currently have no designated place – such as those buried in TODO comments (as is the case with the `PyodidePrerelease` identifier group comment that I modified in this PR).

@joerick and @henryiii, I'd love to know what you think! Do you approve of this initiative? :) If so, I’d be happy to ping our maintainers from other platforms to see if they would be willing to contribute similar documentation for the code they interact with, too!

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
+# Maintaining Pyodide support
+
+This page describes some short steps on how to update cibuildwheel's Pyodide platform code when the set of supported Pyodide Python versions changes – for example, when:
```
hoodmane:
```suggestion
This page describes how to update cibuildwheel's Pyodide platform code when either:
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+- a new Pyodide alpha release arrives with support for a new [Pyodide ABI](https://pyodide.org/en/latest/development/abi.html) (which may be tied to updates in Emscripten and CPython versions, compiler/linker flags, Rust toolchain support and so on), or
+- when that alpha graduates to a stable release.
+
+## Background
+
+Pyodide ships/may ship two kinds of Python builds at any point in time (note that these numbers may not be up to date):
```
hoodmane:
```suggestion
Pyodide has two types of releases that matter to cibuildheel:
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 1. Add the new Python configuration
+
+In `cibuildwheel/resources/build-platforms.toml`, add an entry under `[pyodide]`:
+
+```toml
+{ identifier = "cp315-pyodide_wasm32", version = "3.15", default_pyodide_version = "315.0.0a1", node_version = "v24" },
+```
+
+`version` is the CPython version string, `default_pyodide_version` is the Pyodide release to use when the user does not pin one explicitly (use the latest available alpha/beta for a prerelease entry), and `node_version` is the minimum Node.js major required by that Pyodide release — check the [pyodide-build FAQ](https://pyodide-build.readthedocs.io/en/latest/faq.html#what-node-js-version-do-i-need) for a rudimentary idea of what the correct value is.
```
hoodmane, line 31: The `node_version` situation is a bit wishy-washy. We should improve our docs on node versions. No action needed here.

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+In `cibuildwheel/selector.py`, update the patterns in the `PyodidePrerelease` guards to match the new identifier:
+
+```python
+if EnableGroup.PyodidePrerelease not in self.enable and fnmatch(
+    build_id, "cp315-pyodide_*"
+):
+    return False
+```
+
+Also, update the comment above it to reflect the new version numbers and the instructions for when this guard should eventually be disabled.
```
hoodmane: Probably not needed to update this docs file, I think people can understand the pattern.

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 3. Generate and pin a constraints file
+
+The easiest way is to run the `update_constraints` `nox` session, which reads `build-platforms.toml` and regenerates all Pyodide constraints files automatically:
```
hoodmane: Let's not say how to run the generator script directly here because the nox code could change and we'd  unlikely to remember to update this. Anyone who wants to know what it does can read the noxfile.
```suggestion
Run the `update_constraints` `nox` session, which reads `build-platforms.toml` and regenerates all Pyodide constraints files automatically:
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+        -o cibuildwheel/resources/constraints-pyodide315.txt
+```
+
+### 4. Update tests and update CI configuration
+
+- Update the unit tests so the new identifier is accepted by the selector with `PyodidePrerelease` enabled and rejected without it. Pyodide-specific integration tests may also need their hardcoded expected-wheel lists extended.
+
+- Run the full test suite with `CIBW_PLATFORM=pyodide` and `CIBW_ENABLE=pyodide-prerelease` environment variables to make sure the new configuration is well exercised in CI.
+
+- In GHA, make sure `CIBW_ENABLE: pyodide-prerelease` is set in the Pyodide CI job so that the prerelease identifier is exercised in CI.
```
hoodmane: Does this need to be changed on every update? Surely we should set `CIBW_ENABLE=pyodide-prerelease` all the time, even when there isn't a prerelease?

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+## When a Pyodide prerelease becomes stable
+
+Pyodide uses a versioning scheme where the stable release for a given CPython version is named `[PythonMajorMinor].0.0`, so the first stable release shipping cp314 will be **`314.0.0`**. See [pyodide/pyodide#6084](https://github.com/pyodide/pyodide/issues/6084) for a rationale of this versioning scheme.
+
+### 1. Update the stable entries, and remove (or replace) the prerelease entry
+
+In `build-platforms.toml`, update the former prerelease entry's `default_pyodide_version` to the new stable release (e.g. `314.0.0`) and remove the prerelease marker from the identifier if present. Remove previous prerelease entries if they are now obsolete, or update them to the next prerelease if one is available. Based on your discretion, you may choose to keep the old stable entry for a while if it is still supported by Pyodide, or remove it immediately if it is already retired.
```
hoodmane, line 72: Can we simplify this to say just whatever the most likely thing is? We're talking to ourselves here, if we later disagree with the advice we can deviate from it or change it.

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 2. Disable or update the prerelease guards
+
+In `selector.py`:
+
+- **If a new prerelease is available**: update the `fnmatch` pattern to the next identifier (e.g. `cp315-pyodide_*`) as described above.
+- **If there is no new prerelease**L remove the `PyodidePrerelease` logic entirely.
```
hoodmane: We want to comment it right?
```suggestion
- **If there is no new prerelease** comment out the `PyodidePrerelease` logic.
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+
+Run `nox -s update_constraints` to regenerate the constraints file for the newly stable version. If the entry was already in `build-platforms.toml` as a prerelease, its constraints file already exists, and this step just refreshes it against the stable Pyodide release and updates the dependencies' versions.
+
+### 4. Update tests and CI configuration
+
+Update the unit tests so the newly stable identifier is accepted by the selector without needing `PyodidePrerelease` in the enable set. Pyodide-specific integration tests may also need their hardcoded expected-wheel lists extended.
+
+- Run the full test suite with `CIBW_PLATFORM=pyodide` environment variables to make sure the new configuration is well exercised in CI.
+
+- Remove `CIBW_ENABLE: pyodide-prerelease` from the GHA Pyodide CI job steps, as no prerelease identifiers should remain after the stable release ships.
```
hoodmane: Why not always set it?

cibuildwheel/resources/build-platforms.toml
```diff
@@ -225,8 +225,8 @@ python_configurations = [
 
 [pyodide]
 python_configurations = [
-  { identifier = "cp312-pyodide_wasm32", version = "3.12", default_pyodide_version = "0.27.7", node_version = "v22" },
```
hoodmane, line 228: Is there a reason to remove 3.12 here?

cibuildwheel/selector.py
hoodmane, line 1: Thanks for documenting this!

cibuildwheel/selector.py
```diff
@@ -96,17 +96,19 @@ def __call__(self, build_id: str) -> bool:
...
+        #      Pyodide constraints files in cibuildwheel/resources/constraints/ as necessary.
+        # When disabling, update the pattern to match the experimental Python version in case
+        # it is bumped to Python 3.15 (likely cp315-pyodide_* but could remain as 3.14 as well).
         # This depends on the CPython version being used in the Pyodide runtime at the time.
-        # if EnableGroup.PyodidePrerelease not in self.enable and fnmatch(
-        #     build_id, "cp313-pyodide_*"
-        # ):
-        #     return False
+        if EnableGroup.PyodidePrerelease not in self.enable and fnmatch(
+            build_id, "cp314-pyodide_*"
```
ryanking13, line 109: I think it is slightly uncomfortable that we need to turn this on and off every time we make an alpha release. Would there be a way to remove this?

As we standardized `pyemscripten` platform now, I think there is no reason to differentiate Pyodide prerelease and stable. Both versions should satisfy the pyemscripten platform constraints and users doesn't have to know whether they are using alpha xbuildenv or not.

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
+# Maintaining Pyodide support
+
+This page describes some short steps on how to update cibuildwheel's Pyodide platform code when the set of supported Pyodide Python versions changes – for example, when:
+
+- a new Pyodide alpha release arrives with support for a new [Pyodide ABI](https://pyodide.org/en/latest/development/abi.html) (which may be tied to updates in Emscripten and CPython versions, compiler/linker flags, Rust toolchain support and so on), or
```
ryanking13:
```suggestion
- a new Pyodide alpha release arrives with support for a new [PyEmscripten Platform](https://pyodide.org/en/latest/development/abi.html) (which is tied to updates in Emscripten and CPython versions, compiler/linker flags, and so on), or
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,109 @@
...
+- when that alpha graduates to a stable release.
+
+## Background
+
+Pyodide ships/may ship two kinds of Python builds at any point in time (note that these numbers may not be up to date):
+
+- **Stable** – the most recent full Pyodide release (e.g., `0.29.x` / cp313). This is enabled by default with no special `CIBW_ENABLE` flag needed.
+- **Prerelease** – an alpha/beta/rc Pyodide release that uses the _next_ CPython version (e.g., `314.0.0a1` / cp314). Users must opt in with `CIBW_ENABLE: pyodide-prerelease` to build against this version. This may or may not be available at any given time, depending on the Pyodide release cycle.
```
ryanking13, line 13: As I mentioned in the previous comment, I think we should remove differentiating Pyodide alpha from stable at least in cibuildwheel-side of view.

cibuildwheel/selector.py
hoodmane: We still need a notion of alpha versus stable I'd say. We can change the platform after alpha releases so people should wait for a stable release to upload wheels.

ryanking13: > We can change the platform after alpha releases

Hmm, my idea was to freeze the ABI as of alpha release, and that is what I was going to say in the blog post (https://github.com/pyodide/pyodide-blog/pull/66).

I think we all agreed freezing the ABI earlier before making a stable release (https://github.com/pyodide/pyodide/issues/5580), do you think we need another, say "beta" release for that then?

docs/_internal/pyodide-maintenance.md
agriyakhetarpal: Agreed

docs/_internal/pyodide-maintenance.md
agriyakhetarpal: You're right, no, I was wrong about this – I thought we would raise an error if `CIBW_ENABLE=pyodide-prerelease` is set and no Pyodide pre-release is available to be enabled, in cibuildwheel, but turns out this is a no-op. We silently don't select any pre-release builds if none are available, we're all good. I'm changing it so that it is always set.

cibuildwheel/resources/build-platforms.toml
agriyakhetarpal: Yes, since I don't think it's useful to have three ABIs enabled at once. We don't support anything but the latest version. Pyodide 0.29.3 and 314.0.0a1 have two different ABIs, which should be enough? Since we don't have a stable 314.0.0 release and IMO cibuildwheel shouldn't be as disruptive to users' workflows, I think it's okay to keep 0.29.3 – but three different ABIs becomes a bit excessive. People can still build for arbitrary (read: older) versions, such as Pyodide 0.27.7, using `CIBW_PYODIDE_VERSION`. Or would you prefer I keep three ABIs?

cibuildwheel/selector.py
agriyakhetarpal: Yes. I agree with @hoodmane here. I think we should still make the difference between alpha and stable. This is because even for PyPI and wheels for conventional (desktop) platforms, it is advised to only _test_ building wheels with an upcoming ABI – not upload them. Uploading to PyPI does not happen until there is a stable CPython release.

Doing this also protects us in the event where may need to change the ABI after our proclamation of stabilising it, due to any bugs or regressions that get caught. We only have to do this process a few times a year, at most, so I would personally be fine with it flipping the switches on and off in this manner.

docs/_internal/pyodide-maintenance.md
agriyakhetarpal: I think we need to decide on this before I change this sentence, because we want to be aligned on a policy for how many Pyodide versions/ABIs we should support at a time in cibuildwheel. So far, it's always been two at a time (or even one, some time back). Please read my other comment: https://github.com/pypa/cibuildwheel/pull/2812#discussion_r3073778240

cibuildwheel/selector.py
ryanking13: I see. I thought PyPI allows uploading Python wheels when CPython alpha or beta release is out. If it doesn't, then it would make sense for us to do the same approach as well.

agriyakhetarpal: Yes, I think CPython does so starting with b1, but it is still discouraged, IIRC. @henryiii, I assume you'd be the right person to know about this topic!

agriyakhetarpal: Even then, freezing the ABI and making the announcement is still fine I think – the more eyes we receive on this, the better our adoption will be for PEP 783, as downstream package authors will be better prepared to make the jump.

henryiii: CPython is not guaranteed to be ABI stable till the last beta release (historically, everyone basically uses RC1 as the first "stable" release, and that's what cibuildwheel does, but officially it's last beta, beta 4, at least in the past).

So you should not upload a wheel build with an early beta, but I don't think anyone stops it. In cibuildwheel you acknowledge it being a bad idea to upload with the `enable` opt-in.

IMO this should be removed whenever the ABI is stable. Doesn't have to match some special "final" release, but it has to be ensured to be stable.

cibuildwheel/resources/build-platforms.toml
henryiii: What's the usage of the old versions? IMO, people can upgrade, and probably should, but it's also easy to forget. I'm sure I have webpages that are using old versions.

I think the `pyemscripten` tag is _much_ more important than the one you can't upload to PyPI, so I think non-pyemscripten are really low priority, I don't mind dropping any of them.

test/utils.py
henryiii: If a patch release of pyodide 3.13 with updated micropip was possible, then I'd go:

* drop 3.12
* switch 3.13 to pyemscripten_2025
* keep 3.14 as "pre-release" until pyemscripten_2026 is declared stable.

If a patch release is hard to do... Well, that's uglier. Are these "normal" enough that `wheel tag` can re-tag?

cibuildwheel/resources/build-platforms.toml
hoodmane: Per jsdelivr, the most-used version is Python 3.12, followed by 3.13 and 3.10 which are almost tied.
```
Python 3.13: 2,033,897
Python 3.12: 6,602,877
Python 3.11:   697,276
Python 3.10: 1,929,739
```

cibuildwheel/platforms/pyodide.py
joerick: I think for this one, I'm happy for cibuildwheel to control the environment. PATH can have all sorts of things on it that cause users issues, actually. 

I also wonder if `PYODIDE_EMSCRIPTEN_DIR` would be a way for advanced users to work around this?

Seems like it'd do what we want-

```console
$ PYODIDE_EMSCRIPTEN_DIR=/tmp/somedir pyodide config get emscripten_dir
/tmp/somedir
```

cibuildwheel/resources/build-platforms.toml
joerick: > but three different ABIs becomes a bit excessive

What's the excess you're considering? The requirement for the user to grab 3 different emscripten toolchains?

I think that supporting 3 current + one prerelease version seems quite reasonable to me going forward. But perhaps we'd like to hide one or two behind a `pyodide-eol` enable flag to keep them off by default.

This current scenario might be an exception, though, as the package consumer is gonna need to upgrade something anyway, so I can see the argument to drop more aggressively here.

In summary, my (weakly held) opinion would be to drop 3.12 now, but going forward, I'm open to supporting more versions than the current behind a `pyodide-eol` flag.

cibuildwheel/selector.py
joerick: This line is required as long as the default_pyodide_version in [build-platforms](https://github.com/pypa/cibuildwheel/pull/2812/changes#diff-5a314b29ee6e9abb1efe8664680deafa073c8223dc351f7c45fbcb9a250a79cbR229) is set to a prerelease version. 

Once the first 'proper' pyodide release is out for 3.14, we'll announce it in the release notes, remove this line, our users will begin building for it, we'll never set that `default_pyodide_version` to a prerelease version again. So this only toggles on/off once per CPython minor release.

Then again, if the ABI will be frozen as of the first alpha, you could argue we don't need the `pyodide-prerelease` flag at all. I'd find that a little surprising - cibuildwheel's users would be getting alpha-quality software without specifically asking for it - but I'm not across the topic enough to know if that's a good idea or not.

docs/_internal/pyodide-maintenance.md
joerick: `bin/update_pythons.py` defines a lot of the logic, especially around default_pyodide_version. It might be cleaner when adding a version, to add a stub entry to `build-platforms.toml` and then run the script. If we need some behaviour that's not captured by the script, we'll have to update it, otherwise the `Update Dependencies` PRs will fight it.

docs/_internal/pyodide-maintenance.md
joerick, line 1: This is nice as a point-in-time capture of the current thinking! I'll just caution that these kinds of maintainer notes often go out-of-date or else require lots of maintenance as time goes on. Two ideas-

- It might be a good idea to chuck a "Last updated: April 2026" note in the file so we know how fresh it is.
- I'd recommend keeping it as short as possible, so we have less to maintain down the road!

But, if you're happy to keep it updated @agriyakhetarpal, and it's helpful for you, this sounds good to me.

joerick (commented):
Great work on this!

henryiii:
@agriyakhetarpal have you had a chance to review the feedback yet?

agriyakhetarpal:
Hi @henryiii, I looked at the feedback but I'm waiting for @hoodmane and @ryanking13 to release a v0.29.4 first so that I can add it for the 2025 ABI, thanks!

hoodmane:
I'm having a really hard time getting the 0.29.X branch into a releasable state still.

cibuildwheel/selector.py
agriyakhetarpal: I believe we will need a second 314.0 alpha, as @ryanking13 found the reason for a regression (or two) caught by a few package builders. I don't think that constitutes a change to the ABI itself, and thus it will be fruitful to wait – the `pyodide-prerelease` flag is still useful. I am also of the opinion that we users should start uploading 2026 ABI packages to PyPI after the alphas.

cibuildwheel/resources/build-platforms.toml
agriyakhetarpal: Okay, so I think `pyodide-eol` sounds like a good idea here – no need to drop support for older Pyodide versions if they're still going to work. I'll bump to Pyodide 0.29.4, that we released today, which allows pyemscripten wheels with the 2025 ABI.

Edit: okay, I have now added 3.12 back, behind a `pyodide-eol` flag. This means we will build for Pyodide 0.29.4/Python 3.13 by default. Users can build for Pyodide 314.0a1/Python 3.14 if they enable `pyodide-prerelease`.

cibuildwheel/platforms/pyodide.py
agriyakhetarpal: Yup, we have it here: https://github.com/pyodide/pyodide-build/blob/720edf797d6ae22cb82c1159bd8d787b716ec863/pyodide_build/config.py#L265. Happy to document this in a comment or so if you suggest, but I'd say we should wait until someone really has this need – I haven't come across anyone 😅

test/utils.py
agriyakhetarpal: There is now a v0.29.4 patch release: https://github.com/pyodide/pyodide/releases/tag/0.29.4. Updating to it!

agriyakhetarpal:
Ready for one more look!

cibuildwheel/resources/constraints-pyodide314.txt
agriyakhetarpal: I added `constraints-pyodide312.txt` back, so I think this is resolved now!

joerick (commented):
I think we might be missing docs on the new enable option in docs/options.md. 

also, I think it might be time for pyodide to lose its 'experimental' status in cibuildwheel? I noticed a note in docs/platforms.md. 

The rest looked good to me, although I only skim-read it!

joerick:
Oh, the other place to remove the experimental language would be the table at the top of the readme

agriyakhetarpal:
Good points @joerick, I've made those changes! Thank you!

agriyakhetarpal:
A bit of a note: since we do have `pyodide-version`/`CIBW_PYODIDE_VERSION` and it still works for any generalised Pyodide version case, people will be able to build for Pyodide beyond EOL if they would like to, so we'd consider `pyodide-eol` as more of a helper to do so.

docs/_internal/pyodide-maintenance.md
```diff
@@ -0,0 +1,113 @@
+# Maintaining Pyodide support
```
ryanking13, line 1: Thanks for writing this!

test/test_dependency_versions.py
```diff
@@ -140,11 +140,20 @@ def test_dependency_constraints(
...
-        "pip": "23.1.2",
-        "build": "1.2.2",
-        "delocate": "0.10.3",
-    }
+    if utils.get_platform() == "pyodide":
+        # pyodide-build 0.34+ requires build~=1.4.0, so we must use a
+        # compatible version here. delocate is macOS-only and not used on pyodide.
+        tool_versions = {
+            "pip": "23.1.2",
+            "build": "1.4.2",
```
ryanking13, line 148: what happens if we use build 1.4 or 1.5 for all cases?

ryanking13 (approved):
Thanks for working on this @agriyakhetarpal! It looks good to me overall. As you mentioned, probably we need a new alpha, but I think this PR can land without it.

test/test_dependency_versions.py
henryiii: It now supports 1.4 and 1.5, but it's highly constrained due to worries about it using build's internals. Those are generally okay, I think. I think the reason for 1.2.2 was likely older Python support, but we are dropping 3.9 so hopefully that's better now.

henryiii (approved):
Probably needs a followup removing pyodide 3.12.

henryiii:
@copilot resolve the merge conflicts in this pull request

(Edit: GitHub offered this as an option, but I don't think it works on cross-org pull requests)

EditEdit: Used Kimi-K2.6 locally.

agriyakhetarpal:
> Probably needs a followup removing pyodide 3.12.

Yes. Just leaving it here for later that it shouldn't be for this release, I think – we'll have 314.0 stable building for the 2026 ABI soon and then we'll be fine moving Pyodide 0.27.7/Python 3.12 to `pyodide-eol`. So I guess that can be for v4.1 or v4.2, based on when we get it done.

ubutatu (approved):

ubutatu (approved):

ubutatu (approved):

ubutatu (approved):

PR #2819, opened 2026-04-10
chore: add PEP 723 metadata for various `bin/` scripts, drop `bin` dependency group

agriyakhetarpal:
See https://github.com/pypa/cibuildwheel/pull/2816#discussion_r3065710834. This PR adds inline script metadata for all our scripts. Some scripts import cibuildwheel, and not all of these scripts need such metadata because they run within a `nox` session that installs the `bin` dependency group. Still, it's nice to see all the dependencies up top from a quick look :)

I had to move the Playwright Chromium installation to a subprocess, which is a bit hacky, but it works all well. I regenerated the image, wherein I realised that Playwright does the exact same thing I did in #2815 (capture a screenshot of the node), except the difference is the font. My macOS machine uses Helvetica Neue, while this one looks like a different one (DejaVu Sans? Not sure). I kinda like this one better!

Edit: I restored the image with what we have on `main`, as suggested.


bin/update_how_it_works_image.py
```diff
@@ -1,24 +1,23 @@
 #!/usr/bin/env python3
```
henryiii: These shabangs need to be updated too. I'd use `#/usr/bin/env -S uv run --script`, though you could use pipx instead if you want.

noxfile.py
```diff
@@ -191,9 +191,7 @@ def update_how_it_works_image(session: nox.Session) -> None:
     """
     pyproject = nox.project.load_toml()
     session.install("-e.", *nox.project.dependency_groups(pyproject, "docs"))
-    session.install("playwright")
-    session.run("playwright", "install", "chromium")
-    session.run("python", "bin/update_how_it_works_image.py")
+    session.install_and_run_script("bin/update_how_it_works_image.py")
```
henryiii, line 187: There might be others we can use install_and_run_script on? That will reduce the installs per session vs. installing `bin` every time. Maybe we can drop bin?

docs/data/how-it-works.png
henryiii, line 1: IMO, I'd be fine to just just leave this to reduce churn.

bin/update_how_it_works_image.py
```diff
@@ -1,24 +1,23 @@
...
 
+# /// script
+# dependencies = [
+#   "mkdocs",
+#   "playwright",
+# ]
+# ///
+
```
joerick, line 8: this is probably wrong, in the sense that to build the mkdocs, you need some extensions too. so the 'docs' dependency group really is the thing we'd need here :/

joerick (commented):
Some of the nox sessions install script deps themselves too, for example update_constraints installs click, that can probably also be DRY'd up with a call to install_and_run_script. Maybe that's the only example actually...

bin/update_how_it_works_image.py
henryiii: This one might be the one case where we don't want to use PEP 723, since it seems to need some extra setup. A little ironic given it was the one that triggered this. :)

Why does it need mkdocs to make an image, though? Couldn't we put this in an html file and then embed that in the docs and read it here too?

agriyakhetarpal: It doesn't – `mkdocs` is actually used only to build the docs. I guess it's probably better to run the docs nox session within the how-it-works session; that would be one way. I remember nox used to have a way to do that. Let me check!

bin/update_how_it_works_image.py
agriyakhetarpal: This works, will update!

bin/update_how_it_works_image.py
joerick: It was in there to get some base styles so it draws like it does in the docs. I had a little play with it and it looks to me that this would be equivalent-

```
            <link href="https://fonts.googleapis.com/css2?family=Lato&display=swap" rel="stylesheet">
            <style>
                html, body {
                  font-family: Lato, "Helvetica Neue", Helvetica, Arial, sans-serif;
                  font-weight: 400;
                  font-size: 16px;
                  color: #404040;
                  background: white;
                  margin: 0;
                  padding: 0;
                }
                * {
                  box-sizing: border-box;
                }
            </style>
```

agriyakhetarpal:
Locally, the session went from 50 seconds to 6 seconds (nice!):

```bash
nox > Running session update_how_it_works_image
nox > Creating virtual environment (uv) using python in .nox/update_how_it_works_image
nox > uv pip install playwright
nox > python bin/update_how_it_works_image.py
nox > Session update_how_it_works_image was successful in 6 seconds.
```

I restored the image with what we have on `main`, though, as suggested.

noxfile.py
```diff
@@ -23,7 +23,7 @@
 import nox
 
 nox.needs_version = ">=2025.2.9"
-nox.options.default_venv_backend = "uv|virtualenv"
+nox.options.default_venv_backend = "uv"
```
agriyakhetarpal: I suppose it's okay to make `uv` the default, since it's the runner that we are going to use for PEP 723 going forward. Also, we're at 0.9.18 now, so 0.1.23 was pretty old!

henryiii:
FWIW, I generally keep setup-python, as it's faster than uv on non-EoL Pythons since python is already installed on the runners, and not installing (<1 second) is faster than installing, even if uv is fast at installing (8 seconds on Windows).

henryiii:
Also setup-python installs universal macOS binaries, while uv doesn't.

agriyakhetarpal:
Thanks, I didn't know about that. Reverted!

noxfile.py
henryiii: This means that a user running our noxfile _must_ have uv installed. And we don't use uv to run it, we use nox (`install_and_run_script`)? `<any PEP 723 runner> script.py` works, the shabang is only for what is needed for `./script.py`.

agriyakhetarpal: Oh, I thought `nox` was delegating to `uv` for its PEP 723 runner implementation, instead of having its own. Thanks for the correction!

agriyakhetarpal: I'd prefer that people use `uv` for the bin scripts, though, or run them through `nox`. Unless you feel otherwise, or if there's a better way.

henryiii: If they use `./script.py`, sure, but they can use whatever they want if they run it themselves.

henryiii (approved):

PR #2825, opened 2026-04-22
feat: multiple files from workspaces

henryiii:
Looking at supporting uv build's multiple file output. This is a first pass at it. I've added a new test with two compiled wheels, and tried to generalize to multiple wheels. Making a PR now so I can look over it later, build in CI, & trigger GH review. I don't have a setup for running the non-unit tests at the moment.

:robot: Assisted-by: OpenCode:Kimi-K2.6 (open-source model)

### DIFF
--- a/cibuildwheel/logger.py
+++ b/cibuildwheel/logger.py
@@ -179,8 +179,8 @@ class Logger:
         print()
         print(f"{c.green}{s.done} {c.end}{self.active_build_identifier} finished in {duration_str}")
 
-        if isinstance(filename, Sequence):
-            if not filename:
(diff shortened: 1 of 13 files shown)


henryiii:
This is a really large change, and I'm not sure I can make it much smaller. One problem is the repair command; that now needs to run on every wheel. However, I did notice that `uv build` actually does have the ability to target one build at a time (in fact "all" is also an option you have to pass). I think that might be a better option for a fairly rare case, and I think it already works. (at least if nothing is standardized about workspaces, there's a little bit of discussion on standardizing them)

PR #2827, opened 2026-04-27
feat: config settings placeholders

henryiii:
Suggested on discord.

:robot: Assisted-by: Copilot:GPT-5.4

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -552,6 +552,9 @@ Specify config settings for the build backend. Each space separated
 item will be passed via `--config-setting`. In TOML, you can specify
 a table of items, including arrays.
 
+You can use the `{project}` or `{package}` placeholders in `config-settings`
+to refer to the project root or package being built, respectively.
+
 !!! tip
     Currently, "build" supports arrays for options, but "pip" only supports
     single values.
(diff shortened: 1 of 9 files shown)


freakboy3742 (commented):
Seems fairly straightforward - the only potential issue I can see on the iOS end (and it would be the same for Android and Emscripten) would be whether there's a need to reference a path that is relative to target machine - but AFAICT, the args being interpreted here are always going to be build arguments, so that won't be a factor.

docs/options.md
```diff
@@ -575,6 +578,14 @@ Platform-specific environment variables also available:<br/>
     CIBW_CONFIG_SETTINGS: "--build-option=--use-mypyc"
     ```
 
+    ```yaml
+    CIBW_CONFIG_SETTINGS_WINDOWS: "setup-args=--cross-file={project}/meson_cross_files/windows-386.ini"
+    ```
```
agriyakhetarpal, line 583: This is a bit too specific to my use case, so maybe we can make it more general? And maybe Linux, like this (also because I'm not sure if we need forward slashes or backward slashes on Windows in this case)?




```suggestion
    ```yaml
    CIBW_CONFIG_SETTINGS_LINUX: "setup-args=--cross-file={project}/cross_file.txt"
    ```
```

agriyakhetarpal (approved):
Thanks, @henryiii! One comment about the example being added, but this should work for my needs pretty well.

docs/options.md
henryiii: Windows generally always works with forward slashes, even in cmd. But that's fine.

PR #2830, opened 2026-05-01
fix: set `UV_PYTHON` for `before_build` on Linux when using `uv`

### LINKED ISSUE #1955: Installing dependencies before build fails when using `uv` and disabling build isolation
### Description

According to the cibuildwheel documentation, it is valid to disable build isolation for the wheel builds via
```toml
build-frontend = { name = "build[uv]", args = ["--no-isolation"] }
```
Naturally, this requires to install the build time dependencies manually.
However, specifying
```toml
before-build = "uv pip install scikit-build-core[pyproject]>=0.8.1 setuptools_scm>=7 pybind11>=2.13"
```
and then running cibuildwheel results in 
```console
Building cp38-manylinux_x86_64 wheel
CPython 3.8 manylinux x86_64

Setting up build environment...

    + which python
    + which uv

                                                              ✓ 0.07s
Running before_build...

    + sh -c 'uv pip install scikit-build-core[pyproject]>=0.8.1 setuptools_scm>=7 pybind11>=2.13'
error: No virtual environment found

                                                              ✕ 3.16s
Error: Command ['sh', '-c', 'uv pip install scikit-build-core[pyproject]>=0.8.1 setuptools_scm>=7 pybind11>=2.13'] failed with code 2. 
```

This only seems to be happening under linux because for all the other operating systems, cibuildwheel explicitly calls `uv venv ...` in the "Setting up build environment..." step.

### Build log

_No response_

### CI config

_No response_

agriyakhetarpal:
Fixes #1955. On Linux, we do not create a virtualenv for the build environment, unlike macOS and Windows which call `virtualenv()` before `before_build` runs. This means `uv pip install` in `before_build` fails with `error: No virtual environment found` when using the `build[uv]` as the build frontend, regardless of whether build isolation is disabled. 

We can set `UV_SYSTEM_PYTHON=1` to allow uv to install directly into the container's Python. The build-step environment can be left unchanged, so that isolated builds continue to manage their own internal venvs as usual.

I am unsure how to add a test for this, so I haven't added one.

### DIFF
--- a/cibuildwheel/platforms/linux.py
+++ b/cibuildwheel/platforms/linux.py
@@ -264,7 +264,15 @@ def build_in_container(
                     project=container_project_path,
                     package=container_package_dir,
                 )
-                container.call(["sh", "-c", before_build_prepared], env=env)
+                before_build_env = env.copy()
+                if use_uv:
+                    # On Linux, no virtualenv is created for the build environment
+                    # (unlike macOS/Windows, where one is set up before before_build
+                    # runs). uv requires either an active venv or UV_SYSTEM_PYTHON=1
+                    # to install packages. Set it here so that `uv pip install` works
+                    # in before_build without requiring users to pass --system.
+                    before_build_env.setdefault("UV_SYSTEM_PYTHON", "1")
+                container.call(["sh", "-c", before_build_prepared], env=before_build_env)
 
             log.step("Building wheel...")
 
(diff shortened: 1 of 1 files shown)


henryiii (approved):
I think this makes sense.

mayeut:
I think this makes sense but I'd rather use [UV_PYTHON](https://docs.astral.sh/uv/reference/environment/#uv_python) to be 100% explicit.

agriyakhetarpal:
Nice, makes sense. Thanks, @mayeut! Addressed via 0d73d519befaed98c7d0a9d0374e53fa19570584

Issue #2828, opened 2026-04-29
Plan for cibuildwheel 4.0


henryiii:
I'd like to propose the following:

Drops:
* Drop pyodide wheels and pyodide 3.12, only make pyemscripten wheels for 3.13 and 3.14 (alpha pyodide) #2812 
* Drop Python 3.8 #2686
* Drop Python 3.13t #2684

Additions
* #2833
* #2805
* #2831

Some assorted changes/fixes included too, those are just the main things.

3.15 and manylinux removing stuff will hit in about a week (5-5-2026), so shortly after that we could release what I'm proposing to call 4.0, due to the fairly big drops and pyodide -> pyemscripten change.

*Maintainers, feel free to edit directly. Anyone can comment.*

agriyakhetarpal:
Since we're considering an audit command to land, let's also consider adding `delvewheel` as a default Windows repair command, as @joerick is interested in doing so in https://github.com/pypa/cibuildwheel/issues/2803#issuecomment-4170162853. I am too. Users can still disable it, obviously. The only problem I can think of with that happening is the things we can't control on the GHA runner + MSVC combo's side, see https://github.com/adang1345/delvewheel/issues/54 for an example. But that's fairly old, and was resolved quickly, so most likely not a big problem.

I'm happy to work on this if you also think it's good to add. 😄 

I would also say we could drop support for macOS Intel, but that's a 🌶️ change and we might need some numbers before we do that.

henryiii:
I'd say yes for windows repair.

For macOS, if 3.15 supports Intel, we should too, otherwise it will be hard to build Intel wheels. This year major Intel runners go away, though, so that's probably when we will be likely to drop too. We could make Intel runners deprecated in 4.0, though, so we have a bette excuse to drop them without a major bump, probably near the end of the year.

About a week ago I got my new laptop and am no longer often on an Intel Mac. :)

agriyakhetarpal:
Okay, great. Let's mark macOS Intel as officially deprecated. It looks like [the macos-15-intel runner disappears next year in Fall 2027](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/#notice-of-macos-x86_64-intel-architecture-deprecation), so we could run CI on them until then.

Congratulations on your new laptop! I hope you enjoy it!

xavier2k6:
There are `macos-26-intel` runner images available:

* https://github.com/actions/runner-images#available-images

[...]

Issue #1567, opened 2023-08-08
Weird behaviour with Windows wheels repair


gnikolov95:
### Description

When I try to run the following git workflow: https://gist.github.com/gnikolov95/fa1f95d39bed741ebc559bbd3b01ec56
It fails when it tries to repair the wheel, raises a stop iteration error.

This is the relevant part from .toml:
```toml
# Use delvewheel on windows
[tool.cibuildwheel.windows]
before-build = [
    "pip install delvewheel",
    "if exist build rmdir /s /q build",
]
repair-wheel-command = ["delvewheel repair -w {dest_dir} {wheel}", "pipx run abi3audit --strict --report {wheel}"
]

[tool.cibuildwheel.linux]
before-build = "rm -rf {project}/build"
repair-wheel-command = [
  "auditwheel repair -w {dest_dir} {wheel}",
  #"pipx run abi3audit --strict --report {wheel}",
]


[tool.cibuildwheel.macos]
before-build = "rm -rf {project}/build"
repair-wheel-command = [
  "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}",
  #"pipx run abi3audit --strict --report {wheel}",
]
```

Everything is pretty much created from the examples on the documentation page. I thought that maybe I am somehow messing up the `dest_dir` and `wheel` dirs but it works for macOS and Linux.

The exception stack

```
2023-08-08T16:39:24.4888381Z repairing C:\Users\runneradmin\AppData\Local\Temp\cibw-run-7_2kent2\cp311-win_amd64\built_wheel\<redacted>-0.0.1-cp311-cp311-win_amd64.whl
2023-08-08T16:39:24.4889297Z finding DLL dependencies
2023-08-08T16:39:24.4889904Z no external dependencies are needed
2023-08-08T16:39:24.8214620Z Traceback (most recent call last):
2023-08-08T16:39:24.8223942Z   File "<frozen runpy>", line 198, in _run_module_as_main
2023-08-08T16:39:24.8224862Z   File "<frozen runpy>", line 88, in _run_code
2023-08-08T16:39:24.8225478Z   File "C:\Program Files (x86)\pipx\.cache\637068be72fe8b2\Scripts\cibuildwheel.exe\__main__.py", line 7, in <module>
[...]

Issue #459, opened 2020-11-06
Repairing Windows wheels


myselfhimself:
Hello,
I am attempting to build a wheel for a [C++ Python binding of G'MIC](https://github.com/myselfhimself/gmic-py/) with .dll dependents of a wheel file on Windows MSYS2 with Github Actions.
This is for python 3.8 for now. Once I have understood how to "repair" the wheel by including the dependent .dll files into the wheel file, I will proceed to use cibuildwheel most probably. My goal is to iterate on Python versions and use my MSYS2 gcc/g++ and bash commands instead of using Visual Studio.
I will give a count of my progress and struggles using cibuildwheel here. I am already using the latter for MacOS Clang builds (though disabled again for now).

Related to #329 and https://github.com/myselfhimself/gmic-py/issues/5 as well as [this gmic-py windows-support pull-request](https://github.com/myselfhimself/gmic-py/pull/74)

myselfhimself:
One thing which I have not yet understood, is if `cibuildwheel` already has a default windows repair command in place, or if it does nothing for now, but just has the slot environment variable for injecting such a repair command.

My idea for repairing under msys2 is to run ldd on the compiled C++/Python .dll and copy all matching /mingw64/bin dll files into a directory, which then would be included within the wheel and if needed loaded again at module import time.

myselfhimself:
I have a bit through related https://github.com/joerick/cibuildwheel/blob/master/cibuildwheel/windows.py and https://github.com/joerick/cibuildwheel/blob/master/examples/github-minimal.yml already.

mayeut:
You might want to have a look at https://discuss.python.org/t/delocate-auditwheel-but-for-windows/2589

myselfhimself:
Hello yes I had found that discussion exactly as well and the last
commenting person Vinayak Mehta created a script named wheel_repair.py and
leveraging machomangler for repairing windows wheels
<https://github.com/vinayak-mehta/pdftopng/blob/6e4a9d589228f66dfbb6f0197740b0b7c1bef9b3/scripts/wheel_repair.py>
where there the dependent dlls that get inserted into the wheel are even
renamed to avoid conflicts with same named preinstalled dll on any target
install.
One issue is that this script is not yet a separate python package for
example, it is in the middle of a pdf2png project.

so 7. 11. 2020 v 12:11 odesílatel Matthieu Darbois <notifications@github.com>
napsal:

myselfhimself:
[...]

PR #2831, opened 2026-05-01
Set up `delvewheel` as the default repair wheel command for Windows

agriyakhetarpal:
This PR adds [`delvewheel`](https://github.com/adang1345/delvewheel) in the usual places for it to the default wheel repair command for our Windows platform in cibuildwheel v4. The current version is v1.12.0, and the project has become fairly stable as of 2026.

I think I covered everything, but please LMK if I missed anything. @joerick, since you expressed interest in adding this in various issues, it'd be great if you could take a look – and by all means, please feel free to push any further changes to my PR!

Also, cc: @adang1345 (I hope it is okay for us to ping you here!)

Thank you! :)

- Closes #2803
- xref #2828
- xref #459, https://github.com/pypa/cibuildwheel/issues/459#issuecomment-830051388
- delvewheel was first suggested in https://github.com/pypa/cibuildwheel/discussions/620 and it's finally happening!
- see also https://github.com/pypa/cibuildwheel/discussions/1798

Also, perhaps we should look at #1567 too while we're at it, based on the open issues that reference "delvewheel".


agriyakhetarpal:
One of the concerns we had was listed in https://github.com/pypa/cibuildwheel/issues/2047#issuecomment-2441734095, about users having to add `--add-path` to the Windows wheel repair. I think that can be resolved if we make a note of this in the docs, but I will wait for preliminary review suggestions to come, first.

nilsnolde:
Thanks, I appreciate that:) I'll just reiterate that IMO this is a more general Windows limitation, they don't have a proper system package manager (EDIT: though I did see a `winget` randomly recently) and the way I always use it is to reference vcpkg/conan dll's (there's probably more scenarios for sure, but this might be the most dominant). Adding it to the docs will be appreciated by everyone. Much better than the current situation:)

agriyakhetarpal:
> Thanks, I appreciate that:) I'll just reiterate that IMO this is a more general Windows limitation, they don't have a proper system package manager (EDIT: though I did see a `winget` randomly recently) and the way I always use it is to reference vcpkg/conan dll's (there's probably more scenarios for sure, but this might be the most dominant). Adding it to the docs will be appreciated by everyone. Much better than the current situation:)

Good point, thanks @nilsnolde. I agree that vcpkg is the closest to this kind of thing out there (and even Meson wraps is one more option). I added a note in 8cfdbf76c0f7aa8d9d027f93a6dd6ba35e9cb45d. Does that suffice? :)

cibuildwheel/platforms/windows.py
```diff
@@ -540,8 +566,19 @@ def build(options: Options, tmp_path: Path) -> None:
...
+                    if (
+                        config.identifier.split("-")[0].startswith("gp")
+                        and "delvewheel" in repair_command
+                    ):
+                        # python-native.dll is GraalPy's runtime library (analogous to
+                        # CPython's python3X.dll) and shouldn't be bundled into the wheel.
+                        # See: https://github.com/oracle/graalpython/blob/eab533503e96441c64e2427a8b81b5b38117c899/scripts/wheelbuilder/repair_wheels.py#L66-L82
+                        repair_command = repair_command.replace(
+                            "delvewheel repair", "delvewheel repair --exclude python-native.dll", 1
+                        )
```
agriyakhetarpal: Hi @timfel, I hope it's okay to ping you here! I was wondering if `python-native.dll` could be handled by/upstreamed into delvewheel itself? This should work for now, but I think https://github.com/adang1345/delvewheel/blob/2e72f3c1b91ec5dd681b4af85877c717a97273f7/delvewheel/_dll_list.py could be the right place for this (probably under a guard that checks that the Python implementation is `graalpy`).

Also, it'd be great if you could skim through this PR in general as well :) When I searched for how GraalPy handles Windows wheel repairs, I came across this bug that someone filed quite recently: https://github.com/oracle/graalpython/issues/762, and I was curious whether this could affect users and if we should have a different `delvewheel` version for GraalPy in that scenario. Thank you!

agriyakhetarpal:
Note to self: `test_abi3` fails with an issue that I've previously seen in `delvewheel`'s issue tracker or elsewhere with antivirus stuff:

```
OSError: [Errno 5] The process cannot access the file because it is being used by another process



During handling of the above exception, another exception occurred:



  File "C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\graalpy-25.0.2-windows-amd64\Lib\runpy.py", line 198, in _run_module_as_main

    return _run_code(code, main_globals, None,

  File "C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\graalpy-25.0.2-windows-amd64\Lib\runpy.py", line 88, in _run_code

    exec(code, run_globals)

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Scripts\delvewheel.exe\__main__.py", line 5, in <module>

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\delvewheel\__main__.py", line 121, in main

    wr.repair(args.target, no_mangles, args.no_mangle_all, args.with_mangle, args.strip, args.lib_sdir, not args.no_diagnostic and 'SOURCE_DATE_EPOCH' not in os.environ, namespace_pkgs, args.include_symbols, args.include_imports, args.custom_patch)

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\delvewheel\_wheel_repair.py", line 849, in repair

    discovered, associated, ignored = _dll_utils.get_all_needed(executable_path, self._exclude, self._wheel_dirs, 'raise', include_symbols, include_imports)[:3]

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\delvewheel\_dll_utils.py", line 454, in get_all_needed

    with PEContext(dll_info[0], None, False) as pe2:

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\delvewheel\_dll_utils.py", line 55, in __init__

    self._pe = pefile.PE(path, data)

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\pefile.py", line 2941, in __init__

    self.__parse__(name, data, fast_load)

  File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\build\venv\Lib\site-packages\pefile.py", line 3038, in __parse__

    raise Exception(

Exception: Unable to access file 'C:\hostedtoolcache\windows\Python\3.13.13\x64\vcruntime140.dll': [Errno 5] The process cannot access the file because it is being used by another process

Error: cibuildwheel: Command delvewheel repair --exclude python-native.dll -w C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\repaired_wheel -v C:\Users\runneradmin\AppData\Local\Temp\cibw-run-f6u_6o0x\gp312_250-win_amd64\built_wheel\spam-0.1.0-graalpy312-graalpy250_312_native-win_amd64.whl failed with code 1.
```

cibuildwheel/platforms/windows.py
timfel: Yes, recent versions of delvewheel regressed for GraalPy. @msimacek fixed that on master, but it'll be a month until our next release.

Regarding that dll, yes, in our own scripts we also just added that exclusion, but it's probably better to upstream that. I'll try to get to it

timfel: also, I think we should add the labels here to test pypy and graalpy, do you agree?

agriyakhetarpal: Agreed, just added!

nilsnolde:
> I added a note in https://github.com/pypa/cibuildwheel/commit/8cfdbf76c0f7aa8d9d027f93a6dd6ba35e9cb45d. Does that suffice? :)

@agriyakhetarpal IMO sounds perfect, thanks!

timfel (approved):

adang1345:
By the way, I just made a new delvewheel release 1.12.1, which automatically excludes the DLLs provided by or required by GraalPy (`vcruntime140.dll`, `vcruntime140_1.dll`, and `python-native.dll`).

agriyakhetarpal:
Awesome, thanks a lot @adang1345!

docs/options.md
```diff
@@ -946,29 +947,35 @@ The command is run in a shell, so you can run multiple commands like `cmd1 && cm
 Platform-specific environment variables are also available:<br/>
 `CIBW_REPAIR_WHEEL_COMMAND_MACOS` | `CIBW_REPAIR_WHEEL_COMMAND_WINDOWS` | `CIBW_REPAIR_WHEEL_COMMAND_LINUX` | `CIBW_REPAIR_WHEEL_COMMAND_ANDROID` | `CIBW_REPAIR_WHEEL_COMMAND_IOS` | `CIBW_REPAIR_WHEEL_COMMAND_PYODIDE`
 
-!!! tip
-    cibuildwheel doesn't yet ship a default repair command for Windows.
+!!! note "Windows: telling delvewheel where to find DLLs"
+    On Windows, delvewheel searches the wheel itself and the directories on `PATH` for external DLL dependencies. If your DLLs are already discoverable via `PATH`, (say, installed by a package manager that adds itself and the relevant directories to `PATH`), the default repair command should be sufficient.
```
adang1345: Actually, delvewheel does not search the wheel itself for external DLL dependencies by default. It just searches `PATH`. The `--ignore-existing` option must be passed if searching the wheel itself is desired. I think you should strike out "the wheel itself" from the documentation.

agriyakhetarpal: Thanks for the correction!

agriyakhetarpal:
```suggestion
    On Windows, delvewheel searches the directories on `PATH` for external DLL dependencies. If your DLLs are already discoverable via `PATH`, (say, installed by a package manager that adds itself and the relevant directories to `PATH`), the default repair command should be sufficient.
```

test/test_windows.py
```diff
@@ -10,6 +10,25 @@
 basic_project = test_projects.new_c_project()
 
 
+def test_delvewheel_default_on_windows(tmp_path: Path) -> None:
```
joerick: I don't see what this test is checking, perhaps we just verify that delvewheel ran by checking its output appears in the logs?

test/test_windows.py
```diff
@@ -10,6 +10,25 @@
...
+    actual_wheels = utils.cibuildwheel_run(
+        project_dir, add_args=["--archs", "native"], single_python=True
+    )
+
+    assert len(actual_wheels) == 1
+    assert "none-any" not in actual_wheels[0]
+    tag = "cp{}{}".format(*utils.SINGLE_PYTHON_VERSION)
+    assert actual_wheels[0].startswith(f"spam-0.1.0-{tag}-{tag}-win_")
+
+
```
joerick, line 82: Another test that would be good to see would be that users can disable delvewheel by setting the repair env var to the empty string. Just because that will be our advice to upgrading users if they get false negatives and I want to make sure it works :)

joerick (commented):
Thanks for working on this agriya!

Looks good from my end overall.

joerick:
While I'm thinking about it, this is a change that needs a good notice in release notes. Things to mention in the release notes

- delvewheel runs by default , it will try to find and bundle dlls into your wheel
- you can configure using repair-wheel-command, or 
- disable it (returning to pre-4.0 behaviour) by setting repair-wheel-command to empty-string. 

I think the crucial thing to mention in release notes will be how to disable it for upgrading users that don't want this change yet.

test/test_windows.py
agriyakhetarpal: This test was checking that delvewheel ran and produced a wheel that cibuildwheel can see when it's coming out of the repair step, i.e., that delocate doesn't consume the wheel and not output one due to any errors, and the build and repair pipeline passes. But yeah, it is a bit of a redundant test, since we have others that do the same. Dropping it!

test/test_windows.py
agriyakhetarpal: Done!

agriyakhetarpal:
FYI, I partially reverted the constraints bumps in 17d38bab10cc21c6ac6ef015b46a09973ba57346 because there was a failure likely due to `pip==26.1.1` in https://github.com/pypa/cibuildwheel/actions/runs/25509576233/job/74864644182?pr=2831, which I was kind of stumped by, where GraalPy's `pip` was unable to install `delvewheel` at all. cc: @timfel for visibility. `pip==26.0.1` works fine.

agriyakhetarpal:
I guess the failure with the newer virtualenv will be fixed by https://github.com/pypa/virtualenv/pull/3144.

joerick (commented):
Only a quick comment on tests- apologies if I wasn't clear in the previous review! But I would feel better with a couple of _integration_ tests for this feature a) verifying that delvewheel runs, and b) verifying that it can be turned off.

One possibility would be to construct a test project that delvewheel can't find the libraries for. Then the default setting would show a failure (where we can assert the delvewheel error message) and then we disable delvewheel using repair-command = "", and that run completes with success.

If that's hard we could just assert in the logs (using pytest's `capfd`) that delvewheel runs by default, and doesn't run when disabled.

(the test could be `single_python` as we're just checking our internal logic really)

I'd also be happy to merge as-is and add the above regression test as a followup, if we're tight on time.

agriyakhetarpal:
> Only a quick comment on tests- apologies if I wasn't clear in the previous review! But I would feel better with a couple of _integration_ tests for this feature a) verifying that delvewheel runs, and b) verifying that it can be turned off.
> 
> One possibility would be to construct a test project that delvewheel can't find the libraries for. Then the default setting would show a failure (where we can assert the delvewheel error message) and then we disable delvewheel using repair-command = "", and that run completes with success.
> 
> If that's hard we could just assert in the logs (using pytest's `capfd`) that delvewheel runs by default, and doesn't run when disabled.
> 
> (the test could be `single_python` as we're just checking our internal logic really)
> 
> I'd also be happy to merge as-is and add the above regression test as a followup, if we're tight on time.

Thanks for the feedback! I think we had a test like this that I then removed in 3187f898bb8cdc7bfda36bfa99280401cd5abf87 after https://github.com/pypa/cibuildwheel/pull/2831#discussion_r3207136014. There's no equivalent test for auditwheel or delocate, so consistency called for doing so, but I think it's probably alright to bring it back with `capfd` if you insist! Give me a moment.

As for a test project, that's easy to do; we can mock the PE file to make it believe that a particular DLL is required (which won't be found because it doesn't exist). I'll add this.

agriyakhetarpal:
Let's see if this works!

agriyakhetarpal:
Something is pulling in 26.1.1 from a venv somewhere, but I can't yet figure out what. It's stopping the tests from proceeding, then moving on to the tests we actually want to see run (the C extension with an unfindable DLL).

joerick (approved):
Brilliant, that's perfect.

joerick:
> There's no equivalent test for auditwheel or delocate, so consistency called for doing so, but I think it's probably alright to bring it back

I'd reckon our integration tests do cover auditwheel, because the 'manylinux' filename is set by that tool. Not checking delocate on macOS is maybe an opportunity for better test coverage :)

henryiii (approved):

PR #2832, opened 2026-05-01
Use `python -V -V` for Windows build diagnostics

### LINKED ISSUE #1218: Do `python -V -V` instead of just `python --version` to help debugging
### Description

This is just a suggestion to do `python -V -V` instead of just `python --version` in https://github.com/pypa/cibuildwheel/blob/35c795e4c6fcb627dc1b07998ee28f3310f9a768/cibuildwheel/windows.py#L190 to be able to display the compiler (e.g. `MSC v.1916`, see tables on https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B for toolset (MSVC++ version) approximate correspondence to use with `CMake -T` option, here it would be `-Tv141`) that was used to build python.exe to help debugging (related to  https://cibuildwheel.readthedocs.io/en/stable/faq/#windows-importerror-dll-load-failed-the-specific-module-could-not-be-found problem and potentially other compatibility and dependencies problems...).

### Build log

_No response_

### CI config

_No response_

Herrtian:
Fixes #1218.

This changes the Windows setup logging from `python --version` to `python -V -V` so the log includes the extra build/compiler details when Python provides them.

Checked with:
- `uv run --with ruff ruff check cibuildwheel/platforms/windows.py`
- `uv run --group test pytest unit_test/get_platform_test.py -q`

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -229,7 +229,10 @@ def setup_env(
                 f"or insert {command} above it."
             )
             raise errors.FatalError(msg)
-        call(command, "--version", env=build_env)
+        if command == "python":
+            call(command, "-V", "-V", env=build_env)
+        else:
+            call(command, "--version", env=build_env)
 
     # Construct an altered environment which simulates running on Android.
     android_env = setup_android_env(config, python_dir, venv_dir, build_env)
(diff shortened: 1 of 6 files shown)


agriyakhetarpal (commented):
Hi @Herrtian, thanks for contributing! Could you please make the same change across all the platforms, besides Windows, for consistency?

Also, it looks like we aren't calling `python --version` at all for Linux, so we could add it there.

Herrtian:
Thanks, I updated this in 912c25e. The version logging now uses python -V -V across the existing platform setup paths, and Linux logs the Python version too. CI is green now.

agriyakhetarpal (approved):
Nice, thanks @Herrtian!

PR #2834, opened 2026-05-02
Document platform-specific before-build config

### LINKED ISSUE #1578: Add examples in docs about setting "per-platform" `before-build` options in `pyproject.toml`
### Description

:wave: Loving using this. Thanks y'all!

One thing I hit:

https://cibuildwheel.readthedocs.io/en/stable/options/#before-build lists "per-platform" env vars (e.g. `CIBW_BEFORE_BUILD_MACOS`), but trying:

```toml
[tool.cibuildwheel]
before-build-macos = "..."
```

I get
```
cibuildwheel.options.ConfigOptionError: "Option 'before-build-macos' not supported in a config file. Perhaps you meant 'before-build'?"
```

### Build log

_No response_

### CI config

_No response_

Herrtian:
Fixes #1578.

This adds a short note that pyproject.toml uses platform tables for platform-specific options, and includes efore-build examples for Linux and macOS.

Checked with:
- `git diff --check`
- `python -m cibuildwheel --help`

I also tried `mkdocs build --strict` on Windows, but the docs macro calls the `cibuildwheel` console script without `.exe`, and the local Windows subprocess lookup failed before rendering the page.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -793,6 +793,9 @@ The command is run in a shell, so you can write things like `cmd1 && cmd2`.
 Platform-specific environment variables are also available:<br/>
  `CIBW_BEFORE_BUILD_MACOS` | `CIBW_BEFORE_BUILD_WINDOWS` | `CIBW_BEFORE_BUILD_LINUX` | `CIBW_BEFORE_BUILD_ANDROID` | `CIBW_BEFORE_BUILD_IOS` | `CIBW_BEFORE_BUILD_PYODIDE`
 
+In configuration files, use platform tables like `[tool.cibuildwheel.macos]`,
+rather than platform-suffixed option names like `before-build-macos`.
+
 #### Examples
 
 !!! tab examples "pyproject.toml"
@@ -816,6 +819,13 @@ Platform-specific environment variables are also available:<br/>
 
     # If cibuildwheel is called with a package_dir argument, it's available as {package}
     before-build = "{package}/script/prepare_for_build.sh"
+
+    # Use a different command on a specific platform
+    [tool.cibuildwheel.linux]
+    before-build = "python scripts/install-linux-deps.py"
+
+    [tool.cibuildwheel.macos]
+    before-build = "python scripts/install-macos-deps.py"
     ```
 
     In configuration files, you can use a array, and the items will be joined
--- a/docs/options.md
+++ b/docs/options.md
@@ -845,6 +845,10 @@ In configuration files, use platform tables like `[tool.cibuildwheel.macos]`.
 
     # If cibuildwheel is called with a package_dir argument, it's available as {package}
     CIBW_BEFORE_BUILD: "{package}/script/prepare_for_build.sh"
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -793,6 +793,9 @@ The command is run in a shell, so you can write things like `cmd1 && cmd2`.
 Platform-specific environment variables are also available:<br/>
  `CIBW_BEFORE_BUILD_MACOS` | `CIBW_BEFORE_BUILD_WINDOWS` | `CIBW_BEFORE_BUILD_LINUX` | `CIBW_BEFORE_BUILD_ANDROID` | `CIBW_BEFORE_BUILD_IOS` | `CIBW_BEFORE_BUILD_PYODIDE`
 
+In configuration files, use platform tables like `[tool.cibuildwheel.macos]`,
+rather than platform-suffixed option names like `before-build-macos`.
+
```
agriyakhetarpal: For brevity and to not confuse the reader any more than they would be when reading our docs :)

```suggestion
In configuration files, use platform tables like `[tool.cibuildwheel.macos]`.

```

agriyakhetarpal (approved):

docs/options.md
```diff
@@ -793,6 +793,8 @@ The command is run in a shell, so you can write things like `cmd1 && cmd2`.
 Platform-specific environment variables are also available:<br/>
  `CIBW_BEFORE_BUILD_MACOS` | `CIBW_BEFORE_BUILD_WINDOWS` | `CIBW_BEFORE_BUILD_LINUX` | `CIBW_BEFORE_BUILD_ANDROID` | `CIBW_BEFORE_BUILD_IOS` | `CIBW_BEFORE_BUILD_PYODIDE`
 
+In configuration files, use platform tables like `[tool.cibuildwheel.macos]`.
```
henryiii, line 796: Why do we have this note here but not on every other config option? (Maybe the argument is that this one is more likely to need per-platform config?)

docs/options.md
```diff
@@ -816,6 +818,13 @@ Platform-specific environment variables are also available:<br/>
 
     # If cibuildwheel is called with a package_dir argument, it's available as {package}
     before-build = "{package}/script/prepare_for_build.sh"
+
+    # Use a different command on a specific platform
+    [tool.cibuildwheel.linux]
+    before-build = "python scripts/install-linux-deps.py"
+
+    [tool.cibuildwheel.macos]
+    before-build = "python scripts/install-macos-deps.py"
```
henryiii, line 827: We usually keep the config settings and environment settings tabs matching. We should have these examples there too, I think.

Herrtian:
Thanks, pushed ea0cb64 to add the matching env-var examples and kept the shorter config note from the earlier suggestion. Checked git diff --check locally; pre-commit also got through the markdown hooks, but cog hit a Windows encoding issue while reading README/changelog, so I left that as local-only.

henryiii (approved):

Issue #1190, opened 2022-07-23
Question: How to get the current target architecture


daquexian:
### Description

Thanks for the great tool! It helps me a lot.

I'm building a wheel for apple m1 and from the documentation I can set `CIBW_ARCHS_MACOS` to `x86_64 universal2`. However, I need to also set `CMAKE_OSX_ARCHITECTURES` to `x86_64;arm64`  in my CMakeLists.txt to actually build arm64 binaries. But in the docs, I don't find how to get the current target arch, so that I can set `CMAKE_OSX_ARCHITECTURES` accordingly.

henryiii:
https://github.com/pybind/cmake_example/blob/88a21c216921d85a17cff34dede78b5d35464455/setup.py#L102

This will work anywhere that cross compiling via setuptools is supported, not just cibuildwheel.

(Or use scikit-build)

daquexian:
Thanks! Maybe it will be better to describe "ARCHFLAGS" in cibuildwheel docs? I have seen a project using a sub-optimal way to build m1 wheel because its maintainers don't know the existence of ARCHFLAGS: https://github.com/nmwsharp/robust-laplacians-py/pull/8/files.

XuehaiPan:
I have run into the same issue. I found `cibuildwheel` sets `ARCHFLAGS` for cross-compiling.

https://github.com/pypa/cibuildwheel/blob/4786f528a9a4d21ad762551b2cc484b11b716f8c/cibuildwheel/macos.py#L219-L232

It would be nice to set `CMAKE_OSX_ARCHITECTURES` as well by `cibuildwheel`.

agriyakhetarpal:
Looks like `CMAKE_OSX_ARCHITECTURES` is indeed popular: https://github.com/search?q=%22CMAKE_OSX_ARCHITECTURES%22+language:Python+&type=code

agriyakhetarpal:
Based on discussions in #2835, I think it's probably alright to mark this issue as resolved with no action required for cibuildwheel. We could still consider adding `ARCHFLAGS` or `CMAKE_OSX_ARCHITECTURES` to the docs, but as the activity here suggests, not a lot of people are bumping into this issue over the years. We can always reopen this again or link to it in other threads where needed.

PR #2835, opened 2026-05-02
Set CMAKE_OSX_ARCHITECTURES for macOS builds

Herrtian:
Refs #1190.

This sets `CMAKE_OSX_ARCHITECTURES` in the macOS build environment for `x86_64`, `arm64`, and `universal2` builds, alongside the existing `ARCHFLAGS` setup. It uses `setdefault`, so projects that already set `CMAKE_OSX_ARCHITECTURES` keep their explicit value.

Checked with:
- `.\.venv\Scripts\python.exe -m pytest unit_test\macos_environment_test.py -q`
- `uv run --with ruff ruff check cibuildwheel\platforms\macos.py unit_test\macos_environment_test.py`
- `git diff --check`

### DIFF
--- a/cibuildwheel/platforms/macos.py
+++ b/cibuildwheel/platforms/macos.py
@@ -205,6 +205,25 @@ def install_graalpy(tmp: Path, url: str) -> Path:
     return installation_path / "bin" / "graalpy"
 
 
+def _setup_arch_environment(python_identifier: str, env: dict[str, str]) -> None:
+    if python_identifier.endswith("arm64"):
+        # macOS 11 is the first OS with arm64 support, so the wheels
+        # have that as a minimum.
+        env.setdefault("_PYTHON_HOST_PLATFORM", "macosx-11.0-arm64")
+        env.setdefault("ARCHFLAGS", "-arch arm64")
+        env.setdefault("CMAKE_OSX_ARCHITECTURES", "arm64")
+    elif python_identifier.endswith("universal2"):
+        env.setdefault("_PYTHON_HOST_PLATFORM", "macosx-10.9-universal2")
+        env.setdefault("ARCHFLAGS", "-arch arm64 -arch x86_64")
+        env.setdefault("CMAKE_OSX_ARCHITECTURES", "arm64;x86_64")
+    elif python_identifier.endswith("x86_64"):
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (commented):
Thanks for making a start on this, @Herrtian! This looks fine to me in principle, but I'm still unsure because https://github.com/pypa/cibuildwheel/issues/1190#issuecomment-1296969359 seemed to suggest we should add `CMAKE_OSX_ARCHITECTURES` to the macOS platform code itself for projects to use, rather than in the docs. We should do it unless the project already explicitly sets this environment variable, since we don't want to change how the build takes place.

This example is also CMake-specific, but we also support other build backends, one being meson-python, and that uses cross files to determine whether cross-compilation should take place. So the docs should probably be build-backend-agnostic.

I think @henryiii would have more coherent thoughts than I do! :)

Herrtian:
Thanks, that makes sense. I switched this from the docs-only example to the code path in dd342b1.

It now sets `CMAKE_OSX_ARCHITECTURES` next to the existing macOS `ARCHFLAGS` defaults for `x86_64`, `arm64`, and `universal2`, while still preserving any value the project already set explicitly. I also dropped the CMake-specific docs example so the PR is focused on the environment setup instead.

I added a small unit test for the arch mapping and for keeping existing values.

mayeut:
IMHO, this is a build backend/custom CMake implementation issue and cibuildwheel should not be working around those. The workaround in cibuildwheel will mask issues building from an SDist or building the project directly without the project knowing they have an issue with their build backend/custom CMake implementation on macOS.
See also https://github.com/pypa/cibuildwheel/issues/1190#issuecomment-1193136840

henryiii:
scikit-build-core should read the Python-specific values and convert them for you. Is it not working?

Herrtian:
Thanks, that feedback makes sense. I agree this should not live in cibuildwheel if the build backend or project should derive the CMake architecture from the Python/macOS platform values itself. I'm closing this rather than keep pushing a workaround in the wrong layer.

agriyakhetarpal:
Sorry for directing you the wrong way @Herrtian! I think I'll probably close the linked issue as resolved for now, and we can revisit it if someone has any trouble cross-compiling on macOS later.

PR #2839, opened 2026-05-05
docs: add Windows NuGet cache guidance

### LINKED ISSUE #1527: Caching NuGet (Windows) Python installations: Doc guidance or option in GitHub Action?
## Preamble

In general, I think a lot of build time could be saved if cibuildwheel had better builtin functionality to make caching easier or automatic, but before I create a generic ticket that doesn't go anywhere because the scope is too large, let me start with something specific that might turn out to be low-hanging fruit:

## Feature request

I think it would be nice if **either**

1. cibuildwheel's GitHub Action came with optional caching (via [actions/cache](https://github.com/actions/cache)) of the NuGet Python installations it performs on Windows, or
2. there was more guidance in the documentation on how to implement such caching oneself, ideally as copy-pasteable as the normal example CI configs (or maybe even part of them?).

In either case, it would of course be even better to extend this to other CI services.

## Example of impact on build time

For small builds on GitHub Actions' Windows runners, cibuildwheel will spend a significant fraction of the total build time installing different Python versions from NuGet. Since cibuildwheel already has [logic to avoid re-downloading available Python versions](https://github.com/smheidrich/cibuildwheel/blob/0a7800c9d9080a7985735f9c01dd364f8a3e9711/cibuildwheel/windows.py#L114), it's not that difficult for users to cache them when using GitHub Actions. I tried this [here](https://github.com/smheidrich/cibuildwheel-cache-nuget/pull/1/files) for a dummy repo building the same sample project that cibuildwheel builds during its own tests. Comparing the runtimes to those without caching, we see that this shaves off around 3 minutes on average:

| Run # | Build time w/o cache in minutes | Build time w/ cache in minutes |
| -- | -- | -- |
| 1 | 9.5 | 7.7 |
| 2 | 12.4 | 8.9 |
| 3 | 14.7 | 7.35 | 
| **Average** | **12.2** | **8.0** |

<sup>Runs with and without cache were alternated to more or less rule out the difference stemming from one group being run at a time when the runners happened to perform better/worse than at the time when the other group was run.</sup>

This is despite GitHub Action's cache being [infamously slow](https://github.com/actions/cache/issues/984#issuecomment-1340896418) to restore (~2 minutes in the example above), so perhaps more could be gained for other CI services.

## Other ideas (future)

There are many other avenues for better caching support or documentation: virtualenvs could be cached as well to hopefully shave even more off the build times above, caching Python installations on MacOS would be a bit more difficult since they use the system installer, but maybe it would be possible by allowing users to choose an alternative installer (Homebrew?), and of course there could be better support for allowing users to cache their own dependencies and additional build tools (which I've only accomplished via [some crazy hacks](https://github.com/smheidrich/py-json-stream-rs-tokenizer/blob/2e9e88ba28cd735b10e231365771b670a668bcff/.github/workflows/build.yml#L57-L112) so far). But there should be separate tickets for all of those (maybe an "overview ticket" on caching in general if you agree that this is something that should be worked on?).

Thoughts?

### Build log

https://github.com/smheidrich/cibuildwheel-cache-nuget/actions/runs/5304286464/jobs/9603413251

### CI config

https://github.com/smheidrich/cibuildwheel-cache-nuget/blob/d41d7835d558a3f28096b763396a651f448e6a11/.github/workflows/build.yml

Herrtian:
Adds a short Windows docs section for caching cibuildwheel's NuGet CPython installs between CI runs. I kept the example GitHub Actions-specific and included the cache poisoning warning called out in the issue.

Closes #1527.

Checks run locally:
- git diff --check
- uv run --with prek prek run --files docs/platforms.md
- uv run --group docs mkdocs build --strict

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -149,6 +149,28 @@ Because the builds are happening without full isolation, there might be some dif
 
 In order to speed-up builds, cibuildwheel will cache the tools it needs to be reused for future builds. The folder used for caching is system/user dependent and is reported in the printed preamble of each run (e.g. `Cache folder: C:\Users\Matt\AppData\Local\pypa\cibuildwheel\Cache`). You can override the cache folder using the ``CIBW_CACHE_PATH`` environment variable.
 
+### Caching CPython installations on CI {: #windows-cpython-cache}
+
+On Windows, cibuildwheel installs CPython from NuGet under the `nuget-cpython` directory in its cache folder. If your CI jobs spend a lot of time reinstalling these Python versions, you can set `CIBW_CACHE_PATH` to a known location and cache that directory between runs.
+
+!!! warning
+    Be careful with caches in release workflows. Only restore and save caches from trusted branches or tags; do not save caches from untrusted pull requests, because a poisoned cache can affect later builds. If in doubt, keep release builds uncached.
+
+For example, on GitHub Actions:
+
+```yaml
+env:
+  CIBW_CACHE_PATH: ${{ runner.temp }}\cibw-cache
+
+steps:
+  - uses: actions/cache@v4
+    with:
+      path: ${{ env.CIBW_CACHE_PATH }}\nuget-cpython
+      key: cibw-nuget-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('.github/workflows/*', 'pyproject.toml') }}
+
+  - uses: pypa/cibuildwheel@v3.4.1
+```
+
 ### Windows ARM64 builds {: #windows-arm64}
 
 `cibuildwheel` supports cross-compiling `ARM64` wheels on all Windows runners, but a native `ARM64` runner is required for testing. On non-native runners, tests for `ARM64` wheels will be automatically skipped with a warning. Add `"*-win_arm64"` to your `test-skip` setting to suppress the warning.


docs/platforms.md
```diff
@@ -149,6 +149,28 @@ Because the builds are happening without full isolation, there might be some dif
...
+  CIBW_CACHE_PATH: ${{ runner.temp }}\cibw-cache
+
+steps:
+  - uses: actions/cache@v4
+    with:
+      path: ${{ env.CIBW_CACHE_PATH }}\nuget-cpython
+      key: cibw-nuget-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('.github/workflows/*', 'pyproject.toml') }}
+
+  - uses: pypa/cibuildwheel@v3.4.1
+```
```
agriyakhetarpal, line 172: Since we are caching the `nuget-cpython` directory, which only exists on Windows, can we remove `runner.os` from the cache key?

agriyakhetarpal (commented):
This looks neat, thanks! Could you please move these changes to after the "Windows ARM64 builds" section? Also, inline comments below.

I will leave it open in case anyone else has comments about the wording.

joerick (commented):
I wonder if this advice could be broadened to caching the whole CIBW_CACHE_DIR? I believe it would also benefit macOS by preventing the redownload of CPython installers there too. Curious if anyone's had experience with doing that.

agriyakhetarpal:
Hmm, it looks like #2842 would be such a PR, it looks like it's more general!

PR #2849, opened 2026-05-13
fix: remove pinning constraints from test installs on Linux

### LINKED ISSUE #2822: Linux pinning in test environments
Our docs claim we `dependency-versions doesn't` affect linux, and we disallow the `linux` platform setting for this, but it actually does affect the test environment. In 3.3.0, we removed pinning from build environments, but it is still present when installing testing environments.

I think we need to select one:

* Remove pinning from all test environments
* Remove pinning from linux test environments
* Update docs and remove the `linux` restriction for `dependency-versions`

It's currently bad because tox released a version that requires a python-discovery version that we pin too low. So unpinned tox no longer works in cibuildwheel test environments. https://github.com/python-lz4/python-lz4/pull/289

I asked Kimi-K2.5 about this, and the report was quite helpful, copied below.

<details><summary>:robot: AI investigation</summary>

You're right to question this. After looking at the code, I can confirm that **dependency pinning DOES affect Linux builds**, contrary to what the documentation might imply.

### 1. Linux code uses `dependency_constraints` (lines 215-223 in `linux.py`):

```python
local_constraints_file = build_options.dependency_constraints.get_for_python_version(
    version=config.version,
    tmp_dir=local_identifier_tmp_dir,
)
if local_constraints_file:
    container_constraints_file = PurePosixPath("/constraints.txt")
    container.copy_into(local_constraints_file, container_constraints_file)
    dependency_constraint_flags = ["-c", container_constraints_file]
```

### 2. These constraints are actively used in Linux for:
- **Line 369**: Installing `virtualenv` for testing
- **Lines 403-409**: Installing the wheel and test requirements

```python
container.call(
    ["pip", "install", "virtualenv", *dependency_constraint_flags], env=env
)
container.call(
    [*pip, "install", str(repaired_wheel) + build_options.test_extras],
    env=virtualenv_env,
)
```

### 3. The default is "pinned" (`resources/defaults.toml` line 10):
```toml
dependency-versions = "pinned"
```

When this default is used, `DependencyConstraints.pinned()` returns the bundled `constraints.txt` lockfile, which is then copied into the container and used.

### 4. The `DISALLOWED_OPTIONS` only affects platform-specific config sections

The code at `options.py:176-177` only prevents setting `dependency-versions` in `[tool.cibuildwheel.linux]` section. But the global setting (default "pinned") still applies, and users can still set it via env var `CIBW_DEPENDENCY_VERSIONS`.

## What the docs probably mean

The changelog entry at line 280 and 45 clarifies:
> "The versions set with `dependency-versions` no longer constrain packages specified by your `build-system.requires`."

This means the constraints don't apply during the **wheel build phase** (when `pip wheel` or `python -m build` is running with build isolation). However, the constraints **are still used** for:
- Installing `virtualenv` 
- Installing your wheel for testing
- Installing `test_requires`

## Summary

**Yes, the lockfile gets used on Linux when `dependency-versions = "pinned"`**, specifically during the testing phase. The documentation is misleading if it suggests dependency-versions doesn't affect Linux at all - it does affect test dependency installation.

</details>

henryiii:
Fix #2822.

:robot: *Human triggered, AI assisted PR. AI text below.* :robot:

## Summary

Removes the build dependency constraint flags from the `pip install virtualenv` call inside the Linux test setup.

## Details

The `dependency_constraint_flags` variable (containing `-c /constraints.txt`) was being unpacked into the command that installs `virtualenv` when setting up the test environment in `cibuildwheel/platforms/linux.py`. These constraints are intended for **build** dependencies only — they pin versions of packages like `delocate`, `build`, `pip`, etc. They should not restrict package versions when installing `virtualenv` for test execution.

macOS, Windows, iOS, Pyodide, and Android were already passing `None` for the dependency constraint when creating test virtualenvs. Linux was the only outlier.

## Change

- Removed the `dependency_constraint_flags` variable and its usage from the test virtualenv setup in `linux.py`.
- Build dependency constraints are still applied to build tool installations as before.

### DIFF
--- a/cibuildwheel/platforms/linux.py
+++ b/cibuildwheel/platforms/linux.py
@@ -6,7 +6,7 @@ import textwrap
 from collections import OrderedDict
 from collections.abc import Iterable, Iterator, Sequence, Set
 from pathlib import Path, PurePath, PurePosixPath
-from typing import TYPE_CHECKING, assert_never
+from typing import assert_never
 
 from cibuildwheel import errors
 from cibuildwheel.architecture import Architecture
@@ -19,9 +19,6 @@ from cibuildwheel.util import resources
 from cibuildwheel.util.file import copy_test_sources
 from cibuildwheel.util.helpers import prepare_command, unwrap
 from cibuildwheel.util.packaging import find_compatible_wheel
-
-if TYPE_CHECKING:
-    from cibuildwheel.typing import PathOrStr
 
 ARCHITECTURE_OCI_PLATFORM_MAP = {
     Architecture.x86_64: OCIPlatform.AMD64,
@@ -212,7 +209,6 @@ def build_in_container(
 
         log.step("Setting up build environment...")
 
-        dependency_constraint_flags: list[PathOrStr] = []
         local_constraints_file = build_options.dependency_constraints.get_for_python_version(
             version=config.version,
             tmp_dir=local_identifier_tmp_dir,
@@ -220,7 +216,6 @@ def build_in_container(
         if local_constraints_file:
             container_constraints_file = PurePosixPath("/constraints.txt")
             container.copy_into(local_constraints_file, container_constraints_file)
-            dependency_constraint_flags = ["-c", container_constraints_file]
 
         env = container.get_environment()
         env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):

henryiii:
Only virtualenv is being installed here, so it should actually be safe.

Issue #2854, opened 2026-05-14
AI review (Kimi-K2.6)


henryiii:
This was generated in OpenCode with Kimi-K2.6. I haven't been able to review yet, some are probably not valid (like shell=True). But worth reviewing. I also reduced the subagents to 2 instead of 5, since I'm not supposed to use more than 2, but it also reduced what it was going to look at due to that, rather than just triggering 2 more, then 1 more.

<details><summary>:robot: _AI report_ :robot:</summary>
<p>

Here is a comprehensive review of the cibuildwheel codebase, organized into **Bugs/Issues**, **Documentation Gaps**, and **Simplification Opportunities**.

---

### Security
| Location | Issue | Severity |
|----------|-------|----------|
| `cibuildwheel/__main__.py:265` | `tarfile.extractall()` is called **without `filter="data"`**. This is the known path traversal vulnerability (CVE-2007-4559). | **Critical** |
| `cibuildwheel/oci_container.py:374` | `copy_into()` uses `shell=True` with f-string interpolation: `subprocess.run(f"tar ... {self.engine.name} exec -i {self.name} ...", shell=True)`. `self.name` and `self.host_tar_format` are **not sanitized**, creating a command injection vector. | **Critical** |
| `cibuildwheel/oci_container.py:532` | `shell=True` used unnecessarily for `docker info`. | Medium |
| `cibuildwheel/util/cmd.py:82` | `shell()` joins user-provided commands with spaces and passes to `shell=True`. Unsafe if any argument contains shell metacharacters. | **High** |
| `cibuildwheel/platforms/android.py:319` | Uses `run_path()` to execute arbitrary Python code from `sysconfigdata` file without validation. | Medium |

### Error Handling & Logic
| Location | Issue | Severity |
|----------|-------|----------|
| `cibuildwheel/projectfiles.py:64` | `except Exception: return None` swallows **all** errors silently during `setup.py` AST parsing. | High |
| `cibuildwheel/oci_container.py:251` | `docker inspect` failure is silently ignored with `except subprocess.CalledProcessError: pass`. | Medium |
| `cibuildwheel/platforms/linux.py:489` | Raises bare `Exception(msg)` instead of a specific `errors.FatalError`. | Medium |
| `cibuildwheel/extra.py:68-69` | `int(e.headers.get("date", 0))` will crash with `ValueError` if the `date` header exists (it's an HTTP-date string, not a timestamp). | Medium |
[...]

PR #2856, opened 2026-05-15
fix: tarfile extraction filter

henryiii:
From pypa/build, with pathlib improvements and adapted for 3.11+. See #2854.

### DIFF
--- a/cibuildwheel/_compat/tarfile.py
+++ b/cibuildwheel/_compat/tarfile.py
@@ -1,3 +1,5 @@
+# Based on https://github.com/pypa/build/blob/f4ebd495cc0c2c74155bd4fe48b76399fb7927ac/src/build/_compat/tarfile.py
+
 from __future__ import annotations
 
 import sys
--- a/cibuildwheel/_compat/tarfile.py
+++ b/cibuildwheel/_compat/tarfile.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import sys
+import tarfile
+
+TYPE_CHECKING = False
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    TarFile = tarfile.TarFile
+
+# Per https://peps.python.org/pep-0706/, the "data" filter will become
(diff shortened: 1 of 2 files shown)


cibuildwheel/_compat/tarfile.py
```diff
@@ -0,0 +1,70 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    TarFile = tarfile.TarFile
+
+# Per https://peps.python.org/pep-0706/, the "data" filter will become
+# the default in Python 3.14. The first series of releases with the filter
+# had a broken filter that could not process symlinks correctly.
+elif (3, 11, 5) <= sys.version_info <= sys.version_info < (3, 14):
```
agriyakhetarpal:
```suggestion
# Per https://peps.python.org/pep-0706/, the "data" filter will become
# the default in Python 3.14. The first series of releases with the filter
# had a broken filter that could not process symlinks correctly.
elif (3, 11, 5) <= sys.version_info < (3, 14):
```

`sys.version_info <= sys.version_info` is a tautology, so we can remove the redundant condition.

agriyakhetarpal (approved):
Thanks, @henryiii! One cleanup below, and one more thing I'd suggest adding is your note about how this is adapted from pypa/build. This is a utility file and by no means a substantial portion, so I don't think we have to reproduce the copyright notice verbatim (too excessive), but we should still add the attribution by mentioning it as a code comment at the top that can be something like:

```python
# Based on https://github.com/pypa/build/blob/f4ebd495cc0c2c74155bd4fe48b76399fb7927ac/src/build/_compat/tarfile.py
# SPDX-License-Identifier: MIT
```

or whatever you think is nicer, so that we can remember beyond the PR where we got this from.

henryiii:
@layday is the original author (along with edits from me and maybe @gaborbernat), so I think if we got permission would be easier than adding a license (BSD-2-clause and MIT).

layday:
I don't remember which part of this I authored, but you have my permission to use it in cibuildwheel without attribution.

PR #2857, opened 2026-05-15
feat: add CPython 3.15 support for iOS

henryiii:
Missing support files; either should be vendored or added to xbuild.

Assisted-by: Copilot:claude-sonnet-4.6


README.md
```diff
@@ -39,13 +39,15 @@ While cibuildwheel itself requires a recent Python version to run (we support th
 
 <sup>**1** PyPy & GraalPy are only supported for manylinux wheels.</sup><br>
 <sup>**2** Windows arm64 support is experimental.</sup><br>
-<sup>**3** Experimental, not yet supported on PyPI, but can be used directly in web deployment. Use `--platform pyodide` to build.</sup><br>
-<sup>**4** manylinux armv7l support is experimental. As there are no RHEL based image for this architecture, it's using an Ubuntu based image instead.</sup><br>
-<sup>**5** Python 3.15 requires opt-in using [`enable`](https://cibuildwheel.pypa.io/en/stable/options/#enable).</sup><br>
+<sup>**3** Not supported on PyPI, uses old tag.</sup><br>
+<sup>**4** Experimental alpha, do not upload to PyPI yet.</sup><br>
```
agriyakhetarpal: This is incorrect, because we need a "✅" for CPython 3.14 as well, for Pyodide. The correct one should be:

- CPython 3.12/Pyodide: Not supported on PyPI, uses old tag (this is for Pyodide 0.27.7). We could also mention that this is enabled only under the `pyodide-eol` flag now, and not built by default.
- CPython 3.13/Pyodide: This one uses the right platform tag and has the 2025 ABI. This is supported on PyPI and corresponds to Pyodide 0.29.4.
- CPython 3.14/Pyodide: This is the experimental alpha (314.0.0a2) with the 2026 ABI, which shouldn't be uploaded to PyPI yet.

So we need one more green check here and the numbers should be moved one below (and a new note about `pyodide-eol`).

agriyakhetarpal (commented):
Thanks @henryiii! LGTM on the other code, but some small corrections needed on the README.

README.md
henryiii: Ahh, should have turned off wrapping, it's so hard to edit with wrapping on.

henryiii: > We could also mention that this is enabled only under the `pyodide-eol` flag now, and not built by default.

I thought it wasn't there yet?

:robot: _No, it does not currently require pyodide-eol. The cp312-pyodide_wasm32 identifier is listed as a default (non-EoL) build target in build-platforms.toml:223. The pyodide-eol guard in selector.py:97 would only apply if it were moved behind that flag — it currently isn't._ :robot:

agriyakhetarpal: Huh. Running `uvx --pre cibuildwheel --platform pyodide` (with `--pre` to grab v4.0.0rc1) on a project I'm working on right now builds cp313-pyodide_wasm32 and finishes.

Running `uvx --pre cibuildwheel --platform pyodide --enable pyodide-prerelease` builds both cp313-pyodide_wasm32 and cp314-pyodide_wasm32.

It is only when I run `uvx --pre cibuildwheel --platform pyodide --enable pyodide-prerelease --enable pyodide-eol` that it builds all three: cp312-pyodide_wasm32, cp313-pyodide_wasm32, and cp314-pyodide_wasm32.

agriyakhetarpal: 85d4d1ab5ee5bd00ee9971167b43583987c7d30e resolves this, thanks!

henryiii: You are correct, checked, I thought that was listed as a followup before.

freakboy3742:
@henryiii Thanks for picking this up. We've already spoken about the `target_python` assertion; looking at the current CI failures, I forgot about the `make_cross_venv` script as well. 

That script isn't part of the offical package either (and shouldn't be); my intention was to break that out into [`xbuild`](https://github.com/beeware/xbuild). xbuild needs a little bit of work before that's possible, plus the changes in cibuildwheel to hook it in. 

Alternatively, (and/or as a stop-gap measure) we could include the `make_cross_venv` script (et al) as a resource in cibuildwheel.

cibuildwheel/platforms/ios.py
```diff
@@ -113,6 +113,69 @@ def get_python_configurations(
     return python_configurations
 
 
+def _inject_support_files(installation_path: Path, python_version: str) -> None:
+    """Inject iOS cross-compilation support files into an XCFramework that
+    doesn't include them (e.g. python.org distributions for 3.15+).
+
+    Beeware's Python-Apple-support packages include ``platform-config/``
```
freakboy3742:
```suggestion
    BeeWare's Python-Apple-support packages include ``platform-config/``
```

cibuildwheel/platforms/ios.py
```diff
@@ -130,6 +193,7 @@ def install_target_cpython(tmp: Path, config: PythonConfiguration, free_threadin
             installation_path.mkdir(parents=True, exist_ok=True)
             call("tar", "-C", installation_path, "-xf", downloaded_tar_gz)
             downloaded_tar_gz.unlink()
+            _inject_support_files(installation_path, config.version)
```
freakboy3742: This approach definitely works, but there's a possible weakness - if we need to make a change to these scrips in cibuildwheel, and there *isn't* a corresponding update to the support package, the "old" version of the injected files will be used.

The likelihood of this is small, but if it does occur, it's likely to be very inconvenient (and confusing) for anyone using cibuildwheel locally. It could also affect cibuildwheel configurations where the CPython files have been cached by CI.

cibuildwheel/platforms/ios.py
```diff
@@ -312,9 +376,8 @@ def setup_python(
         / f"python{python_configuration.version}"
     )
 
-    assert target_python.exists(), (
-        f"{target_python.name} not found, has {list(target_install_path.iterdir())}"
-    )
+    # target_python will never be called, so we aren't validating it (not there
+    # for 3.15+)
```
freakboy3742: We *could* maybe confirm that `target_python.parent` exists - the bin directory will still be there, it just won't have `python3.X` as an executable.

freakboy3742 (changes requested):
This all makes sense to me, and works great in my manual testing. A couple of notes inline. Not sure what's going on with the macOS test failures in CI, though.

I've taken an initial look at doing this with xbuild; it doesn't look too difficult, but there's a couple of fixes needed to xbuild to finalise that. I'm unlikely to get a chance to look at this until post-PyCon; what is in this PR seems like a very workable stop-gap.

cibuildwheel/platforms/ios.py
freakboy3742: Taking a closer look at this - it should be possible to run the `make_cross_venv` script in-situ from the cibuildwheel source directory, rather than needing it to be copied into the cache folder. The `make_cross_venv` script already has support for being pointed at a different "location" for the platform configuration files; a couple of minor tweaks to that script are then needed to handle:
1. the shared location of `make_cross_venv` file itself being one level "higher"
2. the difference between the location of the sysconfigdata file (in the support package) and the venv support files (previously they were essentially co-located; there's now 2 different roots)

joerick:
This looks okay to me! I guess we expect these support files to be included in the next CPython XCframework release? So we'll only be vendoring this stuff for a little while?

freakboy3742:
> This looks okay to me! I guess we expect these support files to be included in the next CPython XCframework release? So we'll only be vendoring this stuff for a little while?

I don't expect they'll ever be in the XCframework - but they *will* (and are already) in [`xbuild`](https://github.com/beeware/xbuild). That set of tools should be able to remove the duplication of "cross platform venv" stuff in Android, iOS and Emscripten - but it needs a little more work before it's ready for serious consideration. 

However, once it *is* ready for consideration, we'll be able to remove the vendoring in cibuildwheel.

freakboy3742 (changes requested):
This doesn't seem quite right yet - the `_inject_support_files` call shouldn't be needed at all. 

I think the problem is that the `platform_config_path` that is being passed in for the 3.15 case doesn't actually exist - and the path to the stdlib (which is where sysconfigdata comes from) isn't being provided.

henryiii:
@freakboy3742 have you had a chance to look at the latest changes? (also waiting on PyPy)

freakboy3742:
> @freakboy3742 have you had a chance to look at the latest changes? (also waiting on PyPy)

My apologies - I wasn't as productive on my flight home as I was hoping. I'm taking a look now.

freakboy3742 (approved):
I've pushed a couple of minor tweaks - mostly clarifying the naming of the paths being referenced in the `make_cross_venv` script. It doesn't change the actual work being done, but the variable/argument names correspond a little closer to the actual purpose of the script. It also tweaks the message printed describing the conversion taking place - previously, it was printing the location of the sysconfig data, not the target platform type.

Otherwise, this looks good to me. I'll be able to revisit this once I've got `xbuild` to a more stable state. That will remove the need to vendor these scripts; but in the meantime, this should work. Thanks @henryiii for pushing this forward.

PR #2859, opened 2026-05-16
fix: replace assert with proper exception in python_build_standalone

henryiii:
See https://github.com/pypa/cibuildwheel/issues/2854.

:robot: *Human triggered, AI assisted PR ([using this skill](https://github.com/henryiii/skills/blob/main/branch-and-pr/SKILL.md)). AI text below.* :robot:

## Summary

Replaces `assert not python_base_dir.exists()` in `cibuildwheel/util/python_build_standalone.py:178` with an explicit check that raises `PythonBuildStandaloneError`. Assert statements are skipped under `python -O`, making them unsuitable for runtime validation.

Identified in #2854 ("Uses assert for runtime validation — It is skipped with `python -O`").

Assisted-by: OpenCode:glm-5

### DIFF
--- a/cibuildwheel/util/python_build_standalone.py
+++ b/cibuildwheel/util/python_build_standalone.py
@@ -175,7 +175,9 @@ def create_python_build_standalone_environment(
     )
 
     python_base_dir = temp_dir / "pbs"
-    assert not python_base_dir.exists()
+    if python_base_dir.exists():
+        msg = f"python-build-standalone directory already exists: {python_base_dir}"
+        raise PythonBuildStandaloneError(msg)
     extract_tar(archive_path, python_base_dir)
 
     return _find_python_executable(python_base_dir)


agriyakhetarpal (approved):
Thanks! I think it might be a good idea to add S101: https://docs.astral.sh/ruff/rules/assert/

joerick (approved):

joerick:
I'd be opposed to S101 - asserts are mighty useful to check for programmer mistakes, and handy to tell the type system something that you as a programmer know to be true.

agriyakhetarpal:
@joerick oh, the idea would be to add S101 for the source directories only – not the test directories, because we need assert statements there. Like this:

https://github.com/pybamm-team/PyBaMM/blob/9c8823baa1a8207ff35f2f4620a5caa3d9148d41/pyproject.toml#L189

https://github.com/pybamm-team/PyBaMM/blob/9c8823baa1a8207ff35f2f4620a5caa3d9148d41/pyproject.toml#L210-L211

henryiii:
Assert is used to narrow type checkers. The AI was able to detect a usage that was not for type checking, but the general ruff check can't do that.

PR #2861, opened 2026-05-17
chore: add agent and copilot setup files

henryiii:
I have no idea what happened to the file names below. :)

Anyway, this are the support files I've been using - if we want them upstream, here they are. (Plus a copilot file if we trigger copilot, copied from scikit-build-core).

I _think _ I used Kimi 2.6 or GLM 5.1 (open source models) when making these originally, and I've manually touched them up since.

This PR adds supporting files for AI-assisted development and GitHub Copilot workspaces:

- **AGENTS.md**: Project-specific developer notes for AI agents, covering testing commands, project layout, generated files, and CI quirks.
- ****: A GitHub Actions workflow to validate that the Copilot setup steps (installing nox, prek, and running pre-install checks) work correctly.
- ****: A local skill for generating changelog entries following the project's established style and conventions.

These are all additive, non-breaking changes that improve the developer experience when using AI-assisted tooling in this repository.


AGENTS.md
```diff
@@ -0,0 +1,57 @@
...
+- Integration tests auto-set a default `CIBW_ENABLE` if the env var is absent.
+- The `build_frontend_env` fixture parameterizes over `pip`, `build`, `build[uv]`, `uv` and skips unsupported combos per platform.
+- Some integration tests require system Python.org installs on macOS; missing them prints a download URL in the error.
+- iOS/Android/pyodide tests have dedicated pytest marks (`ios`, `android`, `pyodide`) and need platform-specific runners/simulators.
+
+## Lint / typecheck
+- Ruff (lint + format) and mypy run via pre-commit. Pylint runs separately via `nox -s pylint`.
+- Mypy is strict (`strict = true`) and targets Python 3.11 for the package, 3.14 for a second check in pre-commit.
+- Ruff config in `pyproject.toml` (`line-length = 100`).
+- Python 3.11 is the minimum supported version for the package itself.
```
agriyakhetarpal, line 45: I think my only concern with this PR is that this kind of information tends to get out of date pretty quickly. Is there a way we can adopt a [Kaizen](https://en.wikipedia.org/wiki/Kaizen)-like approach to these skill files, and add a meta-skill of sorts that will update these files themselves?

agriyakhetarpal (approved):
Thanks! Could you also add `.cursorrules` as a symlink to `AGENTS.md`? Here's an [example](https://github.com/pyodide/pyodide/blob/99f17ee12c72cf214eabd868e1c8f3e5aa383bfb/.cursorrules).

henryiii:
Cursor supports AGENTS.md: https://cursor.com/docs/rules

AGENTS.md
henryiii: If you type `/init` into at least OpenCode, it will update the AGENTS.md file instead of writing a new one. I think the other harness do that too.

agriyakhetarpal: Nice, thanks! Claude Code does the same, but ofc it creates `CLAUDE.md`. Now _that_ is something I guess we could symlink to `AGENTS.md`.

henryiii: Claude also supports AGENTS.md. Though I think it might write a CLAUDE.md instead of updating it - I think copilot writes `.github/copilot-instructions.md` or something like that. But you can tell it what file to use after the `/init`, like `/init update AGENTS.md` I believe.

henryiii: AGENTS.md is an open standard: https://agents.md

henryiii: Ahh, you are right, only Claude does not support it. Of course. However, a user can just add the symlink, don't feel like we need to advertise Claude just because they won't support the standard https://github.com/anthropics/claude-code/issues/6235. :)

henryiii: We can ignore `CLAUDE.md`, though, so it's easy to make a symlink or add one that has `See @AGENTS.md` in it.

agriyakhetarpal: I am fine with this – in practice I have a hard time getting Claude Code to read `AGENTS.md` anyway and I have to be very verbose with my prompts and plans. Thanks :)

PR #2863, opened 2026-05-18
Support new graalpy asset names that include Python version.

timfel:
With https://github.com/oracle/graalpython/pull/810 we will slightly alter upcoming GraalPy release artifact naming. The name will now include the compatible CPython version name in the filename. This change adapts the metadata script to support both the older and newer format. We can drop the older one once support for GraalPy versions older than 25.1 is removed.

/cc @msimacek
Related:  https://github.com/oracle/graalpython/issues/623, https://github.com/pypa/manylinux/pull/1944

### DIFF
--- a/bin/update_pythons.py
+++ b/bin/update_pythons.py
@@ -124,14 +124,20 @@ class GraalPyVersions:
         response.raise_for_status()
 
         releases = response.json()
-        gp_version_re = re.compile(r"-(\d+\.\d+\.\d+)$")
+        gp_asset_re = re.compile(r"^(graalpy(\d+\.\d+)?-(\d+\.\d+\.\d+))-")
         cp_version_re = re.compile(r"Python (\d+\.\d+(?:\.\d+)?)")
         for release in releases:
-            m = gp_version_re.search(release["tag_name"])
-            if m:
-                release["graalpy_version"] = Version(m.group(1))
-            m = cp_version_re.search(release["body"])
-            if m:
+            for asset in release["assets"]:
+                m = gp_asset_re.match(asset["name"])
+                if m:
+                    release["asset_prefix"] = m.group(1)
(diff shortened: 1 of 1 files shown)


bin/update_pythons.py
```diff
@@ -124,14 +124,20 @@ def __init__(self) -> None:
...
-            m = cp_version_re.search(release["body"])
-            if m:
+            for asset in release["assets"]:
+                m = gp_asset_re.match(asset["name"])
+                if m:
+                    release["asset_prefix"] = m.group(1)
+                    release["graalpy_version"] = Version(m.group(3))
+                    if m.group(2):
+                        release["python_version"] = Version(m.group(2))
+                    break
```
henryiii: :robot: _AI text below_ :robot:

Readability nit (non-blocking): named groups make the three captures self-documenting, so the `m.group(1/2/3)` accesses below read in terms of what they mean rather than position. Splitting the now-longer pattern keeps it under the 100-char line limit.

```suggestion
        gp_asset_re = re.compile(
            r"^(?P<prefix>graalpy(?P<cpython>\d+\.\d+)?-(?P<graalpy>\d+\.\d+\.\d+))-"
        )
        cp_version_re = re.compile(r"Python (\d+\.\d+(?:\.\d+)?)")
        for release in releases:
            for asset in release["assets"]:
                m = gp_asset_re.match(asset["name"])
                if m:
                    release["asset_prefix"] = m.group("prefix")
                    release["graalpy_version"] = Version(m.group("graalpy"))
                    if m.group("cpython"):
                        release["python_version"] = Version(m.group("cpython"))
                    break
```

henryiii (approved):
(see suggestion)

joerick (approved):

bin/update_pythons.py
timfel: applied

PR #2866, opened 2026-05-23
chore: use `if TYPE_CHECKING:` blocks

mayeut:
This will become the default in Python 3.14 in any case given the current lint rules.
It ensures typing only imports do not slow down execution (admittedly only useful in edge use cases, such as querying help).

If #2797 makes it, it will further speed-up this one:
```
hyperfine --warmup 10 \
     -n "main"                --prepare "git checkout main"               ".nox/tests-3-15/bin/python -m cibuildwheel --help" \
     -n "Henry's PR"          --prepare "git checkout pr/2797"            ".nox/tests-3-15/bin/python -m cibuildwheel --help" \
     -n "type-checking"       --prepare "git checkout type-checking"      ".nox/tests-3-15/bin/python -m cibuildwheel --help" \
     -n "type-checking-lazy"  --prepare "git checkout type-checking-lazy" ".nox/tests-3-15/bin/python -m cibuildwheel --help"
Benchmark 1: main
  Time (mean ± σ):     308.7 ms ±  16.7 ms    [User: 194.7 ms, System: 44.1 ms]
  Range (min … max):   294.3 ms … 349.3 ms    10 runs
 
Benchmark 2: Henry's PR
  Time (mean ± σ):     106.2 ms ±   9.4 ms    [User: 62.7 ms, System: 17.6 ms]
  Range (min … max):    88.7 ms … 118.2 ms    19 runs
 
Benchmark 3: type-checking (this PR)
  Time (mean ± σ):     306.9 ms ±  15.5 ms    [User: 194.2 ms, System: 43.2 ms]
  Range (min … max):   290.2 ms … 342.8 ms    10 runs
 
Benchmark 4: type-checking-lazy
  Time (mean ± σ):      77.5 ms ±   6.1 ms    [User: 45.6 ms, System: 13.7 ms]
  Range (min … max):    68.9 ms …  86.3 ms    22 runs
 
Summary
  type-checking-lazy ran
    1.37 ± 0.16 times faster than Henry's PR
    3.96 ± 0.37 times faster than type-checking (this PR)
    3.98 ± 0.38 times faster than main
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     exclude: ^cibuildwheel/resources/android/android.patch$
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: 6fec9b7edb08fd9989088709d864a7826dc74e80  # frozen: v0.15.12
+  rev: 0c7b6c989466a93942def1f84baf36ddfcd60c83  # frozen: v0.15.14
   hooks:
   - id: ruff-check
     args: ["--fix"]
--- a/bin/inspect_all_known_projects.py
+++ b/bin/inspect_all_known_projects.py
@@ -24,8 +24,9 @@ This will cache the results to all_known_setup.yaml; you can reprint
 the results without the `--online` setting.
 """
 
+from __future__ import annotations
+
 import ast
-from collections.abc import Iterable, Iterator
 from pathlib import Path
 
 import click
@@ -34,6 +35,10 @@ from github import Github, GithubException
 from rich import print
 
 from cibuildwheel.projectfiles import Analyzer
+
+TYPE_CHECKING = False
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
 
 DIR = Path(__file__).parent.resolve()
 
--- a/bin/projects.py
+++ b/bin/projects.py
@@ -17,13 +17,14 @@ Suggested usage:
     git diff
 """
 
+from __future__ import annotations
+
 import builtins
 import functools
 import textwrap
 import urllib.error
 import urllib.request
 import xml.dom.minidom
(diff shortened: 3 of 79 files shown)


henryiii (approved):
Sounds good to me.

freakboy3742 (approved):
All makes sense to me

mayeut:
After further thoughts, we might want to:
- re-allow `typing.TYPE_CHECKING` (use forbidden in #2864) as typing is always imported by cibuildwheel. If PEP 781 makes it, we can forbid `typing.TYPE_CHECKING` once dropping the Python version before its introduction. Only the fact that there are `TYPE_CHECKING` blocks is important for #2797. Whether we use `TYPE_CHECKING = False` or `typing.TYPE_CHECKING` doesn't change the fact that `typing` is always imported.
- disable TC checks for tests where there's no added value in timings at all I think

Any thoughts ?

mayeut:
I've implemented the proposal from the previous comment in #2869 
Changed the status to draft until there's a consensus on which PR to keep if any.

henryiii:
I like this version fine - `TYPE_CHECKING = False` is accepted by all the type checkers (including the new Rust ones), and is the only proper workaround that works with lazy imports too until PEP 781 makes it. I'd be fine to see it everywhere, might help get PEP 781 in. :)

In timings, `typing` forces an `re` import, which is quite expensive.

I'm fine to keep tests consistent, though I don't really care, we don't have to.

joerick:
I don't have a strong opinion, but the `TYPE_CHECKING = False` thing is fine by me!

mayeut:
> In timings, `typing` forces an re import, which is quite expensive.

Just as a note, `typing` is always imported by cibuildwheel - at least for now.

I closed the other PR and kept this one as I don't have a strong opinion either.

agriyakhetarpal (approved):

PR #2869, opened 2026-05-25
chore: rework `if TYPE_CHECKING:` blocks

mayeut:
This is an alternative to #2866 as proposed in https://github.com/pypa/cibuildwheel/pull/2866#issuecomment-4532756609.

The 1st commit is #2866, it enforces the use of  `if TYPE_CHECKING:` blocks.
The 2nd commit revert #2864 as it's not needed as noted in https://github.com/pypa/cibuildwheel/pull/2866#issuecomment-4532756609 and might borderline regarding standards (it's apparently not even though it works in practice)
The 3rd commit removes enforcing `TC` on tests as it's not likely useful there.

Marking as draft until there's a consensus between #2866 and this one.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     exclude: ^cibuildwheel/resources/android/android.patch$
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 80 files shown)


mayeut:
close in favour of #2866

PR #2870, opened 2026-05-26
Fix Windows Gitlab test running

joerick:
These have been failing on main for a little while.

Don't look like a big issue. `py` is trying to execute the shebang in run_tests.py, which is a posix thing. Trying to use `python` directly...


henryiii:
The python launcher for Windows, `py`, is supposed to be able to read shebangs on Windows and respect common patterns. Ahh, I see, the uv part is probably confusing it.

joerick:
Thanks for finishing this off, henryiii! I had the same suspicion about the startup time as you, but I never managed to verify it. Did you manage to prove that's the problem? Either way, it's probably a bit of an edge case.

PR #2872, opened 2026-05-27
feat: add CIBUILDWHEEL_BUILD_IDENTIFIER environment variable

### LINKED ISSUE #944: Set build-time env vars about the current build for scripts to use
We already set `CIBUILDWHEEL=1` for our builds, so we could add `CIBUILDWHEEL_BUILD_IDENTIFIER` - the `CIBUILDWHEEL` prefix distinguishing these variables from `CIBW` options.

_Originally posted by @joerick in https://github.com/pypa/cibuildwheel/discussions/910#discussioncomment-1633428_

Or just `CIBUILDWHEEL_BUILD` might work. 
An arch variable was also mooted, which might be easier to switch on. In this naming structure, that would be `CIBUILDWHEEL_BUILD_ARCH`

### LINKED ISSUE #2750: Add placeholders containing Python version and/or wheel leafname and system information
### Description

Could cibuildwheel define a new placeholder `{python-version}` set to things like `3.14`, for use in CIBW_TEST_COMMAND?

This would allow gathering of pytest results  with `pytest --junit-xml=test-results{pythonversion}-xml` when cibuildwheel builds and tests for a range of Python versions.

At the moment with something like `CIBW_BUILD='cp310* cp311* cp312* cp313* cp314*'` and `CIBW_TEST_COMMAND='pytest --junit-xml=test-results.xml'`, the invocation of pytest for each python version overwrites the existing `test-results.xml` file, so we can't gather complete information.

It might be useful to have more information than just the python version. In which case how about:

* A `{wheel-leafname}` placeholder that identifies the exact wheel that was tested.
* A `{system}` placeholder containing a wheel name-style `<python-tag>-<abi-tag>-<platform-tag>` string to identify the environment in which tests were run?

### Build log

_No response_

### CI config

_No response_

henryiii:

## Summary

Closes #944, closes #2750.

Sets a new environment variable `CIBUILDWHEEL_BUILD_IDENTIFIER` to the current build identifier (e.g. `cp311-manylinux_x86_64`) in the environment for all per-build steps: `before_build`, the build itself, `repair_command`, `before_test`, and `test_command`.

## Motivation

Users need to inspect the current build identifier inside scripts and commands, for example to write per-build output files without one build overwriting another:

```toml
[tool.cibuildwheel]
test-command = "pytest --junit-xml=results-$CIBUILDWHEEL_BUILD_IDENTIFIER.xml"
```

## Design decisions

- Set **after** the user's `CIBW_ENVIRONMENT` overrides are applied, so it reliably reflects the actual identifier.
- Available **only** for per-build steps — not in `before_all`, where no single identifier applies (that step runs once for a batch of builds).
- Covers all six platforms: **linux**, **macOS**, **Windows**, **pyodide**, **Android**, and **iOS**.

Assisted-by: CopilotCLI:claude-sonnet-4.6

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -140,7 +140,7 @@ def build(options: Options, tmp_path: Path) -> None:
             build_path.mkdir()
             python_dir = setup_target_python(config, build_path)
             build_env, android_env = setup_env(config, build_options, build_path, python_dir)
-            build_env["CIBUILDWHEEL_BUILD_IDENTIFIER"] = config.identifier
+
             state = BuildState(
                 config, build_options, build_path, python_dir, build_env, android_env
             )
@@ -226,6 +226,7 @@ def setup_env(
 
     # Apply custom environment variables, and check environment is still valid
     build_env = build_options.environment.as_dictionary(build_env)
+    build_env["CIBUILDWHEEL_BUILD_IDENTIFIER"] = config.identifier
     build_env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
     for command in ["python"] if use_uv else ["python", "pip"]:
         command_path = call("which", command, env=build_env, capture_stdout=True).strip()
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -140,6 +140,7 @@ def build(options: Options, tmp_path: Path) -> None:
             build_path.mkdir()
             python_dir = setup_target_python(config, build_path)
             build_env, android_env = setup_env(config, build_options, build_path, python_dir)
+            build_env["CIBUILDWHEEL_BUILD_IDENTIFIER"] = config.identifier
             state = BuildState(
                 config, build_options, build_path, python_dir, build_env, android_env
             )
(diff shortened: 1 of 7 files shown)


agriyakhetarpal (approved):
Makes sense to me!

freakboy3742 (approved):
The utility of script-based builds is always going to be slightly limited on iOS, but the example of an argument to `pytest` would definitely apply, and the general approach is straightforward.

cibuildwheel/platforms/android.py
```diff
@@ -140,6 +140,7 @@ def build(options: Options, tmp_path: Path) -> None:
             build_path.mkdir()
             python_dir = setup_target_python(config, build_path)
             build_env, android_env = setup_env(config, build_options, build_path, python_dir)
+            build_env["CIBUILDWHEEL_BUILD_IDENTIFIER"] = config.identifier
```
mhsmith: I think it would be better to set this in `setup_env` after calling `build_options.environment.as_dictionary`. This would keep it together with other  environment overrides such as `ANDROID_API_LEVEL` and `PIP_DISABLE_PIP_VERSION_CHECK`, and also ensure that the variable is set both in `build_env` and `android_env`.

mhsmith (approved):

PR #2873, opened 2026-05-27
feat: add SHA256 verification for direct downloads

### LINKED ISSUE #908: Add hashes for Python downloads
We provide hashes for PyPI downloads, but not for Python downloads. We should be able to capture the hash and include it in our download table. It bothered me that the macOS download from Python 3.10.0 was able to change under us without an error - it was useful in that case, but could be used for nefarious purposes in the future.

henryiii:

## Summary

Implements SHA256 checksum verification for direct downloads in cibuildwheel, addressing #908. SHA256 hashes are stored in resource config files when running maintainer update scripts, and verified at download time to detect unexpected changes to artifacts.

### Runtime verification
- `cibuildwheel/util/file.py`: `download()` now accepts `sha256: str | None = None`; verifies hash after download and raises `FatalError` on mismatch

### Platform wiring
SHA256 is now stored in `PythonConfiguration` dataclasses and passed through to `download()` on:
- **macOS**: CPython, PyPy, GraalPy
- **iOS**: CPython (python.org and BeeWare), including the macOS build Python used for cross-builds
- **Android**: CPython (python.org and Chaquopy/Maven)
- **Windows**: PyPy, GraalPy (CPython uses NuGet, no direct download)
- **virtualenv**: `virtualenv.pyz`
- **python-build-standalone**: used by the test runner

### Update scripts
SHA256 sources vary by provider:
- **python.org CPython**: `sha256_sum` field from the downloads API (free)
- **GraalPy**: `.sha256` sidecar assets from GitHub releases
- **python-build-standalone**: `SHA256SUMS` file in the release
- **PyPy, BeeWare iOS, Chaquopy Android**: no checksums provided upstream — computed by streaming download+hash on first run, then preserved when URLs are unchanged

### Resource files
All URL-bearing entries in the following files now have populated `sha256` fields:
- `cibuildwheel/resources/build-platforms.toml` (macOS, iOS, Android, Windows PyPy/GraalPy)
- `cibuildwheel/resources/virtualenv.toml`
- `cibuildwheel/resources/python-build-standalone-releases.json` (179 assets, tag 20260510)

### Review follow-up
- Require `sha256` for URL-backed Python and virtualenv configs instead of allowing empty / missing values
- Add the two previously empty GraalPy macOS x86_64 SHA256 entries
- Update the GraalPy pin refresh logic to select the newest release that actually contains a matching asset, so macOS x86_64 entries are refreshed instead of skipped when a newer tag omits that artifact
- Add unit coverage for required SHA-bearing resource configs and the GraalPy asset-selection fallback

Assisted-by: copilot-cli:gpt-5.4


bin/update_python_build_standalone.py
```diff
@@ -3,20 +3,31 @@
...
     PythonBuildStandaloneReleaseData,
 )
-from cibuildwheel.util.resources import PYTHON_BUILD_STANDALONE_RELEASES
+
+# Resolve path relative to this script so writes go to the source checkout,
+# not the uv-installed copy of the package.
+DIR: Final[Path] = Path(__file__).parent.parent.resolve()
+PYTHON_BUILD_STANDALONE_RELEASES: Final[Path] = (
+    DIR / "cibuildwheel/resources/python-build-standalone-releases.json"
+)
```
henryiii, line 30: Unrelated fix.

agriyakhetarpal:
Nice; I had started a local branch to take this up recently, but I did not go as far as Copilot has done here. Could we also add SHA256 checksums for Pyodide cross-build environments?

bin/update_python_build_standalone.py
```diff
@@ -3,20 +3,31 @@
...
 
 from cibuildwheel.extra import github_api_request
 from cibuildwheel.util.python_build_standalone import (
     PythonBuildStandaloneAsset,
     PythonBuildStandaloneReleaseData,
 )
-from cibuildwheel.util.resources import PYTHON_BUILD_STANDALONE_RELEASES
+
+# Resolve path relative to this script so writes go to the source checkout,
+# not the uv-installed copy of the package.
```
agriyakhetarpal, line 26: I was just thinking about this yesterday when I found that `[tool.uv.sources]` isn't getting recognised in the PEP 723 metadata. Probably a good follow-up for `nox`'s codebase; just noting it here.

mhsmith (approved):

mhsmith:
LGTM as far as Android is concerned.

cibuildwheel/platforms/android.py
```diff
@@ -72,6 +72,7 @@ class PythonConfiguration:
     version: str
     identifier: str
     url: str
+    sha256: str = ""
```
```suggestion
    sha256: str
```

cibuildwheel/platforms/android.py
```diff
@@ -184,7 +185,7 @@ def setup_target_python(config: PythonConfiguration, build_path: Path) -> Path:
     python_tgz = CIBW_CACHE_PATH / config.url.rpartition("/")[-1]
     with FileLock(f"{python_tgz}.lock"):
         if not python_tgz.exists():
-            download(config.url, python_tgz)
+            download(config.url, python_tgz, sha256=config.sha256 or None)
```
```suggestion
            download(config.url, python_tgz, sha256=config.sha256)
```

mhsmith (commented):
Although, if the hashes are guaranteed to be present in build-platforms.toml, shouldn't we make them required at runtime? Otherwise, if a change to update_pythons accidentally removes the hashes in the future, we would silently lose the security.

This applies to all platforms.

cibuildwheel/util/file.py
```diff
@@ -59,13 +60,20 @@ def download(url: str, dest: Path) -> None:
...
                 dest.write_bytes(response.read())
-                return
+                break
 
         except OSError:
             if i == repeat_num - 1:
                 raise
             time.sleep(3)
 
+    if sha256:
```
mhsmith, line 70: If a hash accidentally became an empty string, that should produce an error.

```suggestion
    if sha256 is not None:
```

unit_test/pyodide_test.py
agriyakhetarpal, line 1: I don't feel very strongly about this, but I think the filename will make it confusing later on when grepping/searching for `pyodide`, especially since the rest of the Pyodide tests are not here and are scattered elsewhere. Could we reorganise `unit_test/pyodide_test.py` and `unit_test/python_build_standalone_test.py` into one `unit_test/test_sha256.py`?

agriyakhetarpal (approved):

bin/update_python_build_standalone.py
henryiii: I believe this is working as intended, this is a uv-specific field and would any be read by `uv run`, not `nox`?

henryiii: We could require uv then use `uv run` inside nox.

agriyakhetarpal: Oh, what I meant was that `nox` itself could probably use this and delegate to `uv run` to run the scripts if `nox[uv]` is installed. More of a feature request, not a bug. Pretty sure I have a recent local branch lingering around somewhere for this; I'll put together a PR in nox a bit to elaborate.

PR #2877, opened 2026-05-29
docs: add delvewheel to diagram, schema defaults, and docs

henryiii:
Noticed we missed these in #2831. ~~The image still needs to be regenerated, I'm not set up for it.~~

## Summary

Since delvewheel is now the default `repair-wheel-command` on Windows, update all places where auditwheel/delocate were referenced but delvewheel was missing.

## Changes

- **diagram.html**: Add delvewheel repair block for Windows (was previously grouped with ios/pyodide as an optional dot)
- **generate_schema.py**: Add Windows delvewheel default to the platform loop so `schema.json` includes the default
- **schema.json**: Regenerated with Windows delvewheel default
- **options.md**: Add Windows to the defaults list, remove outdated tip about delvewheel being early-stage/optional, update examples to reflect delvewheel runs by default
- **contributing.md**: Add delvewheel to the tool list alongside auditwheel, delocate, docker, etc.

Assisted-by: OpenCode:glm-5.1


agriyakhetarpal (approved):
Thanks for catching these!

PR #2881, opened 2026-06-01
fix: remove potential partial extraction in case of error

mayeut:


mayeut:
I intended to work on this in my fork first & only open a PR here when ready. Closing this PR.

PR #2887, opened 2026-06-02
fix: ensure container removal on teardown even if bash hangs or dies

henryiii:
This is a fix for item 4 in https://github.com/pypa/cibuildwheel/issues/2885. First pass tried to introduce a double nested try, I reduced it to a single try.

`OCIContainer.__enter__` was hardened in #2879 to always clean up on
failure, but `__exit__` was left unguarded. If the container/bash had
already died, `bash_stdin.write("exit 0")` raises `BrokenPipeError`; if
bash refuses to exit, `process.wait(timeout=30)` raises `TimeoutExpired`.
Either propagated out of `__exit__` before `_remove_container()` ran,
leaking the container (and, on timeout, the `start` process too).

Wrap the teardown so a broken pipe or timeout instead forces the process
down (kill + wait) and always falls through to container removal, while
still respecting CIBW_DEBUG_KEEP_CONTAINER. Pipe closes are now also
guarded so a flush-on-close against a dead pipe can't mask cleanup.

Add non-docker unit tests for the clean-exit, already-dead-bash, and
shutdown-timeout paths.

Assisted-by: ClaudeCode:claude-opus-4.8

### DIFF
--- a/cibuildwheel/oci_container.py
+++ b/cibuildwheel/oci_container.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import dataclasses
 import io
 import json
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (approved):
I feel this change is narrowly scoped, and since it doesn't change any other behaviour on successes, I am +1 for this. I can't think of any problems with it. LGTM!

PR #2889, opened 2026-06-02
fix: detect musl libc when downloading python-build-standalone

henryiii:
The current check is never true. Using `('', '')` might assume musl for something like portable Python, so used a more complex check.

### Problem

`platform.libc_ver()` only ever reports glibc — it returns `("", "")` on musl systems. The musl detection in `python_build_standalone.py`:

```python
libc_identifier = "musl" if "musl" in (platform.libc_ver() or ("", "")) else "gnu"
```

could therefore never be true, so musl hosts (e.g. Alpine) always selected the `gnu` python-build-standalone asset, which won't run there.

### Fix

Detect musl by shelling out to `ldd --version`, which prints `musl libc` on musl-based systems (and `GLIBC`/`GNU libc` on glibc). This probes the actual system loader rather than the host Python's linkage, so it stays correct even when cibuildwheel runs under a relocatable/statically-linked interpreter (including a python-build-standalone one). Falls back to glibc when `ldd` is unavailable.

Adds unit tests for musl/glibc/missing-`ldd` detection and the resulting libc identifier.

### DIFF
--- a/cibuildwheel/util/python_build_standalone.py
+++ b/cibuildwheel/util/python_build_standalone.py
@@ -5,6 +5,7 @@ import functools
 import hashlib
 import json
 import platform
+import subprocess
 import typing
 
 from filelock import FileLock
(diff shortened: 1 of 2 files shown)


joerick (approved):
Thanks. Not sure how I ended up with the `libc_ver()` approach!

PR #2890, opened 2026-06-02
ci: improve Azure test reliability

henryiii:
Adds reruns to some docker unit tests and the integration tests.

Reduces the reruns for the really long test (40min) to 2 instead of 3 (avoid a timeout)

~~Clear iOS when retrying~~ Just reduce to 1 retry instead of 2 (can timeout) (CC @freakboy3742).

Cache on Azure.

I think the azure runners are supposed to be 3-core, the logs report 4. Reducing concurrency is one way to reduce the flake, so let's see how long it takes. Edit: it does slow it down, going back to auto (4).

I originally bumped the version to 3.13 to get process_cpu_count, but that doesn't affect macOS.

<details><summary>:robot: AI original PR description :robot:</summary>
<p>

Azure has intermittent failures. A recent example ([build 9040](https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=9040)) failed `test/test_before_build.py::test` on macOS with the framework-Python spawn race:

```
python: posix_spawn: /Library/Frameworks/Python.framework/Versions/3.14/Resources/Python.app/Contents/MacOS/Python: Undefined error: 0
```

This is a known intermittent macOS issue with the python.org framework app stub, surfaced under concurrent process spawning — not a cibuildwheel bug. It built cp39–cp313 fine and only flaked on cp314. The `-x` flag means one such blip fails the whole job.

This PR attacks the flakiness from a few angles:

- **Retry integration tests.** `pytest-rerunfailures` is already in `required_plugins` but only used on a handful of tests. Add `--reruns=2 --reruns-delay=5` to the serial and non-serial integration runs so a single transient flake self-heals. It composes cleanly with `-x` (a test only counts as failed once reruns are exhausted).
- **Reduce macOS spawn contention.** The macOS runner was using 4 xdist workers (`os.cpu_count()`). Cap Azure at `--num-processes 3` to lower the spawn-race probability and peak resource pressure. (On macOS `process_cpu_count()` equals `cpu_count()` since there is no affinity mechanism, so an explicit cap is needed here rather than relying on auto-detection.)
- **Cache downloads across runs.** Relocate `CIBW_CACHE_PATH` to a stable `\$(Pipeline.Workspace)/.cibw-cache` and add a `Cache@2` task. This persists interpreter downloads and the per-worker test pip caches (which live under it), cutting the network-download flake surface. Only `CIBW_CACHE_PATH` is cached on purpose — exporting a job-wide `PIP_CACHE_DIR` would defeat the per-worker pip-cache isolation in `test/utils.py` and re-trigger the pip concurrency bug it works around.
- **Bump the Azure host interpreter to 3.13.**

</p>
</details> 


test/test_ios.py
```diff
@@ -41,14 +41,29 @@ def skip_if_ios_testing_not_supported() -> None:
...
 # iOS tests shouldn't be run in parallel, because they're dependent on calling
 # Xcode, and starting a simulator. These are both multi-threaded operations, and
 # it's easy to overload the CI machine if there are multiple test processes
 # running multithreaded processes. Therefore, they're put in the serial group,
 # which is guaranteed to run single-process.
 # This can also fail the first time sometimes.
 @pytest.mark.serial
-@pytest.mark.flaky(reruns=2)
+@pytest.mark.flaky(reruns=1)
+@pytest.mark.usefixtures("clean_ios_simulators")
```
freakboy3742: Reading between the lines here - I presume this is only applied to this single test because in the environment where this is an issue, we'll always be running the full test suite, this test will run first (clearing the iOS simulator state), and future tests are then known to be running on a "working" simulator? And on other environments where users might be running a single test (such as local testing), the iOS simulator is unlikely to be a borked state, and thus won't be a problem?

freakboy3742 (commented):
One request for clarification inline; assuming I've understood the intent, I think this makes sense.

The only question/concern is whether the change will have detrimental performance on non-Azure CI providers. GitHub Actions in particular has had some performance issues with starting new simulators, so I have a mild background concern that explicitly stopping simulators might cause slowness starting them again. However, it's possible the performance issues are only for starting simulators that weren't *previously* active, which shouldn't be an issue here.

However, it looks like the macOS-15 iOS Actions test took around 35 minutes which seems broadly consistent with historical performance (it's highly instance specific. but 25-35 minutes is consistent with other recent runs).

test/test_ios.py
henryiii: This is a parametrized test, and it was breaking on the second run. The goal was to pre-clean it so every run would start in a clean state. Open to ideas if you think something else would be better.

freakboy3742: In that case, I'm a little surprised that it's *only* a problem on the second paramterized run of this specific test, rather than a general problem. That said, Microsoft's suite of runner images have all sorts of weird iOS issues, so this wouldn't be the weirdest one by a long shot.

That said, if it works, I can't argue with success, and I can't say I have any better options, either.

henryiii:
```suggestion
```

test/test_ios.py
```diff
@@ -41,14 +41,29 @@ def skip_if_ios_testing_not_supported() -> None:
...
+    iOS builds occasionally hang for the full pytest timeout when a simulator
+    left booted (or wedged) by a previous build config is reused. Shutting them
+    down first means each parametrized config starts from a clean state. See the
+    180-minute hang on Azure (build 9043), which stalled on the second config.
+    """
+    if utils.get_platform() != "macos":
+        return
+    subprocess.run(["xcrun", "simctl", "shutdown", "all"], check=False, capture_output=True)
+
+
```
henryiii:
```suggestion
```

test/test_ios.py
henryiii: Looks like the latest run hung on Azure iOS. We can drop this for now, the macOS fix is more important and does seem to work.

bin/run_tests.py
```diff
@@ -113,6 +115,8 @@
         f"{' and '.join(['not serial', *marks])}",
         f"--numprocesses={args.num_processes}",
         "-x",
+        "--reruns=2",
+        "--reruns-delay=5",
```
joerick: This is affecting all integration tests, not just the marked ones, and not just on Azure, which doesn't match the PR description. 

I suppose don't love the idea of just blanket reruns as it could hide real flakiness in cibuildwheel (e.g. race conditions). We do see a lot of random network failures in our tests, so there is an argument to be had for this. I just wanted to check that we're all aware that's the change being made.

azure-pipelines.yml
```diff
@@ -19,39 +19,54 @@ jobs:
...
-        pythonVersion: "3.11"
+        pythonVersion: "3.13"
         testSelect: "native"
   timeoutInMinutes: 180
   pool:
     vmImage: $(imageName)
+  variables:
+    # Relocate cibuildwheel's download cache (interpreters) and the per-worker
+    # test pip caches (which live under it) to a stable, cacheable path.
+    CIBW_CACHE_PATH: $(Pipeline.Workspace)/.cibw-cache
```
joerick, line 52: I'd say this is unlikely to be a large portion of the test time or flakiness. I think it's probably ok to use it here (it might catch issues with our cache usage), but we definitely shouldn't be using it across the board with other CIs.

joerick (commented):
I couldn't find any documentation about the supposed 'framework-Python spawn race'. I couldn't find anything on Google about it, and Gemini confidently gave me some irrelevant links in response to a question about it. To me, that looks like an AI hallucination.

It does look like an error that's at a lower level than we're responsible for. Still, I'd rather the AI PR was honest in saying 'I'm not sure what's going on here, but this might help'.

bin/run_tests.py
henryiii: I could limit to just Azure if you prefer. But I think we always rerun if just one thing fails, probably before even checking to see what failed, since docker pull, pip installs, and a few other things can all be flaky.

azure-pipelines.yml
henryiii: It's not across the board with other CIs, just here.

henryiii:
Here's an example: https://dev.azure.com/joerick0429/cibuildwheel/_build/results?buildId=9054&view=logs&jobId=b40ceda2-7479-5d54-729a-f2369a5d22ad&j=b40ceda2-7479-5d54-729a-f2369a5d22ad&t=ea7acb75-f277-5bee-8ee3-3646b20ca533&s=6571a855-781b-52f4-2caa-1239027b30ba

I've been rerunning Azure about 2-3 times to get it to pass. That PR (#2892) I think I've rerun at least 5 times (the failure happens after <10 mins).

For the "Python" stub, I'd assume it's not a core issue, since only Azure has this problem, not GHA for example, which also uses it. But it does seem to be a problem with the launcher on their system. Stuff after the `:robot:` I didn't really touch or edit, though.

bin/run_tests.py
henryiii: I could also just mark tests we see flake out one by one. `test/test_0_basic.py::test[pip]` failed with `python: posix_spawn: /Library/Frameworks/Python.framework/Versions/3.14/Resources/Python.app/Contents/MacOS/Python: Undefined error: 0`.

joerick (approved):

bin/run_tests.py
joerick: Yeah, tbh, I'm open to adding flaky markers across the board on Azure if the posix_spawn thing becomes a persistent reason we need to hit the 'restart' button. Lets keep an eye on it.

PR #2891, opened 2026-06-02
fix: minor 4.0 issues in Android validation, docs, and docstring

henryiii:

A few small fixes found while reviewing for the 4.0 release.

- **Android**: `ANDROID_API_LEVEL` is consumed via `int(...)` in `localized_vars`. If a user sets it to a non-integer through the `environment` option, that previously surfaced as an uncaught `ValueError` deep in the build. Now `setup_env` validates it right after applying the user environment and raises a clear `FatalError`.
- **Docs**: Added a note to `repair-wheel-command` explaining that `delvewheel` is the default on Windows since 4.0, and that a platform-tagged wheel with no extension module may need `repair-wheel-command = ""`. Also updated the now-outdated FAQ wording ("if you're using delvewheel" → "uses delvewheel by default since 4.0").
- **venv**: Corrected the `_parse_pip_constraint_for_virtualenv` docstring — when `marker_env` is `None`, marker-bearing constraints are evaluated against the host's default environment, not skipped.

`prek -a` passes clean.

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -230,11 +230,6 @@ def setup_env(
 
     # Apply custom environment variables, and check environment is still valid
     build_env = build_options.environment.as_dictionary(build_env)
-    try:
-        int(build_env["ANDROID_API_LEVEL"])
-    except ValueError as e:
-        msg = f"ANDROID_API_LEVEL: {e}. This variable must be an integer."
-        raise errors.FatalError(msg) from e
     build_env["CIBUILDWHEEL_BUILD_IDENTIFIER"] = config.identifier
     build_env["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
     for command in ["python"] if use_uv else ["python", "pip"]:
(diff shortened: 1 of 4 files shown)


cibuildwheel/platforms/android.py
```diff
@@ -230,6 +230,12 @@ def setup_env(
 
     # Apply custom environment variables, and check environment is still valid
     build_env = build_options.environment.as_dictionary(build_env)
+    if not build_env["ANDROID_API_LEVEL"].isdigit():
+        msg = (
+            f"ANDROID_API_LEVEL must be an integer, not {build_env['ANDROID_API_LEVEL']!r}. "
+            "Check the value you set in the environment option."
+        )
+        raise errors.FatalError(msg)
```
mhsmith: The environment variable could also be set in cibuildwheel's initial environment, as opposed to the configuration file.

Also, `isdigit` returns true for any Unicode digits, not just ASCII ones.

For those reasons, I think it would be better to handle this using a try/catch at the point where `int` is called, and include the original exception message in the FatalError. For example: `ANDROID_API_LEVEL: invalid literal for int() with base 10: 'hello'. This variable must be an integer.`

henryiii: :robot: _AI text below_ :robot:

Good points. Switched to a try/except around `int()` so it handles env-set values and non-ASCII Unicode digits correctly, and surfaces the original exception message — e.g. `ANDROID_API_LEVEL: invalid literal for int() with base 10: 'hello'. This variable must be an integer.`

henryiii: I'm not sure what makes Claude reply to commits or not. I'd rather it ask me. :) At least it did add the AI text header I set.

mhsmith: I was referring to the existing `int` call in `localized_vars`; there's no need to test it elsewhere.

henryiii: Ah, of course! Much better.

mhsmith (approved):

PR #2892, opened 2026-06-03
fix: remove potential partial cache-population in case of error

mayeut:
This address the "Resource Management & Concurrency" section of #2854

This PR introduces a shared “cleanup on error” mechanism intended to prevent partially-populated cache entries (and related artifacts) from being left behind when downloads/extractions/install steps fail, and applies it across several cache-population code paths.

**Changes:**
- Added a `remove_on_error` context manager in `cibuildwheel.util.file` to delete a target path if an exception occurs.
- Wrapped multiple cache-population operations (downloads, nuget installs, pyodide xbuildenv installs, etc.) with `remove_on_error` to clean up partial cache state on failure.
- Added unit tests covering keep-vs-remove behavior for files and directories.

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -25,7 +25,13 @@ from cibuildwheel.logger import log
 from cibuildwheel.platforms.macos import install_cpython as install_build_cpython
 from cibuildwheel.util import resources
 from cibuildwheel.util.cmd import call, shell, split_command
-from cibuildwheel.util.file import CIBW_CACHE_PATH, copy_test_sources, download, move_file
+from cibuildwheel.util.file import (
+    CIBW_CACHE_PATH,
+    copy_test_sources,
+    download,
+    move_file,
+    remove_on_error,
+)
 from cibuildwheel.util.helpers import prepare_command, unwrap_preserving_paragraphs
 from cibuildwheel.util.packaging import find_compatible_wheel
 from cibuildwheel.venv import constraint_flags, virtualenv
@@ -140,8 +146,9 @@ def install_target_cpython(tmp: Path, config: PythonConfiguration, free_threadin
         if not installation_path.exists():
             downloaded_tar_gz = tmp / ios_python_tar_gz
             download(config.url, downloaded_tar_gz, sha256=config.sha256)
-            installation_path.mkdir(parents=True, exist_ok=True)
-            call("tar", "-C", installation_path, "-xf", downloaded_tar_gz)
+            with remove_on_error(installation_path):
+                installation_path.mkdir(parents=True)
+                call("tar", "-C", installation_path, "-xf", downloaded_tar_gz)
             downloaded_tar_gz.unlink()
 
     return installation_path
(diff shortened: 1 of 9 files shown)


cibuildwheel/platforms/ios.py
```diff
@@ -140,8 +146,9 @@ def install_target_cpython(tmp: Path, config: PythonConfiguration, free_threadin
         if not installation_path.exists():
             downloaded_tar_gz = tmp / ios_python_tar_gz
             download(config.url, downloaded_tar_gz, sha256=config.sha256)
```
freakboy3742, line 148: Should the `download` call have a `remove_on_error` wrapper as well? (macOS has the same pattern)

More generally - is there any reason that `download()` shouldn't incorporate `remove_on_error` handling? Is there any use case for downloading and preserving a failed download?

mayeut: All non-cache persisted `download` cases should be downloads to a temp directory -as is the case here - which is already taken care of when cibuildwheel exits.
I prefer the consistency of `FileLock`+`remove_on_error` always appearing together when it comes to preventing cache pollution, even if the `download` helper was updated with `remove_on_error`.
I consider that updating `download` might be worthwhile but orthogonal to the issue being fixed here and if it were updated, I'd rather keep the `FileLock`+`remove_on_error` consistency and be in a "belt and braces" situation in case like:
```diff
diff --git a/cibuildwheel/platforms/windows.py b/cibuildwheel/platforms/windows.py
index 7dbd3a56..f55fc54d 100644
@@ -112,7 +113,8 @@ def _ensure_nuget() -> Path:
     nuget = CIBW_CACHE_PATH / "nuget.exe"
     with FileLock(str(nuget) + ".lock"):
         if not nuget.exists():
-            download("https://dist.nuget.org/win-x86-commandline/latest/nuget.exe", nuget)
+            with remove_on_error(nuget):
+                download("https://dist.nuget.org/win-x86-commandline/latest/nuget.exe", nuget)
     return nuget
 
 
```

freakboy3742: Ah - I missed the intersection of this with `tmp` usage. In which case - ignore my comment.

cibuildwheel/platforms/pyodide.py
```diff
@@ -110,18 +112,19 @@ def install_emscripten(env: dict[str, str], version: str, xbuildenv_cache_path:
...
+                "xbuildenv",
+                "install-emscripten",
+                "--force",
+                "--version",
+                version,
+                "--path",
+                str(xbuildenv_cache_path),
+                env=env,
+                cwd=CIBW_CACHE_PATH,
+            )
```
agriyakhetarpal, line 127: No action needed here: just noting that with our `--force` argument to install the Emscripten toolchain, `remove_on_error` will be redundant as we always clear and download the artifacts and never use the cache on subsequent builds.

agriyakhetarpal (approved):
Thanks!

cibuildwheel/platforms/pyodide.py
henryiii: Should we remove the context manager here then?

agriyakhetarpal: I'd be okay with keeping it just in case and for consistency; it's harmless.

mhsmith (commented):
The Android changes look fine to me.

cibuildwheel/platforms/pyodide.py
henryiii: Okay

PR #2894, opened 2026-06-04
Drop use of `Unbuffered` wrapper class

agriyakhetarpal:
We can replace it with `sys.stdout.reconfigure(line_buffering=True)`. This takes and completes review point 6 of #2854.

`write_through=True` is not enough, as it only forces the text layer through to the `BufferedWriter`, which still block-buffers a pipe.

The `io.TextIOWrapper` instance check isn't really needed per se, but Mypy raises otherwise, so I added it. Even typeshed mentions the following advice, based on what I see locally in Pylance at `.vscode/extensions/ms-python.vscode-pylance-2026.2.1/dist/typeshed-fallback/stdlib/sys/__init__.pyi`:

```python
# TextIO is used instead of more specific types for the standard streams,
# since they are often monkeypatched at runtime. At startup, the objects
# are initialized to instances of TextIOWrapper, but can also be None under
# some circumstances.
#
# To use methods from TextIOWrapper, use an isinstance check to ensure that
# the streams have not been overridden:
#
# if isinstance(sys.stdout, io.TextIOWrapper):
#    sys.stdout.reconfigure(...)
stdin: TextIO | MaybeNone
stdout: TextIO | MaybeNone
stderr: TextIO | MaybeNone
```

### DIFF
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -344,11 +344,11 @@ def build_in_directory(args: CommandLineArguments) -> None:
     # Add CIBUILDWHEEL environment variable
     os.environ["CIBUILDWHEEL"] = "1"
 
-    # Python is buffering by default when running on the CI platforms, giving
-    # problems interleaving subprocess call output with unflushed calls to
-    # 'print'
+    # Python block-buffers stdout when it isn't a tty, which de-interleaves
+    # `print` output and direct fd-1 writes from subprocesses sharing this
+    # stdout. line_buffering allows each newline to flush all the way to fd-1.
     if isinstance(sys.stdout, io.TextIOWrapper):
-        sys.stdout.reconfigure(write_through=True)
+        sys.stdout.reconfigure(line_buffering=True)
 
     # create the cache dir before it gets printed & builds performed
     CIBW_CACHE_PATH.mkdir(parents=True, exist_ok=True)
--- a/cibuildwheel/__main__.py
+++ b/cibuildwheel/__main__.py
@@ -4,6 +4,7 @@ import argparse
 import contextlib
 import dataclasses
 import functools
+import io
 import os
 import shutil
 import sys
@@ -30,7 +31,7 @@ from cibuildwheel.util.resources import read_all_configs
 TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterable, Sequence
-    from typing import Any, Literal, TextIO
+    from typing import Literal
 
 
 @dataclasses.dataclass
@@ -42,23 +43,6 @@ class GlobalOptions:
 class FileReport:
     name: str
     size: str
-
-
-# Taken from https://stackoverflow.com/a/107717
-class Unbuffered:
-    def __init__(self, stream: TextIO) -> None:
-        self.stream = stream
-
-    def write(self, data: str) -> None:
-        self.stream.write(data)
-        self.stream.flush()
-
-    def writelines(self, data: Iterable[str]) -> None:
-        self.stream.writelines(data)
-        self.stream.flush()
-
-    def __getattr__(self, attr: str) -> Any:  # noqa: ANN401
-        return getattr(self.stream, attr)
 
 
 def main() -> None:
@@ -363,7 +347,8 @@ def build_in_directory(args: CommandLineArguments) -> None:
     # Python is buffering by default when running on the CI platforms, giving
     # problems interleaving subprocess call output with unflushed calls to
     # 'print'
-    sys.stdout = Unbuffered(sys.stdout)
+    if isinstance(sys.stdout, io.TextIOWrapper):
+        sys.stdout.reconfigure(write_through=True)
 
     # create the cache dir before it gets printed & builds performed
     CIBW_CACHE_PATH.mkdir(parents=True, exist_ok=True)


henryiii:
Here's the script Opus 4.8 wrote to test this:

```python
  import io, sys, os

  mode = sys.argv[1]
  if mode == "writethrough":
      if isinstance(sys.stdout, io.TextIOWrapper):
          sys.stdout.reconfigure(write_through=True)
  elif mode == "linebuffer":
      sys.stdout.reconfigure(line_buffering=True)
  # default: no change

  sys.stdout.write("PYTHON-LINE\n")
  # emulate a subprocess writing directly to fd 1
  os.write(1, b"SUBPROC-LINE\n")
  EOF
  echo "=== default (pipe) ==="; python3 bufftest.py default | cat
  echo "=== writethrough (pipe) ==="; python3 bufftest.py writethrough | cat
  echo "=== linebuffer (pipe) ==="; python3 bufftest.py linebuffer | cat)
```
```
=== default (pipe) ===
SUBPROC-LINE
PYTHON-LINE
=== writethrough (pipe) ===
SUBPROC-LINE
PYTHON-LINE
=== linebuffer (pipe) ===
PYTHON-LINE
SUBPROC-LINE
```


```python
  import sys, os
  from typing import Any
  class Unbuffered:
      def __init__(self, stream): self.stream = stream
      def write(self, data):
          self.stream.write(data); self.stream.flush()
      def __getattr__(self, attr): return getattr(self.stream, attr)
  sys.stdout = Unbuffered(sys.stdout)
  sys.stdout.write("PYTHON-LINE\n")
  os.write(1, b"SUBPROC-LINE\n")
  EOF
  echo "=== old Unbuffered (pipe) ==="; python3 bufftest2.py | cat
  echo
  echo "=== python version ==="; python3 --version)
```
```
=== old Unbuffered (pipe) ===
PYTHON-LINE
SUBPROC-LINE

=== python version ===
Python 3.14.5
```

Confirmed a regression. Here's my review.

---

## Review: PR #2894 — Drop use of `Unbuffered` wrapper class

**Overview:** Replaces the hand-rolled `Unbuffered` `TextIO` wrapper (which flushed after every `write`/`writelines`) with `sys.stdout.reconfigure(write_through=True)`, guarded by an `isinstance(sys.stdout, io.TextIOWrapper)` check. Net −15 lines, cleaner. Implements point 6 of #2854.

### 🔴 Correctness — `write_through=True` is not equivalent to the old behavior (likely a regression)

The stated purpose of this code (per the retained comment) is to keep `print()` output correctly interleaved with subprocess output **on CI, where stdout is a pipe and therefore block-buffered**. The old `Unbuffered.write` called `self.stream.flush()` after every write, flushing all the way down to the OS fd. `write_through=True` does **not** do that — it only forces the `TextIOWrapper`'s text layer to hand bytes to the underlying `BufferedWriter` immediately; the `BufferedWriter` still block-buffers when connected to a pipe.

I verified this on Python 3.14 by interleaving a `sys.stdout.write` with a direct `os.write(1, ...)` (emulating a subprocess sharing fd 1), piped through `cat`:

| mode                     | output order                                 |
| ------------------------ | -------------------------------------------- |
| default (no change)      | `SUBPROC` then `PYTHON` ❌                   |
| **`write_through=True`** | `SUBPROC` then `PYTHON` ❌ (same as default) |
| `line_buffering=True`    | `PYTHON` then `SUBPROC` ✅                   |
| old `Unbuffered`         | `PYTHON` then `SUBPROC` ✅                   |

So under the exact scenario this code exists to fix, `write_through=True` behaves like doing nothing, while the old wrapper produced correct ordering.

**Suggested fix** — use `line_buffering=True`, which flushes the binary buffer on each newline and so covers every `print()`:

```python
if isinstance(sys.stdout, io.TextIOWrapper):
    sys.stdout.reconfigure(line_buffering=True)
```

(If you want to match the old "flush on *every* write, even without a newline" semantics exactly, you'd combine `line_buffering=True, write_through=True`, but `line_buffering=True` alone is almost certainly sufficient here since the buffering problem is about `print` output.)

### ✅ Looks good

- **Import cleanup is correct.** `TextIO` and `Any` are no longer referenced and are properly removed; `Iterable` is correctly *kept* (still used at `__main__.py:431` in `detect_errors`). `io` added appropriately.
- **`isinstance` guard is reasonable** and matches the typeshed guidance you quoted. It also makes the code safely no-op when `sys.stdout` has been replaced/redirected or is `None`, which is arguably safer than the old unconditional wrap.

### Minor / process

- **Test coverage:** there's no test asserting interleaving order, so this regression wouldn't be caught by CI. Given it's already untested, I wouldn't block on adding one, but it's worth noting the behavior change is invisible to the suite.
- Worth a quick manual sanity check on a real CI run (e.g., confirm `print` preamble lines still appear before the first subprocess build output) before merging.

### Recommendation
Request changes: swap `write_through=True` → `line_buffering=True` (or verify empirically that `write_through` is sufficient in your target CI environments — my test says it is not). Everything else in the PR is clean and ready.

agriyakhetarpal:
Thanks, I was able to reproduce this with the script! Working to test a few things locally; will update in a bit.

agriyakhetarpal:
Okay, confirmed by running a few builds. Noting this for future reference here: a side-by-side comparison around the first subprocess (`pip install --upgrade pip`) is as follows.

##### `line_buffering=True` is correct

```
+ python -m pip install --upgrade pip            <-- cibuildwheel command echoes
Collecting pip                                   <-- pip subprocess output shows
Installing collected packages: pip
Successfully installed pip-26.1.2
+ python -m pip install --upgrade pip -c file:   <-- next echo is in order
```

##### `write_through=True` is broken

```
Collecting pip                                   <-- subprocess output comes FIRST
Installing collected packages: pip
Successfully installed pip-26.1.2
+ python -m pip install --upgrade pip            <-- echo arrives AFTER its own output
+ python -m pip install --upgrade pip -c file:
```


The cibuildwheel print lines get held in the BufferedWriter and are flushed in chunks, so command echoes land after the subprocess output they were meant to announce.

Thanks!

joerick (approved):

Issue #2883, opened 2026-06-01
4.0rc1: GraalPy fails on delvewheel


henryiii:
### Description

```
  + delvewheel repair -w C:\Users\runneradmin\AppData\Local\Temp\cibw-run-fx0fgtbm\gp311_242-win_amd64\repaired_wheel -v C:\Users\runneradmin\AppData\Local\Temp\cibw-run-fx0fgtbm\gp311_242-win_amd64\built_wheel\boost_histogram-1.7.3.dev13+gf9bf9819f-graalpy311-graalpy242_311_native-win_amd64.whl
  extracting boost_histogram-1.7.3.dev13+gf9bf9819f-graalpy311-graalpy242_311_native-win_amd64.whl to C:\Users\RUNNER~1\AppData\Local\Temp\tmps0m87knd
  repairing C:\Users\runneradmin\AppData\Local\Temp\cibw-run-fx0fgtbm\gp311_242-win_amd64\built_wheel\boost_histogram-1.7.3.dev13+gf9bf9819f-graalpy311-graalpy242_311_native-win_amd64.whl
  finding DLL dependencies
  analyzing package-level extension module boost_histogram\_core.graalpy242-311-native-x86_64-win32.pyd
  Traceback (most recent call last):
    File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-fx0fgtbm\gp311_242-win_amd64\build\venv\Lib\site-packages\pefile.py", line 3030, in __parse__
      self.__data__ = mmap.mmap(self.fileno, 0, mmap.MAP_PRIVATE)
  OSError: [Errno 5] The process cannot access the file because it is being used by another process
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\graalpy-24.2.2-windows-amd64\Lib\runpy.py", line 198, in _run_module_as_main
      return _run_code(code, main_globals, None,
    File "C:\Users\runneradmin\AppData\Local\pypa\cibuildwheel\Cache\graalpy-24.2.2-windows-amd64\Lib\runpy.py", line 88, in _run_code
      exec(code, run_globals)
    File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-fx0fgtbm\gp311_242-win_amd64\build\venv\Scripts\delvewheel.exe\__main__.py", line 10, in <module>
[...]

Issue #2754, opened 2026-03-02
GraalPy wheel build fails when using uv in build frontend on Windows


joerick:
When building GraalPy wheels with uv in the build frontend (`build[uv]` or `uv`), the build fails with a file-not-found error during get_requires_for_build_wheel. This was observed on Windows with both GraalPy 3.11 and 3.12.

Error message:

```
Building wheel...
× Failed to build
│ C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-1\popen-gw3\test_uv_0\project
├─▶ The build backend returned an error
╰─▶ Call to setuptools.build_meta:__legacy__.get_requires_for_build_wheel failed:
    failed to open file D:\_temp\setup-uv-cache\builds-v0\.tmpt0SnPs\get_requires_for_build_wheel.txt:
    The system cannot find the file specified. (os error 2) (exit code: 0)

cibuildwheel: C:\hostedtoolcache\windows\uv\0.10.7\x86_64\uv.EXE build --python=python C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-1\popen-gw3\test_uv_0\project --wheel --out-dir=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-jhxnk6y1\gp312_250-win_amd64\built_wheel failed with code 2.
```

In #2741 I've skipped those tests and documented the known issue. But it would be nice to get to the bottom of this.

cc @timfel

timfel:
I've spent a couple of days trying to reproduce this locally, it does not, even after many iterations with a windows setup similar to the CI (including 2 drives). I'lldebug on github actions in my fork

joerick:
Thanks. Never used it myself, but https://github.com/mxschmitt/action-tmate might be useful...

henryiii:
This is still an issue, I think?

PR #2741, opened 2026-02-13
Debug GraalPy/uv test failures on Windows

joerick:
Debugging windows failures in #2734. Perhaps the error only occurs on GP311...

PR #2895, opened 2026-06-05
feat: remove GraalPy 3.11 (gp311) support

henryiii:
GraalPy 3.11 interacts poorly with delvewheel on Windows. I've seen two errors, #2883, and `ImportError: cannot load C:\Users\runneradmin\AppData\Local\Temp\cibw-run-dwn70_yl\gp311_242-win_amd64\venv-test\Lib\site-packages\python_example.graalpy242-311-native-x86_64-win32.pyd`. AFAICT, both work fine in GraalPy 3.12. In https://github.com/pypa/cibuildwheel/pull/2741#issuecomment-3925628019, @timfel said we could drop 3.11 in March. This is after March, so I think the best path forward is to just drop it before 4.0.0 instead of complicating the default windows repair option. Though we could introduce `graalpy-eol` instead if better.

This gets rid of several workarounds, too.

Removes GraalPy 3.11 / GraalPy 24.2 (`gp311_242`) support, as agreed in #2741 — a ~6-month overlap after the GraalPy 25 (`gp312`) release. GraalPy 3.12 (`gp312_250`) stays.

### What's removed
- All `gp311_242-*` entries from `build-platforms.toml` (linux x86_64/aarch64, macOS x86_64/arm64, win_amd64).
- The GraalPy 24-only workarounds explicitly marked *"Remove when GraalPy 24.x is dropped"*:
  - the Visual Studio compiler-discovery (`vswhere`) block and the build-isolation (`-n`) workaround for oracle/graalpython#491 in the Windows backend;
  - the issue-491 Windows skip/filter hacks in `test_pep518` and `test_dependency_versions` (added in the original GraalPy PR #1538 and unneeded since GraalPy 25, per #2597).
- The `GraalPy 3.11 v24.2` rows in the README and `docs/options.md` support tables, plus the matching test fixtures/IDs.

### Intentionally kept (apply to all GraalPy, not gp311-specific)
- the `uv`-doesn't-support-graalpy skip in `test/conftest.py` (tracked by #2754);
- the `graalpy` enable group;
- the generated `constraints-python31{1,2}.txt` pip-on-Windows pins.

### Testing
- `prek -a` clean; all 790 unit tests pass.

cc @timfel — this is the gp311 drop we discussed in #2741.


joerick (commented):
Looks good to me! Lets see what @timfel thinks, but I think the major version would be a good time to drop gp311.

agriyakhetarpal (approved):

timfel (approved):

PR #2896, opened 2026-06-05
Add instructions for building CUDA wheels

chyunsu3:
In the tips and tricks page, add instructions for building CUDA wheels for Linux.


henryiii:
I tested this in https://github.com/pybind/cmake_example/pull/194, built locally via cibuildwheel on my Mac, and ran on a linux GPU machine. Works great!

henryiii:
Should we only show the env var version, or maybe have tabs with the config version? It's not perfectly suited to config, as often you probably want multiple CUDA versions. I have ideas for that, but I rather want to see where wheel variants go. (I'm okay with it as is)

henryiii (approved):

PR #2898, opened 2026-06-06
chore: use ConfigurationError when package_dir is outside cwd

mayeut:
This resolves one of the "Error Handling & Logic" items in #2854
Mostly, this changes the UX from raw traceback to a simple error log without traceback (by default) and a specific error code.

Test written with the assistance of copilot.

Replace a generic Exception with errors.ConfigurationError when package_dir is not inside the working directory. This makes the error type consistent with the project's error handling and allows the CLI to treat this as a fatal error with the expected exit behavior.


agriyakhetarpal (approved):

henryiii:

Moved the new test out of a standalone `unit_test/linux_build_test.py` and into the existing `unit_test/linux_build_steps_test.py`, which already covers the linux platform build orchestration. This avoids two near-identically-named test files for the same module. The merged test reuses the file's existing import boilerplate and uses the module-qualified `cibuildwheel.platforms.linux.build(...)` call style; also renamed it `..._raises_configuration_error` to match what it asserts.

PR #2904, opened 2026-06-07
Clarify shell used for command options

### LINKED ISSUE #1779: [docs] Make shell interpreter explicit
## Summary

The docs should make it clear what shell is used when running commands such as `before-build`, `before-all`, `before-test`, etc.

## Motivation

`cibuildwheel` allows for an arbitrary shell script to be executed in many places. The problem here is that it is unclear what shell is actually being used.

Naively, I would expect that:

- Linux builds use `/bin/sh` or `/bin/bash`.
- macOS builds use `/bin/zsh` as that is the default shell for macOS, though I would not be surprised if `/bin/bash` was used for consistency with Linux.
- Windows build use either a Windows `bash.exe` for consistency, or the native `cmd.exe` or `pwsh.exe`.

I would rather avoid the trial-and-error, and the docs should make it really clear what shell is used.

### Build log

_No response_

### CI config

_No response_

Himanshuagrawal4:
## Summary
- Add a note explaining which platform shell runs shell command options.
- Mention Linux container command execution and macOS/Windows shell=True behavior.
- Recommend invoking a specific interpreter explicitly when needed.

Closes #1779

## Validation
- git diff --check

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -729,6 +729,17 @@ The placeholder `{package}` can be used here; it will be replaced by the path to
 
 On Windows and macOS, the version of Python available inside `before-all` is whatever is available on the host machine. On Linux, a modern Python version is available on PATH.
 
+!!! note "Shell selection"
+
+    Shell command options such as `before-all`, `before-build`,
+    `repair-wheel-command`, and `before-test` are executed by a platform shell.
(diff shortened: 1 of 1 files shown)


joerick (approved):
Looks good to me. It would be nice if the point could be made in a more general way wrt. the options, but this is nice and neat and reads well.

Issue #2908, opened 2026-06-11
Code review findings: error-handling gaps and minor cleanups


henryiii:
:robot: _AI text below_ :robot:

A code review pass over the codebase (bugs, performance, simplifications, modernizations) turned up the following. All line references are at 55c8985c. Overall the codebase is in very good shape — these are mostly error-handling consistency gaps and small cleanups.

## Error handling

1. **`cibuildwheel/platforms/pyodide.py:474` — repair step failure raises raw `StopIteration`.** Every other platform wraps the repaired-wheel glob (`windows.py:546-549`, `macos.py:598-601`, `ios.py:583-586`, and `linux.py:371-377` via `match`) and raises `RepairStepProducedNoWheelError`. Pyodide alone has a bare `next(repaired_wheel_dir.glob("*.whl"))`. A user-supplied repair command that produces no wheel gives an unhelpful traceback instead of the dedicated error.

2. **Unchecked built-wheel globs.** `linux.py:349` uses `container.glob(built_wheel_dir, "*.whl")[0]`, and `macos.py:569` / `windows.py:525` use bare `next(built_wheel_dir.glob("*.whl"))`. If a frontend exits 0 without producing a wheel (e.g. a misconfigured custom frontend), the user gets `IndexError`/`StopIteration` instead of a clear "build produced no wheel" error.

3. **`cibuildwheel/platforms/android.py:412-417` — `assert` used to validate external command output.** The parsing of `android.py env` output uses `assert token == "export"` / `assert sep == "="`. These vanish under `python -O`, and produce an unhelpful `AssertionError` if CPython's `android-env.sh` output format ever changes.

4. **`cibuildwheel/platforms/ios.py:95-98` — unchecked `matching[0]`** when pairing iOS configs with macOS build URLs. Only triggers if the bundled `build-platforms.toml` resources get out of sync, but since `bin/update_pythons.py` edits those files mechanically, a clear "no matching macOS configuration" error would beat an `IndexError`.

## Performance / cleanups

5. **`cibuildwheel/util/python_build_standalone.py:146`** — `hashlib.sha256(asset_cache_path.read_bytes())` loads the whole ~30-60 MB archive into memory; `hashlib.file_digest()` (3.11+) streams.

6. **`cibuildwheel/logger.py:366-372`** — the `colors`/`symbols` properties construct new `Colors`/`Symbols` objects on every access; the enabling flags never change after init, so they can be cached.

[...]

PR #2909, opened 2026-06-11
fix: raise clear errors when no wheel is produced

henryiii:

This addresses items 1-4 of #2908, replacing several bare tracebacks with clear `FatalError`s when a build does not produce a wheel or platform resources are inconsistent.

Changes:

- Add `errors.BuildProducedNoWheelError` (exit code 10) for the case where a build frontend exits successfully but produces no wheel.
- Guard the built-wheel globs on Linux, macOS, Windows, iOS, and Pyodide so they raise `BuildProducedNoWheelError` instead of a bare `IndexError` / `StopIteration`. (Android already handled this via its `glob1` helper.)
- Wrap Pyodide's repaired-wheel glob in `try/except StopIteration` to raise `RepairStepProducedNoWheelError`, matching the existing macOS/Windows pattern.
- Replace the `assert` statements that validate `android.py env` output with `errors.FatalError`, including the unexpected token in the message.
- Raise `errors.FatalError` naming the missing macOS identifier when `ios.py` cannot find a matching macOS configuration, instead of an `IndexError`.

Part of #2908

### DIFF
--- a/cibuildwheel/errors.py
+++ b/cibuildwheel/errors.py
@@ -109,3 +109,19 @@ class AuditCommandFailedError(FatalError):
     def __init__(self, message: str) -> None:
         super().__init__(message)
         self.return_code = 9
+
+
+class BuildProducedNoWheelError(FatalError):
+    def __init__(self) -> None:
+        message = textwrap.dedent(
+            """
+            Build failed because the build frontend completed successfully but
+            did not produce a wheel.
+
+            This usually indicates a problem with your project configuration or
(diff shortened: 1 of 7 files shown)


agriyakhetarpal:
Oops, sorry, pressed the wrong button on my phone while I was reading this!

agriyakhetarpal:
Please feel free to revert if you like, though I agree with the change and I meant to approve it, not merge it right away 🤦🏻

PR #2910, opened 2026-06-11
chore: minor cleanups and perf tweaks from code review

henryiii:

This PR addresses items 5-9 from the code review checklist in #2908 with five small cleanup/perf fixes:

- **Streaming SHA-256 verification** (`util/file.py`, `util/python_build_standalone.py`): Replace `hashlib.sha256(path.read_bytes()).hexdigest()` with `hashlib.file_digest(f, "sha256").hexdigest()` to avoid loading entire archives (potentially hundreds of MB) into memory at once. `hashlib.file_digest` was added in Python 3.11.0, which is the minimum supported version.

- **`functools.cached_property` for `Logger.colors` / `Logger.symbols`** (`logger.py`): The `Colors` and `Symbols` objects were reconstructed on every property access, but `colors_enabled` and `unicode_enabled` never change after `__init__`. Converting to `cached_property` avoids the repeated allocations.

- **Remove redundant `.strip()`** (`platforms/windows.py`): `where_pip` was already `.strip()`-ed at assignment (line 329); the duplicate call on line 331 was dead code.

- **Remove unreachable `python_base_dir.exists()` guard** (`util/python_build_standalone.py`): Both callers of `create_python_build_standalone_environment` pass a fresh temp subdirectory, so the guard can never trip. Verified by inspecting `pyodide.py` (passes `tmp / "base"`) and `android.py` (passes a newly `mkdir`-ed `build_path`).

- **Comment explaining the `getattr` shim** (`util/file.py`): Notes why the shim is needed (Python 3.11.0-3.11.3 support) and what the cleanup looks like once the minimum is raised to 3.11.4+/3.12.

Part of #2908

### DIFF
--- a/cibuildwheel/util/python_build_standalone.py
+++ b/cibuildwheel/util/python_build_standalone.py
@@ -219,6 +219,7 @@ def create_python_build_standalone_environment(
     )
 
     python_base_dir = temp_dir / "pbs"
+    assert not python_base_dir.exists()
     extract_tar(archive_path, python_base_dir)
 
     return _find_python_executable(python_base_dir)
--- a/cibuildwheel/util/python_build_standalone.py
+++ b/cibuildwheel/util/python_build_standalone.py
@@ -143,7 +143,8 @@ def _download_or_get_from_cache(
         asset_cache_path = cache_dir / asset_filename
         if asset_cache_path.is_file():
             if sha256:
-                computed = hashlib.sha256(asset_cache_path.read_bytes()).hexdigest()
+                with asset_cache_path.open("rb") as f:
+                    computed = hashlib.file_digest(f, "sha256").hexdigest()
                 if computed != sha256:
                     print(
                         f"Cached python_build_standalone SHA256 mismatch for {asset_cache_path}; redownloading."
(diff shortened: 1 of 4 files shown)


cibuildwheel/util/python_build_standalone.py
```diff
@@ -218,9 +219,6 @@ def create_python_build_standalone_environment(
     )
 
     python_base_dir = temp_dir / "pbs"
-    if python_base_dir.exists():
-        msg = f"python-build-standalone directory already exists: {python_base_dir}"
-        raise PythonBuildStandaloneError(msg)
```
joerick, line 223: I'd return to the assert rather than removing the check. The state of the filesystem is the responsibility of the programmer and is a necessary precondition to running this function.
```suggestion
    assert not python_base_dir.exists()
```

joerick (approved):

cibuildwheel/util/python_build_standalone.py
henryiii: > "Applying suggestions on deleted lines is currently not supported. **Refresh and try again.**"

What, GitHub? You don't support it but you want me to refresh and try again until you do support it? 🤣

joerick: Sorry about that. I typed the suggestion tag manually thinking I'd get away with it.

PR #2911, opened 2026-06-12
Fix dead Pyodide env info link, remove mention of alpha ABI

Viicos:
314.0.0 was released, so should be stable and recommended to upload to PyPI.

For the dead link, I went with the published GH pages url (https://github.com/pyodide/pyodide/blob/cf946560d9d1923f87fbcbdc2206cfa62ed41e3b/.github/workflows/pages.yml) to avoid having it breaking again if the repo path changes again.


agriyakhetarpal (approved):
Thanks for catching this, @Viicos!

PR #2913, opened 2026-06-12
Fix quoting of `test-requires` and `audit-requires` options for PEP 508 specifiers

### LINKED ISSUE #2912: Cannot use dependency specifiers in `test-requires`
### Description

Came across this in zarr-developers/numcodecs#529 when testing cibuildwheel v4.1.0.

```bash
  + pip install pytest 'pyzstd;' python_version '>=' ''"'"'3.14'"'"'' msgpack crc32c zfpy 'zarr>=3' importlib_metadata
  ERROR: Invalid requirement: '>=': Expected package name at the start of dependency specifier
      >=
      ^
```

This happens when I do something like:

```
select = "*-pyodide_wasm32"
config-settings = {setup-args = ["-Dsse2=disabled", "-Davx2=disabled"]}
test-requires = [
  "pytest",
  "pyzstd; python_version >= '3.14'",
  "msgpack",
  "crc32c",
  "zfpy",
  "zarr>=3",
  "importlib_metadata",
]
```

---

I have a workaround for now, where I can do:

```
# pyzstd depends on backports-zstd (no wasm32 wheel) on Python < 3.14.
# On 3.14+ it has no extra deps and uses the stdlib compression.zstd
select = "cp314-pyodide_wasm32"
test-requires = [
  "pytest",
  "pyzstd",
  "msgpack",
  "crc32c",
  "zfpy",
  "zarr>=3",
  "importlib_metadata",
]
```

for Pyodide 314.0.0, but obviously it causes some repetition.

### Build log

https://github.com/zarr-developers/numcodecs/actions/runs/27441719109/job/81117125162?pr=529

### CI config

https://github.com/zarr-developers/numcodecs/pull/529

agriyakhetarpal:
`test-requires` and `audit-requires` were formatted with a space-joined `ListFormat(sep=" ")` without any quoting and then split with `.split()`, and that breaks any requirement containing a space, such as a dependency specifier of the form

```python
"pyzstd; python_version >= '3.14'"
```

`test-sources` correctly uses `ListFormat(sep=" ", quote=shlex.quote)` and `shlex.split()`, so this is a trivial fix where the same pattern should be applied for both places.

Closes #2912

- [x] Check if any other option(s) also have the same problem
- [x] Add a test

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -934,9 +934,11 @@ class Options:
             )
             audit_command = audit_command_str.split(" && ") if audit_command_str else []
 
-            audit_requires = self.reader.get(
-                "audit-requires", option_format=ListFormat(sep=" ")
-            ).split()
+            audit_requires = shlex.split(
+                self.reader.get(
+                    "audit-requires", option_format=ListFormat(sep=" ", quote=shlex.quote)
+                )
+            )
 
             return BuildOptions(
                 globals=self.globals,
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -833,9 +833,11 @@ class Options:
                     msg = f"Failed to parse test runtime config. {e}"
                     raise errors.ConfigurationError(msg) from e
 
-            test_requires = self.reader.get(
-                "test-requires", option_format=ListFormat(sep=" ")
-            ).split()
+            test_requires = shlex.split(
+                self.reader.get(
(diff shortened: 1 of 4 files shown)


henryiii (approved):
I feel like I hit this bug before, and seems like I forgot to fix it. Or maybe it was just inside groups - actually, I think that was it.

joerick:
good catch, thanks

PR #2915, opened 2026-06-14
feat: use digests instead of tags for pinned images

mayeut:
A number of improvements regarding the supply-chain security have already made it into v4.0.0
This pull request extends those by using digests rather than tags for pinned images.

The major drawbacks are UX and test maintenance/complexity:
- UX: the summary used to come with a user readable image tag. Digests are not human friendly and the added complexity to manage both tags and digests for this summary might not be worth it. The tags have nevertheless been added as comments in `cibuildwheel/resources/pinned_docker_images.cfg` in order to get the mapping somewhere, similar to what exists in workflow files.
- Test maintenance/complexity: in order to speed-up CI, docker images were pulled once and mutated in order to avoid downloading missing interpreters in every tests and get a cache for the most commonly used wheels. We can't mutate images referenced by digest and thus, to keep this caching behaviour for tests, the fixture now monkey patch the pinned image configuration using the newly introduced `CIBW_INTERNAL_PINNED_DOCKER_IMAGES` environment variable.

Other solution considered: use immutable tags for pypa/manylinux images but, while Quay itself seems to support the feature, quay.io does not offer such setting.


test/conftest.py
```diff
@@ -127,28 +161,37 @@ def docker_warmup(request: pytest.FixtureRequest) -> None:
...
+        fn = root_tmp_dir / "warmup.done"
+        with FileLock(str(fn) + ".lock"):
+            if not fn.is_file():
+                docker_warmup(config)
+                fn.write_text("done")
+
+    with pytest.MonkeyPatch.context() as monkeypatch:
+        for key, value in config.env.items():
+            monkeypatch.setenv(key, value)
+        yield None
```
joerick, line 188: Setting application-level options feels like it risks changing test behaviour. I wonder if a cleaner solution would be to set an (undocumented) option like `CIBW_INTERNAL_PINNED_DOCKER_IMAGES` that referenced a new pinned_docker_images.cfg file produced by the docker warmup fixture. That file would just have pins to the `:cibw_cache_fixture` for each image.

test/conftest.py
```diff
@@ -97,16 +110,35 @@ def docker_warmup(request: pytest.FixtureRequest) -> None:
...
+                    env[f"CIBW_{kind}_{arch_prefix}{arch.upper()}_IMAGE"] = cached_name
+            images.append(DockerWarmUpImage(source_name, cached_name))
+    if images:
+        return DockerWarmUpConfig(images, env)
+    return None
+
+
+def docker_warmup(config: DockerWarmUpConfig) -> None:
     command = (
         "manylinux-interpreters ensure-all &&"
```
joerick, line 128: This is off-topic, but we should also check which interpreters are actually enabled for this test run (`get_enable_groups()`) and only install those. We don't actually test on PyPy and GraalPy on GitHub PRs unless the PR labels are added.

mayeut: If PyPy EOL and GraalPy are not enabled, we should just skip the fixture indeed and I'll add that. I'm not sure we need more complexity (i.e. one-by-one install).
This might relate to the previous comment: we're already setting application-level options. I thought of re-writing the pin file but I'm not sure which is cleaner.

joerick: > we're already setting application-level options 

We do set `CIBW_ENABLE` and `CIBW_PLATFORM`, if that's what you mean? I kinda think of those as test-level options as well as application-level. But none of the CIBW_*_IMAGE options are set before this PR.

Yeah, the change to use a different pin file would only be worth it if we also thought it was cleaner. I kinda think it would be, as it more closely matches the intent - you're changing the default images for the test run, not imagining that the user has configured it differently. Things like test failure logs might make more sense this way.

Not a strongly-held opinion though!

mayeut: > We do set CIBW_ENABLE and CIBW_PLATFORM, if that's what you mean?

indeed.

 I'll add another commit implementing the configuration override.

mayeut: Done, it seems a bit nicer. I also removed caching for musllinux as there's no additional interpreters there.
While testing this, I noticed that the PyPy label did not trigger caching because we're not testing PyPy EOL and there's no label for this one. It's only tested on main. I think we should either add a label or extend the behaviour of the existing one to enable PyPy EOL.

joerick: Yeah, lets make a label. I'll raise a separate PR for that.

joerick (approved):

agriyakhetarpal (approved):

mayeut:
I applied all optional labels to get another CI run using image cache (already tested on my fork while crafting this PR).
`CI: PyPy EoL` won't change a thing until #2930 but `CI: GraalPy` is enough to enable image cache.

PR #2916, opened 2026-06-14
docs: clarify package-dir action input

### LINKED ISSUE #1682: `package-dir` input of the GitHub Action shouldn't imply building only from a Git checkout
### Description

$sbj.
https://github.com/pypa/cibuildwheel/blob/a4f8eb3/action.yml#L5 says it's a dir. But it's actually passed in place where an sdist would work too https://github.com/pypa/cibuildwheel/blob/a4f8eb3/action.yml#L40.

### Build log

_No response_

### CI config

_No response_

Himanshuagrawal4:
Closes #1682

## Summary
- Clarify the GitHub Action `package-dir` input description
- Align the input wording with cibuildwheel's CLI behavior, which accepts a directory or source distribution

## Checks
- `git diff --check`
- Parsed `action.yml` with PyYAML

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -2,7 +2,7 @@ name: cibuildwheel
 description: 'Installs and runs cibuildwheel on the current runner'
 inputs:
   package-dir:
-    description: 'Input directory, defaults to "."'
+    description: 'Path to the package to build, defaults to ".". Can be a directory or an sdist.'
     required: false
     default: .
   output-dir:


joerick (changes requested):
It doesn't make sense to set an arg called 'package-dir' to a file. As noted in the issue, I'd rather rename it to 'package', but while being careful to maintain backward-compat.

PR #2920, opened 2026-06-15
fix: make CIBW_ARCHS parsing case-insensitive and platform-aware

### LINKED ISSUE #2373: Case sensitivity of CIBW_ARCHS for Windows on ARM64
### Description

With:

* `uses: pypa/cibuildwheel@v2.23.2`
* `runs-on: windows-11-arm`
* `CIBW_ARCHS: "arm64"

https://github.com/ultrajson/ultrajson/pull/663/commits/f60b4f665b717ecb6fd8272cad25aa5a7e2ff170

We got an error:

```
cibuildwheel: Invalid archs option {<Architecture.arm64: 'arm64'>}. Windows only supports [<Architecture.AMD64: 'AMD64'>, <Architecture.ARM64: 'ARM64'>, <Architecture.x86: 'x86'>] at the moment. If you want to set emulation architectures on Linux, use CIBW_ARCHS_LINUX instead.
  architectures: arm64
  build_selector: 
    build_config: *
    skip_config: pp39*
    requires_python: >=3.9
    enable: ['cpython-prerelease', 'pypy']
```

Changing it to uppercase fixed it:

```diff
-CIBW_ARCHS: "arm64"
+CIBW_ARCHS: "ARM64"
```

https://github.com/ultrajson/ultrajson/actions/runs/14591266684/job/40927895459?pr=663
https://github.com/ultrajson/ultrajson/actions/runs/14591266684/workflow?pr=663

The docs at https://cibuildwheel.pypa.io/en/stable/options/#archs do indeed show uppercase "ARM64" for Windows and lowercase "arm64" for macOS, but don't mention if they are case-sensitive.

* If they are case sensitive, and it makes sense for them to remain so, please close this.

* If they are case sensitive, but they don't need to be, this is a feature request.

* If they are not meant to be case sensitive, this is a bug report.

I was a little surprised by the case sensitivity, and misread the error as Windows on ARM64 being unavailable for PyPy (which as it happens is true, but not what's being reported here).

Allowing either case will remove a little paper cut and help avoid confusion as others add support for Windows on ARM64.

Thank you!

### Build log

https://github.com/ultrajson/ultrajson/actions/runs/14590225899/job/40925835572?pr=663

### CI config

https://github.com/ultrajson/ultrajson/actions/runs/14590225899/workflow?pr=663

henryiii:

## Problem

`CIBW_ARCHS` architecture parsing was case-sensitive and platform-ambiguous. In `cibuildwheel/architecture.py`, `Architecture.parse_config` used a case-sensitive `StrEnum` lookup (`cls(arch_str)`). The enum carries both a lowercase `arm64` member (value `"arm64"`, used by macOS/Android) and an uppercase `ARM64` member (value `"ARM64"`, used by Windows), plus `AMD64`/`x86_64`, etc.

This caused surprising behavior:

- `Architecture.parse_config("arm64", platform="windows")` returned `{Architecture.arm64}` (the **macOS** member!) with no error, then failed later in `allowed_architectures_check` with a confusing `Invalid archs option` message.
- `Architecture.parse_config("ARM64", platform="windows")` worked.
- `"amd64"` (lowercase) failed to parse entirely, while `"AMD64"` worked.

So Windows-on-ARM users had to use exact-case `ARM64`/`AMD64`.

## Fix

The name lookup is now case-insensitive and resolves to the member valid for the target `platform`. A new helper `Architecture._parse_arch_name` first matches case-insensitively against `Architecture.all_archs(platform)`, then falls back to a case-insensitive match across all members so the existing clear `ConfigurationError` still fires for genuinely invalid names. The special tokens (`auto`, `native`, `all`, `auto64`, `auto32`) are untouched.

After the fix:

- `parse_config("arm64", "windows") == {Architecture.ARM64}`
- `parse_config("ARM64", "macos") == {Architecture.arm64}`
- lowercase `amd64` / `x86` work on Windows
- an actually-invalid name still raises `ConfigurationError`

## Tests

Added parametrized unit tests in `unit_test/architecture_test.py` covering the cases above.

Closes #2373

### DIFF
--- a/cibuildwheel/architecture.py
+++ b/cibuildwheel/architecture.py
@@ -107,12 +107,28 @@ class Architecture(StrEnum):
                 case "auto32":
                     result |= cls.bitness_archs(platform=platform, bitness="32")
                 case _:
-                    try:
-                        result.add(cls(arch_str))
-                    except ValueError as e:
+                    if arch := cls._parse_arch_name(arch_str, platform=platform):
+                        result.add(arch)
+                    else:
                         msg = f"Invalid architecture '{arch_str}'"
-                        raise errors.ConfigurationError(msg) from e
+                        raise errors.ConfigurationError(msg)
         return result
+
+    @classmethod
+    def _parse_arch_name(cls, arch_str: str, platform: PlatformName) -> Self | None:
+        """Resolve an architecture name case-insensitively.
+
+        The same value (e.g. "arm64"/"ARM64") can map to different members
+        depending on the platform, so prefer a member valid for ``platform``
+        before falling back to any case-insensitive match.
+        """
+        for arch in cls.all_archs(platform):
+            if arch.value.lower() == arch_str.lower():
+                return arch
+        for arch in cls:
+            if arch.value.lower() == arch_str.lower():
+                return arch
+        return None
 
     @classmethod
     def native_arch(cls, platform: PlatformName) -> Self | None:
--- a/unit_test/architecture_test.py
+++ b/unit_test/architecture_test.py
@@ -7,6 +7,7 @@ import sys
 import pytest
 
 import cibuildwheel.architecture
+import cibuildwheel.errors
 from cibuildwheel.architecture import Architecture, arch_synonym
 
 TYPE_CHECKING = False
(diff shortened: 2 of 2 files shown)


joerick (approved):
Nice, yeah I imagine this has been a little paper cut for a some people.

agriyakhetarpal (approved):
Thanks! Makes sense to me. Personally, this never hit me because we documented AMD64/ARM64 in capital letters in the docs well, and I used them as such, never questioning why – but I recognise that others have been hitting it.

PR #2922, opened 2026-06-15
fix(macos): make matching python3-config available in build/test venvs

### LINKED ISSUE #2021: macOS: venvs lacking `python3-config` causes some builds to fail
### Description

Some more complex C/C++ libraries need `python3-config` to get ldflags, include directories, etc.

On Linux, because each version of Python is in its own container, this works fine.

On macOS, because `cibuildwheel` relies entirely on virtualenvs and does not put `/Library/Frameworks/Python.framework/Versions/{current_version}/bin` in PATH, `python3-config` defaults to the system-wide Python 3.11/3.12 installation. Meaning any non-trivial C/C++ extensions may fail (and not even outright fail to build- wheels produced may fail to import because of missing symbols due to ABI breaks.)

A solution to this would be to add `/Library/Frameworks/Python.framework/Versions/{current_version}/bin` to `PATH`.

A possible workaround is adding this in your `CIBW_BEFORE_BUILD`:

```sh
ln -s $(dirname $(readlink -f $(which python3)))/python3-config $(dirname $(which python3))/python3-config
```

### Build log

https://github.com/donn/cibw_clobber/actions/runs/11091306625/job/30814862405

### CI config

https://github.com/donn/cibw_clobber/blob/main/.github/workflows/ci.yml

henryiii:

## Problem

On macOS, cibuildwheel builds and tests inside a virtualenv and only adds the venv's `bin` directory to `PATH`. Virtualenvs don't contain `python3-config`, so builds that invoke it (for include dirs / link flags) pick up an unrelated one from elsewhere on `PATH` — typically the system or Homebrew Python — producing broken wheels or ABI mismatches.

## Fix

As suggested in review, the base interpreter's `bin` directory is added to `PATH` right after the venv's (POSIX only), for both build and test venvs. The matching `python3-config`/`python3.X-config` is then found first, and the same holds for anything else shipped next to the base interpreter (e.g. `pypy3-config`, `graalpy3-config`). The venv itself is not modified, and its `bin` still comes first, so `python`/`pip` continue to resolve to the venv.

Callers that need stricter isolation are unaffected: iOS rewrites `PATH` wholesale after venv creation, and Linux builds don't use this code path.

Closes #2021

### DIFF
--- a/cibuildwheel/venv.py
+++ b/cibuildwheel/venv.py
@@ -210,18 +210,8 @@ def virtualenv(
 def _symlink_python_config_scripts(base_python: Path, venv_bin: Path) -> None:
     """
     Symlink the base interpreter's ``python*-config`` scripts into the venv's
-    bin directory.
-
-    Virtualenvs do not ship a ``python3-config`` (or versioned
-    ``python{X.Y}-config``) script, so extensions that call it (e.g. to get
-    include dirs or link flags) would otherwise pick up an unrelated
-    ``python3-config`` from elsewhere on PATH -- typically the system or
-    Homebrew Python. This is most relevant for the python.org framework builds
-    on macOS, which do ship these scripts next to the base interpreter.
-
-    Interpreters that don't ship a config script (e.g. PyPy, GraalPy, or some
-    uv-managed builds) are handled gracefully: nothing is linked and the build
-    is not failed.
+    bin directory if provided and not already linked (virtualenvs don't always
+    provide them).
     """
     # The config scripts live next to the base interpreter, e.g.
     # `python3-config` and `python3.12-config`.
--- a/cibuildwheel/venv.py
+++ b/cibuildwheel/venv.py
@@ -152,6 +152,10 @@ def virtualenv(
     venv. Otherwise, pip is installed.
     """
 
+    # the unresolved path, so e.g. the python.org framework bin directory is
+    # used rather than the internal location of the real binary
+    base_python_bin_dir = None if _IS_WIN else python.parent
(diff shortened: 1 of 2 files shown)


mayeut:
if we aim to fix this, should a symlink be added or do we want to add the base python bin folder to the PATH instead (right after the venv one) ?
IMHO, adding to PATH seems more natural as it's likely what would be happening on a dev local setup rather than tweaking the virtual environment.

agriyakhetarpal:
If pypy3-config and graalpy3-config do exist (based on the AI description at least) I think we should similarly match for them and include them for consistency, otherwise it just becomes yet another case where our builds across interpreters are different from each other (and perhaps something a later AI review will flag).

joerick:
> should a symlink be added or do we want to add the base python bin folder to the PATH instead

Nice idea! It feels a bit less hacky than modding the venv IMO.

henryiii:

Switched to the PATH approach suggested above: `activate_virtualenv` now places the base interpreter's `bin` directory right after the venv's on `PATH`, instead of symlinking config scripts into the venv. Since the whole directory is exposed, this also covers `pypy3-config`/`graalpy3-config` (@agriyakhetarpal's consistency point) without special-casing. Windows is skipped. Tests rewritten accordingly, and the PR description is updated.

henryiii:
Could I get a review? Does a thumbs up count? :)

agriyakhetarpal (approved):
Sorry about holding off on this; this LGTM – it's just been conference season and I haven't been around lately :)

PR #2925, opened 2026-06-19
Validate `pyodide-version` option against Pyodide build identifier

### LINKED ISSUE #2924: Global pyodide-version is applied to mismatched Python targets without a clear error
### Description

When a Pyodide build targets multiple Pyodide identifiers and the `pyodide-version` option (from #2002) is set globally, we silently force every target onto that single xbuildenv version instead of catching the conflict. The build then fails further down with a confusing message that points at the wrong thing:

```
Error: cibuildwheel: The Pyodide xbuildenv version 314.0.0 is not compatible with the pyodide-build version 0.35.0. Please use the 'pyodide xbuildenv search --all' command to find a compatible version.
```

which is not exactly true if read on its own.

Each Pyodide identifier has its own pinned default in `build-platforms.toml`:

```toml
{ identifier = "cp313-pyodide_wasm32", version = "3.13", default_pyodide_version = "0.29.4",  ... }
{ identifier = "cp314-pyodide_wasm32", version = "3.14", default_pyodide_version = "314.0.0", ... }
```

In `cibuildwheel/platforms/pyodide.py`, `setup_python()` resolves the version like this:

```python
pyodide_version = user_pyodide_version or python_configuration.default_pyodide_version
```

So if we set `pyodide-version = "314.0.0"` at the `[tool.cibuildwheel.pyodide]` level, it overrides the per-identifier default for all targets. The `cp313-pyodide_wasm32` build then tries to use the `314.0.0` xbuildenv, which is a Python 3.14 environment, against a Python 3.13 base. There is no check that the resolved version's Python matches the identifier's Python, so the override is applied unconditionally.

The failure surfaces later as either:

- `The Pyodide xbuildenv version 314.0.0 is not compatible with the pyodide-build version 0.35.0. Please use the 'pyodide xbuildenv search --all' command to find a compatible version.` (from `validate_pyodide_build_version`), (as I noted above)
- or, `Expected host Python version to be 3.14 but got version 3.13` during venv setup.

Neither message makes it obvious that the cause is a global `pyodide-version` being applied to a Python version it does not match. The first one is especially misleading, since `pyodide xbuildenv search --all` actually reports `314.0.0` as compatible with `pyodide-build` 0.35.0. The version is fine. It is just being paired with the wrong target.

## To reproduce

We came across this in https://github.com/hynek/argon2-cffi-bindings/pull/129:

```toml
[tool.cibuildwheel.pyodide]
build = ["cp313-pyodide_wasm32", "cp314-pyodide_wasm32"]
pyodide-version = "314.0.0"
```

```
CIBW_PLATFORM=pyodide cibuildwheel
```

Removing the `pyodide-version` line fixes it, since each identifier then falls back to its own compatible defaults (cp313 goes to `0.29.4`, and cp314 to `314.0.0`) and both wheels build.

I think we should raise a `ConfigurationError` when the resolved `pyodide-version` for a given identifier targets a different Python version than that identifier expects. The check can compare the requested xbuildenv's Python version against `python_configuration.version` before installing anything, and fail early with a message that names both the identifier and the mismatched version.

The build environment is as follows:

- cibuildwheel: 4.1.0 (but also reproduces on 4.0.0, where the downstream message differs)
- Platform: pyodide

### Build log

https://github.com/hynek/argon2-cffi-bindings/actions/runs/27815062922/job/82314145503?pr=129

### CI config

https://github.com/hynek/argon2-cffi-bindings/pull/129

agriyakhetarpal:
Closes #2924

Up to now, a global `pyodide-version` option, if set, overrides the per-identifier default for a Pyodide build target, without verifying that the resolved version's Python matches that of the identifier. We then observe the failure surfacing much later, accompanied by a misleading message about pyodide-build's incompatibility with the Pyodide xbuildenv (see details in the issue).

I have added a method which compares the resolved xbuildenv's Python field against `python_configuration.version` and raises a `ConfigurationError`. For example, we raise if `pyodide-version` is set to 314.0.0 and the build identifier at the time is cp313-pyodide_wasm32.

In that sense, it makes sense for the user to remove one of these options. However, a scenario where both are set and are compatible, say, pyodide-version being set to 0.29.4 and the build identifier is cp313-pyodide_wasm32, we won't mind it (but please let me know if you'd like me to change that).

The check occurs after `get_xbuildenv_version_info()`, which requires pyodide-build to be installed, rather than solely from the static config. That's the earliest point we can grab the xbuildenv's Python version, and it's still before any xbuildenv/Emscripten installation, at least.

### DIFF
--- a/cibuildwheel/platforms/pyodide.py
+++ b/cibuildwheel/platforms/pyodide.py
@@ -218,6 +218,37 @@ def validate_pyodide_build_version(
         raise errors.FatalError(msg)
 
 
+def validate_pyodide_target_python(
+    xbuildenv_info: PyodideXBuildEnvInfo, python_configuration: PythonConfiguration
+) -> None:
+    """
+    Validate that the resolved Pyodide xbuildenv targets the same Python version
+    as the identifier being built.
+
+    This is to catch the case where a global ``pyodide-version`` is applied to an
+    identifier whose Python version does not match, such as setting
+    ``pyodide-version = "3Y.0.0"`` (a Python 3.Y environment) while
+    building ``cp3X-pyodide_wasm32``.
+    """
+    xbuildenv_python = xbuildenv_info["python"]
+    xbuildenv_version = xbuildenv_info["version"]
+    identifier = python_configuration.identifier
+    expected_version = python_configuration.version
+    default_version = python_configuration.default_pyodide_version
+    if not xbuildenv_python.startswith(f"{expected_version}."):
+        xbuildenv_python_minor = ".".join(xbuildenv_python.split(".")[:2])
+        msg = unwrap_preserving_paragraphs(f"""
+            The `pyodide-version` option is set to {xbuildenv_version}, which
+            provides Python {xbuildenv_python_minor}, but the {identifier} build
+            needs Python {expected_version}.
+
+            Either remove `pyodide-version` so {identifier} uses its default
+            ({default_version}), or stop building {identifier} so that
+            `pyodide-version` only applies to an identifier it matches.
+        """)
+        raise errors.ConfigurationError(msg)
+
+
 def install_xbuildenv(env: dict[str, str], xbuildenv_cache_path: Path, pyodide_version: str) -> str:
     """Install a particular Pyodide xbuildenv version and set a path to the Pyodide root."""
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
Thanks!

henryiii (approved):

joerick (approved):

PR #2926, opened 2026-06-19
docs: replace outdated note about build frontend and remove mention of ClearLinux

agriyakhetarpal:
- Remove a reference to `pip wheel`, as our build frontend has been pypa/build since v2.x I believe, or at least for a long time anyway. I made the sentence agnostic to the build frontend.
- Drop a mention of ClearLinux, as it is now discontinued and fully abandoned since July 2025: https://endoflife.date/clear-linux

Noticed when reading through https://github.com/scikit-learn/scikit-learn-release/issues/1

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -807,7 +807,7 @@ instead.
 ### `before-build` {: #before-build env-var toml}
 > Execute a shell command preparing each wheel's build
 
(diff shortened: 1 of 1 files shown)


henryiii:
There are wheels for Alpine, though? Should we put FreeBSD or something like that instead?

agriyakhetarpal:
Yes, I think this section is prehistoric and may have been written at a time when MUSL wheels didn't exist on PyPI. Let me change it.

henryiii (approved):

PR #2929, opened 2026-06-26
Use a custom Python shell for our action's cibuildwheel setup

agriyakhetarpal:
Just a nicer way of expressing things instead of a Bash Heredoc. This uses a [Python](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-running-an-inline-python-script) [custom shell](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#custom-shell). I validated this with the GitHub Actions schema locally, and it looks like it's a supported pattern:

- https://github.com/search?q=%22shell:+$%7B%7B+steps%22+language:YAML&type=code
- https://github.com/search?q=%22shell:+python+-u%22+language:YAML&type=code&l=YAML

Let's see if our own tests that use the action pass or not!

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -37,7 +37,6 @@ runs:
     - id: cibw
       run: |
         # Install cibuildwheel and build the command line
-        "$PYTHON" -u << "EOF"
         import os
         import shlex
         import shutil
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -116,10 +115,8 @@ runs:
             f.write(f"cmd-pwsh={cmd_pwsh}\n")
 
         print("::endgroup::")
-        EOF
-      shell: bash
+      shell: ${{ steps.python.outputs.python-path }} -u {0}
```
mayeut, line 118: If for some reason, there's a space in `steps.python.outputs.python-path`, this won't work.
This is definitely not an issue for GitHub-Hosted runners. It could be an issue on Self-Hosted runners, although unlikely.

agriyakhetarpal: Right... I don't think there is any elegant way to do this, unfortunately – I'll close this PR.

PR #2930, opened 2026-06-26
Implement test enabling with the "CI: PyPy EoL" PR label

joerick:
See https://github.com/pypa/cibuildwheel/pull/2915#discussion_r3464858751 for context.

🤖 PyPy EoL builds were only running on main with no way to trigger them on PRs via a label. This adds support for the CI: PyPy EoL label, matching the existing pattern for CI: PyPy and CI: GraalPy.


agriyakhetarpal:
Perhaps we should add the "CI: PyPy EoL" label here and see if it works (hoping nothing is broken), or would you like to do that exercise in another PR?

mayeut:
> Perhaps we should add the "CI: PyPy EoL" label here and see if it works (hoping nothing is broken)

I applied it as it does make sense to test it here first.

I think we should also apply the label in `.github/workflows/update-dependencies.yml`

mayeut:
I rebased/force pushed this PR in order to get a green status and added a [commit to apply "CI: PyPy EoL" to "Update dependencies" PR](https://github.com/pypa/cibuildwheel/pull/2930/commits/0b2265b38a8f46310cca62a03d4cbe2a7ed9b26b)

mayeut (approved):

agriyakhetarpal (approved):

PR #2934, opened 2026-06-28
Use absolute path for `{project}` placeholder in `config-settings`

agriyakhetarpal:
While the `project="."` pattern is consistent across all platforms and all our build stages, it's specifically incorrect for `config-settings`, because that value is passed to `python -m build` and consumed by the build backend, which may run from a different CWD. It's better to pass the absolute path to the project root in `prepare_config_settings` in this case. This only failed on Windows and macOS, as `{project}` on Linux was always absolute (we set it as `/project` via the container there).

I think we should add a test or something for this, but I'm not sure what would be best. Should I add a project that cross-compiles a wheel using a Meson cross file (like I use this functionality)? Or perhaps `test/test_projects/meson.py` can stay as is, but I add a `--native-file` since it's resolved the same way a cross file is?

xref #2827

### DIFF
--- a/cibuildwheel/platforms/android.py
+++ b/cibuildwheel/platforms/android.py
@@ -310,7 +310,9 @@ def setup_env(
         "wheel",
         parse_config_settings(
             prepare_config_settings(
-                build_options.config_settings, project=".", package=build_options.package_dir
+                build_options.config_settings,
+                project=Path.cwd(),
+                package=build_options.package_dir,
             )
         ),
     )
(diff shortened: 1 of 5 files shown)


freakboy3742 (commented):
From an iOS perspective, I don't see any problem with this; explicit full paths are generally safer, as long as we're not dealing with something that will be running on the simulator (which will only be for testing paths).

mhsmith (commented):
The Android changes look good to me.

henryiii (approved):

PR #2945, opened 2026-07-22
fix: don't error when a global build-frontend is set on pyodide

henryiii:
This would cause many valid configuration files to break. Making this just a warning for now, maybe pip/uv could become an error in a future version. Maybe build / build[uv] could always just tap over with a warning.

I'd expect this will only matter if we introduce a new xbuild option eventually.

Currently blocking a patch release.

Since #2609, the pyodide platform requires the `pyodide-build` build frontend. But a global `build-frontend` setting (in TOML or `CIBW_BUILD_FRONTEND`) overrides the platform *default* in the option cascade, so previously-working configs like a global `build-frontend = "build"` now fail with a `ConfigurationError` unless a pyodide-specific override is added. Even `build-frontend = "default"` failed, since `"default"` was hard-mapped to `"build"` before the check.

This resolves `"default"` to the platform default (`pyodide-build` on pyodide), and turns the mismatch error into a warning that uses `pyodide-build` instead (keeping any frontend `args`). That matches pre-#2609 behavior, where pyodide always ran `pyodide build` and the frontend name only affected verbosity flags.

Regression tests were added first and confirmed to fail without the fix.

### DIFF
--- a/cibuildwheel/options.py
+++ b/cibuildwheel/options.py
@@ -43,7 +43,7 @@ from packaging.specifiers import SpecifierSet
 from cibuildwheel import errors
 from cibuildwheel.architecture import Architecture
 from cibuildwheel.environment import EnvironmentParseError, ParsedEnvironment, parse_environment
-from cibuildwheel.frontend import BuildFrontendConfig
+from cibuildwheel.frontend import BuildFrontendConfig, BuildFrontendName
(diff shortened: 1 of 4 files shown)


agriyakhetarpal (approved):
Thanks @henryiii, makes sense and sorry for the trouble! Glad we caught this before a release.

PR #2947, opened 2026-07-24
docs: add missing cp314-pyodide_wasm32 to build-id table

henryiii:
I noticed this when reviewing #2946. Added a test to make sure this stays up to date in the future.

The build-id table in `docs/options.md` was missing `cp314-pyodide_wasm32`. That table is hand-maintained, so an identifier can be dropped whenever `build-platforms.toml` changes.

This adds the missing entry and a unit test that asserts every identifier from `build-platforms.toml` appears in the table, so CI catches a future omission and points at the exact identifier.


joerick (approved):
love the use of a unit test to check this!

PR #2951, opened 2026-07-27
fix(schema): default is fine for pyodide too

henryiii:
The schema claims pyodide-build is required, but default is fine too.

Also adding `.claude`, so users can symlink `.claude/skills` to `.agent/skills`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,4 @@ site/
 
 # This file should be a symlink or contain "See @AGENTS.md"
 CLAUDE.md
+.claude/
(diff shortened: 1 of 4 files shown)


agriyakhetarpal (approved):

henryiii:
This should work too, but doesn't:

```toml
select = ["*pyodide*"]
build-frontend = "pyodide-build"
```

henryiii:
Okay, should be more accurate now. "pyodide-build" is accepted at top level, in overrides, and in the "pyodide" platform, but not other platforms. Kept it fairly clean, I think.

Also fixes a dangling description that Claude noticed while working on it.

mayeut (approved):

PR #2953, opened 2026-07-27
fix: survive short network outages when downloading

henryiii:

A CI runner lost DNS for approximately a minute during a test run, which failed two macOS Intel jobs. Neither layer of retry was long enough to survive it.

- `download()` waited a flat 3 seconds over 3 attempts, so it covered only about 6 seconds of downtime. It now waits 3, 6, 12, 24, then 48 seconds over 6 attempts. This helps users on unreliable networks, not only CI. The PyPy download was the one that failed.
- A 4xx response now fails immediately, because a bad URL will not fix itself. Without this, the longer backoff would delay the report by about 93 seconds. A 5xx response is still retried.
- uv stops after 3 retries. `UV_HTTP_RETRIES` is now set to 6 for the test workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,10 @@ concurrency:
   group: test-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  # runners see short DNS/network outages; uv's default of 3 retries is too few
+  UV_HTTP_RETRIES: "6"
+
 jobs:
   lint:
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):
I hope that these issues don't persist on the runners' side, but having our own exponential backoff makes sense to me, especially since it makes our and our users' lives easier.

PR #2583, opened 2025-09-08
Don't constrain `build-system.requires` with our `dependency-versions`

joerick:
Fix #2570.

Don't constrain the versions installed during the package build using `dependency-versions`. See #2570 for full discussion. AI summary of the change below

---

This change stops cibuildwheel's dependency-versions from being applied as constraints within the isolated build environment used for the package build. This resolves an issue where conflicts between cibuildwheel's internal requirements and a project's build-system.requires would cause cryptic build failures.

**The Problem**
Currently, on macOS, Windows, and Pyodide platforms, cibuildwheel injects its own dependency constraints (from CIBW_DEPENDENCY_VERSIONS or the defaults) into the isolated environment where a project's build-time dependencies are installed.

This leads to an issue: if a project specifies a build requirement in pyproject.toml (e.g., conan) that has a dependency conflict with cibuildwheel's constraints (e.g., on urllib3), the build fails with an unhelpful subprocess.CalledProcessError traceback.

This behaviour is also inconsistent, as it does not occur on Linux, which invokes the build without injecting these internal constraints.

**The Solution**
Remove the logic that sets the PIP_CONSTRAINT and UV_CONSTRAINT environment variables on the build_env before invoking the build frontend (`pip wheel` or `build`).

The effect is:

- The user's project's build-time dependencies are now resolved independently, without interference from cibuildwheel's toolchain constraints.

- The build process behaviour is now the same across all platforms (Linux, macOS, Windows, and Pyodide)

PR #2957, opened 2026-08-02
fix(tests): delete test_overridden_pip_constraint

mayeut:
issue seen in #2952

given we no longer mess with `PIP_CONSTRAINT` since #2583, should we keep this test ?


henryiii:
I'd be okay to drop it.

mayeut:
> I'd be okay to drop it.

I added a commit dropping the test altogether

agriyakhetarpal (approved):

PR #2958, opened 2026-08-02
Add pyodide_315 support

hoodmane:
Guarded by cpython_prerelease. I manually tested building pywavelets and it works as expected.

- [x] Fix cross build environment installation not to try to install native scipy


henryiii:
Quick background:

We went with `cpython_prerelease` because this is supposed to be ABI stable from pyodide's perspective for the new pyemscripten release. We could drop `pyodide_prerelease` if there's not really a reason for it anymore.

We want to get this in before the 3.15rc1 release because numpy won't add pyemcripten as a platform unless the can add it for both CPython 3.15 and pyemscripten 3.15 in the same numpy release. (Seems like a rather odd requirement, and patch releases have added things before, but getting pyodide in sync with CPython is nice anyway).

agriyakhetarpal (dismissed):
Drive-by comment for now: I'd definitely want `pyodide-prerelease` group to stay, instead of Pyodide releases being enabled with `cpython-prerelease`; that is a clearer separation of concerns

hoodmane:
@agriyakhetarpal I don't see why pyodide_prerelease should stay, the goal is to have it at the same time as the cpython prerelease.

agriyakhetarpal:
@hoodmane it conflates CPython and Pyodide. If someone enables cpython-prerelease with the changes here and also has a Pyodide build workflow, I'd lean toward not starting to build pyemscripten alpha ABI wheels. There's a reason we have Pyodide as a separate platform, after all, rather than mixing it with GraalPy.

hoodmane:
Pyodide is CPython so it's correct to conflate them.

hoodmane:
I think if someone is building both Pyodide releases and cpython prereleases it's desirable that that also builds Pyodide for prerelease Python versions. I don't see how that would be a bad thing. If they don't want it they can explicitly disable pre-release Pyodide. I think people won't remember to explicitly enable pre-release Pyodide so it's much better as opt-out than opt-in.

ryanking13:
I am in favor of `cpython_prerelease` change with two reasons:

1) We are aiming to release Pyodide prerelease at the same time of CPython rc from this release
2) Pyodide cibuildwheel should be configured separately from other platforms (at least at this point), so I don't think people will accidentally enable Pyodide builds when they do not want.

unit_test/build_selector_test.py
```diff
@@ -91,15 +91,23 @@ def test_build_filter_pypy_all() -> None:
...
+    )
+    assert build_selector("cp314-pyodide_wasm32")
+    assert build_selector("cp315-pyodide_wasm32")
+
+    pyodide_prerelease_only = BuildSelector(
+        build_config="*",
+        skip_config="",
+        enable=frozenset([EnableGroup.PyodidePrerelease]),
+    )
+    assert not pyodide_prerelease_only("cp315-pyodide_wasm32")
```
ryanking13: Not sure if we need this test

docs/options.md
```diff
@@ -357,13 +357,16 @@ values are:
...
+  beta CPython, which is the case for the current Pyodide prerelease.
 - `pypy`: Enable PyPy.
 - `pypy-eol`: Enable PyPy versions that have passed end of life (if still available).
 - `graalpy`: Enable GraalPy.
 - `pyodide-prerelease`: Pyodide versions that haven't released yet, if one is
   available. Safe if you are shipping a site with an early build, not for
-  general distribution.
+  general distribution. Note that when a Pyodide prerelease targets a beta
+  CPython, it is gated by `cpython-prerelease` instead, so this group may have
+  no effect.
```
ryanking13: Maybe we should just remove this?

docs/_internal/pyodide-maintenance.md
```diff
@@ -33,17 +33,29 @@ In `cibuildwheel/resources/build-platforms.toml`, add an entry under `[pyodide]`
 
 `version` is the CPython version string, `default_pyodide_version` is the Pyodide release to use when the user does not pin one explicitly (use the latest available alpha/beta for a prerelease entry), and `node_version` is the minimum Node.js major required by that Pyodide release — check the [pyodide-build FAQ](https://pyodide-build.readthedocs.io/en/latest/faq.html#what-node-js-version-do-i-need) for a rudimentary idea of what the correct value is. `sha256` is the checksum of the Pyodide xbuildenv tarball.
 
-### 2. Update the prerelease guards in the selector
+### 2. Check the prerelease guards in the selector
```
ryanking13: This section needs to be updated or removed.

cibuildwheel/selector.py
```diff
@@ -104,17 +107,18 @@ def __call__(self, build_id: str) -> bool:
             return False
         if EnableGroup.PyodideEoL not in self.enable and fnmatch(build_id, "cp312-pyodide_*"):
             return False
-        # NOTE: Re-enable this when we have a new Pyodide prerelease (e.g., 315.0.0a1+)
-        # When doing this, also:
+        # NOTE: The `pyodide-prerelease` guard is currently inert. The Pyodide
```
ryanking13: These codes/comments should be removed

ryanking13 (approved):

docs/_internal/pyodide-maintenance.md
```diff
@@ -11,17 +11,26 @@ This page describes how to update cibuildwheel's Pyodide platform code when eith
 
 Pyodide has three types of releases that matter to cibuildwheel:
 
-- **Stable** – the most recent full Pyodide release (e.g., `0.29.x` / cp313). This is enabled by default with no special `CIBW_ENABLE` flag needed.
-- **Prerelease** – an alpha/beta/rc Pyodide release that uses the _next_ CPython version (e.g., `314.0.0a1` / cp314). Users must opt in with `CIBW_ENABLE: pyodide-prerelease` to build against this version. This may or may not be available at any given time, depending on the Pyodide release cycle.
-- **End-of-life (EoL)** – older Pyodide stable releases that are no longer the current stable. These are kept available behind `CIBW_ENABLE: pyodide-eol` so that users who still need to build for older Pyodide versions can do so.
+- **Stable** – the most recent full Pyodide release (e.g., `0.29.x` / cp313).
```
ryanking13:
```suggestion
- **Stable** – the most recent full Pyodide release (e.g., `314.0.x` / cp314).
```

docs/_internal/pyodide-maintenance.md
```diff
@@ -11,17 +11,26 @@ This page describes how to update cibuildwheel's Pyodide platform code when eith
...
+  This is enabled by default with no special `CIBW_ENABLE` flag needed.
+- **Prerelease** – an alpha/beta/rc Pyodide release that uses the _next_ CPython
+  version (e.g., `314.0.0a1` / cp314). Users must opt in with `CIBW_ENABLE:
```
ryanking13:
```suggestion
  version (e.g., `315.0.0a1` / cp315). Users must opt in with `CIBW_ENABLE:
```

joerick (approved):
No test regressions with the alpha is very encouraging :) I'm okay with using cpython-prerelease for this, as long as we're expecting the timelines between pyodide and cpython to be the same.

Issue #2950, opened 2026-07-26
Ubuntu 26.04 test failures


mayeut:
### Description

Ubuntu 26.04 is now in Preview in GitHub Actions.
Here's a short summary of failures that might impact workflows once the `ubuntu-latest` tag moves to `ubuntu-26.04`:
1. test_podman_vfs
2. ~~android emulator does not work~~

For 1, I have no clue how to fix this

EDIT: For 2., Android Java 25 compatibility should be fixed by https://github.com/pypa/cibuildwheel/pull/2962.

mhsmith:
I've made the Android testbed compatible with Java 25 in https://github.com/python/cpython/pull/155012, which should hopefully be included in next week's Python releases.

mhsmith:
Android Java 25 compatibility should be fixed by #2962.

mayeut:
I can confirm Android Java 25 compatibility is fixed.

PR #2962, opened 2026-08-08
Update Android Python versions

mhsmith:
This should fix the CI issue discussed at https://github.com/pypa/cibuildwheel/pull/2933#issuecomment-4821042715.

These Python versions have also fixed the content of their .pc files, allowing a workaround to be removed.

And they update the Android testbed Gradle version, making it compatible with Java 25 (#2950).


mayeut (approved):
Just a note: the CI issue discussed at https://github.com/pypa/cibuildwheel/pull/2933#issuecomment-4821042715 did not occur with the update to 3.15.0rc1 for some reason (I can confidently say it was not disk size related so probably has something to do with the testbed update).

PR #2966, opened 2026-08-12
Respect configured NuGet package sources

### LINKED ISSUE #2961: Allow skipping of -FallbackSource in cpython NuGet installs
### Description

Hey folks,

I'm from the Azure Artifacts team and we're looking at doing some additional network isolation for our build environments used to create the [artifacts-keyring](https://pypi.org/project/artifacts-keyring) package. We're able to nearly achieve this using Azure DevOps Artifacts feeds and ACR with `CIBW_MANYLINUX_*_IMAGE` variables, but the cpython installs force a fetch to `api.nuget.org` due to [def get_nuget_args](https://github.com/pypa/cibuildwheel/blob/1828c10ab37f080699c7b81cea34097c684a7074/cibuildwheel/platforms/windows.py#L67) in `cibuildwheel/platforms/windows.py`. This is due to an index fetch for any primary sources in NuGet's [InitializeResourcesAsync](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.PackageManagement/Resolution/ResolverGather.cs#L588).

Would it be possible to have this hard-coded FallbackSource parameter removed with some environment variable or config option? I'd be happy to contribute a change here if there's some consensus on the best approach 😄 

### Build log

https://dev.azure.com/artifacts-public/PublicBuilds/_build/results?buildId=265&view=logs&j=b6d58243-b3ca-5571-e7ac-13f842808d7c&t=2d85dffa-bace-5cf0-464d-9b6623c2d7ca&l=88

### CI config

https://github.com/microsoft/artifacts-keyring/blob/master/pipelines/build-wheels.yml

CAOShurong:
Closes #2961.

## Summary

- Remove the hard-coded nuget.org `-FallbackSource` from Windows CPython installs.
- Let NuGet honor the package sources selected by its configuration.
- Add regression coverage for regular AMD64 and free-threaded ARM64 package names.

## Testing

- `pytest unit_test/get_platform_test.py` (7 passed)
- `PYTHONUTF8=1 nox -s tests -- unit_test` (854 passed, 68 skipped, 3 xfailed)
- `pre-commit run --files cibuildwheel/platforms/windows.py unit_test/get_platform_test.py`
- `pylint cibuildwheel/platforms/windows.py`
- Exercised `install_cpython` with an empty local-only NuGet source: it reported only that source and failed as expected.
- Exercised `install_cpython` with nuget.org explicitly configured: it installed CPython 3.14.1 successfully.


unit_test/get_platform_test.py
```diff
@@ -30,6 +34,22 @@ def patched_environment(
         yield
 
 
+@pytest.mark.parametrize(
+    ("arch", "free_threaded", "package_name"),
+    [("AMD64", False, "python"), ("ARM64", True, "pythonarm64-freethreaded")],
+)
+def test_nuget_args_use_configured_sources(
```
joerick:
```suggestion
def test_get_nuget_args(
```

joerick (commented):
CI is passing so I'm happy

unit_test/get_platform_test.py
CAOShurong: Applied in 51c5d3fc; the focused Windows tests still pass. Thanks!

joerick (approved):

PR #2967, opened 2026-08-12
[inherit-extend] refactor: validate inherit rules eagerly, more standard naming

henryiii:
This is based on an AI review of #2928. @joerick, I can trim out any changes you don't like. The simplification pass was based on comparisons with main, so some parts of the diff simplify the total change more than the diff in this PR itself.

* `CIBW_INHERIT_<PLATFORM>` does seem to be more natural, do you like it better?

The failing CI is because the target branch needs rebasing.

Follow-ups from a review of #2928, targeting `inherit-extend`.

**Validation.** `parse_inherit` now validates rule keys against the known option names (with "Perhaps you meant ...?" suggestions), requires a single token per rule, and converts shlex `ValueError` / non-string TOML values into `OptionsReaderError`. Overrides and `CIBW_INHERIT` are parsed eagerly in `OptionsReader.__init__` again, so malformed config fails fast even on `--print-build-identifiers`. An append/prepend rule on a non-mergeable option (e.g. `container-engine`) now raises a clean error instead of a bare `OptionFormat.NotSupported` traceback, and an inherit rule on `xbuild-tools` no longer leaks the `"\u0000"` sentinel default into the tool list.

**Interface change (worth a look).** Platform-scoped env rules now use `CIBW_INHERIT_<PLATFORM>` (matching the `CIBW_<OPTION>_<PLATFORM>` convention) instead of a `-<platform>` suffix inside `CIBW_INHERIT` keys — with key validation, the suffix form would have collided with option-name checking, and `CIBW_INHERIT_LINUX` is what users will try first anyway. `CIBW_INHERIT` rules also apply to the platform variables unless a platform rule overrides them. Docs updated.

**Schema.** `generate_schema.py` derives the `inherit` keys from the option list instead of a hand-copied block (it was missing `enable`, `archs`, etc.), and each platform/overrides section gets an inherit table restricted to the options that section can set.

**Cleanup.** `parse_arbitrary_key_value_string` is now a thin wrapper over `parse_key_value_string` rather than a near-copy of it.

Regression tests were added for each fix and confirmed to fail beforehand.

One open question: the runtime also accepts the string form (`inherit = "before-all: append"`) in TOML, but the docs and schema only describe the table form there. Left as-is; happy to restrict or document it either way.

### DIFF
--- a/bin/generate_schema.py
+++ b/bin/generate_schema.py
@@ -336,10 +336,14 @@ def inherit_ref_table(option_names: dict[str, Any]) -> dict[str, Any]:
 
 
 # any option can carry an inherit rule; derive the keys from the option list
-# so the schema stays in sync with the runtime
+# so the schema stays in sync with the runtime (must run before the overrides
+# and platform sections are merged into the properties below)
 schema["properties"]["inherit"]["properties"] = inherit_ref_table(schema["properties"])
 
 non_global_options = {k: {"$ref": f"#/properties/{k}"} for k in schema["properties"]}
+# placeholder pinning the key's position; each section's real (narrower)
+# inherit table is filled in below, once the section's option set is final
+non_global_options["inherit"] = {}
 del non_global_options["build"]
 del non_global_options["skip"]
 del non_global_options["test-skip"]
--- a/bin/generate_schema.py
+++ b/bin/generate_schema.py
@@ -43,25 +43,8 @@ properties:
   inherit:
     type: object
     additionalProperties: false
-    properties:
-      audit-command: {"$ref": "#/$defs/inherit"}
-      audit-requires: {"$ref": "#/$defs/inherit"}
-      before-all: {"$ref": "#/$defs/inherit"}
-      before-build: {"$ref": "#/$defs/inherit"}
-      xbuild-tools: {"$ref": "#/$defs/inherit"}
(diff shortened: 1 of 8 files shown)


joerick:
> * `CIBW_INHERIT_<PLATFORM>` does seem to be more natural, do you like it better?

I didn't go with that as to me it implied that this is the 'CIBW_INHERIT' rule for one platform, not that it applied to the platform-specific variables. As in,

```
CIBW_TEST_COMMAND: echo testing
CIBW_TEST_COMMAND_MACOS: echo testing macos
CIBW_INHERIT: before-all
CIBW_INHERIT_MACOS: test-command
```

Here it's kinda ambiguous if we're saying that the inherit test-command rule only happens on macos, or if the inherit applies to the CIBW_TEST_COMMAND_MACOS setting. Also, it raises the question - how do the inherit rules inherit?!

The alternative I used was to say 

```
CIBW_TEST_COMMAND: echo testing
CIBW_TEST_COMMAND_MACOS: echo testing macos
CIBW_INHERIT: before-all test-command-macos
```

which I think is clearer.

joerick:
i'll look at the rule validation concerns though!

joerick:
i've addressed the review in the main PR #2928

PR #2968, opened 2026-08-13
chore: faster and stricter mypy

henryiii:
I'm not sure it's possible to reuse a wheel in pyodide, but it was a logical bug anyway.

Two mypy changes:

- `native_parser = true`, which makes mypy about 26% faster from a cold cache.
- More error codes, including `possibly-undefined` and `exhaustive-match`, plus fixes for everything they report.

Two of the fixes are real bugs:

- Pyodide set `built_wheel` instead of `repaired_wheel` when it reused a compatible wheel, so the test step failed with a `NameError`.
- The OCI container read loop had no EOF guard, so it looped forever if the container shell exited.

`bashlex` has no stubs, so `no-any-unimported` is disabled for `cibuildwheel.bashlex_eval` only.

### DIFF
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -540,7 +540,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     f"that is compatible with {config.identifier}. "
                     "Skipping build step..."
                 )
-                test_wheel = compatible_wheel
+                repaired_wheel = test_wheel = compatible_wheel
             else:
                 if build_options.before_build:
                     log.step("Running before_build...")
--- a/cibuildwheel/platforms/ios.py
+++ b/cibuildwheel/platforms/ios.py
@@ -540,7 +540,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     f"that is compatible with {config.identifier}. "
                     "Skipping build step..."
                 )
-                repaired_wheel = test_wheel = compatible_wheel
+                repaired_wheel = compatible_wheel
             else:
                 if build_options.before_build:
                     log.step("Running before_build...")
@@ -632,8 +632,6 @@ def build(options: Options, tmp_path: Path) -> None:
 
                 run_audit(tmp_dir=tmp_path, build_options=build_options, wheel=repaired_wheel)
 
-                test_wheel = repaired_wheel
-
             if build_options.test_command and build_options.test_selector(config.identifier):
                 if not config.is_simulator:
                     log.step("Skipping tests on non-simulator SDK")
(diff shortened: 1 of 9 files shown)


freakboy3742 (changes requested):
I agree there's a potential endless loop on EOF; but as the test suite bears out, theres either a testing gap or different handling required.

Other than that, the changes all makes sense to me.

henryiii:
The other PRs are stuck in the 3-hour timeout.

The test failure is an environment regression on the new GitHub runner images, not a behavior change from the EOF fix — the fix is what made it visible.

- On runner image 20260720 (podman 4.9.3), `test_podman_vfs` passed with this branch (attempt 1 of the CI run).
- On images 20260729+/20260810 (podman 5.8.4), it fails at container start with `crun: unknown version specified`. Podman 5.x generates an OCI spec v1.2.x config, which needs crun ≥ 1.14.3 (containers/podman#27272). Only this test breaks because it sets `CONTAINERS_CONF`, which makes podman ignore its normal config and fall back to the Ubuntu 24.04 system crun 1.14.1.

Before this change, that scenario busy-looped forever on EOF; now it errors immediately.

Pushed a fix to the test: it queries `podman info --format '{{.Host.OCIRuntime.Path}}'` and carries that runtime into the generated `containers.conf`, so the test uses the same runtime binary as podman's default configuration.

cibuildwheel/platforms/ios.py
```diff
@@ -540,7 +540,7 @@ def build(options: Options, tmp_path: Path) -> None:
                     f"that is compatible with {config.identifier}. "
                     "Skipping build step..."
                 )
-                test_wheel = compatible_wheel
+                repaired_wheel = test_wheel = compatible_wheel
```
joerick: in fact, `test_wheel` is not a useful variable in this file, it can just be removed and replaced with `repaired_wheel`.

joerick (approved):
Looks good to me. we can reduce complexity in ios.py if you fancy it.

joerick:
> potential endless loop on EOF; but as the test suite bears out, theres either a testing gap or different handling required

i don't think it's something possible without fiddling with the internals of the OCIContainer class, but I came up with this test that does check it-

```python
def test_abnormal_exit(container_engine: OCIContainerEngineConfig) -> None:
    container = OCIContainer(
        engine=container_engine, image=DEFAULT_IMAGE, oci_platform=DEFAULT_OCI_PLATFORM
    )
    with container:
        # force the container to exit abnormally
        container.bash_stdin.write(b"exit")
        with pytest.raises(RuntimeError):
            container.call(["echo", "hello"])
```

freakboy3742 (approved):
👍 All makes sense to me

PR #2972, opened 2026-08-19
feat: add log fold mode override

### LINKED ISSUE #2971: [FR] Can we have a toggle to opt out of CI-specific log folding per-run, please?
I've recently learned that running `cibuildwheel` on some emulated platforms introduces some UX/performance penalty in the log view, forcing people to fool the tool into thinking that it's not running in a CI env: https://github.com/aio-libs/aiohttp/pull/12814. I think it's silly to do so and this project should instead support an option to skip outputting the log-folding markers instead of forcing the end-users to resort to such a runtime theater.
It'd be nice to keep using the built-in standard CI-related tweaks that are auto-activated instead of loosing them because we have to resort to hacks like pretending there's no CI around in the first place.

Himanshuagrawal4:
## Summary

- add `CIBW_LOG_FOLD_MODE` as a per-run override for CI log folding
- support `azure`, `github`, `travis`, and `disabled`
- reject invalid values with a clear error
- document the environment variable and cover every accepted mode

Closes #2971

## Validation

- `pytest -q unit_test/wheel_print_test.py` (7 passed)
- Ruff check and format checks pass
- relevant pre-commit hooks pass; unrelated Windows/environment checks still report existing failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,12 +166,13 @@ The following diagram summarises the steps that cibuildwheel takes on each platf
 |  | [`test-skip`](https://cibuildwheel.pypa.io/en/stable/options/#test-skip) | Skip running tests on some builds |
 |  | [`test-environment`](https://cibuildwheel.pypa.io/en/stable/options/#test-environment) | Set environment variables for the test environment |
 |  | [`test-runtime`](https://cibuildwheel.pypa.io/en/stable/options/#test-runtime) | Controls how the tests will be executed. |
-| **Debugging** | [`debug-keep-container`](https://cibuildwheel.pypa.io/en/stable/options/#debug-keep-container) | Keep the container after running for debugging. |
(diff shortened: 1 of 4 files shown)


CAOShurong (commented):
Verified on the current head (7f62200) by installing the exact PR tree and exercising the Logger directly:

- all four accepted values (azure, github, travis, disabled) override the auto-detected fold mode;
- an invalid value raises exactly the documented ValueError ("CIBW_LOG_FOLD_MODE must be one of azure, github, travis, or disabled; got ...").

Two small observations, neither blocking:

1. Since the override is applied unconditionally in Logger.__init__, setting the variable outside CI also emits fold markers for a local run. That reads as intended ("override automatic detection"), just noting the side effect.
2. The docs entry is correctly marked env-var-only, so there is no pyproject-vs-env confusion — good.

Nice minimal scope for the feature; tests cover both branches. (Formal approval isn't permitted for my account on this repo, hence the comment review.)

joerick (approved):

PR #2978, opened 2026-08-31
fix: don't run call() through cmd.exe on Windows

r3wretrhy:
## Summary

- stop passing `shell=True` from `call()` on Windows
- keep `shutil.which` so PATH/PATHEXT lookup is unchanged
- add unit tests for `&`, `|`, `%VAR%`, `^`, and paths with spaces

## Why

`call()` already resolves the executable with `shutil.which` (CPython issue 52803), then still ran the argv list through `cmd.exe` on Windows. `cmd.exe` treats `&` and `|` as command separators and expands `%VAR%`, so a path or extra flag containing those characters is rewritten before the child starts.

Reproduced locally: `call(python, "-c", "print(sys.argv[1])", "a&b")` printed `a` and then tried to run `b` as a separate command. Python 3.11+ can start the resolved executable without a shell.

This is the leftover `shell=True` / Windows quoting case next to #2886 (`{project}` path spaces) and #2913 (`test-requires` spaces).

## Validation

- `uv run pytest unit_test/cmd_test.py -q` (7 passed)
- `uvx prek run --files cibuildwheel/util/cmd.py unit_test/cmd_test.py`
- `git diff --check`


burgholzer:
We also hit this error in one of our repositories. Below is the summary from Codex (got-5.6-sol). I added a small follow-up PR that could be merged into this one that also fully addresses the problems we have been encountering.

🤖 *AI text below* 🤖

I hit this root cause through `test-groups` on Windows: a generated `uv pip install ... mqt-core[pennylane,qiskit]<5,>=3.9.2` command was parsed by cmd.exe as input redirection and failed before uv started.

The central fix here is exactly right. I opened r3wretrhy/cibuildwheel#1 with two small completeness updates: add the actual `<`/`>` regression case, and remove shell-only quotes from the existing Windows `python -c` architecture diagnostic so it still executes after `shell=True` is removed.

r3wretrhy:
Folded those two cases into this PR (efbcbc3): the PEP 508 `<`/`>` extra, and the Windows `python -c` architecture check without cmd.exe quotes.

joerick:
I'd love this to be the right way forward.

I was initially concerned about backward-compat, but it seems most user-supplied commands run through `shell()` not `call()`. So I'm not too concerned there.

The other conundrum is why we were using `shell=True` on windows in the first place. I think I traced it to https://github.com/pypa/cibuildwheel/pull/978 - @mayeut, do you remember the particular issue you were trying to avoid? I seem to remember something about PATH resolution, but perhaps that's covered by the shutil.which stuff above.

r3wretrhy:
That matches what I could find in-tree. `shell=True` on Windows was the PATH/PATHEXT workaround for https://bugs.python.org/issue8557 (now https://github.com/python/cpython/issues/52803) — `subprocess` with `shell=False` did not search PATH the way cmd.exe does. You wrote that down in https://github.com/pypa/cibuildwheel/pull/672#issuecomment-877998123:

> execute in a shell on Windows to workaround a PATH resolution bug

`call()` later started resolving argv[0] with `shutil.which(..., path=PATH)` for the same issue, then still handed the already-absolute path to cmd.exe. Once `which()` has done the lookup, that original reason for the shell is gone. User-supplied command strings still go through `shell()`.

mayeut:
@joerick, https://github.com/pypa/cibuildwheel/pull/978 was just a refactor, `shell=True` was there before that: see #399 & #401

joerick:
Ah! But yes, we didn't have the shutil.which code then. So the only remaining mystery is why shell=True was readded back in https://github.com/pypa/cibuildwheel/pull/2076 ...

joerick:
I have a theory... this will probably fail when pypy-eol is enabled.

joerick:
Relevant failure is https://github.com/pypa/cibuildwheel/actions/runs/33867502794/job/101005921718?pr=2978#step:12:2828

The issue is https://github.com/pypy/pypy/issues/4958 - a bug that occurs in venv when pypy is called with an uppercase suffix - "python.EXE".

joerick:
With the help of Codex, have maybe found a solution. We'll see what CI makes of it.

joerick:
The previous CI run failed on the Sample Build, so I think my fix worked :) but all tests on `main` are now failing. I think we're waiting on https://github.com/pypa/build/pull/1176 and a build 1.6.1 release, and on https://github.com/pypa/manylinux/pull/1989 and then some updated docker images.

joerick (approved):

PR #2979, opened 2026-08-31
fix: preserve backslashes in container-engine create_args

r3wretrhy:
## Summary

- keep backslashes literal in `parse_key_value_string` (used by `container-engine` / `create_args`)
- add unit tests for unquoted and quoted Windows volume paths
- note the behaviour in the `container-engine` docs

## Why

`parse_key_value_string` tokenizes with POSIX `shlex`, which treats `\` as an escape. An unquoted Windows bind-mount such as

```
docker; create_args: --volume=C:\Users\foo\bar:/data
```

was parsed as `--volume=C:Usersfoobar:/data`. Quoted values with spaces already worked; unquoted Windows paths did not.

Same quoting class as #2886 (`{project}` / `{package}` backslashes eaten by a later `shlex.split`) and #2913 (`test-requires` split on spaces). Spaces inside a `create_args` value still need quotes, as documented.

## Validation

- `uv run pytest unit_test/utils_test.py::test_parse_key_value_string unit_test/oci_container_test.py::test_parse_engine_config unit_test/options_test.py::test_container_engine_option -q` (23 passed)
- `uv run pytest unit_test/utils_test.py unit_test/options_test.py -q` (142 passed, 3 existing xfails)
- `uvx prek run --files cibuildwheel/util/helpers.py unit_test/utils_test.py unit_test/options_test.py docs/options.md`
- `git diff --check`

### DIFF
--- a/cibuildwheel/util/helpers.py
+++ b/cibuildwheel/util/helpers.py
@@ -116,6 +116,9 @@ def parse_key_value_string(
     shlexer = shlex.shlex(key_value_string, posix=True, punctuation_chars=";")
     shlexer.commenters = ""
     shlexer.whitespace_split = True
+    # POSIX shlex treats `\` as an escape, so unquoted Windows paths in
+    # create_args become C:Usersfoo. Spaces still need quotes.
+    shlexer.escape = "\0"
     parts = list(shlexer)
     # parts now looks like
     # ['docker', ';', 'create_args:', '--some-option=value', 'another-option']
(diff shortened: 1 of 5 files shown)


joerick:
Ah, that's a little unfortunate. But it's pretty hard to fix this way, as backslash does already serve a purpose in escaping other characters, and is probably already in use. The solution for you would be to quote your argument, like this-

```yaml
CIBW_CONTAINER_ENGINE: docker; create_args: --volume="C:\Users\foo\bar:/data"
```

joerick:
Feel free to reopen if you want to continue discussion.

PR #2990, opened 2026-09-12
Fallback to system Python when setup-python has no toolcache

### LINKED ISSUE #2989: action.yml unconditionally calls actions/setup-python, which fails on native riscv64 self-hosted runners (no toolcache binary)
### Problem

When `pypa/cibuildwheel` is used via the GitHub Action on a native riscv64 self-hosted runner, the action immediately fails because `actions/setup-python` has no pre-built riscv64 Python binary in GitHub's toolcache:

```
Error: The version '3.11 - 3.14' with architecture 'riscv64' was not found for this operating system.
```

This matters now because riscv64 was just enabled by default in v4.1.1 (#2509) after PyPI accepted riscv64 uploads. Native riscv64 build hardware exists (e.g. RISE project runners, vendor self-hosted boards), and QEMU emulation is too slow for large projects. The recommended workflow — point a native riscv64 runner at the action and build — silently fails at the very first step.

The same problem affected aarch64 (#1188, #1623) and was never fixed at the action level; those issues were closed with "call cibuildwheel directly from the CLI instead." A workaround for Pyodide (#1868) was merged — but that fixed cibuildwheel's *internal* Python discovery, not the action's unconditional `setup-python` call.

Looking at `action.yml`, the call is hardcoded with no bypass path:

```yaml
- uses: actions/setup-python@...    # unconditional; no continue-on-error
  id: python
  with:
    python-version: "3.11 - 3.14"
    update-environment: false
```

The `$PYTHON` that the `cibw` step uses comes entirely from `steps.python.outputs.python-path`. If `setup-python` fails, the whole action fails.

Current workarounds all push complexity onto users: pre-install cibuildwheel system-wide and invoke it as a CLI, or populate the runner's toolcache with a hand-crafted fake venv (per #1623). Neither is acceptable as a permanent state for a newly-default-enabled architecture.

---

### Proposals

**Option 1 — Automatic fallback to system Python (minimal change, fully general)**

Add `continue-on-error: true` to the `setup-python` step and fall back to the system `python3` when setup-python doesn't produce a path. riscv64 runners running Ubuntu 24.04 ship Python 3.12 as the system Python, which is well within the `3.11 - 3.14` range cibuildwheel needs. This transparently fixes riscv64 today and any future arch that GitHub's toolcache doesn't cover, with zero user-facing API change.

```yaml
- uses: actions/setup-python@...
  id: python
  continue-on-error: true       # ← new
  with:
    python-version: "3.11 - 3.14"
    update-environment: false

# in the cibw step:
env:
  PYTHON: ${{ steps.python.outputs.python-path || 'python3' }}
```

Effort: ~3 lines. Risk: minimal — existing behavior is unchanged on all platforms that have toolcache support.

**Option 2 — Optional `python-path` action input (clean API, explicit opt-in)**

Add a `python-path` input that, when non-empty, skips `setup-python` entirely and uses the provided interpreter directly. Self-hosted runner users set `python-path: python3` (or a full path); GitHub-hosted users do nothing different. No behavioral change for existing workflows.

```yaml
inputs:
  python-path:
    description: >-
      Path to the host Python interpreter. When set, skips actions/setup-python.
      Useful on self-hosted runners (e.g. riscv64) where no toolcache binary exists.
    required: false
    default: ''
```

Effort: small (new input + conditional on the `setup-python` step). This is the cleanest API for users who want explicit control.

**Option 3 — Skip `setup-python` when `runner.arch` is unsupported (automatic, conservative)**

Gate the `setup-python` call on `runner.arch` being one of the arches that GitHub's toolcache actually covers (`X64`, `ARM64`). For anything else (`RISCV64`, `PPC64LE`, `S390X`, ...), skip directly to the system Python fallback. No input needed; no `continue-on-error` mask needed.

```yaml
- uses: actions/setup-python@...
  id: python
  if: runner.arch == 'X64' || runner.arch == 'ARM64'   # ← new
  with:
    python-version: "3.11 - 3.14"
    update-environment: false
```

Effort: one-line condition. Slightly more conservative than Option 1 (explicit allowlist rather than error-catching), but requires updating the list if GitHub adds new toolcache arches.

---

Of the three, Option 1 is the easiest to ship and the most future-proof. Option 2 is the right long-term API if users want explicit control. Option 3 is a reasonable middle ground if the maintainers prefer not using `continue-on-error`.

---

### Build log

The error message quoted above is verbatim output from `actions/setup-python` on an Ubuntu 24.04 riscv64 self-hosted runner. The full log is from a private CI run, but the failure is deterministic and reproducible on any riscv64 runner without a matching toolcache entry: `actions/setup-python` fetches [the versions manifest](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json), finds no `riscv64` entry, and exits with code 1, aborting the cibuildwheel action before it starts.

Native riscv64 GitHub Actions runners are available through the [RISE project](https://www.riseproject.dev/) (label `ubuntu-24.04-riscv`).

### CI config

https://github.com/juenglin/pykdtree/commit/9319d6b8fc10d0b868ac7824e7d60f8c787fb676

This public commit shows the workaround users are currently forced to adopt: add a separate `if: matrix.cibw_archs == 'riscv64'` step that installs cibuildwheel into a manual venv and invokes it as a CLI, bypassing `uses: pypa/cibuildwheel` entirely. The split between the action path (non-riscv64) and the CLI path (riscv64) is the maintenance burden this issue proposes to eliminate.

GruffElixir:
## Summary

On runners whose architecture has no `setup-python` toolcache build, the action skips `setup-python` and uses the runner's system `python3`. Callers can set `python-path` to select a Python 3.11+ interpreter with pip; setting it also skips `setup-python`. The Action documentation describes both paths.

Fixes #2989.

## Validation

- The config-file sample in the GitHub Actions test workflow sets `python-path: python3` on Linux to exercise the explicit path.
- `pre-commit run --files action.yml .github/workflows/test.yml docs/ci-services.md` (passed)
- PyYAML parse of `action.yml` and `.github/workflows/test.yml` (passed)
- `git diff --check` (passed)


henryiii:
I'd generally not recommend making a PR for something that requires choosing options and a maintainer has chosen an option yet. (Though, in this case, looking at the implementation was rather helpful to me. I think I'd at least open it as a draft until a path has been chosen, though!).

I don't think we are going this way since it can cause errors in setting up Python (which do happen) to silently change behavior.

GruffElixir:
Yep, updated this to the Option 2 + 3 combo from the issue thread and pushed `34709ff8`. `setup-python` now runs normally on X64/ARM64, unsupported arches skip it explicitly, and a `python-path` input lets callers choose a host interpreter without any silent `continue-on-error` fallback. The action YAML/pre-commit checks pass, and I refreshed the PR description too.

action.yml
```diff
@@ -30,6 +34,7 @@ runs:
   steps:
     - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       id: python
+      if: inputs.python-path == '' && (runner.arch == 'X64' || runner.arch == 'ARM64')
```
joerick, line 37: This threw me when I read it - the logic seemed reversed.

Perhaps a comment here would be nice e.g. `Only run if the user didn't set python-path and we're on a platform where setup-python is supported`.

joerick (approved):
Looks good to me. @henryiii ?

henryiii:
Quick comparison.

Near-identical solutions to #2989 (skip `setup-python` on non-X64/ARM64, add `python-path` input). The `action.yml` `if:` condition is the same in both. Differences:

|                      | #2990 (GruffElixir)                                 | #2991 (yours)                                                 |
| -------------------- | --------------------------------------------------- | ------------------------------------------------------------- |
| Fallback interpreter | `runner.os == 'Windows' && 'python' \|\| 'python3'` | `python3` always                                              |
| CI coverage          | none                                                | config-file Action job passes `python-path: python3` on Linux |
| Input description    | one line                                            | states 3.11+, pip, default behavior                           |
| Docs                 | appended to existing paragraph                      | new paragraph in the Action tab                               |
| Status               | approved by joerick                                 | reviewed (commented) by you                                   |

Substantive points:

- **Windows fallback**: #2990 is more correct in principle — `python3.exe` is not a standard Windows install name (it comes only from the Store stub or a venv). In practice the branch is unreachable, since Windows runners report `X64` or `ARM64`, so both behave the same today. Cheap to adopt regardless.
- **CI**: only #2991 exercises the new path. `runner.os` is valid in a step `with:`, so that line works, but it tests the explicit `python-path` input, not the arch-based auto-fallback (which cannot be tested without a riscv64 runner).
- Neither validates that the fallback interpreter is 3.11+; #2991 at least documents the requirement.

Recommendation: merge one and cherry-pick the other's better half — take #2990's Windows-aware fallback expression into #2991, or add #2991's CI line and input description to #2990. Since #2990 is already approved and from an outside contributor, pushing your two improvements there is the lower-friction path.

joerick:
Apologies, I missed the other PR. I'm easy. If we don't need the extra logic on Windows then simpler is better.

GruffElixir:
Thanks, I simplified the fallback to `python3`; the Windows-specific branch was unnecessary for the supported setup-python architectures. I also added the Linux config-file CI case to exercise `python-path` and documented the Python 3.11+ / pip requirement. Pre-commit, YAML parsing, and workflow validation pass locally; the updated CI run is underway.

action.yml
GruffElixir: Added a short comment above the condition to clarify when setup-python runs. The targeted pre-commit checks pass.

PR #2991, opened 2026-09-14
feat(action): add python-path input and skip setup-python on unsupported archs

### LINKED ISSUE #2989: action.yml unconditionally calls actions/setup-python, which fails on native riscv64 self-hosted runners (no toolcache binary)

henryiii:

Closes #2989

The action called `actions/setup-python` unconditionally, which fails on self-hosted runners for architectures without a toolcache build (riscv64, ppc64le, s390x).

- Skip `setup-python` when `runner.arch` is not X64 or ARM64 and fall back to `python3` on `PATH`.
- Add a `python-path` input to use a specific interpreter and skip `setup-python` explicitly.
- The host interpreter must be Python 3.11+ with pip, documented in the Action tab of the CI services page.
- CI now runs the config-file sample build through `python-path: python3` on Linux to cover the system-Python path.


action.yml
```diff
@@ -30,6 +37,7 @@ runs:
   steps:
     - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       id: python
+      if: inputs.python-path == '' && (runner.arch == 'X64' || runner.arch == 'ARM64')
```
henryiii, line 40: This is the option 3 part.

joerick (approved):

